PR #22, opened 2015-03-13
Make the bob example slightly more idiomatic

bitborn:
First pull request of hopefully many. :smile: 
The `is_empty` function appears to be used in more rust code, and might be the slightest bit faster as it's just an in-lined length check.

### DIFF
--- a/bob/example.rs
+++ b/bob/example.rs
@@ -6,7 +6,7 @@ pub fn reply(message: &str) -> &str {
 }
 
 fn is_silence(message: &str) -> bool {
-    message == ""
+    message.is_empty()
 }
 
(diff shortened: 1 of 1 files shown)


bitborn:
Uff da.
Looks like small individual pull requests aren't the way to go. :disappointed: 
I'll add some more commits to get travis on my side.

etrepum:
Yeah, the Rust track hasn't had passing tests in a while. It would be best to fix the existing exercises and tests to work with the current version of Rust (and make sure that's the Rust that Travis is trying to use) before trying to improve anything.

bitborn:
Sounds like a plan. Will do!

PR #24, opened 2015-03-31
Fetch appropriate configlet binary

dkinzer:
I noticed that when I am working on my Mac the fetched configlet binary fails.

  This pull request adds a limited but smarter configlet binary fetcher that
  tries to determine the OS and architecture and fetch the appropriate
  configlet binary.


kytrinyx:
The failure is unrelated to this change.

PR #25, opened 2015-03-31
Update Examples to the lastest Rust nightly

bitborn:
I updated the examples and tests to at very least build with the latest Rust nightly (build 2015-03-31). Hopefully this will let get the tests passing.

Obviously there's a lot more work to do, but I think just getting it to compile is a good start.

### DIFF
--- a/parallel-letter-frequency/example.rs
+++ b/parallel-letter-frequency/example.rs
@@ -1,5 +1,4 @@
-#![allow(unstable)] // for entry, as_slice, etc
-
+use std::ascii::AsciiExt;
 use std::collections::HashMap;
 use std::collections::hash_map::Entry;
 use std::sync::Future;
@@ -12,7 +11,7 @@ pub fn frequency(texts: &[&str], num_workers: usize) -> HashMap<char, usize> {
     let rem = texts.len() % num_workers;
     let part_size = if rem > 0 { part_size_floor + 1 } else { part_size_floor };
     let mut parts: Vec<Vec<String>> = Vec::with_capacity(part_size);
-    for _ in (0..num_workers) {
+    for _ in 0 .. num_workers {
         parts.push(Vec::with_capacity(part_size));
     }
     let mut i = 0;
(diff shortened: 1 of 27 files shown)


kytrinyx:
Oh, my, a passing build :) It's been a while!

@pminten, @etrepum I'll leave this to your discretion.

parallel-letter-frequency/example.rs
```diff
@@ -41,9 +40,9 @@ pub fn frequency(texts: &[&str], num_workers: usize) -> HashMap<char, usize> {
 fn count(lines: Vec<String>) -> HashMap<char, usize> {
     let mut results: HashMap<char, usize> = HashMap::new();
     for line in lines.iter() {
-        for c in line.as_slice().chars() {
+        for c in line.chars() {
             if c.is_alphabetic() {
-                match results.entry(c.to_lowercase()) {
+                match results.entry(c.to_ascii_lowercase()) {
```
pminten: The test contains some non-7-bit-ASCII characters. I'm sure the tests pass with this but I'm not sure that's by accident or because it does the right thing.

pminten:
Looks good to me, though I don't have enough experience to do give anything but the most cursory review.

parallel-letter-frequency/example.rs
bitborn: I believe the [ASCII lowercase](http://doc.rust-lang.org/std/ascii/trait.AsciiExt.html#tymethod.to_ascii_lowercase) method does a check to see if the char is actually ASCII. Though if we included a test with both a upper and lower case Unicode character (example: "Θ θ"), I believe it would count these as separate characters...

I will update using the [Unicode version](http://doc.rust-lang.org/std/primitive.char.html#method.to_lowercase) which should, if I read the docs correctly, count Unicode upper and lower ("Θ θ") as the same character.

bitborn:
Just added the Unicode method in the parallel frequency example. 

I think there may be an edge case we need to worry about down the road. Basically, in future versions the `to_lowercase` method may return an iterator of more than one character (which I believe is called case folding). Theres a little blurb about it in the docs for [`to_uppercase`](http://doc.rust-lang.org/std/primitive.char.html#method.to_uppercase).

I also updated this PR to the latest nightly.

pminten:
I wouldn't worry about case folding returning multiple codepoints, that should be pretty rare outside pathological constructed cases.

The rest looks good. Since nobody else has commented I'll just merge this.

pminten:
Merged.

Thanks for the massive work.

PR #26, opened 2015-04-04
Use cargo for builds

### LINKED ISSUE #18: Modifying all exercises to work with cargo
It looks like cargo will be the preferred way to run tests going forward.  Since it comes with Rust, we should probably set up the exercises so that `cargo run` and `cargo test` can be used.  I can help by making the changes.

bitborn:
Hey all,

I moved some stuff around so that now all the examples can be built using cargo! This should close #18. 

I also added a few quick changes to get the builds working with the beta channel. (e.g. removing all the feature flags, which was rough :smile:) 

I'd love to get some feedback. Obviously the current examples aren't the most efficient or succinct code, (e.g. the `lowercase` methods) but I'm thinking once this lands we can focus on rewriting the examples to be more clear.

Thanks


bitborn:
Whew. I messed up the shebang for Travis CI, all should be well now.

If someone could give the `_tests/check_exercises.sh` script a thorough looking over, that would be super helpful. I'm not exactly fluent in bash...

pminten:
Can't say I'm terribly good at bash, but I'd change it to this, which uses a subshell to avoid "cd forward; cd back":

```bash
#!/bin/bash

set -e
tmp=${TMPDIR:-/tmp/}

# a 'tests' directory
for exercise in */tests; do
    # This assumes that exercises are only one directory deep
    # and that the primary module is named the same as the directory
    directory=$(dirname "${exercise}");

    workdir=$(mktemp -d "${tmp}${directory}.XXXXXXXXXX")

    cp -R -T $directory $workdir

    # Run in subshell to change workdir without affecting current workdir.
    (
        cd $workdir
        cp example.rs "${directory}.rs"

        # Forcibly strip all "ignore" statements from the testing files
        for test in tests/*.rs; do
            sed -i '/\[ignore\]/d' $test
        done

        # Run the test and get the status
        cargo test
    )

    status=$?

    if [ $status -ne 0 ]
    then
        echo "Failed";
        return 1;
    fi
done
```

bitborn:
Oh, that's clever. I'll update.

anagram/tests/anagram.rs
```diff
@@ -1,11 +1,6 @@
-#![crate_name = "anagram_test"]
-#![crate_type = "lib"]
-#![feature(collections)]
-
-mod anagram;
+extern crate anagram;
 
 #[test]
-#[ignore]
```
EduardoBautista: We should probably leave the ignore tags in the file.  The user is supposed to enable the tests one by one as they get them to pass.

lpil: Will the users know to do this?

kytrinyx: We could put this as instructions in the SETUP.md file which gets included in each problem's README.md

lpil: Sounds good.

Other languages are defaulting to having all the tests enabled and then letting the user disable them. Should we be consistent here?

EduardoBautista: @lpil Some, not all.  The Ruby exercises have the first one enabled and the rest on "skip".

kytrinyx: The languages are not consistent at all here, and I don't necessarily think that we should strive for consistency in this respect.

The javascript and ruby tracks have a lot more beginners than other tracks, and beginners tend to get overwhelmed by seeing a lot of failures.

The Go track is my favorite in terms of the behavior of the test suite. It generally uses table-tests and fails fast so that you get only the first failure, but you don't have to manually change the test suite to get the next test.

pminten: It's a bit of a matter of language style though. Table driven testing is fairly embraced in Go's culture but it doesn't really chime with styles like BDD (which saddens me, since table driven is hugely effective test design). BDD is what the javascript track uses since BDD frameworks like Jasmine are very popular there.

kytrinyx: Yeah, I do wish table-driven tests were more common in other language communities. You're right though. I don't really see it happening in JavaScript or Ruby. And since exercism does seem to be a little bit about getting a feel for the rhythm and process of TDD it seems like it would be wrong to introduce table-driven tests in a language track where nobody does that out in the wild.

EduardoBautista:
So what's the status on this PR?

kytrinyx:
That's a great question. I suspect that it fell off the radar. @pminten @etrepum -- is this ready to go in your opinion?

pminten:
@kytrinyx Negative, when running the tests with rust 1.0.0 I get compile failures at the robot-name exercise (when running `cargo test`):

```
  Compiling libc v0.1.5
   Compiling log v0.3.1
   Compiling rand v0.3.7
/home/peter/.cargo/registry/src/github.com-1ecc6299db9ec823/rand-0.3.7/src/distributions/mod.rs:20:16: 20:21 error: unresolved import `std::num::Float`. There is no `Float` in `std::num`
/home/peter/.cargo/registry/src/github.com-1ecc6299db9ec823/rand-0.3.7/src/distributions/mod.rs:20 use std::num::{Float, Int};
                                                                                                                  ^~~~~
/home/peter/.cargo/registry/src/github.com-1ecc6299db9ec823/rand-0.3.7/src/distributions/mod.rs:20:23: 20:26 error: unresolved import `std::num::Int`. There is no `Int` in `std::num`
/home/peter/.cargo/registry/src/github.com-1ecc6299db9ec823/rand-0.3.7/src/distributions/mod.rs:20 use std::num::{Float, Int}
```

Seems to come from the `extern crate rand` in the robot-name example.

That said, I don't see any reason not to push this to live once the tests are green.

EduardoBautista:
I can go through these commits, cherry-pick the ones that I believe still work, and open a new PR.  This is blocking some issues from closing.

EduardoBautista:
These changes were included in #27 and I also got it to work on Rust 1.0.0

EduardoBautista:
This can be closed @etrepum since the other PR got merged.

lpil:
:tada: :dancers::dancers::dancers::dancers: :tada:

kytrinyx:
Thanks so much @EduardoBautista!

PR #27, opened 2015-05-19
Use Cargo for builds

### LINKED ISSUE #18: Modifying all exercises to work with cargo

EduardoBautista:
Continuing the work from: https://github.com/exercism/xrust/pull/26

@pminten I also included your bash script and fixed that compilation error.

Closes #18


EduardoBautista:
Looking into the failure...

EduardoBautista:
Looking into the failure...

EduardoBautista:
Looking into the failure...

EduardoBautista:
And we are green!

etrepum:
Green! Woot :) Nice work!! :+1:

PR #31, opened 2015-05-30
New exercises: raindrops and difference-of-squares

adolfosilva:
Implementation of two new exercises: raindrops and difference-of-squares.
Any problem, let me know.


EduardoBautista:
Looks good to me!  Thanks again @adolfosilva!

PR #34, opened 2015-06-01
Added hexadecimal exercise.

razielgn:
:cherry_blossom:

### DIFF
--- a/config.json
+++ b/config.json
@@ -18,6 +18,7 @@
     "bob",
     "grade-school",
     "phone-number",
+    "hexadecimal",
     "beer-song",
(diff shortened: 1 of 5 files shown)


EduardoBautista:
The asserts should probably be in separate tests, here is an example in Ruby:

https://github.com/exercism/xruby/blob/master/hexadecimal/hexadecimal_test.rb

razielgn:
Yes, it's better now. :smiley: :+1:

EduardoBautista:
Looks good!  Thanks!

razielgn:
Thank you. :)

PR #35, opened 2015-06-04
Added forth.

razielgn:
Can't say it wasn't fun!
I went ahead and added an exercise skeleton, to guide the student.
Let me know if it's ok or not. :)


pminten:
I like it. The code is very readable, feels idiomatic and doesn't suffer from unwrapitis.

Thanks for putting in the effort to translate such a big exercise.

razielgn:
Thank you! :bow:

PR #36, opened 2015-06-07
Add rectangles exercise

pminten:


EduardoBautista:
Some of the tests are missing the ignore flag (`#[ignore]`) to be consistent with the rest of the exercises.

pminten:
Added the #[ignore] flags.

EduardoBautista:
This looks good to me.  If you can't think of anything else to improve we should merge this.

pminten:
Thanks for the review.

PR #39, opened 2015-07-16
add test in anagram, that bAnana is not an anagram of banana

stevejb71:
Seems reasonable to ignore case when checking for anagrams.


etrepum:
Thanks! :+1:

PR #40, opened 2015-07-18
Ignore the final test on anagram (initially, only the first test should run)

stevejb71:
Ignore the final test on anagram (initially, only the first test should run)


etrepum:
Thanks! :+1:

PR #41, opened 2015-07-24
Missing word

yonson2:
`test_does_not_detect_a_word_as_its_own_anagram` should contain "banana" inside `outputs` instead of being empty.


kytrinyx:
Thank you.

"banana" is not an anagram of "banana" -- for it to be an anagram, two criteria must be fulfilled:

1. same letters
2. in a different order to produce a new word

From wikipedia:

> An anagram is a type of word play, the result of rearranging the letters of a word or phrase to produce a new word or phrase, using all the original letters exactly once

yonson2:
Oh, I didn't notice that, my bad.

kytrinyx:
No worries! Thanks for jumping in :)

PR #43, opened 2015-07-30
Improve tests in grade school.

stevejb71:
Make names more explicit (to show that values are returned sorted from
grades and grade).
Add more tests for the empty cases.

### DIFF
--- a/grade-school/tests/grade-school.rs
+++ b/grade-school/tests/grade-school.rs
@@ -5,22 +5,70 @@ fn stringvec_to_strvec(v: &Vec<String>) -> Vec<&str> {
 }
 
 #[test]
-fn test_add_student() {
+fn test_grades_for_empty_school() {
+    let s = school::School::new();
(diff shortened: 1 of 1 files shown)


EduardoBautista:
@kytrinyx How often do test cases differ from other languages?

kytrinyx:
@EduardoBautista pretty often. We're trying to get better at two things:

1. Making sure that great test cases spread into all the languages (https://github.com/exercism/todo/issues/13), and
2. Making sure problems in each language follows that language's idioms and patterns

EduardoBautista:
Well, these tests look fine to me.  @kytrinyx so is it OK to merge this in your opinion even though the tests are a bit different from other languages?

kytrinyx:
@EduardoBautista Yepp, go for it. As long as it won't be confusing when they look at the README, this is totally fine.

EduardoBautista:
Thanks @stevejb71!

PR #44, opened 2015-07-31
Use rust style error handling in phone number

stevejb71:
Using 00000000 as an invalid number is rather odd - I think Option is much better style in Rust and properly uses the type system.

What do you think?

### DIFF
--- a/phone-number/example.rs
+++ b/phone-number/example.rs
@@ -1,4 +1,4 @@
-pub fn number(s: &str) -> String {
+pub fn number(s: &str) -> Option<String> {
     let digits: String = s
         .chars()
         .filter(|&c| c.is_digit(10))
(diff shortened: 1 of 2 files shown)


kytrinyx:
Yeah, 00000000 was the invalid number when I originally came across the exercise, but it's not really a great choice in all languages. I think Option is probably better here.

phone-number/example.rs
```diff
@@ -10,19 +10,18 @@ pub fn number(s: &str) -> String {
...
+    number(s).map(|n| n[..3].to_string())
 }
 
-pub fn pretty_print(s: &str) -> String {
-    let n = number(s);
-    format!("({area}) {prefix}-{exchange}",
-            area=&n[..3],
-            prefix=&n[3..6],
-            exchange=&n[6..])
+pub fn pretty_print(s: &str) -> Option<String> {
```
EduardoBautista: I am thinking that maybe `pretty_print` should return just a string with "Invalid Phone Number" in the case of `number` returning `None`.

stevejb71: Sounds reasonable. And gives a chance to use or_else on the option. I'll update the PR if there are no further comments.

EduardoBautista: @stevejb71 Yeah, that is my only comment.  Thanks for the help!

EduardoBautista:
Thanks!

Issue #45, opened 2015-08-09
handle different lengths error case in point-mutations


ekse:
The current tests for the point-mutations exercise assumes that both strings are of the same length. `hamming_distance()`  should return an error when strings of different lengths are passed. I think the idiomatic way to this in Rust would be to change the function signature to return a Result

```rust
pub fn hamming_distance(a : &str, b: &str) -> Result<u32, &'static str> 
```
For expected tests cases a call to unwrap() needs to be added, for  error cases we can compare with the error.

```rust
#[test]
fn test_no_difference_between_empty_strands() {
    assert_eq!(dna::hamming_distance("", "").unwrap(), 0);
}

#[test]
fn test_second_string_is_longer() {
    assert_eq!(dna::hamming_distance("A", "AA"), Result::Err("inputs of different length"));
}
```

I will send a pull request for this issue.

PR #46, opened 2015-08-09
point-mutations: handle error case for different lengths ; robot-name: fix name format

ekse:
This is related to issue #45. hamming_distance() now returns a Result. I added 2 test cases to test for this error.


EduardoBautista:
Thank you!

PR #48, opened 2015-09-07
Add minesweeper exercise

stevejb71:
Here's the minesweeper exercise from the xhaskell stream.

My first attempt at contributing an exercise, and I'm a Rust beginner, so pull with care :).

### DIFF
--- a/config.json
+++ b/config.json
@@ -21,6 +21,7 @@
     "phone-number",
     "hexadecimal",
     "beer-song",
+    "minesweeper",
     "dominoes",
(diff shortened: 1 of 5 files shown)


EduardoBautista:
Thanks for helping out!  In Rust, we prefer to use 4 spaces instead of tabs, so if you can make that change it would be appreciated.

stevejb71:
I've replaced tabs with spaces.

kytrinyx:
Thanks so much, @stevejb71!

PR #49, opened 2015-09-08
Fixed false positive with test_multiple_anagrams

medek:
If the order returned by anagrams_for is different from the test
case's expected output, the test would fail while being correct.

```
---- test_multiple_anagrams stdout ----
        thread 'test_multiple_anagrams' panicked at 'assertion failed: `(left == right)` (left: `["largely", "regally", "gallery"]`, right: `["gallery", "regally", "largely"]`)', tests/anagram.rs:42
```
Ordering both arrays before comparison fixes the false positive.


kytrinyx:
Good catch, thanks!

PR #50, opened 2015-09-23
Dominoes tests

petertseng:
I mostly explained these tests in the individual commit messages, so take a look there. I had to fix example.rs to give correct results for the new tests (it was incorrect on a few).

I believe Rust is the only language that is using the dominoes exercise, hopefully I'm not wrong on that point (otherwise, it would probably be prudent to port these tests to the other languages as well, right?)

### DIFF
--- a/dominoes/tests/dominoes.rs
+++ b/dominoes/tests/dominoes.rs
@@ -96,9 +96,51 @@ fn no_repeat_numbers() {
 
 #[test]
 #[ignore]
+fn can_reverse_dominoes() {
+    let input = vec![(1, 2), (1, 3), (2, 3)];
+    assert_correct(&input);
+}
+
+#[test]
+#[ignore]
 fn invalid_input() {
     let input = vec!((1, 2), (4, 1), (2, 3));
     assert_eq!(dominoes::chain(&input), None);
+}
+
+#[test]
+#[ignore]
+fn disconnected_simple() {
+    let input = vec![(1, 1), (2, 2)];
+    assert_eq!(dominoes::chain(&input), None);
+}
+
+#[test]
+#[ignore]
+fn disconnected_double_loop() {
+    let input = vec![(1, 2), (2, 1), (3, 4), (4, 3)];
+    assert_eq!(dominoes::chain(&input), None);
+}
+
+#[test]
+#[ignore]
+fn disconnected_single_isolated() {
+    let input = vec![(1, 2), (2, 3), (3, 1), (4, 4)];
+    assert_eq!(dominoes::chain(&input), None);
+}
+
+#[test]
+#[ignore]
+fn need_backtrack() {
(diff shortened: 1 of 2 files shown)


kytrinyx:
Yepp, according to http://x.exercism.io/problems/dominoes only Rust has the dominoes exercise. I'll leave this for the track maintainers to review. Thanks!

EduardoBautista:
Hmmm, is this `assert_correct` function actually making an assertion?  Never took a look into this exercise until now.  But it does look to have improved quite a bit.  I am also thinking that we should return a Result instead of an Option in the `chain` function.  What's your opinion on this @petertseng.

dominoes/tests/dominoes.rs
```diff
@@ -64,6 +64,16 @@ fn check(input: &Vec<Domino>) -> CheckResult {
     }
 }
 
+fn assert_correct(input: &Vec<Domino>) {
+    match check(&input) {
+        Correct => (),
+        GotInvalid => panic!("Unexpectedly got invalid on input {:?}", input),
+        ChainingFailure(output) => panic!("Chaining failure for input {:?}, output {:?}", input, output),
+        LengthMismatch(output) => panic!("Length mismatch for input {:?}, output {:?}", input, output),
+        DominoMismatch(output) => panic!("Domino mistmatch for input {:?}, output {:?}", input, output),
```
petertseng: typo here, mistmatch. Can take this opportunity to fix that.

petertseng:
> Hmmm, is this assert_correct function actually making an assertion?

I would say it is, based on the fact that it will panic if the path returned is 1) using dominoes that were not provided, 2) not correctly connected `[(1, 2), (3, 1)]` for example, 3) an incorrect length. this is mostly done by the `check` function; the `assert_correct` is just interpreting the results. The two functions could even be combined, really. I could make that happen if that is desired.

> I am also thinking that we should return a Result instead of an Option in the `chain` function.

By this, do you mean that we should ask the `chain` function to explain why a chain cannot be created? What would be the possible error values that it might return, and would you suggest that specific error values be tested for in the tests?

I think adding in tests for "the `chain` function differentiates between the cases where a Euclidean cycle is impossible because a vertex has odd degree versus the cases where the graph is disconnected" would give away too much information about how this should be solved, so I currently am in favor of keeping it as an Option, but I think I want to hear your proposal for what possible values the `Result` might contain to see if it will change my mind.

petertseng:
I hope it is not too imposing to ask about any further thoughts on this. I think I'm still waiting to hear back on the ideas of what kind of `Err` the `Result` of `chain` might return (if we want to make `chain` return a `Result`)

EduardoBautista:
Oops, forgot to check this.  Everything looks good in this PR.  I am still not 100% convinced about the changes I proposed so I prefer not to fix what isn't necessarily broken. Thanks and sorry it took so long @petertseng!

PR #51, opened 2015-10-06
dominoes: add a test with 2 disjoint domino chains

ekse:
I suggest adding this test case as this situation is not currently covered by the tests. I actually had a bug in my code that is triggered by it but not by the current tests, I suspect it could elicit bugs in the code of other participants.


petertseng:
This had been added at https://github.com/exercism/xrust/blob/master/dominoes/tests/dominoes.rs#L121 (basically, I noticed the same thing too)

petertseng:
Ahhh I get it now, we talked about this on your submission

ekse:
my bad, I thought I had compared with the latest version before submitting but I did a git pull of my fork (oops).

PR #52, opened 2015-10-08
add exercise circular-buffer

ekse:
The tests were ported from the Ruby version, the example implementation is similar to the Go example.

### DIFF
--- a/circular-buffer/tests/circular-buffer.rs
+++ b/circular-buffer/tests/circular-buffer.rs
@@ -1,0 +1,90 @@
+extern crate circular_buffer;
+
+#[allow(unused_must_use)]
+mod tests {
+
+    use circular_buffer::{CircularBuffer, Error};
+
+    #[test]
+    fn error_on_read_empty_buffer() {
+        let mut buffer = CircularBuffer::new(1);
+        assert_eq!(Err(Error::EmptyBuffer), buffer.read());
+    }
+
+    #[test]
+    #[ignore]
+    fn write_and_read_back_item() {
+        let mut buffer = CircularBuffer::new(1);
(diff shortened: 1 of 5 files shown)


circular-buffer/tests/circular-buffer.rs
```diff
@@ -0,0 +1,79 @@
...
+    fn full_buffer_error() {
+        let mut buffer = CircularBuffer::new(2);
+        buffer.write('1');
+        buffer.write('2');
+        assert_eq!(Err(Error::FullBuffer), buffer.write('3'));
+    }
+
+    #[test]
+    #[ignore]
+    fn overwrite_item_in_full_buffer() {
```
petertseng: Maybe could add a test for using overwrite in a non-full buffer too (like in Ruby) - create with size 2, write 1, overwrite 2, read should return 1 instead of 2

ekse: good point, I will add this test tonight.

ekse: I added the test. I actually had a bug in there due to a typo, oopsies.

EduardoBautista:
We don't usually include the `lib.rs` file.  Must have gotten accepted by accident in the "forth" exercise.

ekse:
I can remove it, the reason I included was to provide the Error enum definition.

EduardoBautista:
@ekse I don't think it's too unreasonable to expect the Enum to be defined as a part of the exercise.  If you can remove it, please do.

ekse:
I removed the lib.rs, I also merged the commits together.

EduardoBautista:
Looks good to me!  Thanks!

PR #53, opened 2015-10-12
refactored example.rs for exercise circular-buffer

ekse:
I did a couple changes to the example code of circular-buffer based on suggestions by user petertseng. I think this version is much more readable.


EduardoBautista:
Cool!  Thanks!

PR #55, opened 2015-12-04
leap: Check another exceptional century

JIghtuse:
It is possible to have green tests and not exactly correct solution. See, for example: http://exercism.io/submissions/a5147caadb714d428816389ec6fa62f0


EduardoBautista:
Nice catch!  Thanks!

PR #56, opened 2015-12-07
Normalize YAML in Travis CI config

kytrinyx:
The YAML was not quite valid. Travis can read it
and wasn't complaining, but this ensures that the YAML
is clean.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
+---
 language: rust
 script:
-  - './_test/check-exercises.sh'
-  - './bin/fetch-configlet'
-  - './bin/configlet .'
+- "./_test/check-exercises.sh"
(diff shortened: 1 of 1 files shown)


EduardoBautista:
So it's not supposed to be indented?  That's the first time I have seen it like that actually.

kytrinyx:
I was actually thinking about the `---` at the very top. I ran it through a tool that removed the indenting (I don't think that indentation is right or wrong, I just went with whatever the tool said).

EduardoBautista:
Oh, didn't notice that haha.

EduardoBautista:
@kytrinyx can this be merged or are you going to add something?

kytrinyx:
It's fine. I didn't want to put it straight into master in case someone else caught something I missed.

PR #58, opened 2015-12-13
Add ABOUT.md and LEARNING.md

### LINKED ISSUE #57: Need ABOUT.md
We need a short, friendly introduction to Rust for the website as specified [here] (http://exercism.io/languages/rust). 

The description should at minimum reference Rust's focus on safety and the combination of low-level control with functional features such as pattern matching. 

An ABOUT.md file should be made and put into docs/.

outkaj:
Fixes #57

### DIFF
--- a/docs/ABOUT.md
+++ b/docs/ABOUT.md
@@ -1,0 +1,7 @@
+Rust is a fast compiled language which supports functional, imperative, and object-oriented
+programming. It has built-in memory safety without garbage collection and is ideal for
+systems and low-level programming, while still retaining the abstractions of a higher-level
+language.
+
+Rust has excellent documentation and a beginner-friendly community. Dive right in at [the Rust
+  website] (https://www.rust-lang.org/).
--- a/docs/LEARNING.md
+++ b/docs/LEARNING.md
@@ -1,0 +1,1 @@
+Please see the "Useful Rust Resources" tab.


kytrinyx:
I like this. What is Rust typically used for right now?

johanhelsing:
Large projects I know about:

* The rust compiler
* The Servo browser engine

Seems like things that need performance and benefit from memory safety guarantees.

kytrinyx:
This is helpful. What kind of small projects would people typically use Rust for these days?

kytrinyx:
@johanhelsing @petertseng @EduardoBautista I'd like to be able to say something about what you or I might use Rust for in the documentation. Any suggestions?

I'd love to get this merged.

johanhelsing:
There has popped up a large project recently, Redox os, which may be worth mentioning.

Even though it's a large project, it has lots of small parts that are good examples of what to use rust for. I.e. command line applications, drivers, a display server, file system implementations.

As for smaller projects, I would also consider using rust through ffi from slower languages when I need a performance boost for something computation intensive.

petertseng:
Small note, we're going to deal with the fact that a [different ABOUT.md](https://github.com/exercism/xrust/blob/master/docs/ABOUT.md) got merged while this PR was waiting. What parts should we take from this one to combine with the other one?

Should LEARNING.md be different from RESOURCES.md? I see from https://github.com/exercism/x-api/blob/master/CONTRIBUTING.md README that LEARNING should be geared toward learning from scratch whereas RESOURCES is more general. Should the Rust book be moved from RESOURCES into LEARNING? Anything else?

When I think about why I'd use Rust, my answer is the somewhat vague answer of "when I want more assurance that I've done it right" - basically when I want the compiler to have my back. A very often heard thing for languages with more compile-time checks. I would like to add some clarity to that.

kytrinyx:
I don't think that this ABOUT has anything that the other one is missing. I do think that it would be nice to add something about why someone might want to learn it (compiler has your back) and what it's currently being used for (large projects and also what someone would use it for day-to-day, e.g. command line applications, drivers, a display server, file system implementations, etc).

The Rust book seems like it would be more geared towards learning than ongoing development, so it probably should be moved.

I'm going to go ahead and close this, since I don't think that it should be merged directly, but rather this discussion might be used to improve what's there.

PR #59, opened 2015-12-14
Add hello-world exercise

winterchord:
This commit fixes https://github.com/exercism/todo/issues/47.

Disclaimer, I do _not_ have a "working" knowledge of rust, like it
requires in the reference issue. All feedback is definitely welcome.

### DIFF
--- a/config.json
+++ b/config.json
@@ -4,6 +4,7 @@
   "repository": "https://github.com/exercism/xrust",
   "active": true,
   "problems": [
+    "hello-world",
     "leap",
     "anagram",
     "difference-of-squares",
--- a/hello-world/Cargo.lock
+++ b/hello-world/Cargo.lock
@@ -1,0 +1,4 @@
+[root]
+name = "hello-world"
+version = "0.0.0"
+
--- a/hello-world/Cargo.toml
+++ b/hello-world/Cargo.toml
@@ -1,0 +1,3 @@
+[package]
+name = "hello-world"
+version = "0.0.0"
--- a/hello-world/GETTING_STARTED.md
+++ b/hello-world/GETTING_STARTED.md
@@ -1,0 +1,160 @@
+# Getting Started
+
+These exercises lean on Test-Driven Development (TDD), but they're not
+an exact match.
+
+The following steps assume that you are in the same directory as the exercise.
+
+You must have rust installed.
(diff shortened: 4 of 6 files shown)


EduardoBautista:
Sending an empty string as an argument looks a bit weird.  I was thinking that maybe using an [Option](https://doc.rust-lang.org/std/option/) might be better in this situation although I can't recall it being used that often as a parameter value.  I will wait for someone else to comment on this, but it looks great!

winterchord:
@EduardoBautista, ah, that looks useful. I agree that it looks weird. I was looking for default function arguments (ala ruby or python). Guess I wasn't looking hard enough!

I'll play around with it; wondering how if-clauses deal with `Option`s. It's entirely possible that people new to rust will go for an if-else solution versus a `match` one.

winterchord:
Using `Option` I end up with tests that look like:

```rust
assert_eq!("Hello, World!", hello_world::hello(None));
assert_eq!("Hello, Alice!", hello_world::hello(Some("Alice")));
```

and an implementation that is either:

```rust
pub fn hello(name: Option<&str>) -> String {
    match name {
        Some(ref n) => format!("Hello, {}!", n),
        None => "Hello, World!".to_string(),
    }
}
```

or

```rust
pub fn hello(name: Option<&str>) -> String {
    if let Some(n) = name {
        return format!("Hello, {}!", n);
    }

    "Hello, World!".to_string()
}
```

I'm not sure if this is any better from a rust-readability standpoint and whether it's the right topics for a beginner to learn (`Option`, `if let`) for the first exercise. If it is the right approach, I can submit a follow-up  with the changes. Does this project prefer that I squash (and force push) or add on another commit?

EduardoBautista:
@winterchord 
```rust
pub fn hello(name: Option<&str>) -> String {
    match name {
        Some(ref n) => format!("Hello, {}!", n),
        None => "Hello, World!".to_string(),
    }
}
```

Is preferred.  Also, squashing plus a force push is preferred.  Thanks!

winterchord:
@EduardoBautista Ok. Squashed and force pushed the suggested changes. Can you or another collab give it a once over, when you have a moment?

Went with the `Option` implementation with the accompanying test and `GETTING_STARTED` changes.

EduardoBautista:
@winterchord Looks good!  Thanks!

PR #60, opened 2015-12-16
Detect false positive from `eq_ignore_ascii_case`

chancancode:


EduardoBautista:
I think this is OK haha.  Not too familiar with greek letters.  I am assuming that `γ` is the lowercase version of `Γ` is that correct?

chancancode:
Yes, it is based on another test case in the file:

```rust
#[test]
#[ignore]
fn test_unicode_anagrams() {
    // These words don't make sense, they're just greek letters cobbled together.
    let inputs = ["ΒΓΑ", "ΒΓΔ", "γβα"];
    let outputs: Vec<&str> = vec!["ΒΓΑ", "γβα"];
    assert_eq!(anagram::anagrams_for("ΑΒΓ", &inputs), outputs);
}
```

EduardoBautista:
OK, looks good, thanks!

chancancode:
In the ascii case, there are two tests about lowercase letters:

1. `Carthorse` is an anagram of `Orchestra`
2. `bAnana` is not an anagram of `banana`

This adds the second case to the unicode (greek letters) version

kytrinyx:
Weird, it seems like (1) and (2) are incompatible with each other.

I guess the question boils down to whether or not anagrams are case insensitive or not. (I believe that they are, but I could probably be swayed by an argument against it).

chancancode:
@kytrinyx I don't think they are incompatible. To be clear, the first case is expressing that anagrams are case-insensitive; and the second case is expressing that `banana` is not an anagram of `banana` (because they are the same word) and that the implementation should not be fooled simply because `bAnana` has an upper-case `A`.

kytrinyx:
oh, hah. Yes, you're absolutely right. Thank you.

PR #63, opened 2015-12-19
Switch to container-based Travis CI infrastructure

kytrinyx:
This gets us off the legacy build system, which means
that the builds will start more quickly and also likely
run more quickly (and with more consistent build times).

See https://docs.travis-ci.com/user/migrating-from-legacy.


EduardoBautista:
Oh!  This looks cool, I should start using this on my builds for other projects.

Also, I see no issues in the build so I will merge.

PR #77, opened 2016-02-29
Rename point-mutations to hamming

### LINKED ISSUE #74: "point-mutations" is deprecated in favor of hamming
This happened a while back, and it was for really weird legacy reasons.

I've since fixed the underlying issues that caused the problem, but for consistency
it would be nice to rename point-mutation to hamming, so that all the tracks are using
the same exercise name.

Once the problem has been renamed, I can run a script on the website to point people's
existing point-mutations solutions to the new hamming exercise so that they'll be able
to review solutions to hamming, and people who solve the new hamming exercise can see
all the old ones.

IanWhitney:
Fixes #74

A simple renaming of the old exercise to the new name. I also made sure
the tests still pass with the example solution.


kytrinyx:
Thanks Ian, I'll do the migration to make sure everyone has access to the right stuff on the site.

PR #80, opened 2016-03-06
Implement Roman Numerals exercise

IanWhitney:
I've implemented the standard tests and provided an example solution
that I think isn't terrible. I pretty much stole it from the Ruby
example!


IanWhitney:
Ah. I hadn't noticed the need to change `config.json`. That's why my builds were breaking. Fixing that.

EduardoBautista:
Looks good, thanks!

PR #81, opened 2016-03-08
rna-transcription: Remove as_ref test

### LINKED ISSUE #76: Remove the `as_ref` check in RNA?
This might be because I'm new to Rust, but I don't see the value of [this test](https://github.com/exercism/xrust/blob/04b665e989b578c4e46ff100b1b0a00d5df9bb0d/exercises/rna-transcription/tests/rna-transcription.rs#L41) in the RNA transcription exercise. It doesn't seem germane to the problem, and [the commit that introduced it](https://github.com/exercism/xrust/commit/312c462a19c07c2bc3f4c7d28893339374bbb7fb) doesn't really explain why it's there.

Its presence seems to require the solver to implement `AsRef` for their Struct (as shown in [the example solution](https://github.com/exercism/xrust/commit/312c462a19c07c2bc3f4c7d28893339374bbb7fb)) just so that the test can be passed. Implementing AsRef doesn't add any other value, as far as I can see. All the other tests are passable without implementing AsRef.

But, again, new to Rust here so I'm probably missing something. My solution, which passes all tests but the `as_ref` one is here: https://gist.github.com/IanWhitney/b2ac83ae8195f5fc1a75

petertseng:
This was introduced in 7877d6389ed47d8c3e6631274c1c66a0b4c4ec54 back
when the exercises was ported to xrust, but there's no clear reason why
we would want the RNA to be convertible to a string.

* It's not relevant to the problem statement.
* There's no clear use case for why this is something that should be
  required.
* Other languages don't do this.
* Other exercises in xrust don't do this.

Given these reasons, it makes more sense to remove it.

Closes #76

### DIFF
--- a/exercises/rna-transcription/example.rs
+++ b/exercises/rna-transcription/example.rs
@@ -11,12 +11,6 @@ impl RibonucleicAcid {
     }
 }
 
-impl AsRef<str> for RibonucleicAcid {
-    fn as_ref(&self) -> &str {
-        self.nucleotides.as_ref()
-    }
-}
-
 #[derive(PartialEq, Eq, Debug)]
 pub struct DeoxyribonucleicAcid {
     nucleotides: String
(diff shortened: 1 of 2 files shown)


EduardoBautista:
Sounds fair.

PR #84, opened 2016-03-22
allergies: use Allergies::new for test interface

### LINKED ISSUE #64: Function interfaces for tests
I was having a lot of issues trying to guess what kind of object is required to be passed in the allergies exercise.
I figured this is what it looked like:

    pub struct Allergies {
        score: i32,
    }

    impl Allergies {
        pub fn new(n: i32) -> Allergies {
            Allergies { score: 0 }
        }
    }

So I was basically trying to figure out what the `Allergies(0)` function did and where it came from.
In the end I had to resort to looking into the solution in `example.rs`, which is kind of upsetting.

Turns out the struct has the form `pub struct Allergies(pub usize);` and the `Allergies(0)` is... I still don't understand what it is and where it comes from.

Do you think there is a way to expose the interface without giving out the solution?

petertseng:
As mentioned in #64, using Allergies(3) or similar constrains the
implementation to using tuple structs, whereas there could be a range of
possible implementations. By using `new` as the test interface,
implementations can choose what internal representation they'd like. For
instance, the example solution still uses a tuple struct, but you could
imagine solutions using regular structs.

Closes #64 (and I believe this is the only exercise that constrains the
implementation in such a way)

### DIFF
--- a/exercises/allergies/example.rs
+++ b/exercises/allergies/example.rs
@@ -4,6 +4,10 @@ pub struct Allergies(pub usize);
 pub enum Allergen { Eggs, Peanuts, Shellfish, Strawberries, Tomatoes, Chocolate, Pollen, Cats }
 
 impl Allergies {
+    pub fn new(score: usize) -> Allergies {
+        Allergies(score)
+    }
+
     pub fn is_allergic_to(&self, allergen: &Allergen) -> bool {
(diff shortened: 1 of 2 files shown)


EduardoBautista:
Looks good to me.

PR #85, opened 2016-03-22
travis: test beta and nightly

petertseng:
A suggestion from https://docs.travis-ci.com/user/languages/rust

The motivation here is simply that we get alerted ahead of time if an
upcoming release of Rust would break our tests for whatever reason. In
such a case we'd like to be alerted of it ahead of time.


EduardoBautista:
Good idea!

PR #89, opened 2016-03-26
Implement Queen Attack

IanWhitney:
There is no common set of tests for this exercise, so implementation is
a bit....free form. I think there are two interesting things to look at,
what tests to use, and how to use them.

Which Tests

I first surveyed all of the languages that implement this exercise and
looked at what they tested. In order of prevalence, the five most common
test types cover:

1. Can 2 queens attack each other?
2. [Can 2 pieces occupy the same
space](https://github.com/exercism/xelixir/blob/0965c4e107de68979a493b82f9f0859773fcb83b/exercises/queen-attack/queen_attack_test.exs#L26)?
3. [Comparing string representations of the
chessboard](https://github.com/exercism/xelixir/blob/0965c4e107de68979a493b82f9f0859773fcb83b/exercises/queen-attack/queen_attack_test.exs#L33)
4. [Does a chessboard start with the queens in their default
positions?](https://github.com/exercism/xelixir/blob/0965c4e107de68979a493b82f9f0859773fcb83b/exercises/queen-attack/queen_attack_test.exs#L12)
5. [Errors are raised if an invalid position is
used](https://github.com/exercism/xpython/blob/b9dfe291845a31301029d947b3cf3a2ffc85b0b6/exercises/queen-attack/queen_attack_test.py#L55)

Then there are some uncommon edge-case tests. Can there be more than one
chessboard? Can the board be empty? Algebraic chess notation? Etc.

Of the top five, I found that only numbers 1 and 5 really cover the
exercise as stated -- if I place two pieces on the board, can they attack
each other?

So, those are the tests that I've implemented here.

Implementing the other tests requires the introduction of a Chessboard
object. And also a way to identify pieces (which team they belong to,
how they are represented as strings, etc.)  That can be interesting to
do, but I find it beyond the scope of this exercise as written.

At best, I think these tests should be suggested as Extra Credit.

What Order

I was very particular in my test order. Some languages, for example,
start with the board representation tests. This requires students to
write a lot of code in order to get just one test to pass.

I have found that the faster I can get the first test to pass, the
better. So my first test requires nothing more than a `can_attack`
function that returns false. The tests then build on that, adding
complexity. Then, once `can_attack` is fully implemented, I add the
curveball test of invalid chess positions, which will get students
thinking about bad inputs and Rust's error handling.

### DIFF
--- a/exercises/queen-attack/example.rs
+++ b/exercises/queen-attack/example.rs
@@ -1,0 +1,48 @@
+#[derive(Debug, PartialEq)]
+pub struct Queen {
+    position: (i8, i8)
+}
+
+#[derive(Debug, PartialEq)]
+pub struct InvalidQueen;
+
+impl Queen {
+    pub fn new(position: (i8, i8)) -> Result<Queen, InvalidQueen> {
+        if (position.0 < 0 || position.0 > 7) || (position.1 < 0 || position.1 >7 ) {
+            Err(InvalidQueen)
+        } else {
+            Ok(Queen {position: position})
+        }
+    }
+}
+
+pub struct Queens {
+    first: Queen,
+    second: Queen
+}
+
+impl Queens {
+    pub fn new(first: Result<Queen, InvalidQueen>, second: Result<Queen, InvalidQueen>) -> Queens {
+        Queens {
+            first: first.unwrap(),
+            second: second.unwrap()
+        }
+    }
+
+    pub fn can_attack(&self) -> bool {
+        self.first.position.0 == self.second.position.0 ||
+            self.first.position.1 == self.second.position.1 ||
+            (
+                (self.first.position.0 + self.first.position.1)
+                ==
+                (self.second.position.0 + self.second.position.1)
+            )
+            ||
+            (
+                (self.first.position.0 - self.first.position.1)
+                ==
+                (self.second.position.0 - self.second.position.1)
+            )
+
+    }
+}
--- a/exercises/queen-attack/example.rs
+++ b/exercises/queen-attack/example.rs
@@ -1,33 +1,71 @@
 #[derive(Debug, PartialEq)]
-pub struct Queen { position: (i8, i8) }
+pub struct Queen { position: ChessPosition }
 
 impl Queen {
-    pub fn new(position: (i8, i8)) -> Queen {
-        Queen { position: position }
-    }
-
-    pub fn can_attack(&self, piece: Queen) -> Result<bool, ()> {
-        if self.position_valid() && piece.position_valid() {
-            Ok(self.rank() == piece.rank() ||
-               self.file() == piece.file() ||
-               self.sum() == piece.sum() ||
-               self.difference() == piece.difference())
+    pub fn new(position: (i8, i8)) -> Result<Queen, ()> {
+        let position = ChessPosition::new(position);
+        if position.valid() {
+            Ok(Queen { position: position })
         } else {
             Err(())
         }
     }
 
-    fn position_valid(&self) -> bool {
+    pub fn can_attack(&self, piece: Queen) -> bool {
+        self.horizontal_from(&piece) ||
+            self.vertical_from(&piece) ||
+            self.diagonal_from(&piece)
+    }
+
+    fn horizontal_from(&self, piece: &Queen) -> bool {
+        self.rank() == piece.rank()
+    }
+
+    fn vertical_from(&self, piece: &Queen) -> bool {
+        self.file() == piece.file()
+    }
+
+    fn diagonal_from(&self, piece: &Queen) -> bool {
+        self.sum() == piece.sum() ||
+            self.difference() == piece.difference()
+    }
+
+    fn rank(&self) -> i8 {
+        self.position.rank()
+    }
+
+    fn file(&self) -> i8 {
+        self.position.file()
+    }
+
+    fn sum(&self) -> i8 {
+        self.position.sum()
+    }
+
+    fn difference(&self) -> i8 {
+        self.position.difference()
+    }
+}
+
+#[derive(Debug, PartialEq)]
+struct ChessPosition { coordinates: (i8, i8) }
+
+impl ChessPosition {
+    fn new(coordinates: (i8, i8)) -> ChessPosition {
+        ChessPosition {coordinates: coordinates}
+    }
+
+    fn valid(&self) -> bool {
         (self.rank() >= 0 && self.rank() <= 7) &&
         (self.file() >= 0 && self.file() <= 7)
     }
 
     fn rank(&self) -> i8 {
-        self.position.0
+        self.coordinates.0
     }
 
     fn file(&self) -> i8 {
-        self.position.1
+        self.coordinates.1
     }
 
     fn sum(&self) -> i8 {
--- a/exercises/queen-attack/example.rs
+++ b/exercises/queen-attack/example.rs
@@ -1,48 +1,40 @@
 #[derive(Debug, PartialEq)]
-pub struct Queen {
-    position: (i8, i8)
-}
-
-#[derive(Debug, PartialEq)]
-pub struct InvalidQueen;
+pub struct Queen { position: (i8, i8) }
 
 impl Queen {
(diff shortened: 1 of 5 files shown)


IanWhitney:
If you're interested in the tests implemented by other languages, I've summarized it here: https://gist.github.com/IanWhitney/dd34f8a16c730f3cd77c

IanWhitney:
@petertseng, it sounds like some of the survey I did for this problem overlaps well with your recent work (as outlined in today's newsletter). So if I can contribute to a canonical version of the Queen Attack tests, let me know. Happy to help.

petertseng:
Many thanks for conducting the survey @IanWhitney !!

Since I will be traveling, remind me if I haven't gotten back to this PR in 72 hours from now. Of course anyone else is free to beat me to the punch on this!

In the meantime, my remarks on the cross-track consistency work:

A great first step will be to open a PR in https://github.com/exercism/x-common adding `queen-attack.json` with what you think are the tests that are in fact relevant to the exercise, as you described in your commit message/PR description. It will be good if your PR references https://github.com/exercism/todo/issues/133. In that PR will probably discuss with @kytrinyx and any interested track maintainers to get agreement on the set of tests.

After that is merged, then the next thing is to get tracks in line with it, adding or removing tests.

We have a tool https://github.com/exercism/blazon for filing issues across all tracks with a given problem, we can help draft an issue text if you like.

Of course you are free to file PRs actually making the changes for some (or even all?!) tracks if you desire, but it's totally fine to load-balance across multiple contributors.

All of the above are absolutely open for you to take charge of currently, since I should probably only focus on one problem at a time and my current efforts are focused on a different problem. So it's yours if you want it!

exercises/queen-attack/example.rs
```diff
@@ -0,0 +1,48 @@
...
+    }
+}
+
+pub struct Queens {
+    first: Queen,
+    second: Queen
+}
+
+impl Queens {
+    pub fn new(first: Result<Queen, InvalidQueen>, second: Result<Queen, InvalidQueen>) -> Queens {
```
petertseng: I'm thinking a bit about the interface here; seems a bit odd to have those `Result<Queen, InvalidQueen>` as the argument types.

What if it were `(i8, i8)` here instead of the `Result<Queen, InvalidQueen>`, and returning `Result<Queens, InvalidQueen>` (or maybe `InvalidQueens`? Or just `Option<Queens>`?)

I can understand why the `Result<Queen, InvalidQueen>` is done (so that the result of `Queen::new` can be directly passed in, and then `Queen::new` can be tested individually instead of `Queens::new` which has more decision-making to do). Just seems an unclear interface.

What if it were just `queen1.can_attack(queen2)` ?

I could be convinced either way on this, feel free to discuss it while I'm traveling.

IanWhitney: I'm totally open to changing the interface. If I were writing this from scratch in Ruby I probably would do `queen1.can_attack?(queen2)`.

I think my API was mostly being led by the other implementations of this exercise, which tend to look like 

```ruby
queens = Queens.new(white: [2, 3], black: [4, 7])
assert !queens.attack?
```

But since I'm trying to re-think which tests are appropriate, I might as well also re-think what API is appropriate. Let me take a shot at the `queen1.can_attack(queen2)` approach.

kytrinyx:
FYI @petertseng I merged the PR in x-common: https://github.com/exercism/x-common/issues/211

petertseng:
Yeah, I guess either `can_attack` or `Queen::new` has to deal with the possibility of an invalid queen.

What would you say to having that be in `Queen::new` instead? So `Queen::new` could return a `Option<Queen>` or maybe a `Result<Queen, Something>` (I think a error type of `()` isn't all that useful and one might as well use `Option` in that case, but if there is some sensible errror type we coud consider it).

I realize that yes, you'd still have to `unwrap()` the result of that `Queen::new` in the tests, so it's not like it's saving us any unwrapping. But it seems useful to me to raise errors as close to the site of the error as possible.

Another reason for that would be that then it's obvious what queen caused the problem. If I have `queenA.can_attack(queenB)` which is an Error, how do I know what queen caused the problem? I could have a known-good Queen to call against both queens and see which one errors, but that seems cumbersome. Seems easier for clients of the code if we can know at construction time that we will either get a valid Queen, or an error right then.

(By the way the commit message of 2a74be0b50f803e0e7ac4514c44e7d1f16a0eb0f implies that `Queen::new` is returning the `Result` like I suggest, but in fact it is `can_attack` that has the `Result`)

This is just what I think but I could be convinced otherwise, totally open to hearing what @IanWhitney or @EduardoBautista (or anyone else!) have to say about that.

IanWhitney:
Doing it on Queen::new does seem more in line with the [common test suite](https://github.com/IanWhitney/x-common/blob/fix_queen_attack/queen-attack.json)

But if I do it that way, then every test would have lines like:

```rust
  let white_queen = Queen::new((2,4)).unwrap();
```

Is there a way in Rust to have students avoid using a Result value until they reach the test that introduces error handling? Because that's what I think would be best in terms of learning curve. I'd love students' initial implementations to be simple,

```rust
pub fn new(position: (i8,i8)) -> Queen {
  Queen {position: position}
}
```

Because the simple implementation gets tests passing faster, which I always think is a good thing.

And then when the student hits the error handling test they have to figure out how to modify their simple implementation to handle this extra complexity.

IanWhitney:
Oh, and my example is returning an empty-tuple error because I didn't see a good reason for it to return anything more complex. Students could return any error type they like. It just has to properly handle `is_err()`.

petertseng:
> But if I do it that way, then every test would have lines like:

```rust
let white_queen = Queen::new((2,4)).unwrap();
```

Hmm, is that better or worse than there being lines of the form:

```rust
assert!(white_queen.can_attack(black_queen).unwrap());
```

One of the two has to have that `unwrap` if we are to have error handling at all.

> Is there a way in Rust to have students avoid using a Result value until they reach the test that introduces error handling? Because that's what I think would be best in terms of learning curve. I'd love students' initial implementations to be simple,

This is a good thought.

We'd have to figure out how to get this accepted by the type checker if `Queen::new` should start out returning just a `Queen` but later should change to returning a `Result<Queen, _>`. It seems to imply that the earlier tests expecting a `Queen` would have to be deleted. Or at least not compiled, using some sort of conditional compilation thing? You could explore the conditional compilation route if you think it will be better. If it works out we may like to think about how other problems in this track can make use of it (separately from this PR of course).

Alternatively, is there a way that the *test* could provide the implementation of `some_queen.unwrap()` (which would presumably just be the identity function)? That way `Queen::new((2,4)).unwrap()` will work both when `Queen::new` is expected to return a `Queen` and when it's expected to return a `Result<Queen, _>`

If it turns out the above suggestions are working too hard against the type system, maybe it's a sign that in Rust, the type system means you have to check errors earlier than you might have to in other languages. Although I realize that statement is more about when you would check errors when you're using a library that provides a fixed interface. The question we're dealing with here is "when should we add error handling when iteratively developing, and when we add error handling, how are our tests affected?"

> Oh, and my example is returning an empty-tuple error because I didn't see a good reason for it to return anything more complex. Students could return any error type they like.

Ah I see! I missed the fact that it just needs to be an error of some kind instead of being forced to be `()`.  Great!

IanWhitney:
Thinking about it more, I may be trying to implement Null Object Pattern in Rust. And you're right, that may be an idea that is incompatible with how Rust wants me to work. The pattern exists for OO languages which can return nulls, and Rust can't do that. 

Maybe the best approach is to have `Queen::new` return a `Result`. And then I'll re-order the tests a bit to make initial progress easier.

IanWhitney:
Ok. I think we're pretty good. Or at least I hope so, since I won't have nearly as much free time to hack on this stuff after today.

petertseng:
I think this is pretty good as well! Given that `Queen::new` is returning the `Result` now rather than `can_attack` I think it definitely made sense to reorder the tests in the fashion given.

If it turns out that students are not figuring out that the return value should be of a `Result` type we can point it out to them via a comment in the test file or something. May want to periodically examine submissions to queen-attack to see whether anyone has such trouble.

Given I'm happy with this and nobody else had objections over the weekend let's get this merged. Thanks again for the work.

PR #90, opened 2016-03-26
Added LEARNING.md

### LINKED ISSUE #87: Add LEARNING.md with at least the Rust book
Discussion in #58 indicates that it would be good to move the Rust book from RESOURCES.md to LEARNING.md.

As https://github.com/exercism/x-api/blob/master/CONTRIBUTING.md states, LEARNING is for starting out and learning the language from scratch, while RESOURCES is more general.

IanWhitney:
Closes #87

Links to 5 useful resources for students learning Rust from scratch.

### DIFF
--- a/docs/LEARNING.md
+++ b/docs/LEARNING.md
@@ -1,0 +1,7 @@
+## Learning Rust
+
+* [The Rust Programming Language](http://doc.rust-lang.org/stable/book/) is a great resource for getting started with Rust as well as diving deeper into specific features of Rust.
+* [Rust by Example](http://rustbyexample.com) shows you examples of the most common things you will be writing in Rust.
+* [Rustlings](https://github.com/carols10cents/rustlings) "Small exercises to get you used to reading and writing Rust code."
+* [#rust-beginners](https://client00.chat.mibbit.com/?server=irc.mozilla.org&channel=%23rust-beginners) "an IRC channel that loves answering questions at any depth"
+* The [Rust User Forum](http://users.rust-lang.org) answers questions of all levels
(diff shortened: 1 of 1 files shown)


IanWhitney:
Ah, I put that file in the wrong place. Let me fix that.

IanWhitney:
Ok, now it's in docs.

docs/LEARNING.md
```diff
@@ -0,0 +1,7 @@
+## Learning Rust
+
+* [The Rust Programming Language](http://doc.rust-lang.org/stable/book/) is a great resource for getting started with Rust as well as diving deeper into specific features of Rust.
+* [Rust by Example](http://rustbyexample.com) shows you examples of the most common things you will be writing in Rust.
+* [Rustlings](https://github.com/carols10cents/rustlings) "Small exercises to get you used to reading and writing Rust code."
+* [#rust-beginners](https://client00.chat.mibbit.com/?server=irc.mozilla.org&channel=%23rust-beginners) "an IRC channel that loves answering questions at any depth"
```
petertseng: Any strong reason to explicitly specify `client00` here instead of using `https://chat.mibbit.com/` and letting Mibbit load-balance among its `clientNN` subdomains itself? I'm not sure, but in the absence of strong reason to do otherwise I would prefer to omit `client00`

petertseng:
Thanks, looking good at first glance, just one small question I think. I'd be okay with merging this when that is addressed, someone else may need to do it since I will be traveling a few days.

docs/LEARNING.md
IanWhitney: I'll fix that. I grabbed that link from rust-lang's community page, and it hard-codes the client00 part.

kytrinyx:
Thank you, @IanWhitney!

PR #95, opened 2016-04-02
Change API to borrow piece to can_attack

IanWhitney:
Transferring ownership to can_attack is poor design. It prevents you
from using the piece after the function has been called. A more Rust-ish
approach is to borrow.

I know this exercise is now available, so I'm not sure how easy it is to change the API now. But I sure hope we can!

### DIFF
--- a/exercises/queen-attack/example.rs
+++ b/exercises/queen-attack/example.rs
@@ -11,7 +11,7 @@ impl Queen {
         }
     }
 
-    pub fn can_attack(&self, piece: Queen) -> bool {
+    pub fn can_attack(&self, piece: &Queen) -> bool {
         self.horizontal_from(&piece) ||
             self.vertical_from(&piece) ||
             self.diagonal_from(&piece)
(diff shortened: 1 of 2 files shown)


EduardoBautista:
Agreed.  Looks good.  👍

petertseng:
That's OK to change the API since everyone sees the API as it was when they downloaded the exercise.

The disadvantage is that some reviewers may get confused when looking at others' submissions. "Hey, why did you not borrow the queen when calling `can_attack`?"

Some tracks implement test versioning to deal with this problem (known tracks doing this are xgo and xruby), where each test file declares a test version and submitted solutions have to declare the test version they tested against (enforced by, of course, a test case that checks the solution's declared test version!). The expected test version gets incremented whenever there is a behavior change or an API change in tests (anything that might cause old solutions to fail, basically). This way, reviewers can easily figure out when a solution being reviewed might not pass the current tests.

If we so desire, we can implement such a test versioning scheme. That can come in a separate PR. For now let's get this merged and we can see if it's something we want later on.

EduardoBautista:
@petertseng Ah, I didn't think of that.  That versioning system seems interesting though.

PR #97, opened 2016-04-06
Implement Sieve exercise

IanWhitney:
Tests are from the sieve.json file in xcommon

https://github.com/exercism/x-common/blob/master/sieve.json

Though I have re-ordered them to follow a path I thought better. Start
with no primes, then work up to the simplest results (1 and 2), then get
more complex.

In the config.json I've chosen to put this exercise before Allergies. I
picked this spot for two reasons:

1. It has some interesting iterator usage, which I thought step up after
doing iterator work in Anagrams, and the folding in Difference of
Squares.
2. I think Allergies comes too early. Or, maybe it just came too early
for me, as I found it baffling.

But I'm flexible on placement. Looking at the other tracks it doesn't
seem to have a consistent spot.


exercises/sieve/example.rs
```diff
@@ -0,0 +1,33 @@
+pub struct Sieve { limit: usize }
+
+impl Sieve {
+    pub fn new(limit: usize) -> Sieve {
+        Sieve { limit: limit }
+    }
+
+    pub fn primes(&self) -> Vec<usize> {
+        let mut primes = (0..self.limit).map ( |x| x + 1).collect::<Vec<usize>>();
```
petertseng: Were I very concerned about performance I would wonder whether this could be expressed as (0..(self.limit+1)).skip(1) , though I am not so concerned here.

exercises/sieve/example.rs
```diff
@@ -0,0 +1,33 @@
...
+                }
+            }
+        }
+
+        primes
+    }
+
+}
+
+pub fn prime(num: usize) -> bool {
```
petertseng: do we care very much about a very strict interpretation of the sieve algorithm's operation? If we did, we'd want to check primality via whether `num` hasn't already been marked off (whether it's still in `primes`, perhaps?), rather than re-iterating through all numbers less than `num`

petertseng:
Test ordering seems good.

exercises/sieve/example.rs
IanWhitney: Let me take a closer look at the algorithm and update my example. Since implementing the algorithm is the main reason for this exercise, I want to make sure the example does it right.

IanWhitney: Yeah, I totally mis-read the Sieve algorithm. Re-implementing.

IanWhitney:
Ok. I've now implemented the real algorithm and learned something about the Sieve of Eratosthenes. First thing I learned -- if you have a function that checks if a number is prime, then you're doing it wrong.

I'm still on the fence about using a Sieve struct with a new function. Coming from Ruby that's my style, but it may not be idiomatic here.

exercises/sieve/example.rs
```diff
@@ -0,0 +1,26 @@
...
+        let mut p = 2;
+        let mut done = false;
+
+        while !done {
+            integers.retain( |&x| (x == p) || (x % p != 0));
+
+            let next_possible_prime = integers.iter().find( |&x| *x > p);
+
+            if next_possible_prime.is_none() {
+                done = true
```
petertseng: I don't feel particularly strongly about this, but what would you feel about just doing `break` here instead of setting this boolean?

exercises/sieve/example.rs
```diff
@@ -0,0 +1,26 @@
...
+        let mut integers = (1..self.limit).map ( |x| x + 1).collect::<Vec<usize>>();
+        let mut p = 2;
+        let mut done = false;
+
+        while !done {
+            integers.retain( |&x| (x == p) || (x % p != 0));
+
+            let next_possible_prime = integers.iter().find( |&x| *x > p);
+
+            if next_possible_prime.is_none() {
```
petertseng: also don't feel particularly strongly, but what if we used the `let Some(next_prime) = ...`  form here? Could possibly even fold line 16 into this. Would avoid the `unwrap`, so I think looks slightly better

petertseng:
Gonna think a bit about problem placement. Go and Ruby also have it relatively early. My initial instinct was to say "I think sieve might come too early", but I don't have a concrete reason for thinking that. I would say though that if allergies is too early, we could consider putting it later. Anyway, I'm not opposed to trying out sieve in this position and seeing what comes of it.

It would be great if we could get student feedback on problem ordering, but that is a separate discussion, one such as https://github.com/exercism/exercism.io/issues/1081

petertseng:
Oh, and regarding:

> I'm still on the fence about using a Sieve struct with a new function. Coming from Ruby that's my style, but it may not be idiomatic here. 

Ah, so you think it might just be better as a bare function `primes_up_to` or even an associated function `Sieve::primes_up_to` ? Possible. I have no strong opinion here so all seem acceptable.

IanWhitney:
#### Problem Placement

Maybe we should make an issue to review our problem ordering? I'd like to talk to @kytrinyx and other maintainers about this as well. My preference is to order the exercises in a way that teaches the language, but that's due to how I use exercism. Others do exercises in languages they already know, so their ordering needs are probably different.

#### API

I took a look at [Primal](http://huonw.github.io/primal/primal/), which seems to be the preferred prime crate for Rust. Based on its api, I think the `Sieve::primes_up_to` is more appropriate. I'll update the code.

#### Example

I'm never sure how good the example code needs to be. I do want it to be correct, and I also want it to be idiomatic (since I tend to use the example solutions as a reference for learning the language). I'm not sure what the correct idioms are here. The use of `let` seems right to me, though, so I'll get that change in.

Once everything is in and looks good, I'll squash all this down to a single commit.

kytrinyx:
> My preference is to order the exercises in a way that teaches the language, but that's due to how I use exercism.

I think this is the ideal ordering. For people who already know the language, it doesn't hurt, and for people who don't it's incredibly helpful. For people who already know the language, they'll get whatever they get out of it, no matter what the ordering is.

> I'm never sure how good the example code needs to be.

So far I've not cared too much about good code in the example; I think about it more as a thing that will test the test suite (just make sure that the test suite makes sense, and that it is solvable).

exercises/sieve/tests/sieve.rs
```diff
@@ -0,0 +1,48 @@
...
+#[test]
+#[ignore]
+fn limit_is_the_first_prime() {
+    assert_eq!(Sieve::primes_up_to(2), [2]);
+}
+
+#[test]
+#[ignore]
+fn primes_up_to_10() {
+    assert_eq!(Sieve::primes_up_to(10), [2,3,5,7]);
```
petertseng: since I noticed you did a `rustfmt` on your code, I suppose I might as well ask - did it say anything about commas in this list? There are no spaces after commas here, but spaces after commas elsewhere. Just curious.

petertseng:
I say :+1: to this (one nit that you can fix, or not). I'll leave you to do the merge whenever you want, since you indicated you'd like to squash first.

> My preference is to order the exercises in a way that teaches the language, but that's due to how I use exercism. Others do exercises in languages they already know, so their ordering needs are probably different.

I think the people who have the most different ordering needs may be those who don't know programming at all. One might reasonably say that this isn't too bad because the algorithm is laid out in the instuctions, but I think there's still a chance it will be a trying experience.

We can make a separate issue to review it. I think I had other things to say at some point but have forgotten them, so maybe some time will remind me of them. I don't currently feel like holding up this particular PR for it.

> I'm never sure how good the example code needs to be.

My personal thought is that even though it doesn't necessarily need to use the best algorithm, it should at least be idiomatic for the target language, for whatever algorithm it does choose to use.

I do like how the `while let` cleaned up the code a lot.

exercises/sieve/tests/sieve.rs
IanWhitney: Ah, I didn't rustfmt the tests. I'll do that too before merge.

PR #101, opened 2016-04-08
dominoes: Rename invalid_input -> no_chains

### LINKED ISSUE #100: dominoes: Don't call the "invalid input" test invalid input?
Concerning https://github.com/exercism/xrust/blob/master/exercises/dominoes/tests/dominoes.rs#L113

It's valid input for the function (it's a list of dominoes), it's just that it can't be chained. I would think we wouldn't want to call it `invalid_input`.

I'm just thinking about this as I figure out what to call it in the common JSON file

petertseng:
The input is valid (it's a list of dominoes), it's just that it can't be
chained.

Closes #100

### DIFF
--- a/exercises/dominoes/tests/dominoes.rs
+++ b/exercises/dominoes/tests/dominoes.rs
@@ -110,7 +110,7 @@ fn can_reverse_dominoes() {
 
 #[test]
 #[ignore]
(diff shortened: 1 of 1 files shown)


IanWhitney:
Ok. I think you can merge this whenever you want.

PR #102, opened 2016-04-13
Make module qualifier lowercase.

peap:
Hope I'm not getting ahead of myself, but the previous test files all had correct spellings of their module qualifiers, so this looked like an innocent mistake :)

### DIFF
--- a/exercises/sieve/tests/sieve.rs
+++ b/exercises/sieve/tests/sieve.rs
@@ -4,26 +4,26 @@ use sieve::*;
 
 #[test]
 fn limit_lower_than_the_first_prime() {
-    assert_eq!(Sieve::primes_up_to(1), []);
+    assert_eq!(sieve::primes_up_to(1), []);
 }
 
 #[test]
 #[ignore]
 fn limit_is_the_first_prime() {
-    assert_eq!(Sieve::primes_up_to(2), [2]);
+    assert_eq!(sieve::primes_up_to(2), [2]);
 }
 
 #[test]
 #[ignore]
 fn primes_up_to_10() {
-    assert_eq!(Sieve::primes_up_to(10), [2, 3, 5, 7]);
+    assert_eq!(sieve::primes_up_to(10), [2, 3, 5, 7]);
 }
 
 
 #[test]
 #[ignore]
 fn limit_is_prime() {
-    assert_eq!(Sieve::primes_up_to(13), [2, 3, 5, 7, 11, 13]);
+    assert_eq!(sieve::primes_up_to(13), [2, 3, 5, 7, 11, 13]);
 }
 
 #[test]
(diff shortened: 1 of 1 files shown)


petertseng:
Ah, thanks for watching out! I think in this case, because of the `use sieve::*` above, the `sieve::` is implied. The `Sieve` you see there is a type, of which `primes_up_to` is an associated function.

... However we may even consider just putting `primes_up_to` as a bare function, since there's not that much reason to have a type with only one associated function and no methods or anything else.

You can see the evolution of the (very recently-added) problem in #97 and it appears there may be more work to do here. What would you say @IanWhitney ? And @peap ?

petertseng:
(Edit: My personal gut reaction is to remove the `impl Sieve { }` and just have `primes_up_to` be a top-level function, but let's discuss for a bit!)

peap:
Ah, I see. I thought I might be missing something! Thanks for pointing that out, and I see it in the example file now.

I've been working through the exercises from the beginning, so I'm not sure which pattern is more common, but I agree with you that a single top-level function makes sense here.

IanWhitney:
This API definitely reflects my OO background. I find the idea of bare functions very odd. But maybe that's just an OO thing and bare functions are totally normal in a systems language.

When we were discussing this before I took a inspiration from [Primal](http://huonw.github.io/primal/primal/), which seems to be the main crate for dealing with primes (slow_primes, which I've seen used a lot now points to Primal). 

Primal's sieve API looks like:

```rust
// find the primes up to this upper bound
let sieve = primal::Sieve::new(hi as usize);
```

Our API did originally use the `new` function, but we dropped it. Since we're no longer creating Struct instances, maybe we no longer need the struct at all. 

If we do drop the Struct, then our tests look like this, right?

```rust
assert_eq!([2,3,5,7], sieve::primes_up_to(10));
```

petertseng:
> This API definitely reflects my OO background. I find the idea of bare functions very odd. But maybe that's just an OO thing and bare functions are totally normal in a systems language.

Ooh I think this is a good discussion point, to see why some languages have it some way and some languages another. We can talk about why it seems odd to have bare functions. One answer I might have is "they clutter up the global namespace". C tends to deal with the problem by manually prefixing the function name, so we might get things like `list_prepend` and `list_get`. Here every module gets its own namespace, so even if we have a `queue::prepend` and a `list::prepend` there wouldn't be a risk of collision.

We can see that Primal does indeed have some functions that would get called like `primal::is_prime(0)`. On the other hand, I imagine they also have a `primal::Sieve::new` because they have some methods on a `Sieve` that presumably use some cached result (it even also supports `sieve.is_prime(0)`). For our uses it's not so clear that we need caching or something of the sort.

> If we do drop the Struct, then our tests look like this, right?
>
> ```rust
> assert_eq!([2,3,5,7], sieve::primes_up_to(10));
> ```

yup.

IanWhitney:
I think I'm fine with either approach. I don't think the Struct provides anything useful here, so we can drop it and use a bare module method.

PR #103, opened 2016-04-13
Update Roman API to use From and ToString traits

### LINKED ISSUE #99: Revisiting Roman Numerals API
So, @IanWhitney [agreed](http://exercism.io/submissions/3d5665d6626747989e3cf9328e73f761) that API used for Roman Numerals is unfortunate. [Test suite](http://exercism.io/exercises/rust/roman-numerals) has this:

    assert_eq!("I", Roman::from(1));

Which looks like [`trait From`](https://doc.rust-lang.org/std/convert/trait.From.html).This trait is supposed to convert some type to the type `From` implemented for (`i32` to `Roman` in this case). Instead, the result of `from` is used like `&str` or `String`. It is still possible to use `trait From` if we combine it with, say, [`PartialEq`](https://doc.rust-lang.org/std/cmp/trait.PartialEq.html). But then implementation looks verbose and unidiomatic.

The question is, how should a new API look like? Any suggestions?

Here is my proposal:

    let roman = Roman::new(4);
    assert_eq!("IV", roman.to_string());

Which hints to implement `Roman::new()` like many Rust structs do, and [`trait Display`](https://doc.rust-lang.org/std/fmt/trait.Display.html). Thoughts?

IanWhitney:
Fixes #99 

As per our discussion, @JIghtuse and @petertseng

### DIFF
--- a/exercises/roman-numerals/example.rs
+++ b/exercises/roman-numerals/example.rs
@@ -18,12 +18,14 @@ pub struct Roman {
     num: usize
 }
 
-impl Roman {
-    pub fn from(source: usize) -> String {
-        Roman::new(source).convert()
+impl From<usize> for Roman {
+    fn from(i: usize) -> Self {
+        Roman::new(i)
     }
+}
 
-    fn convert(&self) -> String {
+impl ToString for Roman {
+    fn to_string(&self) -> String {
         let mut start = self.num.clone();
         let mut result = String::new();
         for &(numeric, roman_string) in ROMAN_MAP.into_iter().rev() {
@@ -34,7 +36,9 @@ impl Roman {
         }
         result
     }
+}
 
+impl Roman {
     fn new(num: usize) -> Roman {
         Roman {
             num: num,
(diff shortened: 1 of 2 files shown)


IanWhitney:
Since the conversation in #99 was in favor of this change, I'll let this PR hang out until tomorrow for comment from @petertseng, @JIghtuse or anyone else. If no problems crop up, I'll merge.

JIghtuse:
@IanWhitney Thanks! Looks definitely better.

Implementation in the example is a bit off (it is [recommended](https://doc.rust-lang.org/std/string/trait.ToString.html) to implement trait `Display` instead of `ToString`), but it doesn't affect the API.

petertseng:
It may be good to use `Display` if we want examples to follow idiomatic and recommended behaviors (in general I think I do, though I'm not sure Exercism as a whole has decided on hard rules for that). If it's too  cumbersome I suppose at least a comment that says "it's recommended to use Display but it was not done here because (something)". I could be convinced of either of those two options

IanWhitney:
Display makes sense. I'll update the example.

IanWhitney:
Display implemented. Let me know what you think.

petertseng:
Ah, it wasn't even that big of a change, I see! Seems good to me. Were it up to me I would have put the `rustfmt` in a separate commit, but I'm not a huge stickler on it.

PR #104, opened 2016-04-15
Remove Sieve struct from API

IanWhitney:
Fixes #102


petertseng:
OK, I say yes to this.

PR #109, opened 2016-04-16
DO NOT MERGE: warnings

petertseng:
Staging ground to test some refinements to #107 as well as intentionally adding some warnings to test whether the changes are working as expected. Do not merge.

### DIFF
--- a/_test/check-exercises.sh
+++ b/_test/check-exercises.sh
@@ -1,8 +1,15 @@
 #!/bin/bash
 
-set -e
+# In DENYWARNINGS mode, do not set -e so that we run all tests.
+# This allows us to see all warnings.
+if [ -z "$DENYWARNINGS" ]; then
+    set -e
+fi
+
 tmp=${TMPDIR:-/tmp/}
 mkdir "${tmp}exercises"
+
+exitcode=0
 
 # An exercise worth testing is defined here as any top level directory with
 # a 'tests' directory
@@ -39,7 +46,8 @@ for exercise in exercises/*/tests; do
 
     if [ $status -ne 0 ]
     then
-        echo "Failed";
-        return 1;
+        exitcode=1
     fi
 done
+
+exit $exitcode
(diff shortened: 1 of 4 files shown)


_test/check-exercises.sh
```diff
@@ -27,14 +34,28 @@ for exercise in exercises/*/tests; do
...
+            # By the way, at the time of writing --quiet actually isn't quiet!
+            # It runs tests and prints output. Not what we want.
+            # Wonder if that behavior is intentional???
+            set +e
+            cargo test --no-run
+            status=$?
+            if [ $status -ne 0 ]; then
+                # Instead of immediate exit, we keep going.
+                # This allows us to see all warnings.
+                exitcode=1
```
petertseng: hmm, isn't actually causing exit. wonder why.

petertseng: subshells can't affect their parent shell's vars, of course. have to `return` it.

_test/check-exercises.sh
```diff
@@ -27,14 +34,33 @@ for exercise in exercises/*/tests; do
...
+            # No-run mode so we see no test output.
+            # Quiet mode so we see no compile output
+            # (such as "Compiling"/"Downloading").
+            # Compiler errors will still be shown though.
+            # Both flags are necessary to keep things quiet.
+            set +e
+            cargo test --quiet --no-run
+            status=$?
+            set -e
+            return $status
```
petertseng: NOPE, bash doesn't let you `return` out of a subshell (but `sh` does). Fine, I have other ideas.

exercises/word-count/example.rs
```diff
@@ -1,3 +1,5 @@
+use std::convert::AsRef;
```
petertseng: as luck would have it, this is not a warning. word_count has a legitimate use of `as_ref`. Doesn't matter, I've seen what I need to see.

Issue #111, opened 2016-04-18
Circular Buffer Question


Jonah-Horowitz:
I was wondering if there is a reason that the "circular buffer" exercise seems to be set up to discourage use of generics.

petertseng:
Good question. Looking at #52 where it was added I don't think there is any specific reason. If I had to guess a reason, it would be that the tests were copied from Go (where generics are unsupported) and/or Ruby (where generics are not really necessary)

I definitely have seen someone submit a solution using generics.

It could be possible to make the tests less discouraging in this regard. What are your ideas on how to do this?

I'd like to investigate a few other problems in this track to see if they would be appropriate for generics too. Seems like a good time to do that now that we are having this conversation.

Jonah-Horowitz:
I'd say that it would be more encouraging for generics if we simply changed line 11 in the first test to read `let mut buffer = CircularBuffer::<char>::new(1);` or `let mut buffer: CircularBuffer<char> = CircularBuffer::new(1);` instead of `let mut buffer = CircularBuffer::new(1);` as the latter made it far too confusing to create a `new` method that returned type `CircularBuffer<T>` and I had to hard-code a return type of `CircularBuffer<char>` there.
All the other tests work fine without any changes, but we could encourage generics even more by including more tests which store non-`char` values such as values of type `isize` and/or `String`.
I don't recall seeing any other problems in the rust track which seemed to be asking for generics like this one, but I might have forgotten about them.

petertseng:
For @EduardoBautista and @IanWhitney let's discuss whether generics is something we want. Any reason we would not want it, such as too much for students to think about? I do not think it is too bad, besides circular-buffer is currently the last exercise in the track. The Rust book even has [a nice section about generics](https://doc.rust-lang.org/book/generics.html) that we could link to in the comments of the test if we think there will be trouble. It's a good thing to learn at some point.

@Jonah-Horowitz would you like to do the honors of making the change and sending in the PR? I'm sure I could but for various reason I will not be able to do it myself for a few days so it would be a good chance for you.

IanWhitney:
Generics are part of the custom set exercise as well.

As with other discussions about our problem order, there is probably a good way to order problems to introduce generics.

PR #113, opened 2016-04-19
Update circular-buffer.rs

Jonah-Horowitz:
Empty buffer test now compatible with generics, added tests for Strings, ints.

### DIFF
--- a/exercises/circular-buffer/example.rs
+++ b/exercises/circular-buffer/example.rs
@@ -16,7 +16,7 @@ impl<T: Default> CircularBuffer<T> {
     // when the buffer is full. 
     pub fn new(size: usize) -> CircularBuffer<T> {
         CircularBuffer { 
-            buffer: vec![T::Default; size + 1], 
+            buffer: vec![T::default(); size + 1], 
             size: size + 1, 
             start: 0, 
             end: 0 
--- a/exercises/circular-buffer/example.rs
+++ b/exercises/circular-buffer/example.rs
@@ -4,14 +4,14 @@ pub enum Error {
     FullBuffer,
 }
 
-pub struct CircularBuffer<T: Default> {
+pub struct CircularBuffer<T: Default + Clone> {
     buffer: Vec<T>,
     size: usize,
     start: usize,
     end: usize,
 }
 
-impl<T: Default> CircularBuffer<T> {
+impl<T: Default + Clone> CircularBuffer<T> {
     // this circular buffer keeps an unallocated slot between the start and the end
     // when the buffer is full. 
     pub fn new(size: usize) -> CircularBuffer<T> {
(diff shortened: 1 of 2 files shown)


IanWhitney:
I think the example.rs file needs to be updated to handle these new tests. Our CI system runs the tests using the example code, just to make sure everything works.

IanWhitney:
Oh, just for easy reference, this is for #111

Jonah-Horowitz:
Sorry, it's my first pull request. Should I update example.rs then submit another pull request?

IanWhitney:
No problems! I break the CI all the time. Go ahead and update example.rs and then push it up to the same branch. The PR will automatically update and try the build again.

exercises/circular-buffer/example.rs
```diff
@@ -4,36 +4,36 @@ pub enum Error {
...
 
-impl CircularBuffer {
+impl<T: Default + Clone> CircularBuffer<T> {
     // this circular buffer keeps an unallocated slot between the start and the end
     // when the buffer is full. 
-    pub fn new(size: usize) -> CircularBuffer {
+    pub fn new(size: usize) -> CircularBuffer<T> {
         CircularBuffer { 
-            buffer: vec!['0'; size + 1], 
+            buffer: vec![T::default(); size + 1],
```
petertseng: Ah good idea about using `default`... when I was thinking about how I would tackle this I thought I might have to use `Option` which I suppose wouldn't be a terrible thing, but now you have taught me about the `Default` trait. Very good, thanks.

petertseng:
I think the final result is good!

I would recommend a squash of these commits before merge - individual commits would make sense if the build was passing on each intermediate commit, whereas CI is passing on the combined result of these commits. We do have [a Squashing section in the x-common guide](https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md#squashing) if you would like to do that yourself, or we can do that on your behalf if you prefer (Github has a "Squash" option behind the "Merge pull request" button)

Jonah-Horowitz:
I was actually doing the editing through the browser interface, and I'm not sure how to go about squashing commits, so it's probably better if you can do it on my behalf.

exercises/circular-buffer/example.rs
Jonah-Horowitz: The downside is that now it can't be used for anything that doesn't have the `Default` trait, which means it can't be used for borrowed content.

petertseng: this is true, but this only affects the example solution - students are still free to use solutions that do not have the `Default` constraint.

Can even avoid the `Clone` constraint by not using the `vec![e; n]` macro.

Yeah, since these disadvantages (which are definitely true) are only specific to the example I consider them fine.

petertseng:
Not a problem, there we have it!

I believe you mentioned that this was your first pull request. A hearty welcome from us! :tada:

PR #114, opened 2016-04-20
Update Custom Set tests to follow standard, update API

IanWhitney:
I'm addressing two things in this commit.

#### Bringing Tests in line with the Standard

Our suite of tests for Custom Set did not follow the [standard json file in x-common](https://github.com/exercism/x-common/blob/d6cd42db16fce5090bb486821a36d0fb1bfb7df2/custom-set.json) They tests have different function names, don't cover symmetric_difference and introduced some functions (`is_superset`) that aren't in the standard test file.

I have not implemented the entire standard test suite as it covers sets of [mixed types](https://github.com/exercism/x-common/blob/d6cd42db16fce5090bb486821a36d0fb1bfb7df2/custom-set.json#L475) which I don't think is even possible in Rust. Vectors have to be homogeneous. Tuples can mix types, but can't change size. Mixed type sets may be possible, but implementing that seems way beyond the scope
of this exercise.

I think [this suite of tests is too long](https://github.com/exercism/discussions/issues/10). But until I get a PR in to shorten the standard test suite, I figured I might as well follow it.

#### Updating the API

I had a couple of problems with the current approach. First, the tests hide the actual API in a helper method. This was probably done to reduce duplication, but I find that it obscures the function under test. So I removed those helpers.

The tests also pushed towards a bunch of implementation that I found unnecessary, specifically the whole [Iter implementation](https://github.com/exercism/xrust/blob/04b665e989b578c4e46ff100b1b0a00d5df9bb0d/exercises/custom-set/example.rs#L118). There might be ways to get the current test suite passing without implementing Iter, but I got pretty stuck. My new implementation does not require this implementation, since equality comparisons are now handled by implementing PartialEq.

### DIFF
--- a/exercises/custom-set/example.rs
+++ b/exercises/custom-set/example.rs
@@ -1,149 +1,84 @@
-use std::iter::{IntoIterator, FromIterator};
-use std::slice;
-use std::option::Option;
-use std::vec::Vec;
-
 #[derive(Debug)]
 pub struct CustomSet<T> {
-    elements: Vec<T>
+    collection: Vec<T>,
 }
 
-/// Create an empty set.
-impl <T> CustomSet<T> where T: Eq + Clone {
-    pub fn new() -> CustomSet<T> {
-        CustomSet { elements: Vec::new() }
+impl<T: Ord + Clone> PartialEq for CustomSet<T> {
+    fn eq(&self, other: &Self) -> bool {
+        self.collection.iter().all(|x| other.contains(&x)) &&
+        other.collection.iter().all(|x| self.contains(&x))
+    }
+}
+
+impl<T: Ord + Clone> CustomSet<T> {
+    pub fn new(inputs: Vec<T>) -> CustomSet<T> {
+        let mut s = CustomSet { collection: Vec::new() };
+        for input in inputs {
+            s.add(input);
+        }
+        s
     }
 
-    pub fn len(&self) -> usize {
-        self.elements.len()
+    pub fn add(&mut self, element: T) {
+        if !self.contains(&element) {
+            self.collection.push(element)
+        }
+    }
+
+    pub fn delete(&mut self, element: &T) {
+        self.collection.retain(|x| x != element)
+    }
+
+    pub fn contains(&self, other: &T) -> bool {
+        self.collection.contains(other)
     }
 
     pub fn is_empty(&self) -> bool {
-        self.len() == 0
+        self.collection.is_empty()
     }
 
-    pub fn contains(&self, needle: &T) -> bool {
-        for element in self.elements.iter() {
-            if needle == element {
-                return true;
-            }
-        }
-        false
+    pub fn size(&self) -> usize {
+        self.collection.len()
     }
 
-    pub fn insert(&mut self, element: T) -> bool {
-        if !self.contains(&element) {
-            self.elements.push(element);
-            return true;
-        }
-        false
+    pub fn is_subset(&self, other: &Self) -> bool {
+        self.collection.iter().all(|x| other.contains(x))
     }
 
-    pub fn remove(&mut self, needle: &T) -> bool {
-        let mut index: Option<usize> = None;
-        for (i, element) in self.iter().enumerate() {
-            if needle == element {
-                index = Some(i);
-                break;
-            }
-        }
-        match index {
-            Some(index) => {
-                self.elements.remove(index);
-                true
-            }
-            None => false
-        }
-    }
-    
-    pub fn difference(&self, other: &CustomSet<T>) -> CustomSet<T> {
-        let mut difference: CustomSet<T> = CustomSet::new();
-        for element in self.elements.iter() {
-            if !other.contains(element) {
-                difference.insert(element.clone());
-            }
-        }
-        difference
+    pub fn is_disjoint(&self, other: &Self) -> bool {
+        !self.collection.iter().any(|x| other.contains(x))
     }
 
-    pub fn intersection(&self, other: &CustomSet<T>) -> CustomSet<T> {
-        let mut intersection: CustomSet<T> = CustomSet::new();
-        for element in self.elements.iter() {
-            if other.contains(element) {
-                intersection.insert(element.clone());
-            }
-        }
-        intersection
+    pub fn intersection(&self, other: &Self) -> CustomSet<T> {
+        CustomSet::new(self.collection
+                           .iter()
+                           .cloned()
+                           .filter(|c| other.contains(c))
+                           .collect())
     }
 
-    pub fn union(&self, other: &CustomSet<T>) -> CustomSet<T> {
-        let mut union = CustomSet::new();
-        for element in self.elements.iter().chain(other.elements.iter()) {
-            union.insert(element.clone());
-        }
-        union
+    pub fn union(&self, other: &Self) -> CustomSet<T> {
+        CustomSet::new(self.collection
+                           .iter()
+                           .cloned()
+                           .chain(other.collection.iter().cloned())
+                           .collect())
     }
 
-    pub fn is_disjoint(&self, other: &CustomSet<T>) -> bool {
-        for element in self.elements.iter() {
-            if other.contains(element) {
-                return false;
-            }
-        }
-        true
+    pub fn difference(&self, other: &Self) -> CustomSet<T> {
+        CustomSet::new(self.collection
+                           .iter()
+                           .cloned()
+                           .filter(|c| !other.contains(c))
+                           .collect())
     }
 
-    pub fn is_subset(&self, other: &CustomSet<T>) -> bool {
-        for element in self.elements.iter() {
-            if !other.contains(element) {
-                return false;
-            }
-        }
-        true
-    }
-    
-    pub fn is_superset(&self, other: &CustomSet<T>) -> bool {
-        other.is_subset(self)
-    }
-
-    pub fn clear(&mut self) {
-        self.elements.clear();
-    }
-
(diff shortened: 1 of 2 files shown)


IanWhitney:
Before merge, I'd like:

- To hopefully get a reduced `custom-set.json` file into x-common, so that I can delete about 40% of those tests.
- Discuss the API, which I think is fine, but I'm probably wrong.
- Improve the implementation in the example file. I'm sure that can be better.

IanWhitney:
PR submitted to reduce the size of the test suite. 

https://github.com/exercism/x-common/pull/232

IanWhitney:
@petertseng 

since the PR to slim down the test suite is in limbo, I'm wondering if you have any thoughts on my new API or implementation?

exercises/custom-set/example.rs
```diff
@@ -1,149 +1,100 @@
...
+impl<T: Ord + Clone> PartialEq for CustomSet<T> {
+    fn eq(&self, other: &Self) -> bool {
+        self.collection.iter().all(|x| other.contains(&x)) &&
+        other.collection.iter().all(|x| self.contains(&x))
     }
+}
 
-    pub fn len(&self) -> usize {
-        self.elements.len()
+impl<T: PartialEq + Ord + Clone> CustomSet<T> {
```
petertseng: out of curiosity: since https://doc.rust-lang.org/std/cmp/trait.Ord.html tells me `Ord` depends on `PartialEq`, does having `PartialEq` here make a difference?

exercises/custom-set/example.rs
```diff
@@ -1,149 +1,100 @@
...
+    pub fn new(inputs: Vec<T>) -> CustomSet<T> {
```
petertseng: I don't understand fully, but can `inputs` be borrowed rather than owned?

I'm unsure because all its elements get added to `s` so there may be lifetime problems if it's just borrowed, but I've been surprised before

exercises/custom-set/example.rs
```diff
@@ -1,149 +1,100 @@
...
     }
 
-    pub fn contains(&self, needle: &T) -> bool {
-        for element in self.elements.iter() {
-            if needle == element {
-                return true;
-            }
-        }
-        false
+    pub fn delete(&mut self, element: T) {
```
petertseng: I have a feeling the `element` can be borrowed, right?

exercises/custom-set/example.rs
```diff
@@ -1,149 +1,100 @@
...
+    pub fn intersection(&self, other: &Self) -> CustomSet<T> {
+        let mut s = Vec::new();
+        for element in self.collection.iter() {
             if other.contains(element) {
-                return false;
+                s.push(element.clone());
             }
         }
-        true
+        CustomSet::new(s)
```
petertseng: could possibly be written in terms of `filter`, but I don't see this as a hard requirement

exercises/custom-set/example.rs
```diff
@@ -1,149 +1,100 @@
...
-    pub fn clear(&mut self) {
-        self.elements.clear();
-    }
-
-    pub fn iter(&self) -> Iter<T>{
-        Iter { iter: self.elements.iter() }
+        for element in other.collection.iter() {
+            s.push(element.clone());
+        }
+        CustomSet::new(s)
```
petertseng: similarly, it may be possible to use https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.chain here, though not strictly necessary

exercises/custom-set/example.rs
```diff
@@ -1,149 +1,100 @@
...
+        let mut s = Vec::new();
 
-    fn into_iter(self) -> Iter<'a, T> {
-        self.iter()
+        for element in self.collection.iter() {
+            if !other.contains(element) {
+                s.push(element.clone());
+            }
+        }
+        CustomSet::new(s)
```
petertseng: possible `filter` here

petertseng:
May be possible to use more iterator methods rather than creating a vector and conditionally pushing things onto it, but I admit this may be my style bias talking. There is a desire that the example implementation be idiomatic; my current shortcoming is I'm not heavily involved with Rust enough to know whether it is idiomatic to always use iterator methods where possible instead of conditionally pushing onto a mutable vec.

IanWhitney:
I definitely try to favor iterators. Let me poke at the code and see what I can do.

exercises/custom-set/example.rs
IanWhitney: PartialEq is not necessary. Thanks for noticing that.

exercises/custom-set/example.rs
IanWhitney: Yep, lifetime problems everywhere. I think it might be interesting to have some of our exercises focus on exposing students to lifetimes. Is this a good candidate for that?

exercises/custom-set/example.rs
IanWhitney: You can borrow `element` for `delete`. But I think it's odd to only borrow it for `delete` and not `add`. Doing it for `add` is probably tricker (see: lifetimes)

IanWhitney:
If this looks good I'd like to merge it. Depending on the outcome of https://github.com/exercism/x-common/pull/232, there might be another PR that tweaks this exercise further.

exercises/custom-set/example.rs
```diff
@@ -1,149 +1,95 @@
...
-impl<'a, T> Iterator for Iter<'a, T> {
-    type Item = &'a T;
 
-    fn next(&mut self) -> Option<&'a T> {
-        self.iter.next()
+    pub fn difference(&self, other: &Self) -> CustomSet<T> {
+        CustomSet::new(self.collection
+                           .iter()
+                           .filter(|c| !other.contains(c))
+                           .map(|c| c.clone())
```
petertseng: Is this a time where https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.cloned would be useful?

exercises/custom-set/example.rs
```diff
@@ -1,149 +1,95 @@
...
-
-impl<'a, T> IntoIterator for &'a CustomSet<T> where T: Eq + Clone {
-    type Item = &'a T;
-    type IntoIter = Iter<'a, T>;
-
-    fn into_iter(self) -> Iter<'a, T> {
-        self.iter()
+    pub fn union(&self, other: &Self) -> CustomSet<T> {
+        let mut s = CustomSet::new(vec![]);
+        for element in self.collection.iter() {
```
petertseng: I forget whether I've mentioned https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.chain already - could the `collection` of the two sets be `chain`-ed together?

exercises/custom-set/example.rs
```diff
@@ -1,149 +1,95 @@
...
+        !self.collection.iter().any(|x| other.contains(x))
     }
 
-    pub fn iter(&self) -> Iter<T>{
-        Iter { iter: self.elements.iter() }
+    pub fn intersection(&self, other: &Self) -> CustomSet<T> {
+        CustomSet::new(self.collection
+                           .iter()
+                           .filter(|c| other.contains(c))
+                           .map(|c| c.clone())
```
petertseng: Seems like a place for https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.cloned

petertseng:
I have some small comments, otherwise I feel OK.

Check whether all the `collect::<Vec<_>>()` could be changed to just `collect()` as well - I feel like type inference should work there, but I could be wrong.

Just so it's clear what I reviewed, I didn't examine the entire test suite for matching with the exercism common tests so I'm pretty much assuming that if everything's passing it's good. This will mean that if there's a test that's missing I won't know.

IanWhitney:
The tests follow the current standard json file, though not in the exact same order. I did leave out the parts of the current standard that implement mixed-type sets because I didn't think that was appropriate for a Rust implementation. But all the other tests are there. All 64 of them.

exercises/custom-set/example.rs
petertseng: Well I note that HashSet sees no problem with that: [insert](https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.insert) takes an owned T, and [remove](https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.remove) borrows

exercises/custom-set/example.rs
petertseng: Hmm, I know `anagram` did. In general I don't have a good sense of how much we want to show people at once.

However, in this case it doesn't seem like we would be able to do it anyway, so I feel like I don't need to spend too much time thinking about this.

petertseng:
All right, my last question is just about [whether `delete` should borrow the element it's deleting](https://github.com/exercism/xrust/pull/114#discussion_r61370284) - I had forgotten about that one and didn't expand comments on outdated diffs

IanWhitney:
Looks like there are two conversations that are folded away in outdated diffs

1. [Should `delete` borrow](https://github.com/exercism/xrust/pull/114#discussion-diff-61013282)
2. [Should `new` borrow its inputs](https://github.com/exercism/xrust/pull/114#discussion-diff-61013282)

I think we're ok with saying 'No' to the 2nd question. However, I would like to pursue the idea of structuring 2-3 exercises so that they introduce lifetimes. That is an unique concept to Rust (afaik) and it is one people struggle with.

As for question 1, now that I see the similarity to HashSet, that change makes sense to me.

petertseng:
👍

IanWhitney:
Ok. I'll rebase down to a single commit & merge.

PR #115, opened 2016-04-24
Move Gigasecond to the second exercise

IanWhitney:
Based on conversation in https://github.com/exercism/xrust/issues/112,
and by looking at the solutions that students have provided, the
Gigasecond exercise is very straightforward. Other than introducing
Crates it doesn't teach much about Rust.

Also, having it come between Hamming and Nucleotide is odd, since those exercises share domain knowledge.

### DIFF
--- a/config.json
+++ b/config.json
@@ -5,6 +5,7 @@
   "active": true,
   "problems": [
     "hello-world",
+    "gigasecond",
     "leap",
     "anagram",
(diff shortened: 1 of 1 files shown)


petertseng:
Well I'm OK with it. I would even consider merging it without further discussion but I'll give others a chance, juuuuuuust in case.

PR #116, opened 2016-04-27
Move Sieve to before Minesweeper

### LINKED ISSUE #112: Sieve. Too much too soon? Gigasecond. Who struggles?
As a Rust beginner, I found the `Sieve` exercise to be quite tough at that point in the track.

My solution ended up being quite terse, but that doesn't convey the sense of struggle involved, and it feels like an unexpected holiday when you hit `Gigasecond` (aka `HelloWorld 2: the revenge which is even less scary than the first time around`).

Probably the biggest offender in my opinion is `Gigasecond`. The ordering of `Sieve` related to other subsequent exercises is quite hard to pinpoint, and perhaps I was having a bad brain day.

IanWhitney:
Fixes #112

If a student is progressing through the exercises in order, this seems
like a reasonable place to put this exercise. They should be familiar
enough with Rust's iterator functions that they can arrive at the
solution more idiomatically.


petertseng:
I say 👍

PR #119, opened 2016-04-30
Queen Attack: Replace tuple with ChessPosition

### LINKED ISSUE #118: Passing a tuple to new in Queen attack exercise feels weird
Passing a pair to a `new` function feels unidiomatic.

There is a discussion started on the site: 
http://exercism.io/submissions/05c66333b1b44c12a224e8fe4884a50d

We can continue here.

IanWhitney:
Fixes #118

`Queen::new` now expects a `ChessPosition` instead of a tuple. `ChessPosition::new` returns a Result, indicating if the position is valid or not.

I've added a few tests at the start to capture the behavior of `ChessPosition`.  The tests of Queen have their API usage updated, and have been renamed for readability.

### DIFF
--- a/exercises/queen-attack/example.rs
+++ b/exercises/queen-attack/example.rs
@@ -1,78 +1,69 @@
-#[derive(Debug, PartialEq)]
-pub struct Queen { position: ChessPosition }
+pub struct Queen {
+    position: ChessPosition,
+}
+pub trait ChessPiece {
+    fn position(&self) -> &ChessPosition;
+    fn can_attack<T: ChessPiece>(&self, other: &T) -> bool;
+}
 
-impl Queen {
-    pub fn new(position: (i8, i8)) -> Result<Queen, ()> {
-        let position = ChessPosition::new(position);
-        if position.valid() {
-            Ok(Queen { position: position })
-        } else {
-            Err(())
-        }
+impl ChessPiece for Queen {
+    fn position(&self) -> &ChessPosition {
+        &self.position
     }
 
-    pub fn can_attack(&self, piece: &Queen) -> bool {
-        self.horizontal_from(&piece) ||
-            self.vertical_from(&piece) ||
-            self.diagonal_from(&piece)
-    }
-
-    fn horizontal_from(&self, piece: &Queen) -> bool {
-        self.rank() == piece.rank()
-    }
-
-    fn vertical_from(&self, piece: &Queen) -> bool {
-        self.file() == piece.file()
-    }
-
-    fn diagonal_from(&self, piece: &Queen) -> bool {
-        self.sum() == piece.sum() ||
-            self.difference() == piece.difference()
-    }
-
-    fn rank(&self) -> i8 {
-        self.position.rank()
-    }
-
-    fn file(&self) -> i8 {
-        self.position.file()
-    }
-
-    fn sum(&self) -> i8 {
-        self.position.sum()
-    }
-
-    fn difference(&self) -> i8 {
-        self.position.difference()
+    fn can_attack<T: ChessPiece>(&self, piece: &T) -> bool {
+        self.position.horizontal_from(&piece.position()) ||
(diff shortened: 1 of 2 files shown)


exercises/queen-attack/example.rs
```diff
@@ -1,78 +1,69 @@
...
-    fn rank(&self) -> i8 {
-        self.position.rank()
+pub struct Queen {
+    position: ChessPosition,
+}
+pub trait ChessPiece {
+    fn position(&self) -> &ChessPosition;
+    fn can_attack<T: ChessPiece>(&self, other: &T) -> bool;
+}
+//
```
petertseng: do you mean for this comment to be here?

exercises/queen-attack/tests/queen-attack.rs
```diff
@@ -3,79 +3,84 @@ extern crate queen_attack;
...
+#[ignore]
+fn queen_is_created_with_a_ok_position() {
+    Queen::new(ChessPosition::new(2, 4).unwrap());
+}
 
-    let white_queen = Queen::new((5,8));
-    assert!(white_queen.is_err());
+#[test]
+#[ignore]
+#[should_panic(expected = "Invalid Position")]
```
petertseng: I'm a little unsure about placing an expectation on the exact error message, but maybe I'd feel better if it were just "panic message includes this string" rather than "panic message is exactly this string". I'm just saying that because I don't want to stop anyone who wants to provide error messages such as "rank can't be negative" or something

petertseng:
Yeah seems like a good way to address what's discussed in #118.

Tell me about the changing of the `assert!(boolean_condition)` to `assert_eq!(true, boolean_condition)` - Is it for clarity? For consistency with the `false` case? Or does the error message look better? Or something else?

exercises/queen-attack/tests/queen-attack.rs
```diff
@@ -3,79 +3,84 @@ extern crate queen_attack;
...
+    assert_eq!(true, white_queen.can_attack(&black_queen));
 }
 
 #[test]
 #[ignore]
-fn test_can_attack_on_first_diagonal() {
-    let white_queen = Queen::new((2,2)).unwrap();
-    let black_queen = Queen::new((0,4)).unwrap();
-    assert!(white_queen.can_attack(&black_queen));
+fn queens_on_the_same_diagonal_can_attack_one() {
```
petertseng: Hmm, while we are changing the test names, maybe we can take the opportunity to bikeshed. I could see these being `upper_right`, `upper_left`, etc. rather than just `one`, `two`, etc.

But I don't actually care, so you can pick whatever makes the most sense.

exercises/queen-attack/tests/queen-attack.rs
IanWhitney: I'm not committed to it, but I thought it was a way to introduce that optional part of `should_panic`. I didn't know that option existed until about 20 minutes ago.

exercises/queen-attack/tests/queen-attack.rs
IanWhitney: Yeah, I couldn't think of a good name for them either.

IanWhitney:
`assert` vs `assert_eq`: probably a copy & paste error. Testing booleans should always use `assert`, I think. Hopefully Rust implements `refute!`, but I don't see an RFC for it. Perhaps I'll make one.

petertseng:
Okay I say :+1:  with squash (I think you'll agree on the squash)

petertseng:
Would like to hear what others involved think, of course.

exercises/queen-attack/tests/queen-attack.rs
petertseng: I read https://doc.rust-lang.org/book/testing.html and it says "The test harness will make sure that the failure message **contains** the provided text." (emphasis mine) so that addresses what I was thinking about.

PR #121, opened 2016-05-01
Set Ignore flag on 2nd-last test

IanWhitney:
I accidentally left out the 'ignore test' flag. Fixing.


IanWhitney:
Oops. Deleted the branch before Travis finished. Restored the branch and retried the build. No reason to think it'll fail, but who wants that red x?

IanWhitney:
Well, I dunno. Still complaining about missing git refs. I'm not going to worry about it anymore.

PR #125, opened 2016-05-17
Add Scrabble Score

IanWhitney:
Implements the Scrabble Score exercise.

As of right now there is no official test suite for Scrabble Score. There is a PR (https://github.com/exercism/x-common/pull/161), but it is not yet merged.

I am following the test suite in that PR, with one exception. I thought it was likely that a common naive implementation would be to map through the chars, find the score and then simply unwrap the value. Like this:

```rust
word.chars().map(|c| values.get(&c).fold(0, |sum, v| sum + v.unwrap())
```

So I have added a test that uses a non-English character that will cause this implementation to panic.

I've put this exercise between the Nucleotides and Roman Numerals, figuring that it did similar character iteration and value mapping.

My example implementation probably leaves a lot to be desired. I wanted to make the values hash a static, but you can't do that with a hash.

[Lazy Static](http://rust-lang-nursery.github.io/lazy-static.rs/lazy_static/index.html) does let you do this, but I'm not going to require a crate in my example.

### DIFF
--- a/config.json
+++ b/config.json
@@ -15,6 +15,7 @@
     "rna-transcription",
     "nucleotide-count",
     "nucleotide-codons",
+    "scrabble-score",
     "roman-numerals",
     "robot-name",
     "etl",
--- a/exercises/scrabble-score/Cargo.lock
+++ b/exercises/scrabble-score/Cargo.lock
@@ -1,0 +1,4 @@
+[root]
+name = "scrabble-score"
+version = "0.0.0"
+
--- a/exercises/scrabble-score/Cargo.toml
+++ b/exercises/scrabble-score/Cargo.toml
@@ -1,0 +1,3 @@
+[package]
+name = "scrabble-score"
+version = "0.0.0"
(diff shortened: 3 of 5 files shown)


IanWhitney:
I'm going to wait for https://github.com/exercism/x-common/pull/255 to merge before finishing this PR

IanWhitney:
Updated to follow https://github.com/exercism/x-common/pull/255

I'll wait a few days for comments. I'd like to merge this by May 24th, if there are no problems.

petertseng:
I dropped the ball.

But I don't see anything I would change.

Honestly I may have done the inserts with iterating over a `Vector<(char, u16)>` or something (avoids repeating `insert` so many times) but that's really not important enough to be really worth mentioning I guess

IanWhitney:
Now that I found https://doc.rust-lang.org/std/iter/trait.FromIterator.html, I'd probably use that in the example. But it doesn't really matter now.

Issue #117, opened 2016-04-28
Provide function signatures 


ijanos:
When solving Rust exercises the first thing I do is figuring out the name and the right type signature and I'm wondering about the usefulness of this. This process is busywork, I doubt anybody is learning from it and in some cases it requires reverse engineering  the thought process of the exercise creator.
I think the the signatures could be added to the readme or maybe a skeleton `src/lib.rs` could be provided with the exercises. 

Look at the *Parallel letter frequency* exercise for example. The solution has the signature of `frequency(&[&str], u32) -> HashMap<char, u32>` but nowhere in the readme or in the test cases are mentioned what is the second parameter supposed to be. It can be assumed from the name of one test case that it is the number of workers.
I was confused by it at first. This confusing could have been removed if the readme had a line similar to this:

> Your task is to implement the following function: `fn frequency(text: &[&str], workers: u32) -> HashMap<char, u32>`

What is your opinion on this?

IanWhitney:
I'm entirely on the fence. And I think whichever way we go there's going to be a group of students that don't like it. 

For me, I like not having the signature. I'm entirely new to Rust and static typing and I've found that I like puzzling out what I have to implement by looking at the tests. But that might just be me!

As for `parallel letter frequency`, I think the API could be improved. I find the current implementation to be hard to parse. Again, that might just be me.

kytrinyx:
I'm also on the fence. In the Go track we've provided stub files early on, and then we phase them out after the first few. For the most part reverse engineering function signatures doesn't seem to be difficult in Go, though, and I don't have experience with Rust, so I can't speak to how much busywork that adds here.

ijanos:
Take a look at the output of `cargo test` if I run an exercise with empty source file and another one with an empty function with a wrong type signature:  

https://gist.github.com/ijanos/59644c68af57e38d5fd81ba6b1baee27

Now compare it to the output once you get the type signature right:

https://gist.github.com/ijanos/53cb04393b6c4e80e5ecbcd908bae0c2

Interpreting the errors is definitely not hard but it can be intimidating at first and mildly annoying if someone doesn't get the types right at the first try. 

I think extending the README is a good middle ground to this issue. People still have the option to figure out the function signature while others can just copy paste it and start solving the real exercise.

kytrinyx:
We can extend the README by adding a `HINTS.md` file in the exercise directory.

IanWhitney:
[...]

PR #129, opened 2016-05-27
Ignore extraneous Rust files

IanWhitney:
Just so that we don't accidentally commit backup files, compiled code or
source files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 *.swp
-*.beam
 *.bk
 .DS_Store
 **/src
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 *.swp
 *.beam
+*.bk
 .DS_Store
+**/src
+**/target
 tmp
 bin/configlet
 bin/configlet.exe


IanWhitney:
Nothing controversial in here, I think. But I'll let it sit until Tuesday, May 31 before merging.

.gitignore
```diff
@@ -1,6 +1,9 @@
 *.swp
 *.beam
```
petertseng: (I know this isn't realted to this PR but this cuased me to notice) - BEAM? As in Erlang BEAM? Is that actually useful in a Rust repo?

.gitignore
```diff
@@ -1,6 +1,9 @@
...
+*.bk
```
petertseng: can you explain what .bk is? I didn't find such files in my rust repos, but I may not be looking hard enough and/or some rust repos may have different requirements than others

.gitignore
```diff
@@ -1,6 +1,9 @@
...
 .DS_Store
+**/src
```
petertseng: well there are five `src` dirs in this repo, but we'll consider them to be the exception not the rule

petertseng:
Ultimately I'd be OK with this as is but I'm curious about what `.bk` is

IanWhitney:
Apparently bk files come from rustfmt

https://github.com/rust-lang-nursery/rustfmt/issues/796

I have my editor setup to run rustfmt after every save. Apparently I can [turn off the bk files](https://github.com/rust-lang-nursery/rustfmt/issues/873), but we should probably ignore them as well.

.gitignore
IanWhitney: Probably not. Might as well git rid of it.

.gitignore
IanWhitney: Odd. Looking at the github files, I only see `src` in the Forth exercise. 

the `src/lib.rs` file in Forth sets up an Enum and stubs out some functions. That's something we talked about doing in a `hints.md` file in #117.

So we could move that forth file to a `hints.md` file and settle on a standard of "don't provide a stub src/lib.rs file". Or we could follow Forth's example and try to solve #117 by providing stub functions.

For student ease, I think the stubbed out `lib.rs` file is best. There are several exercises that require students to either comment-out tests or write their own stubs just so that they can get the first test to pass.

For example, Allergies. If you focus on the first test and implement `is_allergic_to`, then your tests will fail because you haven't yet implemented `allergies`. Even though that function isn't even introduced until the 4th test.

A stub file would solve that problem, right?

IanWhitney:
A useful conversation got lost in a outdated diff. so I'm going to pull it out here. @petertseng pointed out that he saw a few exercises with `/src` directories. To which, I said:

> Odd. Looking at the github files, I only see src in the Forth exercise.

> the src/lib.rs file in Forth sets up an Enum and stubs out some functions. That's something we talked about doing in a hints.md file in #117.

> So we could move that forth file to a hints.md file and settle on a standard of "don't provide a stub src/lib.rs file". Or we could follow Forth's example and try to solve #117 by providing stub functions.

> For student ease, I think the stubbed out lib.rs file is best. There are several exercises that require students to either comment-out tests or write their own stubs just so that they can get the first test to pass.

> For example, Allergies. If you focus on the first test and implement is_allergic_to, then your tests will fail because you haven't yet implemented allergies. Even though that function isn't even introduced until the 4th test.

> A stub file would solve that problem, right?

petertseng:
I don't mind forth's current `src/lib.rs` stub

The other `src` dirs were me having built some exercises locally (symlinking the example to src/lib.rs), so that's my problem

I think I'm OK with having `src` in gitignore since we'll still be able to add `src` dirs if we really want to

So I say :+1: to this

PR #131, opened 2016-05-30
Implement wordy

IanWhitney:
This implements the Wordy exercise and follows the canonical test suite.

I'm placing this problem before custom-set. Fairly late in the current problem track, but before some of the most complex exercises. This is a hard problem to place, because I think the implementations will vary. My initial solution is not that complex, but it is extremely brittle. Some students might go farther and implement a real-ish lexer/parser

But, by the time students get to this part of the track there's no question that they'll know enough Rust to be able to complete this problem

I would like to discuss the API I'm using. Specifically...

### Should WordProblem be a Struct?

In the most Rust-like languages, [Go](https://github.com/exercism/xgo/blob/master/exercises/wordy/wordy_test.go) does not implement a `WordProblem` concept, but [Swift](https://github.com/exercism/xswift/blob/master/exercises/wordy/WordyTest.swift) does. 

I went with a struct because that is my natural tendency. But I'm open to change.

### What is the right API?

I have given the struct an `answer` function that returns a `Result`. I think whatever the function is, it has to return a `Result` because of the last two tests. It's either that or have the last two tests use `should_panic`, which.... 😬 That doesn't seem awesome to me.

### DIFF
--- a/exercises/wordy/example.rs
+++ b/exercises/wordy/example.rs
@@ -1,0 +1,82 @@
+#[derive(Debug, PartialEq)]
+pub struct WordProblem {
+    command: String,
+}
+
+#[derive(Debug, PartialEq)]
+struct Token {
+    value: String,
+}
+
+impl Token {
+    fn is_valid(&self) -> bool {
+        !self.value.is_empty() && (self.is_operand() || self.is_operator())
+    }
+
+    fn is_operand(&self) -> bool {
+        self.value.chars().all(|c| c.is_numeric() || c == '-')
+    }
+
+    fn is_operator(&self) -> bool {
+        self.value == String::from("plus") || self.value == String::from("minus") ||
+        self.value == String::from("multiplied") || self.value == String::from("divided")
+    }
+}
+
+impl WordProblem {
+    pub fn new(c: &str) -> Self {
+        WordProblem { command: String::from(c) }
+    }
+
+    pub fn answer(&self) -> Result<isize, ()> {
+        let mut t = self.tokens();
+        let mut result: isize = 0;
+        let mut opr = "plus".to_string();
+
+        if t.len() <= 1 {
+            Err(())
+        } else {
+            while t.len() > 1 {
+                result = self.evaluate(result, opr, self.operand(&t.remove(0)));
+                opr = self.operator(&t.remove(0));
+            }
+            result = self.evaluate(result, opr, self.operand(&t.remove(0)));
+            Ok(result)
+        }
+    }
+
+    fn evaluate(&self, mut r: isize, operator: String, operand: isize) -> isize {
+        if operator == String::from("plus") {
+            r += operand;
+            r
+        } else if operator == String::from("minus") {
+            r -= operand;
+            r
+        } else if operator == String::from("multiplied") {
+            r *= operand;
+            r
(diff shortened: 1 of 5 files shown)


IanWhitney:
And another one is ready for you, @exercism/rust

IanWhitney:
I think I'm debating between two API options:

The one I have now is: 

```rust
WordProblem::new(command).answer().unwrap()
```

And the other option is:

```rust
wordy::answer(command).unwrap()
```

IanWhitney:
Ok, this has been hanging out a week without comment. Merging. Students will tell me what I got wrong.

exercises/wordy/example.rs
```diff
@@ -0,0 +1,82 @@
...
+        } else if operator == String::from("divided") {
+            r /= operand;
+            r
+        } else {
+            r
+        }
+    }
+
+    fn operand(&self, t: &Token) -> isize {
+        let x: isize = t.value.parse().unwrap();
```
petertseng: is the reason this line is here (rather than just `t.value.parse().unwrap()` and omit the following line) to make it an `isize`? I would have thought it would be possible without explicitly specifying, but  maybe there's another reason

exercises/wordy/example.rs
```diff
@@ -0,0 +1,82 @@
...
+        WordProblem { command: String::from(c) }
+    }
+
+    pub fn answer(&self) -> Result<isize, ()> {
+        let mut t = self.tokens();
+        let mut result: isize = 0;
+        let mut opr = "plus".to_string();
+
+        if t.len() <= 1 {
+            Err(())
```
petertseng: stylistically I may have considered making this a `return`, then unindenting the rest (removing the `else`), but not something I'd urgently change

exercises/wordy/tests/wordy.rs
```diff
@@ -0,0 +1,99 @@
...
+
+use wordy::*;
+
+#[test]
+fn addition() {
+    let command = "What is 1 plus 1?";
+    assert_eq!(2, WordProblem::new(command).answer().unwrap());
+}
+
+#[test]
```
petertseng: our other exercises `#[ignore]` all but the first test, should we do so here too?

petertseng:
`Result` was definitely the right choice, panic didn't make sense. `Option` may have been considered, but `Result` leaves the possibility of having the error case specify why it's an error, so I too would have picked `Result`.

I'm not strongly opinionated about the interface, though I have an inkling that whatever the interface, at least this impementation was right to use a struct. I think even if had chosen the `wordy::answer(command).unwrap()` interface it would have made the most sense to keep the `WordProblem` struct internally.

exercises/wordy/example.rs
IanWhitney: I'm not especially familiar with `parse()`, but this is the only way I could get the parsing of a string into an isize to work.

petertseng: hmm. now you're getting me curious. what happened when you only say `t.value.parse().unwrap()` ? As far as I can tell, that compiles and passes the tests for me...

IanWhitney: Huh. I was definitely getting complier errors earlier, but now I'm not. No idea.

PR #133, opened 2016-05-30
Update test to prevent panic with missing value

cjschneider2:
The `check_word_count` function tries remove elements from the `HashMap`
which may not be there and then `.unwraps()` them directly causing a
panic if they are not there. This is changed to `unwrap_or(0)` to show
that no values were found to improve the failure message in this case.


IanWhitney:
Looks good. Thanks!

PR #134, opened 2016-05-30
Check that result contains expected allergens

### LINKED ISSUE #130: Allergies: don't test for the order of allergies
The order of allergies should be irrelevant, but since `Allergies.allergies()` returns a vector, and order is important when comparing vectors, `fn test_allergic_to_everything()` fails when it shouldn't.

Maybe expecting `Allergies.allergies()` to return a `HashSet` is a better choice here?

taravancil:
Fixes #130

Testing for all allergies would fail if the allergens weren't in the
same order as the test vector.

This checks that the result contains all the expected allergens rather
than expecting the result and test vector to be equal.

### DIFF
--- a/exercises/allergies/tests/allergies.rs
+++ b/exercises/allergies/tests/allergies.rs
@@ -40,10 +40,14 @@ fn test_just_to_peanuts() {
 #[test]
 #[ignore]
 fn test_allergic_to_everything() {
-    assert_eq!(vec![Allergen::Eggs, Allergen::Peanuts, Allergen::Shellfish,
-                    Allergen::Strawberries, Allergen::Tomatoes, Allergen::Chocolate,
-                    Allergen::Pollen, Allergen::Cats],
-        Allergies::new(255).allergies());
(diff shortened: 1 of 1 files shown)


IanWhitney:
Good change!

I would add one more test: that `all` and `allergies` have the same number of elements. If `allergies` contains `Eggs` twice, for example, it'd pass that test but still have a failure somewhere in its implementation.

kytrinyx:
Are we missing this test in https://github.com/exercism/x-common/blob/master/allergies.json?

taravancil:
Oh wait, do you mean check that `all` and `allergies` have the same number of elements in a completely different test? I think that makes more sense.

IanWhitney:
Same test or different test, I don't think I have a strong opinion. I'm happy to merge this as-is. 

@kytrinyx, I think the tests in the json file are fine. This is just a way to make the canonical tests work for Rust, while still handling the weird edge case that could emerge because of our testing approach.

taravancil:
Great! I'll leave it alone then. No need to over-complicate things.

IanWhitney:
Fireworks. Marching Bands. Etc. Thank you!

PR #137, opened 2016-06-01
Reorder Problems to (hopefully) improve learning

### LINKED ISSUE #127: Problem Ordering
As brought up in #126 (and elsewhere), our problem ordering is weird. I've certainly put some problems in the wrong spots, for sure.

I ran through the example solutions for all of our problems made a table of what skills each problem (probably) requires. Implementations can vary, obviously. 

problem | topics
----- | -----
hello-world |  Some/None. Really? We did that in Hello World?
gigasecond |  Crates, type stuff. 
leap |  Math, booleans
anagram |  lifetimes, str vs string, loops, iter, vector
difference-of-squares |  fold & map
allergies |  struct, enum, bitwise (probably), vectors, filter
word-count |  hashmap, str vs string, chars, entry api
hamming |  result, chars, filter
rna-transcription |  match, struct, str vs string
nucleotide-count |  filter, entry api, mutablity, match
nucleotide-codons |  struct, hash map, lifetimes, Result
scrabble-score |  chaining map/fold. Hashmap (maybe)
roman-numerals |  mutable, results, loops, struct, traits
robot-name |  struct, slices, randomness, lifetimes, self mut
etl |  btree? 
raindrops |  case (or formatting). Mutable string
bob |  chars, string functions
grade-school |  struct, entry api, Vec, Option
phone-number |  option, format, unwrap_or, iters, match
hexadecimal |  Option, zip/fold/chars, map
queen-attack |  struct, trait (optional), Result
beer-song |  case, vector (?), loop
sieve |  vector, map, while let (optional)
minesweeper |  board state, Vec, heavy logic
dominoes |  I do not even know, man
parallel-letter-frequency | multi threading? heavy
sublist |  enum, generic over type
custom-set |  generic over type, vector, equality, struct
tournament |  enum,  file io, try!, result, hashmap, struct
rectangles |  traits and structs, enum
forth |  like, everything but lifetimes
circular-buffer |  same

So, some stuff is at the bottom that clearly should be (rectangles, forth, circular-buffer). But the start of our problem list is not well geared towards newcomers. Let's discuss better options.

IanWhitney:
Fixes #127

This commit follows the problem order I suggested in #127, with one
exception. I've put sublist ahead of allergies. Both use enum, but the
enums in allergies can be more of a hassle, as evidenced by all of the
ways I've seen students implement them.

Sublist does have the disadvantage of also requiring a solution that is
generic over multiple types. But I think the best thing to do is keep an
eye on students that reach this part of the track.

Or the later introduction of new exercises could improve this flow.


IanWhitney:
#127 has had this proposed solution for a while, and there haven't been any complaints. So I plan on merging this PR on Friday, June 3 unless I hear otherwise.

And when I do.... 🎆 🎉 🎈

kytrinyx:
I think this looks great!

I think you're right that we might need an extra exercise that helps introduce some of the enum stuff earlier than sublist and allergies.

PR #138, opened 2016-06-02
Point students to the xrust Repository

IanWhitney:
I was trying to figure out how how I even found this repository. I think
I stumbled into it via the contributing guide on the main exercism repo.

I thought it would be nice to have each exercise contain a pointer to
the repo, so that students can easily see where all this code comes
from.

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -20,5 +20,11 @@ if you wish.
 Make sure to read the [Crates and Modules](https://doc.rust-lang.org/stable/book/crates-and-modules.html) chapter if you
 haven't already, it will help you with organizing your files.
 
+## Feedback, Issues, Pull Requests
+
+The [exercism/xrust](https://github.com/exercism/xrust) repository on GitHub is the home for all of the Rust exercises. If you have feedback about an exercise, or want to help implement new exercises, head over there and create an issue. Members of the [rust track team](https://github.com/orgs/exercism/teams/rust) are happy to help!
+
+If you want to know more about Exercism, take a look at the [contribution guide](https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md).
+
 [help-page]: http://exercism.io/languages/rust
 [crates-and-modules]: http://doc.rust-lang.org/stable/book/crates-and-modules.html


SETUP.md
```diff
@@ -20,5 +20,11 @@ if you wish.
 Make sure to read the [Crates and Modules](https://doc.rust-lang.org/stable/book/crates-and-modules.html) chapter if you
 haven't already, it will help you with organizing your files.
 
+## Feedback, Issues, Pull Requests
+
+The [exercism/xrust](https://github.com/exercism/xrust) repository on GitHub is the home for all of the Rust exercises. If you have feedback about an exercise, or want to help implement new exercises, head over there and create an issue. Members of the [rust track team](https://github.com/orgs/exercism/teams/rust) are happy to help!
```
petertseng: rust team *might* only visible to those who are already members of the org. will check with a logged out win shortly

petertseng: yeah asks me to log in. Not sure what's the best way to deal with this

petertseng:
hmm, seems like a thing that might be good to have in all the repos. wonder if we can make the site automatically generate this for all

SETUP.md
IanWhitney: Dang. I'll remove that part, I guess.

IanWhitney:
Hey @kytrinyx, @petertseng has a good idea. Can we do this automatically? 

Also, any ideas on how we publicly share the track maintainer team link?

kytrinyx:
I've added an issue to create an overview of important repositories here: https://github.com/exercism/exercism.io/issues/2924

The idea is that we can link to it from the new contributing page that's being worked on here:
https://github.com/exercism/exercism.io/pull/2918

Also, we should probably stick something at the bottom of the main code page that links to the relevant repository, as well as make sure that if you go to the exercism.io/languages/TRACK_ID page, that it is easy to find the repository.

I'm not sure about the gitter thing—I think we should definitely figure that out and standardize it (and script it) in some way. Want to open something in https://github.com/exercism/discussions/issues so that we can get everyone who cares about this involved over there?

I'll ask some folks at GitHub about a public link to a team. That would be helpful.

PR #142, opened 2016-06-06
Update Allergies tests to canonical set

### LINKED ISSUE #135: Update Allergies to use canonical tests
Our tests currently differ from [the canonical tests](https://github.com/abeger/x-common/blob/50d691afa763a084eab6529ef08d76972a5481c1/allergies.json)

Update the tests to the standard, while also being aware of the issue raised in #130. Many of the new canonical tests will also need to handle ordering. Either use the fix introduced in #134 for those tests, or derive some new way of comparing results.

IanWhitney:
Fixes #135

- Introducing tests that were missing
- Handling the ordering problem in all tests that check vectors,
  encapsulating the solution introduced in #134
- Updating test names to not start with `test_`
        - Cargo starts each test result line with "test ", followed by
          the test's name.. So these tests looked like "test
          test_allergic_to_peanuts".
- Trying to give the final test a useful name. "ignore non allergen
  score parts" didn't make much sense to me when I worked on this
  exercise. I think the new name better describes what we're trying to
  capture.

### DIFF
--- a/exercises/allergies/tests/allergies.rs
+++ b/exercises/allergies/tests/allergies.rs
@@ -2,8 +2,16 @@ extern crate allergies;
 
 use allergies::*;
 
+fn compare_allergy_vectors(expected: &Vec<Allergen>, actual: &Vec<Allergen>) {
+    assert_eq!(actual.len(), expected.len());
+
+    for allergy in expected {
+        assert!(actual.contains(&allergy));
+    }
+}
+
 #[test]
-fn test_no_allergies_means_not_allergic() {
+fn is_not_allergic_to_anything() {
     let allergies = Allergies::new(0);
     assert_eq!(false, allergies.is_allergic_to(&Allergen::Peanuts));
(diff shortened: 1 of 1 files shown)


IanWhitney:
While this works, the new `compare_allergy_vectors` function does not return useful error messages when it fails. Any ideas how to improve this?

IanWhitney:
@taravancil Thanks for getting this started! I found a more Rusty way to compare two vectors that I thought you might be interested in.

petertseng:
Are sure this is doing what it should? If I replace `allergies` with:

```rust
        let mut v: Vec<_> = Allergies::allergens().into_iter().filter(|allergen| self.is_allergic_to(allergen)).collect();
        v.reverse();
        v   
```

Then I get (among others):

```
---- allergic_to_eggs_and_peanuts stdout ----
	thread 'allergic_to_eggs_and_peanuts' panicked at 'Expected [Eggs, Peanuts], got [Peanuts, Eggs]', tests/allergies.rs:7
```

PR #144, opened 2016-06-16
Copy track icon into language track repository

martinsvalin:
See exercism/exercism.io#2925 for more details.


config.json
```diff
@@ -44,7 +44,8 @@
   ],
   "ignored": [
     "_test",
-    "docs"
+    "docs",
+    "img"
```
petertseng: note for us, https://github.com/exercism/configlet/pull/5/files made this explicit add unnecessary - do we still want it anyway?

martinsvalin: Ah. I did the same PR for another language yesterday, before that PR to configlet was merged. Do you want me to remove the change?

petertseng: I thought about it a bit. I'm not hugely concerned. In fact it might be less suprising if we keep it. Imagine if we removed this change, then later down the road I take a look and wonder "well the img dir is here, but why is it not being excluded...?"

petertseng: I thought about it a bit. I'm not hugely concerned. In fact it might be less suprising if we keep it. Imagine if we removed this change, then later down the road I take a look and wonder "well the img dir is not here, but why is it being ignored despite its absence..?"

PR #146, opened 2016-06-27
Implement Robot Simulator

IanWhitney:
Implement Robot Simulator

This follows the [current standard test suite](https://github.com/exercism/x-common/commit/183934754b1847612809db410a8aeb640678f188#diff-7dba901ac525b4fbb77d291df9c5ff9e)

The stub file, as we discussed in #117, is just enough to run `cargo test` and not get any errors from ignored tests.

### DIFF
--- a/exercises/robot-simulator/example.rs
+++ b/exercises/robot-simulator/example.rs
@@ -1,0 +1,101 @@
+#[derive(PartialEq, Debug, Copy, Clone)]
+pub enum Direction {
+    North,
+    East,
+    South,
+    West,
+}
+
+impl Direction {
+    pub fn previous_clockwise(&self) -> Self {
+        match *self {
+            Direction::North => Direction::West,
+            Direction::East => Direction::North,
+            Direction::South => Direction::East,
+            Direction::West => Direction::South,
+        }
+    }
+
+    pub fn next_clockwise(&self) -> Self {
+        match *self {
+            Direction::North => Direction::East,
+            Direction::East => Direction::South,
+            Direction::South => Direction::West,
+            Direction::West => Direction::North,
+        }
+    }
+}
+
+#[derive(Clone, Copy)]
+struct Position {
+    x: i32,
+    y: i32,
+}
+
+impl Position {
+    fn new(x: i32, y: i32) -> Self {
+        Position { x: x, y: y }
+    }
+
+    fn advance(&self, direction: &Direction) -> Self {
+        match *direction {
+            Direction::North => Self::new(self.x, self.y + 1),
+            Direction::South => Self::new(self.x, self.y - 1),
+            Direction::East => Self::new(self.x + 1, self.y),
+            Direction::West => Self::new(self.x - 1, self.y),
+        }
+    }
+}
+
+#[derive(Clone)]
+pub struct Robot {
+    position: Position,
+    direction: Direction,
+}
+
+impl Robot {
+    pub fn new(x: i32, y: i32, d: Direction) -> Self {
+        Robot::build(Position::new(x, y), d)
+    }
+
+    fn build(position: Position, direction: Direction) -> Self {
+        Robot {
+            position: position,
+            direction: direction,
+        }
+    }
+
+    pub fn turn_right(&self) -> Self {
+        Self::build(self.position, self.direction.next_clockwise())
+    }
+
+    pub fn turn_left(&self) -> Self {
+        Self::build(self.position, self.direction.previous_clockwise())
+    }
+
+    pub fn advance(&self) -> Self {
+        Self::build(self.position.advance(&self.direction), self.direction)
+    }
+
+    pub fn instructions(&self, instructions: &str) -> Self {
+        instructions.chars().fold(self.clone(),
+                                  |robot, instruction| robot.execute(instruction))
+    }
+
+    pub fn position(&self) -> (i32, i32) {
+        (self.position.x, self.position.y)
+    }
+
+    pub fn direction(&self) -> &Direction {
+        &self.direction
+    }
+
+    fn execute(self, command: char) -> Self {
+        match command {
+            'R' => self.turn_right(),
+            'L' => self.turn_left(),
+            'A' => self.advance(),
+            _ => self,
+        }
+    }
+}
--- a/exercises/robot-simulator/src/lib.rs
+++ b/exercises/robot-simulator/src/lib.rs
@@ -1,0 +1,44 @@
+// The code below is a stub. Just enough to satisfy the compiler.
+// In order to pass the tests you can add-to or change any of this code.
+
+#[derive(PartialEq, Debug)]
+pub enum Direction {
+    North,
+    East,
+    South,
+    West,
+}
+
+pub struct Robot;
+
+impl Robot {
+    #[allow(unused_variables)]
+    pub fn new(x: isize, y: isize, d: Direction) -> Self {
+        unimplemented!()
+    }
+
+    pub fn turn_right(self) -> Self {
+        unimplemented!()
+    }
+
+    pub fn turn_left(self) -> Self {
+        unimplemented!()
+    }
+
+    pub fn advance(self) -> Self {
+        unimplemented!()
+    }
+
+    #[allow(unused_variables)]
+    pub fn instructions(self, instructions: &str) -> Self {
+        unimplemented!()
+    }
+
+    pub fn position(&self) -> (isize, isize) {
+        unimplemented!()
+    }
+
+    pub fn direction(&self) -> &Direction {
+        unimplemented!()
+    }
+}
--- a/exercises/robot-simulator/tests/robot-simulator.rs
+++ b/exercises/robot-simulator/tests/robot-simulator.rs
@@ -1,0 +1,147 @@
+extern crate robot_simulator;
+
+use robot_simulator::*;
+
+#[test]
+fn robots_are_created_with_position_and_direction() {
+    let robot = Robot::new(0, 0, Direction::North);
+    assert_eq!((0, 0), robot.position());
+    assert_eq!(&Direction::North, robot.direction());
+}
+
+#[test]
+#[ignore]
+fn positions_can_be_negative() {
+    let robot = Robot::new(-1, -1, Direction::South);
+    assert_eq!((-1, -1), robot.position());
+    assert_eq!(&Direction::South, robot.direction());
+}
+
+#[test]
+#[ignore]
+fn turning_right_does_not_change_position() {
+    let robot = Robot::new(0, 0, Direction::North).turn_right();
+    assert_eq!((0, 0), robot.position());
+}
+
+#[test]
+#[ignore]
+fn turning_right_from_north_points_the_robot_east() {
+    let robot = Robot::new(0, 0, Direction::North).turn_right();
+    assert_eq!(&Direction::East, robot.direction());
+}
+
+#[test]
+#[ignore]
+fn turning_right_from_east_points_the_robot_south() {
+    let robot = Robot::new(0, 0, Direction::East).turn_right();
+    assert_eq!(&Direction::South, robot.direction());
+}
+
+#[test]
+#[ignore]
+fn turning_right_from_south_points_the_robot_west() {
+    let robot = Robot::new(0, 0, Direction::South).turn_right();
+    assert_eq!(&Direction::West, robot.direction());
+}
+
+#[test]
+#[ignore]
+fn turning_right_from_west_points_the_robot_north() {
+    let robot = Robot::new(0, 0, Direction::West).turn_right();
+    assert_eq!(&Direction::North, robot.direction());
+}
+
+#[test]
+#[ignore]
+fn turning_left_does_not_change_position() {
+    let robot = Robot::new(0, 0, Direction::North).turn_left();
+    assert_eq!((0, 0), robot.position());
+}
+
+#[test]
+#[ignore]
+fn turning_left_from_north_points_the_robot_west() {
+    let robot = Robot::new(0, 0, Direction::North).turn_left();
+    assert_eq!(&Direction::West, robot.direction());
+}
+
+#[test]
+#[ignore]
+fn turning_left_from_west_points_the_robot_south() {
+    let robot = Robot::new(0, 0, Direction::West).turn_left();
+    assert_eq!(&Direction::South, robot.direction());
+}
+
+#[test]
+#[ignore]
+fn turning_left_from_south_points_the_robot_east() {
+    let robot = Robot::new(0, 0, Direction::South).turn_left();
+    assert_eq!(&Direction::East, robot.direction());
+}
+
+#[test]
+#[ignore]
+fn turning_left_from_east_points_the_robot_north() {
+    let robot = Robot::new(0, 0, Direction::East).turn_left();
+    assert_eq!(&Direction::North, robot.direction());
+}
+
+#[test]
+#[ignore]
+fn advance_does_not_change_the_direction() {
+    let robot = Robot::new(0, 0, Direction::North).advance();
+    assert_eq!(&Direction::North, robot.direction());
+}
+
+#[test]
+#[ignore]
+fn advance_increases_the_y_coordinate_by_one_when_facing_north() {
+    let robot = Robot::new(0, 0, Direction::North).advance();
+    assert_eq!((0, 1), robot.position());
+}
+
+#[test]
+#[ignore]
+fn advance_decreases_the_y_coordinate_by_one_when_facing_south() {
+    let robot = Robot::new(0, 0, Direction::South).advance();
+    assert_eq!((0, -1), robot.position());
+}
+
+#[test]
+#[ignore]
+fn advance_increases_the_x_coordinate_by_one_when_facing_east() {
+    let robot = Robot::new(0, 0, Direction::East).advance();
+    assert_eq!((1, 0), robot.position());
+}
+
+#[test]
+#[ignore]
+fn advance_decreases_the_x_coordinate_by_one_when_facing_west() {
+    let robot = Robot::new(0, 0, Direction::West).advance();
+    assert_eq!((-1, 0), robot.position());
+}
+
+#[test]
+#[ignore]
+fn follow_instructions_to_move_west_and_north() {
+    let robot = Robot::new(0, 0, Direction::North).instructions("LAAARALA");
+    assert_eq!((-4, 1), robot.position());
+    assert_eq!(&Direction::West, robot.direction());
+}
+
+#[test]
+#[ignore]
+fn follow_instructions_to_move_west_and_south() {
+    let robot = Robot::new(2, -7, Direction::East).instructions("RRAAAAALA");
+    assert_eq!((-3, -8), robot.position());
+    assert_eq!(&Direction::South, robot.direction());
+}
+
+#[test]
+#[ignore]
+fn follow_instructions_to_move_east_and_north() {
+    let robot = Robot::new(8, 4, Direction::South).instructions("LAAARRRALLLL");
+    assert_eq!((11, 5), robot.position());
+    assert_eq!(&Direction::North, robot.direction());
+}
--- a/config.json
+++ b/config.json
@@ -22,6 +22,7 @@
     "roman-numerals",
     "hexadecimal",
     "grade-school",
+    "robot-simulator",
     "queen-attack",
     "sublist",
     "allergies",
--- a/exercises/robot-simulator/Cargo.lock
+++ b/exercises/robot-simulator/Cargo.lock
@@ -1,0 +1,4 @@
+[root]
+name = "robot-simulator"
+version = "0.0.0"
+
--- a/exercises/robot-simulator/Cargo.toml
+++ b/exercises/robot-simulator/Cargo.toml
@@ -1,0 +1,3 @@
+[package]
+name = "robot-simulator"
+version = "0.0.0"


exercises/robot-simulator/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+
+pub struct Robot {
+}
+
+impl Robot {
+    pub fn new(x: isize, y: isize, d: &Direction) -> Self {
+        Robot {}
+    }
+
+    pub fn turn_right(&mut self) {}
```
jonasbb: I would pre-fill the functions with `unimplemented!()` just to make sure, that you know where to put the implementation. This is more important for functions which return values, because `unimplemented!()` always typechecks, no matter the type.

exercises/robot-simulator/src/lib.rs
```diff
@@ -0,0 +1,32 @@
+#[derive(PartialEq, Debug)]
+pub enum Direction {
+    North,
+    East,
+    South,
+    West,
+}
+
+pub struct Robot {
```
jonasbb: I would personally use `pub struct Robot;`, because it feels a bit less pushed towards "normal" structs, because a tuple struct like `Robot(Direction, i32, i32)` would also be possible.

exercises/robot-simulator/src/lib.rs
```diff
@@ -0,0 +1,32 @@
+#[derive(PartialEq, Debug)]
```
jonasbb: Is there a reason to not derive `Eq` as well?

jonasbb:
Your test suite will push them to a certain solution in any case. If you set `mut` it is more likely to change the state of the struct, but if you remove `mut` and return a new struct, you need to adapt your test cases to support this.

exercises/robot-simulator/src/lib.rs
IanWhitney: I was only adding what I needed in order to get the compiler to compile. I imagine the real solution will have more derived traits.

exercises/robot-simulator/src/lib.rs
IanWhitney: Cool. Done. Thanks!

IanWhitney:
Ah, right, if I remove the `mut`  then the signature would change to 

```rust
fn turn_right(&self) -> Self {
  Robot::New{...}
}
```
Which might be interesting. I don't think any of our exercises take that approach.

IanWhitney:
Two tweaks:

- Stub & Tests now use immutable robots.
- I'm silencing some warnings

IanWhitney:
I'm unsure what the `direction` function should return --  `Direction` or `&Direction`?

Currently the Robot `new` function uses `&Direction`, which makes sense to me. Students can choose to handle the lifetimes that come with a struct using a borrowed reference, or they can derive `Clone` and just clone the Direction. That works for me.

But when getting the `Direction` back out, what is the idiomatic thing to do? My instinct says that the function should be

`fn direction(&self) -> &Direction`

But my instinct has been wrong a lot.

steveklabnik:
usually, accessor methods return `&T`. If they returned `T`, then they would destroy the owner, or you'd be making a clone every time.

IanWhitney:
Ooh, my instinct was right for once. Go team instinct! 👏

IanWhitney:
What about primitives, @steveklabnik ?

For example, given the struct

```rust
struct Robot {
  x: isize,
  y: isize
}
```

Would you expect a `position` function to be:

```rust
fn postion(&self) -> (isize, isize)
```

or 

```rust
fn postion(&self) -> (&isize, &isize)
```

Alternatively we could create a whole new Position type, like we did with [queen attack](https://github.com/exercism/xrust/blob/master/exercises/queen-attack/tests/queen-attack.rs)

steveklabnik:
If the type is `Copy`, then they're basically the same, so yeah, returning `T` is fine. Consider http://doc.rust-lang.org/stable/collections/vec/struct.Vec.html#method.len as an example

IanWhitney:
That's what I was thinking as well. I'm clearly on a roll today.

Dr-Emann:
Taking by value and returning a copy just doesn't feel rusty-y to me. I expect functions to take the most general type of arguments they can, `&T` if no mutation, `&mut T` if mutation needed, and `T` only if ownership is _required_.

IanWhitney:
Working example in place. Now to figure out where to place this problem.

The solution will involve:

- Enums (though the stub file provides them)
- Structs (either a tuple struct, or a `normal` struct)
- Some matching on `Direction` variant
- Iteration

And, if the student chooses not to clone the `Direction`, they may end up in a world of Lifetimes. That, though, is optional.

I'm thinking it should go around `grade-school`. It's got a pretty straightforward use of enums, so it should come before Allergies (which has slightly more complex enum usage). But beyond enums, students will be familiar with all the tools they need to solve this by the time they reach `grade-school`

IanWhitney:
So what API are you proposing, @Dr-Emann?

Dr-Emann:
I would expect `turn_right`, `turn_left`, `advance`, and `instructions` to take `&mut self`, instead of `self`, and return nothing.

IanWhitney:
That was the initial implementation. And it might be the right one. But I'm leaning towards immutable because

- Immutability is the default in Rust
- No other problem (that I know of) really features Rust's default immutability
- Immutability is not my natural inclination

I figure if I don't expect immutability then other programmers with an OO-focused background probably don't expect immutability. So a problem that forces immutability may make me (and them) think about immutability. Which I think is good.

Immutability brings other benefits, though they aren't exposed by the tests. It would be very easy to trace your Robot's path, for example. Re-winding to a specific point on the path would be trivial. All very nice, but not part of the test suite.

IanWhitney:
Ok. This now passes CI and is basically ready to go. Once we settle on API and improve the example code a bit, it'll be ready to merge.

IanWhitney:
Thinking about the immutability thing more... 

As I mentioned earlier, I think the immutable design advantages become clear if you want to rewind to a point. Maybe we could have an optional test that's commented out:

```rust
// Optional:
// #[test]
// fn robot_simulator_can_return_robot_at_any_point() {
// let robot = Robot::new(...);
// let simulator = RobotSimulator::new(robot);
// simulator.simulate("LAAAR");
// assert_eq!((-1,2), simulator.at_step(4).robot_position);
// assert_eq!((0,2), simulator.at_step(3).robot_position);
// etc.
// }
```

IanWhitney:
I think this is now ready. Leaving the PR open until Tuesday, July 5th, after which I'll merge if there are no issues.

Ryman:
@IanWhitney Is the intention for the example implementation to be showing idiomatic rust, or just a potential solution to pass the tests?

IanWhitney:
@Ryman That is a good question without a clear answer! 

Track maintainers discussed this recently: https://github.com/exercism/discussions/issues/27 

Unsurprisingly, people had different opinions. I think everyone is agreed that the first job of the example code is to pass the tests (otherwise our CI fails).

Beyond that first requirement, there was general agreement (though not unanimous) that the code should be:

- not clever
- idiomatic
- a good example

Probably my favorite description of the example code was [this one](https://github.com/exercism/discussions/issues/27#issuecomment-225046387):

> [T]ry to solve the new proposed exercise as if the track maintainer were a real exercism user.

Which I most certainly am. I ended up writing Rust exercises because I wanted to use Exercism to help me learn Rust. My example code for this exercise is probably pretty close to what I would submit if I did this exercise.

IanWhitney:
That's not to say my example code is exactly idiomatic, though. I'm still quite new to Rust and am learning new stuff all the time. Which is one of the reasons I encourage discussion in these PRs.

exercises/robot-simulator/example.rs
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    fn build(position: Position, direction: Direction) -> Self {
+        Robot {
+            position: position,
+            direction: direction,
+        }
+    }
+
+    pub fn turn_right(self) -> Self {
```
petertseng: I'll not comment on whether we want to go with mutable or immutable interface.

But I want to say that if we want an immutable interface, I suspect we can do `turn_right(&self) -> Self` instead of `turn_right(self) -> Self`. Can you check?

(And if it were mutable, it would be `turn_right(&mut self) -> Self`)

petertseng: (mutable should not return. Editing previous comment)

petertseng: I'll not comment on whether we want to go with mutable or immutable interface.

But I want to say that if we want an immutable interface, I suspect we can do `turn_right(&self) -> Self` instead of `turn_right(self) -> Self`. Can you check?

(And if it were mutable, it would be `turn_right(&mut self)`)

exercises/robot-simulator/src/lib.rs
petertseng: The takeaway I get from this is: "If students think that they can't add more traits, we should add a comment telling them they can". I don't know whether the conditional is true (I can't read minds).

exercises/robot-simulator/example.rs
```diff
@@ -0,0 +1,102 @@
...
+        Self::build(self.position, self.direction.previous_clockwise())
+    }
+
+    pub fn advance(self) -> Self {
+        Self::build(self.position.advance(&self.direction), self.direction)
+    }
+
+    pub fn instructions(self, instructions: &str) -> Self {
+        let mut r = Self::build(self.position, self.direction);
+        for c in instructions.chars() {
```
petertseng: Despite `fold` being possible here, this is straightforward. So I would probably stick with this, unless someone really thinnks we should use `fold` here.

exercises/robot-simulator/example.rs
IanWhitney: That results in compiler errors. Can not move out of borrowed content. There are additional changes we'd need to make to get that to work, though I don't know what they are off the top of my head.

exercises/robot-simulator/src/lib.rs
IanWhitney: For my example code I ended up not needing Eq or Copy. But I did require clone. Other students may need different traits, though.

What we might want at the top of all of our stub files is a general comment to say "This file is here to get you started. Change/add/remove code as you see fit."

exercises/robot-simulator/example.rs
petertseng: This comment surprised me enough that I checked out your branch to take a look.

My diagnosis: Make your `Position` `#[derive(Clone, Copy)]` and it will work.

petertseng: same with `advance` and `instructions` - If your `Direction` is `Copy` then they can take `&self`

Can't do the same for `execute` though since it can potentially return `self`, which means a `&self` wouldn't be able to become a `Self`.

exercises/robot-simulator/example.rs
IanWhitney: I tried `fold` for a bit but was running into problems. Ownership, I think. This solution was my fallback. Once the problem is being solved by students I plan on checking their solutions to see how I should have been using fold. 

That's what I do with most problems. Get it working myself, then review other solutions to figure out how I _should_ be doing it.

exercises/robot-simulator/example.rs
IanWhitney: Huh. I never would have guessed that from the compiler error. Clearly. 

Based on this quote from The Book, I'll update these to take `&self`

> We should default to using &self, as you should prefer borrowing over taking ownership, as well as taking immutable references over mutable ones.

exercises/robot-simulator/example.rs
IanWhitney: Fold may have been failing because of the Clone/Copy problems you already diagnosed. I'll try it again.

exercises/robot-simulator/src/lib.rs
IanWhitney: Comment added.

exercises/robot-simulator/example.rs
petertseng: well, to be fair, making them `Copy` was not the only solution - could have only made them `Clone`, not `Copy`, and then explicitly call `clone()` on them. But I suggested `Copy` simply because https://doc.rust-lang.org/std/marker/trait.Copy.html said "Generally speaking, if your type *can* implement `Copy`, it should"

petertseng: I don't always do this myself (maybe I should if I run into a compiler output I don't understand), but rustc should have outputted ```src/lib.rs:67:21: 67:25 help: run `rustc --explain E0507` to see a detailed explanation```

And I found that one of the solutions  `rustc --explain E0507` does say you can make it `Copy`, so that could have been a way to understand.

exercises/robot-simulator/example.rs
petertseng: well, looks like you did want to use `fold`, and well I'll not complain about you using it! I think I only hesitate because I'm not sure if the typical Rust user is comfortable thinking in `fold`s, so I'm never sure whether it's idiomatic and whether I should just use a (relatively more) straightforward `for` instead. Maybe I should have more faith. It is true that it can result in shorter code.

exercises/robot-simulator/example.rs
```diff
@@ -0,0 +1,100 @@
...
+    pub fn turn_left(&self) -> Self {
+        Self::build(self.position, self.direction.previous_clockwise())
+    }
+
+    pub fn advance(&self) -> Self {
+        Self::build(self.position.advance(&self.direction), self.direction)
+    }
+
+    pub fn instructions(&self, instructions: &str) -> Self {
+        instructions.chars().fold(Self::build(self.position, self.direction),
```
petertseng: I haven't tested, but if you had `Robot` derive `Clone`, I think this `Self::build(...)` could just be `clone()` instead. But I don't see this as a change that has to be done.

exercises/robot-simulator/src/lib.rs
petertseng: Seems good. It makes me wonder that if we are going to be in the habit of providing stubs, maybe this comment should be a standard that goes on top of all of them...

clearly an idea for another PR, but neverthelss worth bringing up while we are gthinking about it here

petertseng:
For me, I think I've said all I want to say :+1:

exercises/robot-simulator/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+    East,
+    South,
+    West,
+}
+
+pub struct Robot;
+
+impl Robot {
+    #[allow(unused_variables)]
+    pub fn new(x: isize, y: isize, d: &Direction) -> Self {
```
Ryman: The `Direction` argument being a reference here is a bit out of place when compared to the rest of the interface, for if we assume the user may want to store `&Direction` in their `Robot` type, the other methods will not allow this to be updated to anything reasonable, i.e.:
```rust
fn turn_right(&self) -> Self {
     Robot { 
         position: self.position,
         // nowhere to tie this lifetime to as it's not an input
         direction: &self.direction.next_clockwise() 
     }
}
```

exercises/robot-simulator/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+    #[allow(unused_variables)]
+    pub fn instructions(self, instructions: &str) -> Self {
+        unimplemented!()
+    }
+
+    pub fn position(&self) -> (isize, isize) {
+        unimplemented!()
+    }
+
+    pub fn direction(&self) -> &Direction {
```
Ryman: `Direction` should probably be `Copy` which would make this idiomatically `-> Direction`

exercises/robot-simulator/tests/robot-simulator.rs
```diff
@@ -0,0 +1,147 @@
...
+#[ignore]
+fn turning_left_does_not_change_position() {
+    let robot = Robot::new(0, 0, &Direction::North).turn_left();
+    assert_eq!((0, 0), robot.position());
+}
+
+#[test]
+#[ignore]
+fn turning_left_from_north_points_the_robot_west() {
+    let robot = Robot::new(0, 0, &Direction::North).turn_left();
```
Ryman:
```rust
let robot = Robot::new(..);
let robot = robot.turn_left();
assert ...
```
If these tests were written in the above form then users might be able to implement the methods with either of these signatures: 
- `fn builder_style(self) -> Self`
- `fn with_copies(&self) -> Self` 
- `fn mutably(&mut self) -> &mut Self`

exercises/robot-simulator/example.rs
```diff
@@ -0,0 +1,101 @@
...
+pub enum Direction {
+    North,
+    East,
+    South,
+    West,
+}
+
+impl Direction {
+    pub fn previous_clockwise(&self) -> Self {
+        match self {
```
Ryman: nit: These matches are generally written in the form: 
```rust
match *self {
    Direction::North => Direction::West,
    ...
}
```

(Note the lack of reference sigils in the patterns)

exercises/robot-simulator/example.rs
```diff
@@ -0,0 +1,101 @@
...
+    pub fn turn_right(&self) -> Self {
+        Self::build(self.position, self.direction.next_clockwise())
+    }
+
+    pub fn turn_left(&self) -> Self {
+        Self::build(self.position, self.direction.previous_clockwise())
+    }
+
+    pub fn advance(&self) -> Self {
+        Self::build(self.position.advance(&self.direction), self.direction)
```
Ryman: This is personal preference but I find these one liners rather noisy and think separating it out is clearer to see what's happening:
```rust
let mut next = self.clone();
next.position = next.position.advance(self.direction);
next
```

Or to pattern match it into components and rebuild:
```rust
let Robot { mut position, direction } = self.clone();
position = position.advance(direction);
Self::build(position, direction)
```

Or with functional update syntax:
```rust
Robot {
    position: self.position.advance(&self.direction),
    // makes it more clear that only the position property is changing
    .. self.clone() 
}
```

exercises/robot-simulator/example.rs
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    pub fn instructions(&self, instructions: &str) -> Self {
+        instructions.chars().fold(self.clone(),
+                                  |robot, instruction| robot.execute(instruction))
```
Ryman: nit: formatting of the closure is a bit unexpected, usually I see this formatted as:

```rust
instructions.chars().fold(self.clone(), |robot, instruction| {
    robot.execute(instruction)
})

File: exercises/robot-simulator/example.rs [hunk 13]
```diff
@@ -0,0 +1,101 @@
...
+    pub fn direction(&self) -> &Direction {
+        &self.direction
+    }
+
+    fn execute(self, command: char) -> Self {
+        match command {
+            'R' => self.turn_right(),
+            'L' => self.turn_left(),
+            'A' => self.advance(),
+            _ => self,
```
Ryman: Should there be a test exercising 'unknown' commands? (I would expect this to panic, or propagate an error value)

Ryman:
Sorry for the delayed response, I left some feedback on things which may affect the users code and just left some `nits` on the example itself.

exercises/robot-simulator/tests/robot-simulator.rs
IanWhitney: In my testing (which may have missed something) if we want to allow mutable implementations, then the tests you suggest don't work.

For example, if I do this test:

```rust
fn turning_right_does_not_change_position() {
    let robot = Robot::new(0, 0, &Direction::North);
    let robot = robot.turn_right();
    assert_eq!((0, 0), robot.position());
}
```

and this implementation

```rust
pub fn turn_right(&mut self) -> &mut Self {
        self.position = Position::new(1, 1);
        self
}
```

Then I get the compiler error

```
tests/robot-simulator.rs:8:17: 8:22 error: cannot borrow immutable local variable `robot` as mutable
tests/robot-simulator.rs:8     let robot = robot.turn_right();
                                           ^~~~~
tests/robot-simulator.rs:7:9: 7:14 help: to make the local variable mutable, use `mut` as shown:
tests/robot-simulator.rs:      let mut robot = Robot::new(0, 0, &Direction::North);
```

And the compiler is right, if I change the test to

```rust
let mut robot = Robot::new(0, 0, &Direction::North);
```

Then we're cool. 

The current test structure does already allow for the first two function signatures. But I don't know of a way to also allow mutability without mutability being declared in the tests.

exercises/robot-simulator/example.rs
IanWhitney: That's a good suggestion for the [common test suite](https://github.com/exercism/x-common/blob/master/robot-simulator.json)

exercises/robot-simulator/example.rs
IanWhitney: All the formatting is being done automatically by [rustfmt](https://github.com/rust-lang-nursery/rustfmt). I try not to fight it.

exercises/robot-simulator/src/lib.rs
IanWhitney: I don't think there's anything in the tests that forces Direction to be Copy. If there is, then I agree that this signature makes more sense. But if not, then I think we should stick with &T

exercises/robot-simulator/src/lib.rs
IanWhitney: It's preferred for structs to take ownership of their values, so I'm OK with pushing students towards that.

exercises/robot-simulator/tests/robot-simulator.rs
Ryman: Ah yes, you're totally correct! I forgot that would be required!

I guess that silencing the "unnecessary mut" warnings with an attribute, or even having `mut` in the tests at all whenever the user chooses to do things with copies could be confusing so it's probably best to leave things as they are then. It's a shame as I feel the mutable implementation can be more idiomatic here.

exercises/robot-simulator/example.rs
Ryman: The suite mentions the possibility to have an optional test for ["an invalid instruction throws an error"](https://github.com/exercism/x-common/blame/master/robot-simulator.json#L6).

exercises/robot-simulator/example.rs
Ryman: Fair enough! :)

exercises/robot-simulator/src/lib.rs
Ryman: To be clear, that means `fn new(x: isize, y: isize, d: Direction) -> Self` rather than the current signature with the push towards the student calling `clone` to take ownership?

IanWhitney:
Unless there's more discussion on this, I'll merge tonight.

IanWhitney:
And in it goes. Thanks for all your input!

PR #147, opened 2016-06-27
Variable Length Quantity

jonasbb:
I tried to solve the variable length quantity exercise, because I quite like the encoding scheme. Unfortunately, there are no other implementations of this problem yet, so I am looking for feedback for the function names and test design.

If you know a better way to store bytes back to front into a `Vec` I would be glad to know this, because allocating a vector and then reversing it seems a bit weird.

Values larger than 0x0FFFFFFF could also be tested in `to_bytes`. The `from_bytes` is lacking test for error conditions, like numbers which cannot be represented in `u32` and incomplete byte sequences.

### DIFF
--- a/exercises/variable-length-quantity/tests/variable-length-quantity.rs
+++ b/exercises/variable-length-quantity/tests/variable-length-quantity.rs
@@ -1,0 +1,102 @@
+extern crate variable_length_quantity as vlq;
+
+#[test]
+fn to_single_byte() {
+    assert_eq!(&[0x00], vlq::to_bytes(&[0x00]).as_slice());
+    assert_eq!(&[0x40], vlq::to_bytes(&[0x40]).as_slice());
+    assert_eq!(&[0x7f], vlq::to_bytes(&[0x7f]).as_slice());
+}
+
+#[test]
+#[ignore]
+fn to_double_byte() {
+    assert_eq!(&[0x81, 0x00], vlq::to_bytes(&[0x80]).as_slice());
+    assert_eq!(&[0xc0, 0x00], vlq::to_bytes(&[0x2000]).as_slice());
+    assert_eq!(&[0xff, 0x7f], vlq::to_bytes(&[0x3fff]).as_slice());
+}
+
+#[test]
+#[ignore]
+fn to_triple_byte() {
+    assert_eq!(&[0x81, 0x80, 0x00], vlq::to_bytes(&[0x4000]).as_slice());
+    assert_eq!(&[0xc0, 0x80, 0x00], vlq::to_bytes(&[0x10_0000]).as_slice());
+    assert_eq!(&[0xff, 0xff, 0x7f], vlq::to_bytes(&[0x1f_ffff]).as_slice());
+}
+
+#[test]
+#[ignore]
+fn to_quadruple_byte() {
+    assert_eq!(&[0x81, 0x80, 0x80, 0x00],
+               vlq::to_bytes(&[0x20_0000]).as_slice());
+    assert_eq!(&[0xc0, 0x80, 0x80, 0x00],
+               vlq::to_bytes(&[0x0800_0000]).as_slice());
+    assert_eq!(&[0xff, 0xff, 0xff, 0x7f],
+               vlq::to_bytes(&[0x0fff_ffff]).as_slice());
+}
+
+#[test]
+#[ignore]
+fn to_quintuple_byte() {
+    assert_eq!(&[0x81, 0x80, 0x80, 0x80, 0x00],
+               vlq::to_bytes(&[0x1000_0000]).as_slice());
+    assert_eq!(&[0x8f, 0xf8, 0x80, 0x80, 0x00],
+               vlq::to_bytes(&[0xff00_0000]).as_slice());
+    assert_eq!(&[0x8f, 0xff, 0xff, 0xff, 0x7f],
+               vlq::to_bytes(&[0xffff_ffff]).as_slice());
+}
+
+#[test]
+#[ignore]
+fn from_bytes() {
+    assert_eq!(&[0x7f], vlq::from_bytes(&[0x7f]).unwrap().as_slice());
+    assert_eq!(&[0x2000],
+               vlq::from_bytes(&[0xc0, 0x00]).unwrap().as_slice());
+    assert_eq!(&[0x1f_ffff],
+               vlq::from_bytes(&[0xff, 0xff, 0x7f]).unwrap().as_slice());
+    assert_eq!(&[0x20_0000],
+               vlq::from_bytes(&[0x81, 0x80, 0x80, 0x00]).unwrap().as_slice());
+    assert_eq!(&[0xffff_ffff],
+               vlq::from_bytes(&[0x8f, 0xff, 0xff, 0xff, 0x7f]).unwrap().as_slice());
+}
+
+
+#[test]
+#[ignore]
+fn to_bytes_multiple_values() {
+    assert_eq!(&[0x40, 0x7f], vlq::to_bytes(&[0x40, 0x7f]).as_slice());
+    assert_eq!(&[0x81, 0x80, 0x00, 0xc8, 0xe8, 0x56],
+               vlq::to_bytes(&[0x4000, 0x12_3456]).as_slice());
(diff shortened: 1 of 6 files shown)


IanWhitney:
Thanks!

New exercises are awesome. I think the normal first step for a new exercise is to add the markdown, yaml and json files to x-common https://github.com/exercism/x-common

That's where I'd normally expect conversations about test design. And other track maintainers are more likely to see conversations in x-common.

Nothing about the Rust implementation jumps out at me. But I've only given it a quick glance.

jonasbb:
[md file](https://github.com/exercism/x-common/blob/master/variable-length-quantity.md) and [yml file](https://github.com/exercism/x-common/blob/master/variable-length-quantity.yml) are already present in x-common. It is just that no other language has yet implemented it.

IanWhitney:
Ok. We can make a first attempt at a test suite here and then extract it to a common json file.

exercises/variable-length-quantity/tests/variable-length-quantity.rs
```diff
@@ -0,0 +1,89 @@
...
+#[test]
+fn to_single_byte() {
+    assert_eq!(&[0x40], vlq::to_bytes(&[0x40]).as_slice());
+    assert_eq!(&[0x7f], vlq::to_bytes(&[0x7f]).as_slice());
+}
+
+#[test]
+#[ignore]
+fn to_single_byte_zero() {
+    assert_eq!(&[0x00], vlq::to_bytes(&[0x00]).as_slice());
```
IanWhitney: Should this test go first? It seems like it's the simplest to pass, and I usually put the simplest-to-pass tests first.

exercises/variable-length-quantity/tests/variable-length-quantity.rs
```diff
@@ -0,0 +1,89 @@
+extern crate variable_length_quantity as vlq;
+
+#[test]
+fn to_single_byte() {
+    assert_eq!(&[0x40], vlq::to_bytes(&[0x40]).as_slice());
+    assert_eq!(&[0x7f], vlq::to_bytes(&[0x7f]).as_slice());
```
IanWhitney: Is there a value to having both of these tests? Is there an implementation detail that's tested in the second assert that is not tested in the first assert?

exercises/variable-length-quantity/tests/variable-length-quantity.rs
```diff
@@ -0,0 +1,89 @@
...
+fn to_double_byte() {
+    assert_eq!(&[0x81, 0x00], vlq::to_bytes(&[0x80]).as_slice());
+    assert_eq!(&[0xc0, 0x00], vlq::to_bytes(&[0x2000]).as_slice());
+    assert_eq!(&[0xa6, 0x37], vlq::to_bytes(&[0x1337]).as_slice());
+    assert_eq!(&[0xff, 0x7f], vlq::to_bytes(&[0x3fff]).as_slice());
+}
+
+#[test]
+#[ignore]
+fn to_tripple_byte() {
```
IanWhitney: little spelling thing: `triple`

exercises/variable-length-quantity/tests/variable-length-quantity.rs
```diff
@@ -0,0 +1,89 @@
...
+#[test]
+#[ignore]
+fn chained_execution_is_identity_10000() {
+    let test: Vec<u32> = (0..10000).collect();
+    assert_eq!(test, vlq::from_bytes(vlq::to_bytes(&test).as_slice()));
+}
+
+#[test]
+#[ignore]
+fn chained_execution_is_identity_lfsr() {
```
IanWhitney: What is `lfsr`?

exercises/variable-length-quantity/tests/variable-length-quantity.rs
```diff
@@ -0,0 +1,89 @@
...
+#[test]
+#[ignore]
+fn to_single_byte_zero() {
+    assert_eq!(&[0x00], vlq::to_bytes(&[0x00]).as_slice());
+}
+
+#[test]
+#[ignore]
+fn to_double_byte() {
+    assert_eq!(&[0x81, 0x00], vlq::to_bytes(&[0x80]).as_slice());
```
IanWhitney: Similar to my earlier comment, are there edge cases that these additional tests are capturing? I've found that tests that focus on one thing work best. So if each of these asserts is testing a unique implementation detail, then I'd make this 4 tests.

But if getting the first assert to pass will also make the remaining 3 asserts pass, then I'd get rid of the additional asserts.

exercises/variable-length-quantity/tests/variable-length-quantity.rs
jonasbb: It's an edge case in my solution, where the general does not work. So I did not put it first.

exercises/variable-length-quantity/tests/variable-length-quantity.rs
jonasbb: Linear feedback shift register. I use it here to create pseudo random numbers. With the correct XOR value you can create a sequence of all numbers of size x-bits which wraps around.

jonasbb:
Due to the lack of other test cases to compare I took all the examples from the md file and made an assert for each of it. The examples include the smallest, a random, and the largest number representable in x bytes (1 to 4). This might be more than necessary, but does not seem unreasonable to me.

You can definitely pass the single byte cases without passing the multi byte ones. However, I do not think there is much different between the multibyte test case, so that it's likely to pass them all at once. At least if you try to implement a generic solution instead of maybe a more optimal one, which has hard-coded path for each length.

Both identity tests should be redundant too with respect to all of the above. The only new thing they are testing are large vectors (again, general solution). The difference between them is the source of numbers, first 10000 vs pseudo random.

IanWhitney:
I've not found large vector tests to be very useful. We have some in the [`sublist` exercise](https://github.com/exercism/xrust/blob/master/exercises/sublist/tests/sublist.rs#L143) and I turned them off when worked on that problem. All they showed me was that my solution wasn't performant, not that it was invalid.

I do like the idea of random vector tests as an ersatz form of generative testing. I just don't think the number of random elements needs to be large.

Basing the tests off of the Readme makes sense. I hadn't noticed that you'd done that.

In short, what I think we need to do before merging:

- Remove the large vectors
- Have 1 test that uses a small collection of random inputs

IanWhitney:
Looking a lot better!

I don't think we need to make this the last problem in the track. Looking at the examples, it seems like students will need to use:

- Encoding
- Slices
- Vector
- Bitwise operators
- Result

When I look at other problems it seems like it would fit well after Allergies (which also uses bitwise).  I don't know of a problem that features slices, but students may have come across them by this point. If not, this problem would be a nice introduction to `as_slice`.

@kytrinyx, would a file named `example2.rs` be copied to the student's machine when they fetch this exercise? I'm not sure how that works.

jonasbb:
Hi, thanks for the comment.
I merged example.rs with example2.rs to avoid possible problems. I also changed the order in the config file. I just put it in somewhere without thinking much about it.

IanWhitney:
Great! I think this is looking good. I'll let it sit for a bit and try to get feedback from @petertseng and others. Barring problems I intend to merge this Friday, July 15.

petertseng:
My only question is do we want to be specific about the messages in the `Err` or just check that they are `is_err`? Advantages to both.

I think everything else that I would say has been said

jonasbb:
It seems to be more common to only check for `is_err()` instead of a specific string. Notable exceptions are Forth and Circular Buffer, but they provide `Error` enums for different cases. This would leave Hamming to be the sole exception.

IanWhitney:
@jonasbb Do you want to rebase this down to a commit or should I do the Squash & Commit?

kytrinyx:
Quick note: it looks like this got rebased (we don't get notifications when new code is pushed up, so @IanWhitney may not have seen this).

kytrinyx:
> would a file named example2.rs be copied to the student's machine when they fetch this exercise

Nope, anything with `example` in the path will be ignored (case insensitive regex).

IanWhitney:
Problem number 36! Thanks very much @jonasbb! 🍰

PR #148, opened 2016-06-27
Test that bob responds properly to non-ascii prompts.

gregcline:
This is just pulled directly from the Elixir bob exercise. Since strings in Rust are UTF8 by default why not have a test phrase with more than ASCII in mind.

### DIFF
--- a/exercises/bob/tests/bob.rs
+++ b/exercises/bob/tests/bob.rs
@@ -53,3 +53,8 @@ fn test_silent_treatment() {
     assert_eq!("Fine. Be that way!", bob::reply(""));
 }
 
+#[test]
+#[ignore]
+fn test_shouting_in_russian() {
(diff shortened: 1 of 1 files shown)


IanWhitney:
Thanks for submitting!

There are [canonical tests](https://github.com/exercism/x-common/blob/master/bob.json) for Bob that the Rust track currently does not follow. These tests include some tests that could cover [similar edgecases](https://github.com/exercism/x-common/blob/master/bob.json#L79).

I think we're better off updating our tests to follow the standard first. Then, if we can cover some special UTF8-only edge cases, we should.

gregcline:
Ah, I didn't realize that. Feel free to close this then if you'd like. Thanks!

IanWhitney:
No problem. Thanks again.

PR #151, opened 2016-07-02
Add transpose exercise

andrey-gvrd:
An easy exercise based on this [daily programmer challenge](https://www.reddit.com/r/dailyprogrammer/comments/4msu2x/challenge_270_easy_transpose_the_input_text).

Not sure where to put the README.


IanWhitney:
Awesome! Thanks.

When adding a whole new exercise to Exercism, we [need to do some additional work](https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md#implementing-a-completely-new-exercise)

So, the next steps here are:

[ ] Add `transpose.md` and `transpose.yml` files to https://github.com/exercism/x-common
[ ] Update this PR to add `transpose` to the `config.json` file
  - This will determine where the problem appears as students work through the Rust track

I'd also like to create a `transpose.json` file in x-common, which will give the problem a canonical suite of tests. But that's not a requirement to get this problem merged.

andrey-gvrd:
`transpose.json` is probably not happening due to the nature of test data.

IanWhitney:
The testing approach here is similar to the Tournament exercise, which we recently discussed: https://github.com/exercism/xrust/issues/122

When I looked at some other implementations of Tournament, I noticed that their tests defined the inputs/outputs in-line, instead of in separate files. For example, [Go](https://github.com/exercism/xgo/blob/master/exercises/tournament/tournament_test.go).

Would it make sense to do the same thing here? Is there an advantage to having large input/output files? Are they exposing edge cases that can't be exposed by smaller inputs/outputs?

andrey-gvrd:
I couldn't figure out a way to make the string formatting work in-line in rust (leading spaces specifically) so I went with the file approach.

kytrinyx:
Sweet! @andrey-gvrd do you happen to know what the licensing is on those challenges?

andrey-gvrd:
@kytrinyx 
I've reached out to /r/dailyprogrammer admins, they basically say that its up to the original suggester of the question. The only contact of his that I have is his inactive reddit [account](https://www.reddit.com/user/gommie) so I'm not able to get in touch with him.
Given the simplicity of the challenge, I'd think we wouldn't need the explicit permission anyway.

IanWhitney:
Thanks. I hope to dig into this more over the weekend.

kytrinyx:
@andrey-gvrd OK, let's be sure to give them proper credit in the yaml file in x-common.

kytrinyx:
... which I've checked and we're all good 👍

IanWhitney:
As @petertseng found in #152 and #153, tournament's output files allowed students to pass all the tests with an empty implementation. Is that the case here?

Also in #152, we settled on a pretty good way of replacing the input files with inline text

https://github.com/exercism/xrust/pull/152/files#diff-747e29dabc2baacda24a6125a9c768ceR5

Could the same approach be used here to remove the need for input & output files?

exercises/transpose/tests/output-2.txt
```diff
@@ -0,0 +1,16 @@
+TT
+hh
+ee
+
```
ErikSchierboom: Is this correct? I would expect that this line would have two spaces, as the original input also has two spaces.

exercises/transpose/tests/output-2.txt
```diff
@@ -0,0 +1,16 @@
...
+fs
+ie
+rc
+so
+tn
+ d
+l
```
ErikSchierboom: Similarly, I expected a space as the second character.

IanWhitney:
It looks like this one hasn't been touched for a while. If @andrey-gvrd wants to finish it up, fantastic. If not, we'll decide to either close or finish it up ourselves.

From what I can see, the remaining items are

- [ ] Move from input files to inline test input
- [ ] Address @ErikSchierboom questions

Probably also want to rebase this branch off of master so that we can add this problem in `problems.md`

IanWhitney:
Thoughts on closing this one? Either that or one of us finishes it off.

petertseng:
The "Address questions" item probably expands to "use the tests in https://github.com/exercism/x-common/blob/master/exercises/transpose/canonical-data.json "

There are currently enough of those that I won't have the time to do this anytime soon, so it is up to our brave contributors. If anyone wishes to use these commits directly and just build on top of them feel free to check out the branch, maintaining the original author to give the credit. Or obviously one could start afresh if truly desired.

Issue #122, opened 2016-05-06
Discuss tournament exercise


ijanos:
I see the following issues with the tournament exercise: 

- At first I had no idea what the columns supposed to mean  `| MP |  W |  D |  L |  P`. OK, I'm kind of ignorant about sports and I made an educated guess that the W is for Win and L is Loss but still it baffled me at first. 
- No mention of a win worth 3 points, a draw 1 and a loss 0. I had to reverse engineer this from the test outputs. Again this could be do to my ignorance towards sports, maybe this is a common scoring system, but I believe exercism puzzles should be self-contained. 
- the file formats are not explained:
 - the input format can contain comments, lines starting with `#`
 - what is a valid line?
 - should I expect invalid lines?
 - if the input line `team1;team2;win` does it mean team1 or team2 won?
- Why should the `tally` function return the number of parsed lines? What purpose does it serve? Why does it need a return value at all? I think the author intended the use of `Result` but it is not expressed anywhere and because tests use `unwrap` the `Optional` type can also be used. 

And another issue, thinking about the "big picture": this exercise requires reading files, parsing them and writing data to files. I think smaller exercises with only reading and only writing files should precede this one.

IanWhitney:
Thanks! I haven't done this exericse yet, but it sounds like it has some issues. I will dig into it this weekend.

IanWhitney:
Points 1 & 2 could be addressed in the [common readme](https://github.com/exercism/x-common/blob/master/tournament.md), I think. A change there would propagate to all tracks that implement Tournament, which would be good.

I think the common Readme should also address the format of the input file, and include examples. I'm equally baffled by it. I can not tell you which team won a game.


I also don't see a reason for `tally` to return anything. I'm not even sure it needs to be public. If the goal is to compare two text files then I think we'd want to compare `Display` output or something.

Based on the current description of the problem, I also suggest removing the test of bad lines. But I only think that because it's unclear to me what problem this exercise is trying to get me to solve. Is it:

- Parsing of text files
- Scoring Football results
- Sorting and displaying tabular data

Or some combination of the three?

If the exercise's goal is "Parse text files", then I think the tests of invalid/commented lines are valid. That is exactly the sort of edge case you want to make students think about.

But if not, then invalid lines are an unnecessary wrinkle in this exercise.

Since this exercise is not widely supported, it's pretty easy to do a quick survey of the other implementations and see what they focus on:

[...]

PR #152, opened 2016-07-02
tournament: Put inputs/expectations inline, not in files

petertseng:
The previous iteration of reading/writing files is a valuable experience
in itself, but dilutes the focus of this exercise. We should consider an
exercise dedicated to that task if necessary, as discussed in #122.

Note that this preserves all behavior of existing test files. In
particular, the expectation that an invalid line is ignored is
preserved. If you wish for this behavior to be changed, you should
discuss that at https://github.com/exercism/x-common/issues/286 before
we make that change here.

The function `tally` now takes only a `&str` (it previously took two
`Path`, one for input and one for output) and now returns only a
`String` (it previously returned a `Result<u32>` where the `Ok`
counted the number of valid lines).

This commit is necessary but not sufficient for the closing of #122.
Points not covered by this commit but requested by #122 include:
* Explain the input format. Covered by
  https://github.com/exercism/x-common/pull/254
* Explain behavior on invalid lines. Covered by
  https://github.com/exercism/x-common/issues/286


exercises/tournament/example.rs
```diff
@@ -77,13 +67,12 @@ fn write_tally(results: &HashMap<String, TeamResult>, output_filename: &Path) ->
...
-        try!(writeln!(&mut f, "{:30} | {:2} | {:2} | {:2} | {:2} | {:2}",
-                      team, games, r.wins, r.draws, r.losses, points));
-    }
-    Ok(())
+    let header = format!("{:30} | MP |  W |  D |  L |  P\n", "Team");
+    let lines: Vec<_> = v.iter().map(|&(ref team, games, r, points)| {
+        format!("{:30} | {:2} | {:2} | {:2} | {:2} | {:2}",
+                team, games, r.wins, r.draws, r.losses, points)
+    }).collect();
+    header + &lines.join("\n")
```
petertseng: Note that for an empty input, this prints the header and nothing after. No test tests this behavior, so I have nothing to say about whether this appropriate.

exercises/tournament/tests/tournament.rs
```diff
@@ -1,51 +1,57 @@
...
-    assert_eq!("\n".to_string() + output.as_ref(), "\n".to_string() + expected.as_ref());
-    
-}
-
-
 #[test]
 fn test_good() {
-    assert_eq!(tournament::tally(&Path::new("tests/input1.txt"), &Path::new("tests/output1.txt")).unwrap(), 6);
-    file_equal("tests/output1.txt", "tests/expected1.txt");
+    let input = "Allegoric Alaskians;Blithering Badgers;win\n\
```
petertseng: your comment is accepted on whether you prefer to see `\n\` at the end of every line or not, but if you choose "not", then all other lines will be unindented. Here's an example of what it looks like.

```rust
fn test_good() {
    let input = "Allegoric Alaskians;Blithering Badgers;win
Devastating Donkeys;Courageous Californians;draw";
    let expected = "Team                           | MP |  W |  D |  L |  P
Allegoric Alaskians            |  1 |  1 |  0 |  0 |  3"; // other expected lines omitted, this is just an example
    assert_eq!(tournament::tally(&input), expected);
```

petertseng:
amusing fact: Since the files output1.txt, output2.txt, output3.txt were pre-provided before this commit, I believe this means an empty implementation that does nothing will pass the tests.

exercises/tournament/example.rs
```diff
@@ -30,12 +27,9 @@ impl TeamResult {
     }
 }
 
-pub fn tally(input_filename: &Path, output_filename: &Path) -> Result<u32> {
-    let reader = BufReader::with_capacity(2048, File::open(input_filename).unwrap());
-    let mut count = 0;
+pub fn tally(input: &str) -> String {
```
petertseng: I would like a reminde ron the idiomatic-ness of `str` and `String`. I will see if I can find an answer myself.

exercises/tournament/tests/tournament.rs
petertseng: then again, maybe that way is less confusing to students. maybe that is the way to go.

IanWhitney:
This is good. I want to go through it a bit more carefully, but I don't think there's anything controversial here. As I mentioned in #122, no other languages use input/output files for this exercise. I don't think they are necessary.

exercises/tournament/tests/tournament.rs
petertseng: I think this can be alleviated by adding a comment explaining to students that the leading newlines will be stripped.

petertseng: I think this can be alleviated by adding a comment explaining to students that the leading whitespaces will be stripped.

IanWhitney: Does this work?

```
   let input = "Allegoric Alaskians;Blithering Badgers;win\n" +
                 "Devastating Donkeys;Courageous Californians;draw\n" +
                 "Devastating Donkeys;Allegoric Alaskians;win\n" +
                 "Courageous Californians;Blithering Badgers;loss\n" +
                 "Blithering Badgers;Devastating Donkeys;loss\n" +
                 "Allegoric Alaskians;Courageous Californians;win";
```

petertseng: Yes, with one small modification:

```rust
    let input = "Allegoric Alaskians;Blithering Badgers;win\n".to_string() +
        "Devastating Donkeys;Courageous Californians;draw\n" +
        "Devastating Donkeys;Allegoric Alaskians;win\n" +
        "Courageous Californians;Blithering Badgers;loss\n" +
        "Blithering Badgers;Devastating Donkeys;loss\n" +
        "Allegoric Alaskians;Courageous Californians;win";
```

Else, we would get:

```
tests/tournament.rs:5:17: 5:63 error: binary operation `+` cannot be applied to type `&'static str` [E0369]
tests/tournament.rs:5     let input = "Allegoric Alaskians;Blithering Badgers;win\n" +
                                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Anyway I think your suggestion combines the criteria of "Doesn't require you to realise that the trailing \ strips out the leading whitespace" and "doesn't make the code look weird", so I am going to apply it to all the strings!

petertseng:
squashed in the string suggestion since I'm pretty sure it's the way to go

petertseng:
a small rebase since both this and #155 removed some `use` lines (I just had to tell git "yup, all of them should remain removed")

IanWhitney:
With the removal of the File IO stuff, this problem could probably now be moved to a different spot in the order. It still requires a fair amount of work, but it's mostly just string parsing and some comparisons. Maybe it should move to somewhere in these problems?

- hexadecimal
- grade-school
- queen-attack

petertseng:
Yeah now the list of skills is "string parsing, custom sorting, string formatting". I'm going to say after grade-school; both are problems that may see use of a HashMap to solve (store students by grade, store reults by team). Of course, if we have a desire to not put them together since they have too similar things, just say the word (that's why I mention it!)

petertseng:
Oh, I originally moved between grade-school and queen-attack.

In the meantime, robot-simulator was also added to the exact same position.

I'll add it before robot simulator.

petertseng:
Merging soon since it seems that all that wants to be said has been said.

petertseng:
Amazing. According to https://github.com/exercism/x-common/pull/22 , tournament was **intentionally designed** to include file I/O. The Rust track was the first track to have this exercise.

Looks like that purpose has been lost since then, both in this track and others.

I generally think the file I/O portion is not well-suited for the testing style of Exercism, but I see that above we considered whether there can be an exercise specifically dedicated for file I/O (the grep exercise ostensibly requires file I/O too).

I am sorry to the creator of this exercise for playing my part in the exercise's diversion from its original purpose. It was in the name of giving the exercise some purpose at all.

petertseng:
C# is the only track with vestiges of that file I/O since it uses streams.

PR #153, opened 2016-07-02
DO NOT MERGE: tournament: don't bother writing the output file

petertseng:
Since the output file is pre-provided, we don't need to write it, the
test will just pass.


petertseng:
Tests pass. This means lazy implementations can get away with not writing files. We should not allow them. #154 will fix. Do not merge this PR.

PR #154, opened 2016-07-02
tournament: rm output files

petertseng:
With these output files present, implementation that do not actually
write files will be able to pass the tests, defeating the purpose. See
PR #153 for an example of such a lazy implementation.

They were added in 65b691e235ff2c0bc908950267917460d765ac48 in what I
can only assume was an accident, since they were added long after the
tournament exercise was added to this track in
ee0a6fde1cc8c6c7d5f984511cc6aaddc6dc75ad.


IanWhitney:
I'm fine with this.

Issue #149, opened 2016-06-29
test cases for nucleotide count should include handing of invalid sequences


lcowell:
In some implementations I've seen of this exercise, if I were to pass in a sequence of ATCGLL, we see include `L: 2` in the results. There is no test to indicate if that behaviour is invalid and it doesn't seem right. Are we in favour of adding a test case to address invalid input or does that feel like it's outside the scope of this exercise?

IanWhitney:
I have noticed the same. 

Currently there is not a standard, cross-track test suite for Nucleotide Count. It would probably be good to  create one, so that all tracks could handle this edge case.

IanWhitney:
(Oh, also, thanks for adding the issue! Much appreciated)

petertseng:
List of tracks that implement this exercise is at http://x.exercism.io/problems/nucleotide-count. I have not checked all of them but I know that at least [Haskell expects that invalid nucleotides result in an error](https://github.com/exercism/xhaskell/blob/master/exercises/nucleotide-count/nucleotide-count_test.hs) ), would be interesting to see how many other tracks expect the same

petertseng:
Info on what other tracks are doing:

xgo now tests for sequences containing invalid nucleotides and expects that the function under test signals an error for them, as of https://github.com/exercism/xgo/pull/413

PR #155, opened 2016-07-03
parallel-letter-frequency, tournament, word-count: Use `Entry#or_insert`

petertseng:
These shorten the functions by removing the need to match on the `Entry` cases.

This seems good to make the examples more idiomatic.

Note the exercises that still match on `Entry` cases but were not changed by this PR:

* ~~grade-school: I judged that pushing an element onto the vector then sorting it was sufficiently different from creating a vector with a single element. I can be convinced otherwise.~~
* nucleotide-count: As we discuss in #149, invalid nucleotides should maybe produce errors, so the decision resulting from that discussion may mean using `or_insert` doesn't make sense.

### DIFF
--- a/exercises/grade-school/example.rs
+++ b/exercises/grade-school/example.rs
@@ -1,5 +1,4 @@
 use std::collections::HashMap;
-use std::collections::hash_map::Entry;
 
 pub struct School {
     grades: HashMap<u32, Vec<String>>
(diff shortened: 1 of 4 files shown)


petertseng:
travis didn't pick this up for testing. kicking it.

petertseng:
I also decided to make the change for `grade-school`.

PR #156, opened 2016-07-03
allergies: replace a match -> bool with a comparison

petertseng:
It feels a bit odd to match when the result is going to be a boolean,
and doesn't feel like an idiomatic use of matching.

### DIFF
--- a/exercises/allergies/example.rs
+++ b/exercises/allergies/example.rs
@@ -11,7 +11,7 @@ impl Allergies {
     pub fn is_allergic_to(&self, allergen: &Allergen) -> bool {
         let allergens = Allergies::allergens();
         let index = allergens.iter().position(|x: &Allergen| x == allergen).unwrap();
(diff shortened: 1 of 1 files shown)


exercises/allergies/example.rs
```diff
@@ -11,7 +11,7 @@ impl Allergies {
     pub fn is_allergic_to(&self, allergen: &Allergen) -> bool {
         let allergens = Allergies::allergens();
         let index = allergens.iter().position(|x: &Allergen| x == allergen).unwrap();
-        match self.0 & 1 << index { 0 => false, _ => true }
+        self.0 & 1 << index != 0
```
petertseng: potential concern that the operator precedence will be unclear. we can make it super clear with:

```rust
(self.0 & (1 << index)) != 0
```

if that is preferred

IanWhitney:
I like the additional parens.

petertseng:
parens it is.

PR #158, opened 2016-07-03
dominoes, forth: Use while let

petertseng:
Instead of `loop` that simply `break` in the `None` case, it's better to
use `while let`.


IanWhitney:
Ok

PR #159, opened 2016-07-03
grade-school: Use `cloned` instead of `map(|k| k.clone())`

petertseng:
They're equivalent, but I figure if there's already a function that does
the clone, might as well use it instead of `map` and making someone read
the body of the map.


IanWhitney:
Ok.

PR #160, opened 2016-07-03
hello-world: No ref in pattern

petertseng:
I couldn't figure out why it was necessary, and I actually didn't know
about it off the top of my head (says something about how much Rust I
know). So I tried to remove it to see what compiler error would happen
if it wasn't there. Nothing, it worked.

Well, if it's not necessary, perhaps we should remove it then... simpler
solution for our first exercise should be good, right?

### DIFF
--- a/exercises/hello-world/example.rs
+++ b/exercises/hello-world/example.rs
@@ -1,6 +1,6 @@
 pub fn hello(name: Option<&str>) -> String {
     match name {
-        Some(ref n) => format!("Hello, {}!", n),
+        Some(n) => format!("Hello, {}!", n),
         None => "Hello, World!".to_string(),
     }
 }


petertseng:
Of course I would like to know if there actually is an observable difference - and/or whether using the `ref` is recommended for some reason or other. If it's recommended, then we should keep it to be a better example, but if it's not then I want to keep it simple.

petertseng:
http://rustbyexample.com/scope/borrow/ref.html

petertseng:
https://doc.rust-lang.org/book/patterns.html#ref-and-ref-mut

petertseng:
I guess the part that confuses me is that the code works both with and without the `ref`. But are they doing different things under the hood?

IanWhitney:
A surprisingly deep question for Hello, World!

I exported the MIR representation of [the ref version](https://gist.github.com/f4a6efcc3436e5075b9ddcc6fa248cc9) and [non-ref](https://gist.github.com/da99514fffaed246713e4e877fbe920d)


[ref_and_no_ref_mir.zip](https://github.com/exercism/xrust/files/346284/ref_and_no_ref_mir.zip)

The relevant diff, I think, is here:

screenshot 2016-07-04 07 46 42

But I'm probably missing some of the details.

petertseng:
Okay, I believe I have gained an understanding of this. I present a gist https://gist.github.com/anonymous/53e77f95fdf28bd221b731efcc8f1f9d that can be reviewed on the playground at https://play.rust-lang.org/?gist=53e77f95fdf28bd221b731efcc8f1f9d&version=stable . The big difference that I see is that without `ref`, you will move the value, which is important for types that aren't `Copy` (notice that the int is `Copy` but the `Hi` isn't)

If we use `ref n`, the type of `n` would be `&&str`, whereas if we don't it will be `&str`, because `name` is an `Option<&str>`

I don't see a need to take a reference of a reference, so I recommend removal.

petertseng:
Commit message updated with what I learned

IanWhitney:
I commend your dedication to Hello, World and to leaving useful commit messages!

PR #161, opened 2016-07-03
wordy: Immediately return in `operand`

petertseng:
The previous code just added a `: isize` type, but this is already
inferrable from the function signature, so it can just be returned
immediately.


IanWhitney:
ok. Someone pointed that out to me in a nitpick as well.

PR #162, opened 2016-07-03
Iterator methods

petertseng:
In general, this is taking advantage of Rust's iterator methods (map, filter, collect) where possible, which I believe to be more idiomatic.

Another potential place not covered here:

* `etl` could use `flat_map`. I wanted to evaluate whether that would make things too complicated. Will investigate and send a PR.


IanWhitney:
Ok

PR #163, opened 2016-07-03
etl: use flat_map

petertseng:
Instead of creating a `BTreeMap` and adding to it, we can just use
`flat_map` on the input and `collect` the result.


petertseng:
question I have about this one is: "is it too confusing? Too clever? Not understandable?"

Not sure how far into iterator methods we should go with these examples

IanWhitney:
I think it's fine. Using a move closure is definitely more conceptually abstract than the old approach, but I don't think it's cleverness for cleverness sake.

petertseng:
I'll throw a link to https://doc.rust-lang.org/book/closures.html#move-closures in the commit message in case anyone is curious and reads that. Looks like the only other place it's used is parallel-letter-frequency,

PR #164, opened 2016-07-04
anagram: Use collect

petertseng:
Instead of creating a String manually, pushing into it in a loop, and
then returning it, we can just `collect` on an iterator to turn the
vector into a String.


IanWhitney:
great.

PR #165, opened 2016-07-04
rectangles: Add new large test

petertseng:
This was added to x-common in
https://github.com/exercism/x-common/pull/288

It originated from the Lua track. Neither this track nor the Python
track had it prior.

### DIFF
--- a/exercises/rectangles/tests/rectangles.rs
+++ b/exercises/rectangles/tests/rectangles.rs
@@ -97,3 +97,19 @@ fn test_not_so_complicated() {
         ];
     assert_eq!(2, count(&lines))
 }
+
+#[test]
+#[ignore]
+fn test_large_input_with_many_rectangles() {
+    let lines = vec![
+        "+---+--+----+",
+        "|   +--+----+",
+        "+---+--+    |",
+        "|   +--+----+",
+        "+---+--+--+-+",
+        "+---+--+--+-+",
+        "+------+  | |",
+        "          +-+"
+        ];
+    assert_eq!(60, count(&lines))
+}


IanWhitney:
Does this hit an edge case that the previous tests did not? I haven't done this problem yet, so I'm not familiar with it.

petertseng:
The main thing I see it providing is just checking that the solution isn't terribly inefficient

petertseng:
> The main thing I see it providing is just checking that the solution isn't terribly inefficient

Then again, I'm not sure how good of a job it does at that either, it's not like there's a timeout built in (other than the student's patience). I may intentionally code an inefficient solution to see how it performs. Probably something like scan every pair of points regardless of whether they're corners

IanWhitney:
How do you encode that in test suite, though? I'm always reluctant to test for performance in these sorts of things. I like tests that prove the code works, and performance is too subjective to fit into unit tests.

Maybe performance tests like this one (and the large `vec` ones in `sublist`, etc.) should be made optional? With an explanation of what they are there to prove. Then students that want to really focus on performance can do so.

petertseng:
My report is that even with an implementation that is O(h^2w^2) (h for height and w for width), `cargo test` is still passing in less than half a second, so even this test is not showing any performance problems. Maybe in interpreted languages... certainly not here. Would have to ask Lua track if that is a concern when adding that test.

Maybe it would reveal problems for solutions that don't cache corner/edge positions (indexing into a string is not constant time!), but I don't seem to have the time right now to make my solution even more inefficient to test the limits of my own patience.

The comment about test intent is good - all other cases in https://github.com/exercism/x-common/pull/288 are named to give an idea of their intent, but this one's just named "large test", leaving all to wonder "OK so it's large. So what?"

petertseng:
> With an explanation of what they are there to prove.

Now would be a good time to look at https://github.com/exercism/x-common/pull/289 and see whether it is explained sanely there, and think about how that will translate to an auto-generated test file when this track auto-generates tests. Is simply putting it in the test description sufficient? Should there be a student-visible comment included with the case? etc.

IanWhitney:
https://github.com/exercism/x-common/pull/289#issuecomment-233128518

Issue #166, opened 2016-07-06
Can `Anagram` be changed to either require or prevent Lifetimes?


IanWhitney:
In #126, Anagram was the main reason we restructured our problem track. In that restructuring I grouped all of the "Lifeteme" problems together and moved them to be much later in the track.

But [you can pass the Anagram test suite without lifetimes](http://exercism.io/submissions/ff6c6d32c9184214a769b12ce28faec4). Which, if I'd thought about it for more than 2 seconds, I may have realized.

I'm wondering if there's a way to change the Anagram test suite to either require or prevent a solution that uses lifetimes. And, if so, would we want to.

#### Pros
- Lets us position the problem more exactly
- Lets us use the problem as a way to teach Lifetimes (assuming we force lifetimes)

#### Cons
- Forces an implementation

If this is possible, then I'm also wondering if we could provide 2 test suites, one with Lifetimes & one without, so that students could see both in action.

kotp:
The possibility to include more than just the test, readme and example file is "built in" because only the example file is not transferred on fetch.

If one test is called `_with_lifetimes.ext` I think that may be as simple as possible.  The `README` for the track's exercise could explain what that is for and how to use it.

The Ruby track is talking about how to use an additional testing framework, and this may be one way to attack that problem.

IanWhitney:
Good to know that it's possible on the Exercism side. I'm just not sure it's possible to do this with the Rust language itself.

petertseng:
Yeah it's probably possible to write tests in such a way that a lifetime annotation is required. I can't think of how to do this right now, but I feel it should be possible given that there are many times when code won't compile without a lifetime annotation. On the other hand, maybe I'm wrong about that, since anything we would do that might eem to require lifetime annotations might be worked around by simply making a copy and taking ownership of the copy. I will need more time to reason about that.

On the other hand, I'm not sure it's possible to make it so that code compile without lifetime annotations and fails to compile with. but I'm also not sure what value that would have anyway.

If we are able to devise something that requires solutions to have a lifetime annotation, it would probably be good to have some exercise dedicated to showing that. Whether anagram has to be the exercise to demonstrate lifetimes remains to be seen. It certainy is a strong candidate since takes in some strings and returns some of the input strings (again, though, someone could just make a copy). That also means https://github.com/exercism/x-common/blob/master/strain.md is a possibility.

jonasbb:
[...]

PR #167, opened 2016-07-10
Add file that documents the thought behind problem order

IanWhitney:
The downside of this file is that we have to keep in in sync with
config.json. I'm not sure of a way to leverage travis.yml to check that for
us. Though maybe there is an approach.

Even with that drawback, I think having this file is useful. The PR that
kicked of our problem ordering process is already out of date. And
having something in the repo (instead of a gist or a blog post) makes it
easier to find.

### DIFF
--- a/problems.md
+++ b/problems.md
@@ -1,0 +1,75 @@
+# Rust Track Problem Order
+
+The actual source of truth of problem order is [config.json](config.json), but this file documents our reasoning behind the problem order in that file.
+
+## Background
+
+- https://github.com/exercism/xrust/issues/126
+- https://github.com/exercism/xrust/issues/127
+- http://designisrefactoring.com/2016/07/09/exercism-shouldnt-make-you-cry/
+
+# The Problems, By Section
+
+## Introduction
+
+The first section contains the sort of stuff you expect when learning any programming languages: conditionals, booleans, looping and some higher-order functions.
+
+problem | topics
+----- | -----
+hello-world |  Some/None. println!
+gigasecond |  crates, math
+leap |  math, booleans, conditionals
+raindrops |  case (or `format`). Mutable string
+bob |  chars, string functions
+beer-song |  case, string concatenation, vector (optional), loop
+difference-of-squares |  fold & map
+
+## Getting Rusty
+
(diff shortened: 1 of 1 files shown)


petertseng:
Seems like a good idea, so I'm supportive. :+1: 

you switched tournament/robot-simulator and sublist/allergies.

I'll try to help out with the dominoes description

problems.md
```diff
@@ -0,0 +1,74 @@
...
+robot-name |  struct, slices, randomness, lifetimes, self mut
+
+## Putting it all Together
+
+These problems don’t necessarily require additional Rust knowledge, but they do require complex solutions.
+
+problem | topics
+----- | -----
+minesweeper |  managing board state, Vec, heavy logic
+dominoes |  The logic here isn't bad, I just don't quite understand the problem
```
petertseng: I'd say "graph theory, searching". Most solutions to this exercise either take the graph-theoretical approach (find a Euclidean cycle) or search solutions until finding one that works. I'm not sure how nuanced you want to be here - one could reasonably solve this problem only knowing one of these two topics, I'm not sure if there should be an "OR" in there somewhere.

problems.md
```diff
@@ -0,0 +1,74 @@
...
+parallel-letter-frequency | multi-threading
+rectangles |  Enum, structs, traits, Lifetimes
```
petertseng: similarly to #166, I need to think about this a bit. My solution to this problem has no lifetime annotations. Since the function under test is `count(lines: &Vec<&str>) -> usize` my feeling is that lifetimes should not be necesary.

IanWhitney: Since I haven't done Rectangles, I based this list on the example which [does use lifetimes](https://github.com/exercism/xrust/blob/master/exercises/rectangles/example.rs#L66)

For these Putting It All Together problems I'm less concerned about Rust features, though. The order of these problems doesn't really matter, as there's no real progression. And by this point the student should be familiar with Rust's feature set.

IanWhitney: So if there's a quick summary of the type of problems you'll have to solve while working on Rectangles, that will do fine.

PR #169, opened 2016-07-12
Hamming: Make error checking identical to other tests

jonasbb:
I noticed this after @petertseng mentioned the error handling in [#147](https://github.com/exercism/xrust/pull/147#issuecomment-232076551).
The Hamming exercise performs error checking with comparison to `Err(.)`. Most tests (Nucleotide Codons, Queen Attack, Wordy) only check with `is_err()`. The exceptions are Circular Buffer and Forth, but they use more complex error reporting with different conditions and provide an `Error` enum with the correct return values.

Because there is only one error conditions I suggest to remove the fixed value and only check with `is_err()`.

### DIFF
--- a/exercises/hamming/tests/hamming.rs
+++ b/exercises/hamming/tests/hamming.rs
@@ -32,11 +32,11 @@ fn test_larger_distance() {
 #[test]
 #[ignore]
 fn test_first_string_is_longer() {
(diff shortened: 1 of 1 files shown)


IanWhitney:
Thanks. I like it.

PR #170, opened 2016-07-12
Implement All Your Bases

jonasbb:
Problem description and test data from x-common repository.

Design decisions:
 * Empty list is equal to 0
 * No leading 0s
 * Invalid input must result in Err(.) but not specified which one

[Problem Description](https://github.com/exercism/x-common/blob/master/all-your-base.md)
[Test data](https://github.com/exercism/x-common/blob/master/all-your-base.json)

This exercise should be the replacement ([1]) for all the individual base conversion exercises like Binary or Octal. Those should probably be marked in the config.json as not to implement.

Can we add an individual Markdown which describes these design decisions as well as the data type for the digits? Digits are just `i64` (or any other number type) instead of something like `char` because there is a test with base 97.

[1]: https://github.com/exercism/x-common/issues/279


IanWhitney:
> Can we add an individual Markdown which describes these design decisions as well as the data type for the digits? Digits are just i64 (or any other number type) instead of something like char because there is a test with base 97.

You could put that in a markdown file, or you could put it in the stub `src/lib.rs` file.

jonasbb:
Sure, that's what I wanted to do. Does the markdown file need a special name?

exercises/all-your-bases/tests/all-your-bases.rs
```diff
@@ -0,0 +1,213 @@
...
+fn second_base_is_one() {
+    let input_base = 2;
+    let input_digits = &[1, 0, 1, 0, 1, 0];
+    let output_base = 1;
+    assert!(ayb::convert(input_digits, input_base, output_base).is_err());
+}
+
+#[test]
+#[ignore]
+fn first_base_is_zero() {
```
IanWhitney: I know that these test names come from the canonical set,  but I think "first" should be replaced with "input", and "second" should be replaced with "output", since that reflects the language used in the tests.

IanWhitney:
> Sure, that's what I want to do. Does the markdown file need a special name?

I don't think so. Looking at the exercises that have extra files (Tournament, Hello World), they naming should not matter.

IanWhitney:
Thoughts on where we place this in the track?

jonasbb:
As I understood the intention behind this exercise it should replace all existing conversion exercises. The Rust track already has the hexadecimal exercise. I think it should deprecate hexadecimal and replace it with all your base. Looking at the sample solutions, all your base is a bit more complex, mostly because both bases are now variables. However, you need the same tools to solve it.

Topics (as mentioned in your pull request)

| Hexadecimal | All your base |
| ------ | -----  |
| Option | Result |
| zip (but can also use enumerate) | enumerate (but can also use zip) |
| fold | fold |
| chars | (not needed) |
| map | map |

I will fix the placement of hexadecimal in my pull request.

petertseng:
is exercism going to get confused by the fact that this here is called `all-your-bases`, but in `x-common` it is called `all-your-base`?

petertseng:
Philosophically speaking... if the function is expected to treat negative bases as invalid, why should it still take `i32` rather than `u32`? What if we let type system ensure for us that we never have a negative base?

(Other than the fact that the x-common test cases have a negative base. We can just ignore those tests cases if we use unsigned inputs)

There are times when I chose to use a signed integer and that was because even though any individual value would never go negative, sometimes I subtract two values and that might go negative. There may be similar reasons here.

jonasbb:
@petertseng Thanks for the note about the typo. I didn't intent to change the name. I shouldn't do this late at night.

I don't know if it is better to use signed or unsigned integers. I struggled with the decision while writing the tests, but then decided it would be better to keep them as close as possible to the shared test data.

Looking at my sample solution I find only one half of the checks could be removed, namely `num < 0 ||` while testing if all digits are in the valid range. The test for the bases would be unchanged. A solution would be to change the mappings of both bases so that the value 0 would correspond to base 2, because it is the first valid base. But this might be unnecessary complex, but being safe.

petertseng:
One more small thing to fix: the test file is still called `all-your-bases.rs`

It's true that any solution would still have to check the bases. It however seems too unclear to subtract two from every base, as it would be confusing. So even if this exercise moved to use unigned integers, I wouldn't touch the values.

I don't feel strongly enough to say it must be either one way or the other, so I leave that to you and others who might feel strongly.

jonasbb:
It is probably unnecessary to have signed numbers so I changed it. I updated all the description texts and renamed the test file.

exercises/all-your-base/tests/all-your-base.rs
```diff
@@ -0,0 +1,177 @@
+extern crate allyourbases as ayb;
```
petertseng: one last allyourbase here

exercises/all-your-base/Cargo.lock
```diff
@@ -0,0 +1,4 @@
+[root]
+name = "allyourbases"
```
petertseng: base

exercises/all-your-base/Cargo.toml
```diff
@@ -0,0 +1,3 @@
+[package]
+name = "allyourbases"
```
petertseng: base

petertseng:
OK, if you have decided on unsigned that seems reasonable, I'm sorry I missed a few lingering bases, after that we should be good to go :+1:

IanWhitney:
Looks like the one thing preventing a merge are the three remaining "bases" that @petertseng noted.  Once those are fixed, we'll merge.

petertseng:
It has been a month, I have merged this as https://github.com/petertseng/xrust/commit/0c10aa3f5b062d2aae4cc1e396281968a8fa70ee - thank you @jonasbb  for the contribution.

petertseng:
Oh wait, hold on, need to put that in the right repo.

petertseng:
Merged as https://github.com/exercism/xrust/commit/9c527e465fd95eef5b4243f123acec76892f212f, thank you!

petertseng:
To examine changes I made, run:

`diff -u <(git diff 6684722ec42e22ca67b10ef93891a0a9728991b8~9 6684722ec42e22ca67b10ef93891a0a9728991b8) <(git diff 9c527e465fd95eef5b4243f123acec76892f212f~ 9c527e465fd95eef5b4243f123acec76892f212f)`

petertseng:
So that more are able to see the changes I made,

I will paste the output of that command for me since it might not be the case that everyone has 6684722ec42e22ca67b10ef93891a0a9728991b8. Note that this is a diff-of-diff, so you mostly care about the prefixed with two symbols either plus or minus:

```diff
$ diff -u <(git diff 6684722ec42e22ca67b10ef93891a0a9728991b8~9 6684722ec42e22ca67b10ef93891a0a9728991b8) <(git diff 9c527e465fd95eef5b4243f123acec76892f212f~ 9c527e465fd95eef5b4243f123acec76892f212f)
--- /proc/self/fd/11	2016-09-05 00:24:21.607889398 -0700
+++ /proc/self/fd/12	2016-09-05 00:24:21.607889398 -0700
@@ -1,5 +1,5 @@
 diff --git a/config.json b/config.json
-index 328573d..cee4327 100644
+index 0775fa7..9ca5b8c 100644
 --- a/config.json
 +++ b/config.json
 @@ -20,7 +20,7 @@
@@ -11,7 +11,7 @@
      "grade-school",
      "tournament",
      "robot-simulator",
-@@ -41,7 +41,7 @@
+@@ -45,7 +45,7 @@
      "circular-buffer"
    ],
    "deprecated": [
@@ -20,7 +20,7 @@
    ],
    "ignored": [
      "_test",
-@@ -49,6 +49,8 @@
+@@ -53,6 +53,8 @@
      "img"
    ],
    "foregone": [
@@ -32,22 +32,22 @@
  }
 diff --git a/exercises/all-your-base/Cargo.lock b/exercises/all-your-base/Cargo.lock
 new file mode 100644
-index 0000000..245d36d
+index 0000000..92cc4a2
 --- /dev/null
 +++ b/exercises/all-your-base/Cargo.lock
 @@ -0,0 +1,4 @@
 +[root]
-+name = "allyourbases"
++name = "allyourbase"
 +version = "0.0.0"
 +
 diff --git a/exercises/all-your-base/Cargo.toml b/exercises/all-your-base/Cargo.toml
 new file mode 100644
-index 0000000..b67d361
+index 0000000..158e712
 --- /dev/null
 +++ b/exercises/all-your-base/Cargo.toml
 @@ -0,0 +1,3 @@
 +[package]
-+name = "allyourbases"
++name = "allyourbase"
 +version = "0.0.0"
 diff --git a/exercises/all-your-base/example.rs b/exercises/all-your-base/example.rs
 new file mode 100644
@@ -159,11 +159,11 @@
 +}
 diff --git a/exercises/all-your-base/tests/all-your-base.rs b/exercises/all-your-base/tests/all-your-base.rs
 new file mode 100644
-index 0000000..fdf2ef5
+index 0000000..9f49cb5
 --- /dev/null
 +++ b/exercises/all-your-base/tests/all-your-base.rs
 @@ -0,0 +1,177 @@
-+extern crate allyourbases as ayb;
++extern crate allyourbase as ayb;
 +
 +#[test]
 +fn single_bit_one_to_decimal() {
@@ -340,3 +340,16 @@
 +    let output_base = 0;
 +    assert!(ayb::convert(input_digits, input_base, output_base).is_err());
 +}
+diff --git a/problems.md b/problems.md
+index 2e39384..e09365a 100644
+--- a/problems.md
++++ b/problems.md
+@@ -39,7 +39,7 @@ etl |  btree
+ sieve |  vector, map, while let (optional)
+ rna-transcription |  match, struct, str vs string
+ roman-numerals |  mutable, results, loops, struct, traits
+-hexadecimal |  Option, zip/fold/chars, map
++all-your-base |  Result, enumerate, fold, map
+ grade-school |  struct, entry api, Vec, Option
+ tournament |  enum, sorting, hashmap, structs
+ robot-simulator | Immutability, enum
```

PR #171, opened 2016-07-13
Add test to capture invalid range-based solutions

### LINKED ISSUE #168: Add pangram test case to cover exclusive ranges
I've seen this implementation a few times now

http://exercism.io/submissions/3e04544e41414e4d9f2c6abee5b79134

If we add a test that does not include 'z' in the sentence then we'll catch this case.

IanWhitney:
Fixes #168


petertseng:
:+1:

PR #172, opened 2016-07-16
Implement ocr-numbers

IanWhitney:
This is not ready for merge. Just getting it up here so that we can start the discussion/bikeshedding.

I've created the tests and the stub file. No working implementation yet. I'm really not sure what implementations people will use for this one. Will be interesting to see.

The one goofy thing I've done here is turn off rustfmt for each test. I don't see a way to turn it off at the file or directory level. I've confirmed that currently a [directory-level ignore (https://github.com/rust-lang-nursery/rustfmt/issues/705) is not supported. But maybe there's a file wide directive that I'm missing?

### DIFF
--- a/exercises/ocr-numbers/example.rs
+++ b/exercises/ocr-numbers/example.rs
@@ -1,0 +1,67 @@
+use std::collections::BTreeMap;
+use std::str::Lines;
+
+fn lines_valid(lines: &Lines) -> bool {
+    lines.clone().count() % 4 == 0 && lines.clone().all(|line| line.chars().count() % 3 == 0)
+}
+
+pub fn convert(input: &str) -> Result<String, ()> {
+    if !lines_valid(&input.lines()) {
+        return Err(());
+    }
+
+    let y = input.lines().collect::<Vec<_>>();
+
+    let mut converted_lines = vec![];
+
+    for char_lines in y.chunks(4) {
+        let mut unparsed_characters = BTreeMap::new();
+
+        for line in char_lines {
+            let line_chars = line.chars().collect::<Vec<_>>();
+
+            for (char_number, char_chunk) in line_chars.chunks(3).enumerate() {
+                let char_chars = unparsed_characters.entry(char_number).or_insert(vec![]);
+                for c in char_chunk {
+                    char_chars.push(*c);
+                }
+            }
+        }
+
+        let mut parsed_characters = String::new();
+
+        for (_, v) in unparsed_characters {
+            parsed_characters.push(convert_character(&v));
+        }
+
+        converted_lines.push(parsed_characters.to_string());
+    }
+
+    Ok(converted_lines.join(","))
+}
+
+fn convert_character(input: &Vec<char>) -> char {
+    if &input[..] == [' ', '_', ' ', '|', ' ', '|', '|', '_', '|', ' ', ' ', ' '] {
+        '0'
+    } else if &input[..] == [' ', ' ', ' ', ' ', ' ', '|', ' ', ' ', '|', ' ', ' ', ' '] {
+        '1'
+    } else if &input[..] == [' ', '_', ' ', ' ', '_', '|', '|', '_', ' ', ' ', ' ', ' '] {
+        '2'
+    } else if &input[..] == [' ', '_', ' ', ' ', '_', '|', ' ', '_', '|', ' ', ' ', ' '] {
+        '3'
+    } else if &input[..] == [' ', ' ', ' ', '|', '_', '|', ' ', ' ', '|', ' ', ' ', ' '] {
+        '4'
+    } else if &input[..] == [' ', '_', ' ', '|', '_', ' ', ' ', '_', '|', ' ', ' ', ' '] {
+        '5'
+    } else if &input[..] == [' ', '_', ' ', '|', '_', ' ', '|', '_', '|', ' ', ' ', ' '] {
+        '6'
+    } else if &input[..] == [' ', '_', ' ', ' ', ' ', '|', ' ', ' ', '|', ' ', ' ', ' '] {
+        '7'
+    } else if &input[..] == [' ', '_', ' ', '|', '_', '|', '|', '_', '|', ' ', ' ', ' '] {
+        '8'
+    } else if &input[..] == [' ', '_', ' ', '|', '_', '|', ' ', '_', '|', ' ', ' ', ' '] {
+        '9'
+    } else {
+        '?'
+    }
+}
--- a/exercises/ocr-numbers/tests/ocr-numbers.rs
+++ b/exercises/ocr-numbers/tests/ocr-numbers.rs
@@ -1,0 +1,211 @@
+extern crate ocr_numbers as ocr;
+
+#[test]
+#[cfg_attr(rustfmt, rustfmt_skip)]
+fn input_with_lines_not_multiple_of_four_is_error() {
+    let input = " _ \n".to_string() +
+                "| |\n" +
+                "   ";
+
+    assert!(ocr::convert(&input).is_err());
+}
+
+#[test]
+#[ignore]
+#[cfg_attr(rustfmt, rustfmt_skip)]
+fn input_with_columns_not_multiple_of_three_is_error() {
+    let input = "    \n".to_string() +
+                "   |\n" +
+                "   |\n" +
+                "    ";
+
+    assert!(ocr::convert(&input).is_err());
+}
+
+
+#[test]
+#[ignore]
+#[cfg_attr(rustfmt, rustfmt_skip)]
+fn unrecognized_chararcters_return_question_mark() {
+    let input = "   \n".to_string() +
+                "  _\n" +
+                "  |\n" +
+                "   ";
+
+    assert_eq!("?", ocr::convert(&input).unwrap());
+}
+
+#[test]
+#[ignore]
+#[cfg_attr(rustfmt, rustfmt_skip)]
+fn recognizes_0() {
+    let input = " _ \n".to_string() +
+                "| |\n" +
+                "|_|\n" +
+                "   ";
+
+    assert_eq!("0", ocr::convert(&input).unwrap());
+}
+
+#[test]
+#[ignore]
+#[cfg_attr(rustfmt, rustfmt_skip)]
+fn recognizes_1() {
+    let input = "   \n".to_string() +
+                "  |\n" +
+                "  |\n" +
+                "   ";
+
+    assert_eq!("1", ocr::convert(&input).unwrap());
+}
+
+#[test]
+#[ignore]
+#[cfg_attr(rustfmt, rustfmt_skip)]
+fn recognizes_2() {
+    let input = " _ \n".to_string() +
+                " _|\n" +
+                "|_ \n" +
+                "   ";
+
+    assert_eq!("2", ocr::convert(&input).unwrap());
+}
+
+#[test]
+#[ignore]
+#[cfg_attr(rustfmt, rustfmt_skip)]
+fn recognizes_3() {
+    let input = " _ \n".to_string() +
+                " _|\n" +
+                " _|\n" +
+                "   ";
+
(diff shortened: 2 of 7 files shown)


exercises/ocr-numbers/tests/ocr-numbers.rs
```diff
@@ -0,0 +1,212 @@
+extern crate ocr_numbers as ocr;
+
+// use ocr_numbers::*;
```
IanWhitney: This should be removed.

IanWhitney:
Any ideas on what a common approach to this will be? Knowing that will help figure out where this problem should be placed.

I'm thinking of doing the example with either an approach like [Swift](https://github.com/exercism/xswift/blob/master/exercises/ocr-numbers/OcrNumbersExample.swift) or hashing each input and comparing it a list of known hashes (which is more or less the same, I guess).

petertseng:
> Any ideas on what a common approach to this will be?

My general thought when looking over Ruby implementations was that an implementation similar to the Swift implementation was the most common. Mine was the only solution I saw that counted segments and determined the character based on the on/off segment locations.

exercises/ocr-numbers/tests/ocr-numbers.rs
```diff
@@ -0,0 +1,212 @@
...
+                "  |\n" +
+                "   ";
+
+    assert_eq!("?", ocr::convert(input).unwrap());
+}
+
+#[test]
+#[ignore]
+#[cfg_attr(rustfmt, rustfmt_skip)]
+fn input_with_lines_not_multiple_of_four_panics() {
```
petertseng: well these aren't really panicking

petertseng: well these aren't really panicking

Since I assume we *do* want to keep the current behavior of error (rather than panic) I assume that means just to change the test name

exercises/ocr-numbers/tests/ocr-numbers.rs
```diff
@@ -0,0 +1,212 @@
...
+                "                              ";
+
+    assert_eq!("1234567890", ocr::convert(input).unwrap());
+}
+
+#[test]
+#[ignore]
+#[cfg_attr(rustfmt, rustfmt_skip)]
+fn numbers_across_multiple_lines_are_joined_by_commas() {
+    let input = "    _  _ ".to_string() +
```
jonasbb: You are missing the terminating newlines here.

IanWhitney:
Well, matching against slices apparently requires a gated feature, so I probably won't use that.

https://doc.rust-lang.org/book/slice-patterns.html

IanWhitney:
I've been working on a production issue the past few weeks. Sorry that this one has stagnated.

I'm wondering what the Rusty way is for breaking the input into 4-character chunks.

For example, when I get this input:

```rust
 "       _     _        _  _ \n".to_string() +
"  |  || |  || |  |  || || |\n" +
"  |  ||_|  ||_|  |  ||_||_|\n" +
"                           ";
```

I'll send each 'character' through the conversion function, right? To do that I need to get the first 4 characters of each line.

But indexing strings by characters is not something that Rust supports. So I can't just say "Get me the first 4 characters of this string"

So I'm not quite sure how to proceed. `nth()`? 

Hopefully someone has a good idea.

petertseng:
Untested ideas:

* [as_bytes](https://doc.rust-lang.org/std/string/struct.String.html#method.as_bytes) if willing to make an assumption that each character is one byte.
* [chars](https://doc.rust-lang.org/std/string/struct.String.html#method.chars) to get an iterator over the characters, then do something with the iterator

IanWhitney:
It's pretty gross, but it currently passes everything but the last test. So it's a start.

I spent some time looking at the examples in other tracks and most of them end up doing something similar, so I guess I'm on the right track.

exercises/ocr-numbers/example.rs
```diff
@@ -1 +1,64 @@
+// The code below is a stub. Just enough to satisfy the compiler.
```
petertseng: can probably not have this line in the example

IanWhitney:
Ok. Tests passing. Finally. Not sure why this one was so hard for me. I don't think I quite get Rust iterators yet.

Starting to think about problem placement. Obviously this one was tough for me, but that might just be my dumb brain. My dumb brain aside, solutions to this will probably use

- iterators & looping
- [`lines`](https://doc.rust-lang.org/std/primitive.str.html#method.lines)
- [`chunks`](https://doc.rust-lang.org/std/primitive.slice.html#method.chunks) (and therefore slices)

The problem of recognizing a single character is pretty straightforward. My head began to hurt when I had to do multiple-character strings. The problem there is that you need to get the 3 characters per line, while retaining which 'character' they belong to. Rust makes this slightly harder because you can't index strings, or simply `chunk` through them.

Because of that, I think this exercise should go later in the track. I'm thinking [after wordy](https://github.com/exercism/xrust/commit/a2e70001ba6f85449663c2ca18496ed1e213f95c#diff-da4a0b4ace8d0d7d90610be217d49347R52). Or maybe even [after anagram](https://github.com/exercism/xrust/commit/a2e70001ba6f85449663c2ca18496ed1e213f95c#diff-da4a0b4ace8d0d7d90610be217d49347R60) since the non-indexing of strings is certainly an aspect of Rust that I find new.

I also want to re-order the tests so that they have a smoother curve:

- Errors
- Individual numbers 0-9
- Garbled individual numbers
- Multi-character, single line
- Multi-line, multi-character

The current order puts the test of a multi-character binary too early. Once this test is passed, nearly all of the remaining tests (except for multi-line) are trivial.

IanWhitney:
Now that I've gotten this working, I think the API is wrong. This exercise is another one that is well suited for the [`From` trait](https://doc.rust-lang.org/std/convert/trait.From.html), right? 

Like [Roman Numerals](https://github.com/exercism/xrust/blob/master/exercises/roman-numerals/tests/roman-numerals.rs) I'm thinking the tests should look like

```rust
 let input = " _ \n".to_string() +
             "| |\n" +
             "|_|\n" +
             "   ";

    assert_eq!("0", Ocr::from(input).unwrap().to_string());
```

petertseng:
Since the conversion can fail, it really makes me wish the [`TryFrom` trait](https://doc.rust-lang.org/std/convert/trait.TryFrom.html) were stable so that it could be used instead!

Someone may rightly point out that converting a roman numberal could have failed too (> 3999, < 0), but I guess the difference is the tests never tested a failing roman numeral conversion, whereas we will be testing failed OCR reads.

It's not certain to me that we can use `From` because of this consideration. Maybe if implementing `From` for `Result<something, something>` - not sure how common this is.

petertseng:
I have discovered when attempting this exercise myself:

If we keep the `convert` function, I would suggest changing the `convert(input)` into `convert(&input)`, since I don't think it needs to take ownership. And this allows me to write `convert(input: &str)` instead of the current `convert(input: String)`

IanWhitney:
> Since the conversion can fail, it really makes me wish the TryFrom trait were stable so that it could be used instead!

Ah, right, because `From` requires a return of `Self`. I missed that. Thanks.

Converting to a borrow also makes sense.

petertseng:
I finished implementing. The harder part is definitely splitting up the characters, iterating over the lines in parallel. I imagine that minesweeper also would have needed string indexing (though I guess there is a possible solution that doesn't). Placement immediately before or immediately after minesweeper could be appropriate.

petertseng:
Until this PR is out I can't submit my solution to exercism.io so I'll put it on https://gist.github.com/petertseng/7ec8c8aec510e3685613d8fb36b34d24 for now (note for anyone reading in the future, I'm deleting the gist once I can submit it).

IanWhitney:
Ok. Updated and rebased.

exercises/ocr-numbers/example.rs
```diff
@@ -0,0 +1,67 @@
+use std::collections::HashMap;
+use std::str::Lines;
+
+fn lines_valid(l: &mut Lines) -> bool {
```
petertseng: explain why the mut? what happens without?

exercises/ocr-numbers/example.rs
```diff
@@ -0,0 +1,67 @@
...
+                let char_chars = unparsed_characters.entry(char_number).or_insert(vec![]);
+                for c in char_chunk {
+                    char_chars.push(c.clone());
+                }
+            }
+        }
+
+        let mut parsed_characters = String::new();
+
+        for k in 0..unparsed_characters.len() {
```
petertseng: can this be `for v in unparsed_characters.iter()` ?

exercises/ocr-numbers/example.rs
```diff
@@ -0,0 +1,67 @@
...
+
+    let y = input.lines().collect::<Vec<_>>();
+
+    let mut converted_lines = Vec::new();
+
+    for char_lines in y.chunks(4) {
+        let mut unparsed_characters = HashMap::new();
+
+        for line in char_lines {
+            let line_chars = line.chars().clone().collect::<Vec<_>>();
```
petertseng: I think `.clone()` is not necessary

exercises/ocr-numbers/example.rs
```diff
@@ -0,0 +1,67 @@
...
+
+            for (char_number, char_chunk) in line_chars.chunks(3).enumerate() {
+                let char_chars = unparsed_characters.entry(char_number).or_insert(vec![]);
```
petertseng: do you have a rule for when you use `vec![]` and when `Vec::new()`? You have used both in this file (not that I actually care, but curious)

exercises/ocr-numbers/example.rs
petertseng: oh... guess not, unless that guarantees iteration order. Okay then, carry on!

petertseng: (This *does* make me wonder whether you could use a `Vec` instead of a `HashMap` for `unparsed_characters`, but I'm not too concerned about that)

exercises/ocr-numbers/example.rs
IanWhitney: I don't have a rule. I think I usually reach for `new` when I'm starting with an empty vec, and the macro when I'm starting with a populated vector. But I'm not following that rule here.

The Book [doesn't seem to have an opinion](https://doc.rust-lang.org/std/vec/). I'll just pick something and stick to it.

exercises/ocr-numbers/example.rs
IanWhitney: Since I `clone` before `all`, this no longer has to be mutable.

exercises/ocr-numbers/example.rs
IanWhitney: HashMap may not be the best choice, but I wanted to use `Entry`, which I can't use with a vec! (not that I can see, at least.)

I'll probably switch this to BTree, since that should solve the problem of ordering.

petertseng: Okay. If you want to do that I can take a look. You might also consider using `map` but I'm uncertain whether that will work out and wouldn't hold up the PR over it.

petertseng:
Well I'm satisfied with what I see here.

If you intend to do the BTreeMap thing I can take a look at that as well when you've done it.

I might have also considered asking that the 4 and 3 be made into constants, but I'm also not going to be super rigorous with non-mission-critical code.

exercises/ocr-numbers/example.rs
```diff
@@ -0,0 +1,67 @@
...
+                for c in char_chunk {
+                    char_chars.push(*c);
+                }
+            }
+        }
+
+        let mut parsed_characters = String::new();
+
+        for (_, v) in unparsed_characters {
```
IanWhitney: BTreeMap change made this much better. It's probably possible to make this a `fold`, but I don't think it's necessary.

petertseng: I would see whether `values()` can be used to make the `(_, v)` just `v`, but I'll not be too picky

petertseng:
:+1:

PR #173, opened 2016-07-27
Use [] instead of () for vec!

lpil:
This is the idiomatic and recommended style for this macro.

> (Notice that unlike the println! macro we’ve used in the past, we use
> square brackets [] with vec! macro. Rust allows you to use either in
> either situation, this is just convention.)
>
> https://doc.rust-lang.org/book/vectors.html


IanWhitney:
Definitely an improvement. Thanks!

PR #175, opened 2016-08-02
Test anagram for same bytes but different chars

quartsize:
One can currently pass the tests by comparing a sorted `into_bytes()`, but the same bytes may represent different `char`s.

### DIFF
--- a/exercises/anagram/tests/anagram.rs
+++ b/exercises/anagram/tests/anagram.rs
@@ -100,3 +100,13 @@ fn test_does_not_detect_a_differently_cased_unicode_word_as_its_own_anagram() {
     let outputs: Vec<&str> = vec![];
     assert_eq!(anagram::anagrams_for("ΑΒΓ", &inputs), outputs);
 }
+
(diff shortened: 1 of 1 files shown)


IanWhitney:
Thanks.

I'm not sure what `latin_ff_a` and `a_sth_else` mean. Can you elaborate?

petertseng:
It would be great if these could be added to https://github.com/exercism/x-common/blob/master/anagram.json - any possibility of such a thing happening?

quartsize:
I think I was trying too hard to embed the meaning in the code itself.

I've amended the commit with something that is hopefully much clearer.

IanWhitney:
Ah! Now that test makes way more sense, thanks for the clarification. 

I like this change. I think adding it to `anagram.json` makes sense as an optional test, since this edge case is going to be very language specific.

petertseng:
I agree it's much clearer. I'm good with this! Will let Ian take a look real quick.

Will put a small note in x-common

PR #176, opened 2016-08-02
Test variable-length-quantity for an incomplete byte sequence of zero

quartsize:
My initial solution to this exercise decided if a byte sequence was incomplete based on whether the accumulator/total is zero when you run out of bytes. This works for the current test case, `0xff`, which leaves `0x7f` in your accumulator, but not for `0x80`.

### DIFF
--- a/exercises/variable-length-quantity/example.rs
+++ b/exercises/variable-length-quantity/example.rs
@@ -71,7 +71,7 @@ fn to_bytes_single(mut value: u32) -> Vec<u8> {
 pub fn from_bytes(bytes: &[u8]) -> Result<Vec<u32>, &'static str> {
     let mut res = vec![];
     let mut tmp = 0;
-    for b in bytes {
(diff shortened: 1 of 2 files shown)


IanWhitney:
Thanks.

I'm not sure what `latin_ff_a` and `a_sth_else` mean. Can you elaborate?

petertseng:
I agree this seems like a good test to have, especially since the example solution seems to fail it too. Could we trouble you to fix the example?

For other reviewers: x-common has no json file for this

quartsize:
> Could we trouble you to fix the example?

I've amended the commit. Let me know if you'd rather have it in two.

petertseng:
I say 👍  on this. If Ian's got nothing in the next 8 hours or so I'm merging it.

I am OK with one commit.

IanWhitney:
Good over here. Merging.

IanWhitney:
Thank you @quartsize!

PR #177, opened 2016-08-11
Implement Space Age

IanWhitney:
First version that passes the tests.

This first commit is intended so we can discuss the crate's API and the tests. There are two decisions I made here that I can see people finding odd.

1. Creating each class as a separate struct.
 - I did this because I want students to learn about traits. Obviously they'll have do deal with the `From` trait, but I can also see implementations grappling with shared behavior between the various planets that can be handled via traits.
2. ~~Testing that a struct is equal to a float~~ This was dumb

I'm not bound to either of these decisions, but they were intentional.

### To Do Before Merge

- [x] Confirm API makes sense
- [x] Add test helper that can compare floats
-  ~~Update example code to implement `From<u64>` and `From<f64>` generically~~ Skipping, not necessary and probably requires the `num` crate anyway.
- [x] Can we add to Readme with links to Traits & From?
- [x] Placing it in the problem order

### DIFF
--- a/exercises/space-age/tests/space-age.rs
+++ b/exercises/space-age/tests/space-age.rs
@@ -1,0 +1,69 @@
+extern crate space_age;
+
+use space_age::*;
+
+fn assert_in_delta(expected: f64, actual: f64) {
+    let diff: f64 = expected - actual.abs();
+    let delta: f64 = 0.01;
+    if diff > delta {
+        panic!("Your result of {} should be within {} of the expected result {}",
+               actual,
+               delta,
+               expected)
+    }
+}
+
+#[test]
+fn earth_age() {
+    let duration = Duration::from(1_000_000_000);
+    assert_in_delta(31.69, Earth::years_during(&duration));
+}
+
+#[test]
+#[ignore]
+fn mercury_age() {
+    let duration = Duration::from(2_134_835_688);
+    assert_in_delta(280.88, Mercury::years_during(&duration));
+}
+
+#[test]
+#[ignore]
+fn venus_age() {
+    let duration = Duration::from(189_839_836);
+    assert_in_delta(9.78, Venus::years_during(&duration));
+}
+
+#[test]
+#[ignore]
+fn mars_age() {
+    let duration = Duration::from(2_329_871_239);
+    assert_in_delta(39.25, Mars::years_during(&duration));
+}
+
+#[test]
+#[ignore]
+fn jupiter_age() {
+    let duration = Duration::from(901_876_382);
+    assert_in_delta(2.41, Jupiter::years_during(&duration));
+}
+
+#[test]
+#[ignore]
+fn saturn_age() {
+    let duration = Duration::from(3_000_000_000);
+    assert_in_delta(3.23, Saturn::years_during(&duration));
+}
+
+#[test]
+#[ignore]
+fn uranus_age() {
+    let duration = Duration::from(3_210_123_456);
+    assert_in_delta(1.21, Uranus::years_during(&duration));
+}
+
+#[test]
+#[ignore]
+fn neptune_age() {
+    let duration = Duration::from(8_210_123_456);
+    assert_in_delta(1.58, Neptune::years_during(&duration));
+}
--- a/config.json
+++ b/config.json
@@ -26,6 +26,7 @@
     "robot-simulator",
     "queen-attack",
     "sublist",
+    "space-age",
     "allergies",
     "variable-length-quantity",
     "phone-number",
--- a/exercises/space-age/Cargo.lock
+++ b/exercises/space-age/Cargo.lock
@@ -1,0 +1,4 @@
+[root]
+name = "space-age"
+version = "0.0.0"
+
--- a/exercises/space-age/Cargo.toml
+++ b/exercises/space-age/Cargo.toml
@@ -1,0 +1,3 @@
+[package]
+name = "space-age"
+version = "0.0.0"
--- a/exercises/space-age/example.rs
+++ b/exercises/space-age/example.rs
@@ -1,0 +1,79 @@
+pub struct Duration {
+    seconds: f64,
+}
+
+impl From<u64> for Duration {
+    fn from(s: u64) -> Self {
+        Duration { seconds: s as f64 }
+    }
+}
+
+impl From<f64> for Duration {
+    fn from(s: f64) -> Self {
+        Duration { seconds: s }
+    }
+}
+
+pub trait Planet {
+    fn orbital_duration() -> Duration;
+    fn years_during(d: &Duration) -> f64 {
+        d.seconds / Self::orbital_duration().seconds
+    }
+}
+
+pub struct Mercury;
+pub struct Venus;
+pub struct Earth;
+pub struct Mars;
+pub struct Jupiter;
+pub struct Saturn;
+pub struct Uranus;
+pub struct Neptune;
+
+impl Planet for Mercury {
+    fn orbital_duration() -> Duration {
+        Duration::from(7600543.81992)
+    }
+}
+
+impl Planet for Venus {
+    fn orbital_duration() -> Duration {
+        Duration::from(19414149.052176)
+    }
+}
+
+impl Planet for Earth {
+    fn orbital_duration() -> Duration {
+        Duration::from(31557600)
+    }
+}
+
+impl Planet for Mars {
+    fn orbital_duration() -> Duration {
+        Duration::from(59354032.69008)
+    }
+}
+
+impl Planet for Jupiter {
+    fn orbital_duration() -> Duration {
+        Duration::from(374355659.124)
+    }
+}
+
+impl Planet for Saturn {
+    fn orbital_duration() -> Duration {
+        Duration::from(929292362.8848)
+    }
+}
+
+impl Planet for Uranus {
+    fn orbital_duration() -> Duration {
+        Duration::from(2651370019.3296)
+    }
+}
+
+impl Planet for Neptune {
+    fn orbital_duration() -> Duration {
+        Duration::from(5200418560.032)
+    }
+}
--- a/exercises/space-age/src/lib.rs
+++ b/exercises/space-age/src/lib.rs
@@ -1,0 +1,35 @@
+// The code below is a stub. Just enough to satisfy the compiler.
+// In order to pass the tests you can add-to or change any of this code.
+#![allow(unused_variables)]
+
+pub struct Duration;
+
+impl From<u64> for Duration {
+    fn from(s: u64) -> Self {
+        unimplemented!()
+    }
+}
+
+pub trait Planet {
+    fn years_during(d: &Duration) -> f64 {
+        unimplemented!();
+    }
+}
+
+pub struct Mercury;
+pub struct Venus;
+pub struct Earth;
+pub struct Mars;
+pub struct Jupiter;
+pub struct Saturn;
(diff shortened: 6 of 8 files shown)


IanWhitney:
I wouldn't spend much time looking at the example code yet. It's not meant to be good. Just there to get the tests passing.

petertseng:
I have observations, but not necessarily positions on how things should be done. I will give the obesrvations, and maybe come back in a day with positions when I have had time to think about them.

> Testing that a struct is equal to a float

Currently, we would have to accept that the example's partial equivalence is not symmetric, which is contrary to https://doc.rust-lang.org/core/cmp/trait.PartialEq.html .

We can see that if we changed one of the assertions from `assert_eq!(age, 31.69);` `assert_eq!(31.69, age);` it would be a compilation error! Well, we would fix the error if the example would also define `impl PartialEq<EarthAge> for f64`, etc.

Although the symmetricity would not get tested unless we explicitly have assertions both ways in our tests. So I guess one question is "the tests don't require that the partial equivalence is symmetric; should they?"

The same goes for transitivity if `f64` defines its equivalence to other types - then for the partial equivalence to be transitive `*Age` would have to define its equivalence to those other types as well. But would we even want to test that?

Huh. This gets quite verbose quite the more types we wish to define as equal to one another, since O(n^2) pairs of `eq` functions need to be defined, it seems...! I wonder how others have dealt with this problem?

I wonder what it would be if the various `*Age` were equatable to just each other, though maybe that would make it `derive`able, contrary to one of the stated goals.

IanWhitney:
Yeah, I had already decided that my Value Object idea was terrible. This is just more evidence to support that. I'll switch that part of the implementation.

petertseng:
BUT YOU FORGOT PLUTO :trollface: 

I am obviously joking, but I am (and I guess that you are also) old enough to have grown up with Pluto being a planet.

I will attempt the exercise myself and see what I think of the idea of using this exercise for traits. The stub file is very verbose. I will attempt to curb the verbosity when I implement, but of course we can't guarantee that all students will think this way. Perhaps that has to be left to discussion on the site.

If I can't find a way to curb the verbosity, then I would question the choice to make each `*Age` a different type.

petertseng:
Life would probably be better if there was a tolerance on the expected values

```thread 'earth_age' panicked at 'assertion failed: `(left == right)` (left: `31.688087`, right: `31.69`)', tests/space-age.rs:8```

I don't think I should be required to round that

**Comment by petertseng** [comment]
the `#[derive(Debug)]` can probably be removed?

**Comment by petertseng** [comment]
The least-repetitive I came up with: https://gist.github.com/petertseng/0b3928f0ee8881e94bf654d3f2153df9

Repeating `X` and `XAge` on the same line is unfortunate, but I didn't find a way around it.

**Comment by IanWhitney** [comment]
Macro rules. Interesting! Never occurred to me.

I think my approach looks something like this:

 https://gist.github.com/9a9ddb50afd08f4ae5a20258c056916b

My least favorite part of it is that the `Planet` trait has to define a `seconds` function that ends up being an alias for the `seconds` value. Maybe there's a way around that, but I don't know what it is.

I will also have to implement `From<u64>` for each individual planet, right? Though maybe there's a way to make that generic over each T: Planet?

Questions like that are why I structured the tests the way I did, honestly. I want to push students towards thinking about generics, traits and code reuse. To me that's how we make this problem interesting and valuable to people learning Rust.

**Comment by petertseng** [comment]
> Macro rules. Interesting! Never occurred to me.

I only use them to take care of the duplication that I could not avoid.
All other duplication was avoided with the `Planet` trait.
Someone could ostensibly write the entire thing in a macro (`From<u64>` and `years` for each individual planet).
I would consider that bad style and too much to put in a macro: If it can be done outside of a macro, do so.
(But enough soap-boxing from me)

> I will also have to implement `From<u64>` for each individual planet, right? Though maybe there's a way to make that generic over each `T: Planet`?

There absolutely is. See the `impl <T: Planet> From<u64> for Age<T>` if you need, though I would encourage a search for other solutions in addition - perhaps there are other ways! The desire to avoid that duplication was what drove me to start my solution.

> Questions like that are why I structured the tests the way I did, honestly. I want to push students towards thinking about generics, traits and code reuse. To me that's how we make this problem interesting and valuable to people learning Rust.

You certainly got me interested and made me learn a few things, so you have me convinced.

**Comment by IanWhitney** [comment]
Over the course of several hours I worked on a long comment explaining why this approach wouldn't work. Turns out I [rubber ducked myself](https://en.wikipedia.org/wiki/Rubber_duck_debugging)

I think this does what I want

https://play.rust-lang.org/?gist=9e1acfeae7ebfdc7dd8176c1c015adb9&version=stable&backtrace=0

It's pretty different from the first approach. After working on this a few ways, I realized that at its core, the problem has a couple of concepts

- Durations (seconds, orbital lengths [i.e., years])
- Planets

The language of the exercise makes you think about "Years" as a property of a person, or some other object separate from a Planet. So I kept trying to put that logic somewhere outside of Planet.

But a "year" is the same as a planetary orbit, and it makes perfect sense to ask a planet how many orbits it's made in a duration. Hence the `orbits_during` function. And the trait can easily provide a default implementation for that function that works for all `<T: Planet>`

I would like to combine the 2 implementations of `From` for `Duration` , but found differing advice on how to declare a function generic over numeric types.

Planets have orbital lengths. And durations-in-seconds can be converted to durations-in-years, but only if you provide the planet.

From (and Into) only convert _one_ type into another. But taking a durations-in-seconds and getting a duration-in-years requires two: the seconds and the planet. So I can't use From to give me:

```rust
YearDuration::from(seconds, planet)
```

And I can't think of a single type that really encapsulates the idea of a number of years spent on a specific planet. Well, I can, but it doesn't seem like a sensible type. It seems like an unnecessary conflation of two types just so that I can use the `from` trait.

The crate [Chrono](https://lifthrasiir.github.io/rust-chrono/chrono/index.html) has some similar concerns, as it has to [create DateTime values from more than one input](https://lifthrasiir.github.io/rust-chrono/chrono/datetime/struct.DateTime.html#method.parse_from_str). So I'm thinking that might be the best approach here.

IanWhitney:
> Life would probably be better if there was a tolerance on the expected values

Looking at the Ruby implementation they can use the `assert_in_delta` method. Rust doesn't have that, but it's easy enough to add.

IanWhitney:
I've updated the implementation. And I've put a To Do list at the [top of the PR](https://github.com/exercism/xrust/pull/177#issue-170561347) so that I can track progress.

IanWhitney:
Problem order specified. I think its current position is OK.

exercises/space-age/tests/space-age.rs
```diff
@@ -0,0 +1,81 @@
...
+// https://doc.rust-lang.org/book/guessing-game.html#generating-a-secret-number
+//
+extern crate space_age;
+extern crate num;
+
+use std::fmt::Display;
+use space_age::*;
+use num::{Float, NumCast};
+
+fn assert_in_delta<T: Float + Display>(expected: T, actual: T) {
```
petertseng: I suppose I'm not afraid to add a dependency if it's necesary (that's what cargo is for) but I want to hear the explanation for why it is necesary, rather than have `fn assert_in_delta(expected: f64, actual: f64)`. What does the bound of `Float` get us that we will not otherwise have? Does it allow students more flexibility in the return type of their `years_during` function?

IanWhitney: Maybe there's an aspect of Rust I'm missing here. I want this function to accept both f32 and f64 and the only way I could see to do that is by using the `Float` trait in `num`

Also, the `NumCast` part of `num` comes in pretty handy.

IanWhitney:
One of the to-do items has baffled me

> Update example code to implement From<u64> and From<f64> generically

I've tried tackling this a variety of ways, all with no luck. Anyone know how to implement From for all numeric types? I would thinks something like:

```rust
impl<T: Num> From<T> for Duration {
  fn from(s: T) -> Self {
    Duration { seconds: s }
  }
}
```

Would work. But no. I get "non-scalar cast: `T` as `f64`". I think I'm getting the syntax wrong, but I'm not sure how.

exercises/space-age/tests/space-age.rs
petertseng: > I want this function to accept both f32 and f64

Is that because you think students may change `years_during` to return an `f32` instead of an `f64` (as it is so defined in the stub file) and you want to be prepared for this situation?

IanWhitney: Hmm. I swear I was having problems with comparisons in some cases, but now I can't re-create them. I'll drop the crate and see if I run into any problems.

IanWhitney:
I'm on the fence about 81fa6eeab0caa8dd2a48729480fc46fa5170dbba. Interested in what others think of the idea.

If we're OK with the API then this should be ready to go.

petertseng:
I know you aren't thinking about the genericity over `u64` and `f64` anymore, but you weren't alone on that "non-scalar cast": https://github.com/rust-lang/rust/issues/31174#issuecomment-174481550

I do suppose the `num` crate would have done it, but then we'd be including the `num` crate for the example only (it's not used by the test), and I wonder whether it would confuse students to see the crate and wonder whether they have to use it. We're [not the only track with this problem](https://github.com/exercism/xhaskell/pull/189#discussion_r70165604) either.

I have no complaints about the problem placement or topics.md

I imagine there may be questions like "why isn't `Planet` just an `enum` and the interface `years_on(p: Planet, seconds: u64)`?!" or "does it make sense that they're separate types if there's only one of each panet", but that will probably have to be the way things be if taking this approach. I'm not opposed.

exercises/space-age/tests/space-age.rs
```diff
@@ -0,0 +1,69 @@
...
+               actual,
+               delta,
+               expected)
+    }
+}
+
+#[test]
+fn earth_age() {
+    let duration = Duration::from(1_000_000_000);
+    assert_in_delta(31.69, Earth.years_during(&duration));
```
petertseng: interesting point that `Earth.years_during` is used rather than `Earth::years_during`. This is possible because `Earth` is a struct with no members (a [unit-like struct](https://doc.rust-lang.org/book/structs.html#unit-like-structs) in the nomenclature). Doesn't seem to me like these functions need to take a `self` though...

petertseng: This allows for the "unusual" solution of https://gist.github.com/petertseng/ec3887c899f0259661847e8fbfba820d

IanWhitney:
> I imagine there may be questions like "why isn't Planet just an enum and the interface years_on(p: Planet, seconds: u64)?!" or "does it make sense that they're separate types if there's only one of each panet", but that will probably have to be the way things be if taking this approach. I'm not opposed.

I think we're always welcoming of suggested changes like this. We can always point to PRs like this one (or the commit history) to say, "Here was the goal of this API". And if people can provide a good alternative that satisfies the same goals, then all the better.

> interesting point that Earth.years_during is used rather than Earth::years_during. This is possible because Earth is a struct with no members (a unit-like struct in the nomenclature). Doesn't seem to me like these functions need to take a self though...

Agreed that `self` is unnecessary here. Commit incoming.
> This allows for the "unusual" solution of https://gist.github.com/petertseng/ec3887c899f0259661847e8fbfba820d

IanWhitney:
Ok. I think this one is winding down so I've rebased down to a single commit. I'll let this PR sit until Friday, August 19th. If nothing comes up, I'll merge it.

PR #178, opened 2016-08-17
Implement Bracket Push

IanWhitney:
I'm on the fence about using `&str`. If we don't provide the stub file, it will increase the difficulty of implementing `from`.

But if we include the stub file, then there's almost no extra work involved. Which also might reduce the amount students learn from the exercise.

Sometimes you gotta work with borrowed stuff, right? Writing a `from` implementation for a borrow seems like a useful thing to learn.

So I think my current wish is to:

- Write the tests to expect `From<&'a str>` (`From<&'static str>` would also work, I think)
- Don't include the stub file

### DIFF
--- a/exercises/bracket-push/example.rs
+++ b/exercises/bracket-push/example.rs
@@ -1,0 +1,79 @@
+use std::collections::HashMap;
+
+pub struct Brackets {
+    raw_brackets: Vec<char>,
+    pairs: MatchingBrackets,
+}
+
+impl<'a> From<&'a str> for Brackets {
+    fn from(i: &str) -> Self {
+        Brackets::new(String::from(i), None)
+    }
+}
+
+impl Brackets {
+    pub fn new(s: String, pairs: Option<Vec<(char, char)>>) -> Self {
+        let p = match pairs {
+            Some(x) => MatchingBrackets::from(x),
+            None => MatchingBrackets::from(vec![('[', ']'), ('{', '}'), ('(', ')')]),
+        };
+
+        Brackets {
+            raw_brackets: s.chars().filter(|c| p.contains(&c)).collect::<Vec<char>>(),
+            pairs: p,
+        }
+    }
+
+    pub fn are_balanced(&self) -> bool {
+        let mut unclosed: Vec<char> = Vec::new();
+
+        for &bracket in self.raw_brackets.iter() {
+            if let Some(last_unclosed) = unclosed.pop() {
+                unclosed.extend(self.pairs.unmatched(last_unclosed, bracket));
+            } else {
+                unclosed.push(bracket);
+            }
+        }
+
+        unclosed.is_empty()
+    }
+}
+
+pub struct MatchingBrackets {
+    collection: HashMap<char, char>,
+}
+
+impl From<Vec<(char, char)>> for MatchingBrackets {
+    fn from(v: Vec<(char, char)>) -> Self {
+        MatchingBrackets { collection: v.into_iter().collect::<HashMap<char, char>>() }
+    }
+}
+
+impl MatchingBrackets {
+    fn contains(&self, other: &char) -> bool {
+        let known = self.collection.keys().chain(self.collection.values()).collect::<Vec<_>>();
+        known.contains(&other)
+    }
+
+    fn closer_for(&self, k: &char) -> Option<&char> {
+        self.collection.get(k)
+    }
+
+    fn closed_by(&self, l: char, r: char) -> bool {
+        match self.closer_for(&l) {
+            Some(&x) => r == x,
+            None => false,
+        }
+    }
+
+    fn unmatched(&self, open: char, potential_close: char) -> Vec<char> {
+        let mut ret: Vec<char> = Vec::new();
+
+        if !self.closed_by(open, potential_close) {
+            ret.push(open);
+            ret.push(potential_close);
+        }
+
+        ret
+    }
+}
--- a/config.json
+++ b/config.json
@@ -24,6 +24,7 @@
     "grade-school",
     "tournament",
     "robot-simulator",
+    "bracket-push",
     "queen-attack",
     "sublist",
     "space-age",
--- a/exercises/bracket-push/Cargo.lock
+++ b/exercises/bracket-push/Cargo.lock
@@ -1,0 +1,4 @@
+[root]
+name = "bracket-push"
+version = "0.0.0"
+
--- a/exercises/bracket-push/Cargo.toml
+++ b/exercises/bracket-push/Cargo.toml
@@ -1,0 +1,3 @@
+[package]
+name = "bracket-push"
+version = "0.0.0"
--- a/exercises/bracket-push/HINTS.md
+++ b/exercises/bracket-push/HINTS.md
@@ -1,0 +1,6 @@
+# Bracket Push in Rust
+
+Reading about these Rust topics may help you implement a solution.
+
+- Lifetimes and Structs: https://doc.rust-lang.org/book/lifetimes.html#impl-blocks
+- From trait: https://doc.rust-lang.org/std/convert/trait.From.html
--- a/exercises/bracket-push/tests/bracket-push.rs
+++ b/exercises/bracket-push/tests/bracket-push.rs
@@ -1,0 +1,82 @@
+extern crate bracket_push;
+
+use bracket_push::*;
+
+#[test]
+fn paired_square_brackets() {
+    assert!(Brackets::from("[]").are_balanced());
+}
+
+#[test]
+#[ignore]
+fn empty_string() {
+    assert!(Brackets::from("").are_balanced());
+}
+
+#[test]
+#[ignore]
+fn unpaired_brackets() {
+    assert!(!Brackets::from("[[").are_balanced());
+}
+
+#[test]
+#[ignore]
+fn wrong_ordered_brackets() {
+    assert!(!Brackets::from("}{").are_balanced());
+}
+
+#[test]
+#[ignore]
+fn paired_with_whitespace() {
+    assert!(Brackets::from("{ }").are_balanced());
+}
+
+#[test]
+#[ignore]
+fn simple_nested_brackets() {
+    assert!(Brackets::from("{[]}").are_balanced());
+}
+
+#[test]
+#[ignore]
+fn several_paired_brackets() {
+    assert!(Brackets::from("{}[]").are_balanced());
+}
+
+#[test]
+#[ignore]
+fn paired_and_nested_brackets() {
+    assert!(Brackets::from("([{}({}[])])").are_balanced());
+}
+
+#[test]
+#[ignore]
+fn unopened_closing_brackets() {
+    assert!(!Brackets::from("{[)][]}").are_balanced());
+}
+
+#[test]
+#[ignore]
+fn unpaired_and_nested_brackets() {
+    assert!(!Brackets::from("([{])").are_balanced());
+}
+
+#[test]
+#[ignore]
+fn paired_and_wrong_nested_brackets() {
+    assert!(!Brackets::from("[({]})").are_balanced());
+}
+
+#[test]
+#[ignore]
+fn math_expression() {
+    assert!(Brackets::from("(((185 + 223.85) * 15) - 543)/2").are_balanced());
+}
+
+#[test]
+#[ignore]
+fn complex_latex_expression() {
+    let input = "\\left(\\begin{array}{cc} \\frac{1}{3} & x\\\\ \\mathrm{e}^{x} &... x^2 \
+                 \\end{array}\\right)";
+    assert!(Brackets::from(input).are_balanced());
+}
--- a/problems.md
+++ b/problems.md
@@ -43,6 +43,7 @@ hexadecimal |  Option, zip/fold/chars, map
 grade-school |  struct, entry api, Vec, Option
 tournament |  enum, sorting, hashmap, structs
 robot-simulator | Immutability, enum
+bracket-push | From trait, stack or recursion
 queen-attack |  struct, trait (optional), Result
 sublist |  enum, generic over type
 space-age | Custom Trait, From Trait, Default Trait implementation


petertseng:
Oh, right, because we're making a `Brackets` from a string, so that will need to annotate its lifetime. Okay.

And what would you say to the suggestion to instead make the interface `brackets::are_balanced("()")` ?

Of course, internally I may use some structures that require me to track lifetimes, so don't worry that that's being lost.

petertseng:
I will work on a solution soon to make sure everything is sane. I don't expect there will be anything groundbreaking in it so I'll simply wait until the exercise goes live to submit it, but I'll report whether there were any problems.

exercises/bracket-push/src/lib.rs
```diff
@@ -0,0 +1,16 @@
+// The code below is a stub. Just enough to satisfy the compiler.
+// In order to pass the tests you can add-to or change any of this code.
+
+pub struct Brackets;
+
+impl<'a> From<&'a str> for Brackets {
+    fn from(i: &str) -> Self {
```
petertseng: is `s` better? or does the `i` stand for "input"?

I don't actually care, but I ask in case it matters to someone

petertseng:
The current tests appear OK.

IanWhitney:
Re: Using the `from` trait. It might be unnecessary here. But recently I've started paying a lot more attention to Primitive Obsession smells. I want to get concepts out of primitives and into named concepts (objects, structs, whatever). 

I think this approach works especially well with strong, static type languages. In Ruby I can get away with passing primitives around forever, thanks to its endless number of type-casting functions (both implicit and explicit).

This gets harder in Rust because it won't do implicit type changes. Eventually you'll end up [with an i8 and a u8 and want to do some math and run into a type error](https://play.rust-lang.org/?gist=aa906daa85c93a3266caf5af8b81337b&version=stable&backtrace=0), for example.

But if I get all primitives into types that I control, then I'm:

- Less likely to run into problems like the i8 vs u8 one.
- Giving the concept a domain name
- Providing a place to hang behavior

In short, I think it's good design.

Which all seems like *extreme overkill* in an exercise like this, I agree. There is no additional behavior, there is no 'domain', there is only passing the tests.

But if people are using these exercises to learn how to write Rust and the exercises encourage smelly Rust code, then won't most of the students go off and write smelly Rust code in production?

petertseng:
Of the two choices of `brackets::are_balanced("[]")` and `Brackets::from("[]").are_balanced()`, I find either acceptable. If only one choice is acceptable to others, it obviously makes sense to proceed with this PR using the one that is acceptable to all. At this point, we could consider this decision made and move on to other aspects of the PR. I'll discuss it because there are points worth discussing.

Since the motivation here is only to gain understanding for myself, there's no need to wait for this discussion before moving forward with this PR, since an acceptable resolution is already known.

---

> But recently I've started paying a lot more attention to Primitive Obsession smells. I want to get concepts out of primitives and into named concepts (objects, structs, whatever). 

I wonder if I jokingly said "For every design principle there is an equal and opposite design principle" whether it would be true. I did find that "It is better to have 100 functions operate on one data structure than 10 functions on 10 data structures" and http://stackoverflow.com/questions/6016271/why-is-it-better-to-have-100-functions-operate-on-one-data-structure-than-10-fun attempted to explain why that would be the case.

How would that apply to the matter at hand? I'm not sure. I think a situation I can imagine is: I have a collection of strings, and I want to first filter out the ones that contain valid brackets, and of the ones that do, I want to do something else with them, maybe `map(|x| x.len())` or something. But I can also imagine that it's not that much harder to do `.filter(|x| Brackets.from(x).are_balanced())` than `.filter(|x| brackets::are_balanced(x))`. Perhaps another idea is: If I have a collection of `Brackets`, I can only do operations on them that `Brackets` allows, whereas if I had a collection of strings I can do many more things. Sometimes I want to be able to do that, sometimes not.

From my reading, a good use of a custom type rather than a primitive is when not all possible values the primitive can take are valid for the custom type. That seems to make sense here: Not all strings are bracketed strings. Unfortunately [TryFrom](https://doc.rust-lang.org/std/convert/trait.TryFrom.html) will not be stable for a while, otherwise I would find that interface very good: `BracketedString::try_from("[[")` would give an error, while `BracketedString::try_from("[]")` would give something suitably successful. Alas, we shall have to make do without for now...

> I think this approach works especially well with strong, static type languages.
> Eventually you'll end up with an i8 and a u8 and want to do some math and run into a type error, for example.

Yes, I think this is often why I am interested in languages with static type systems, and taking advantage of their type systems. I like to avoid mistakes such as the adding of an i8 and a u8, or the mistake of adding two `(int, int)` when one of them is `(x, y)` and the other is `(r, θ)`.

> But if I get all primitives into types that I control, then I'm less likely to run into problems like the i8 vs u8 one.

I thought it would make them more likely, *but* you would be happy that you are running into the problems (happy that the compiler is catching the potential mistakes, rather than the code having weird bugs exposing themselves only at runtime)

> But if people are using these exercises to learn how to write Rust and the exercises encourage smelly Rust code, then won't most of the students go off and write smelly Rust code in production?

Wholeheartedly agree.

IanWhitney:
Thanks for the link to the SO question, that's a good perspective. I'm guessing this the sort of problem that Generics are trying to address? A single function can continue to work on your primitive types and all of your custom types.

"Good Design" certainly isn't a checklist, it's a series of tradeoffs. No doubt. In a real-world application that checks brackets we might find that using `Brackets` a bad idea, or it might be great.  Hard to say. Without further knowledge I'm OK going with my first instinct of reducing direct Primitive usage.

exercises/bracket-push/example.rs
```diff
@@ -0,0 +1,62 @@
+// The code below is a stub. Just enough to satisfy the compiler.
```
petertseng: probably good to remove these from example

exercises/bracket-push/example.rs
```diff
@@ -0,0 +1,62 @@
...
+        d.insert('[', ']');
+        d.insert('{', '}');
+        d.insert('(', ')');
+        d
+    }
+
+    pub fn are_balanced(&self) -> bool {
+        let mut unmatched: Vec<char> = Vec::new();
+
+        for b in self.all.clone() {
```
petertseng: Hmm, why is clone needed? what happens without?

exercises/bracket-push/example.rs
```diff
@@ -0,0 +1,62 @@
...
+            if unmatched.is_empty() {
+                unmatched.push(b);
+            } else {
+                let t = unmatched.pop().unwrap();
```
petertseng: If you like to avoid `unwrap` you can put this in an `if let` as well, but I don't care so much

exercises/bracket-push/example.rs
```diff
@@ -0,0 +1,62 @@
...
+
+                if let Some(c) = self.pairs.get(&t) {
+                    if *c != b {
+                        unmatched.push(t);
```
petertseng: Ah, pushing back what you just popped. I guess you do have to do one of the two following though:

1. Always pop, push it back if it doesn't match
2. Just look, only pop if it does match.

I don't suppose one is better than the other by itself.

But you might consider whether it is possible to only have one `unmatched.push(b)`

exercises/bracket-push/example.rs
petertseng: I needed to review https://doc.rust-lang.org/book/loops.html#for to answer why. because `for b in self.all` uses `IntoIterator` which consumes, whereas `self` is only borrowed. That implies that `for &b in self.all.iter()` will work, and I suggest it to not clone the vector.

exercises/bracket-push/example.rs
```diff
@@ -0,0 +1,62 @@
...
+
+        Brackets {
+            all: all,
+            pairs: p.clone(),
+        }
+    }
+
+    fn default() -> HashMap<char, char> {
+        let mut d = HashMap::new();
+        d.insert('[', ']');
```
petertseng: It's times like these when I wish for some hash map literals. (This is possible today with macros, if one really wishes it, though doesn't seem useful enough for this example implementation)

And maybe a way to share this default against all instances of `Brackets` that get the default map, but this is tricky.

No action needed, only musing about language capabilities.

exercises/bracket-push/example.rs
IanWhitney: I'm not sure what you mean by 'just look'. There's [`peek`](https://doc.rust-lang.org/core/iter/struct.Peekable.html#method.peek). Is there a way I can look at the last element of a vec w/o popping it?

exercises/bracket-push/example.rs
IanWhitney: Good suggestion. Thanks.

exercises/bracket-push/example.rs
petertseng: Probably [`last`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.last)

You are right that I meant to look at the last element without popping it. I wasn't specific at the time since I hadn't examined whether it was possible, now I can be specific about `last`

exercises/bracket-push/example.rs
```diff
@@ -13,46 +11,65 @@ impl<'a> From<&'a str> for Brackets {
...
-                unmatched.push(b);
+                unmatched.push(bracket);
             } else {
-                let t = unmatched.pop().unwrap();
-
-                if let Some(c) = self.pairs.get(&t) {
-                    if *c != b {
-                        unmatched.push(t);
-                        unmatched.push(b);
+                if let Some(last_unmatched) = unmatched.pop() {
```
petertseng: I expect that will always be `Some`, because if you get here, then `unmatched.is_empty()` is false. Since the `else` on this `if let` is the same as what happens if `unmatched` is empty, the two can be combined and deduplicated.

petertseng: edited for accuracy (the `else` for the `if let` was not the push)

exercises/bracket-push/example.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    pub fn are_balanced(&self) -> bool {
+        let mut unmatched: Vec<char> = Vec::new();
+
+        for &bracket in self.raw_brackets.iter() {
+            if unmatched.is_empty() {
+                unmatched.push(bracket);
+            } else {
+                if let Some(last_unmatched) = unmatched.pop() {
+                    if let Some(closing_bracket) = self.pairs.closer_for(&last_unmatched) {
```
petertseng: test whether my comment appears on the diff

exercises/bracket-push/example.rs
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    pub fn are_balanced(&self) -> bool {
+        let mut unmatched: Vec<char> = Vec::new();
+
+        for &bracket in self.raw_brackets.iter() {
+            if unmatched.is_empty() {
+                unmatched.push(bracket);
+            } else {
+                if let Some(last_unmatched) = unmatched.pop() {
```
petertseng: test whether my comment appears on the diff

IanWhitney: Oh, sure. That makes way more sense.

exercises/bracket-push/example.rs
```diff
@@ -0,0 +1,75 @@
...
+impl Brackets {
+    pub fn new(s: String, pairs: Option<Vec<(char, char)>>) -> Self {
+        let p = match pairs {
+            Some(x) => MatchingBrackets::from(x),
+            None => MatchingBrackets::from(vec![('[', ']'), ('{', '}'), ('(', ')')]),
+        };
+
+        Brackets {
+            raw_brackets: s.chars().filter(|c| p.contains(&c)).collect::<Vec<char>>(),
+            pairs: p.clone(),
```
petertseng: Is `clone` needed? At this point no matter whether `pairs` was a `Some` or `None`, you have full ownership over a `MatchingBrackets`, right?

exercises/bracket-push/example.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    pub fn are_balanced(&self) -> bool {
+        let mut unmatched: Vec<char> = Vec::new();
+
+        for &bracket in self.raw_brackets.iter() {
+            if let Some(last_unmatched) = unmatched.pop() {
+                if let Some(closing_bracket) = self.pairs.closer_for(&last_unmatched) {
+                    if *closing_bracket != bracket {
+                        unmatched.push(last_unmatched);
+                        unmatched.push(bracket);
```
petertseng: I think the final thing I want to think about codewise is whether there is a way to deduplicate the `unmatched.push(bracket)` since there are three of them right now.

What do the three have in common? Or maybe it helps to think about the one case (if I read the code correctly, there is only one) when `unmatched.push(bracket)` is *not* done. If it can be deduplicated and the code remains understandable, I think that would be preferable.

petertseng:
placement:

If you would like to make a decision before being biased by mine, I put my final thoughts at the bottom.

topics are: iteration over a string's characters, one of (stack OR recursive descent parser). *maybe* something to do with hash maps (but it's easy to imagine a solution that doesn't use one).

I'll suggest either right between queen-attack and sublist. Can probably be easily convinced to place it anywhere between grade-school and wordy. Anywhere else will need more convincing.

exercises/bracket-push/example.rs
IanWhitney: It gets angry when I remove the clone. Let me look at that some more.

IanWhitney: Huh. Now it's not angry. I must have been doing something wrong or fixed something without knowing.

exercises/bracket-push/example.rs
```diff
@@ -53,23 +68,18 @@ impl Brackets {
...
-            if let Some(last_unmatched) = unmatched.pop() {
-                if let Some(closing_bracket) = self.pairs.closer_for(&last_unmatched) {
-                    if *closing_bracket != bracket {
-                        unmatched.push(last_unmatched);
-                        unmatched.push(bracket);
-                    }
-                } else {
-                    unmatched.push(bracket);
+            if let Some(last_unclosed) = unclosed.pop() {
+                for c in self.pairs.unmatched(last_unclosed, bracket) {
```
petertseng: take a look at whether this can be expressed as an [`extend`](https://doc.rust-lang.org/std/iter/trait.Extend.html) which I think cn work because the docs say `impl<T> Extend<T> for Vec<T>`.

the idea of letting the `pairs`  determine what's unmatched seems OK.

exercises/bracket-push/example.rs
```diff
@@ -0,0 +1,84 @@
...
+            pairs: p,
+        }
+    }
+
+    pub fn are_balanced(&self) -> bool {
+        let mut unclosed: Vec<char> = Vec::new();
+
+        for &bracket in self.raw_brackets.iter() {
+            if let Some(last_unclosed) = unclosed.pop() {
+                for c in self.pairs.unmatched(last_unclosed, bracket) {
```
petertseng: take a look at whether this can be expressed as an [`extend`](https://doc.rust-lang.org/std/iter/trait.Extend.html) which I think cn work because the docs say `impl<T> Extend<T> for Vec<T>`.

the idea of letting the `pairs`  determine what's unmatched seems OK.

IanWhitney: Huh! Extend! Learning things all the time. Good find.

petertseng:
Everything seems to be coming together nicely. You mentioned you wanted to do something about `unmatched` (the duplication, I presume) which I would support. I think once that is done this will be good to go.

IanWhitney:
Ok, I have been torturing this example code for far too long. It's fine. Really, it's just fine. I can save the code torture for when I submit solutions to exercism.

I think everyone is ok with the API (at least no one has complained openly). I still want to:

- [ ] Add a HINTS.md file
- [ ] Remove the stub file, as I think it trivializes one of the things this problem can teach
- [ ] Restore the `#[ignore]` directive on all-but-the-first test.

petertseng:
I don't suppose you would find it too much to see if that `MatchingBrackets`' `HashMap` initialization can be done with `collect`?

```rust

fn main() {
    let v = vec![(1, 2), (3, 4), (5, 6)];
    let h1 = v.iter().cloned().collect::<HashMap<_, _>>();
    let h2 = v.into_iter().collect::<HashMap<_, _>>();
    println!("{:?}", h1);
    println!("{:?}", h2);
}
```

petertseng:
I'm pretty sure I've said all I have to say :+1:

PR #186, opened 2016-08-24
nucleotide-count: Use `Entry#or_insert`

petertseng:
In #155, this operation was performed for other exercises (tournament,
parallel-letter-frequency, word-count), but nucleotide-count was skipped
because #149 indicates that invalid nucleotides maybe should produce
errors.

We don't need to wait for x-common to get a nucleotide-count.json before
deciding to make the current implementation more idiomatic though.

If it's eventually decided that `or_insert` doesn't make sense, we can
change it then.


IanWhitney:
ok

PR #187, opened 2016-08-26
Use lowercase name for romans module

bradleybeddoes:
The previous uppercase module name would output the following
warning when running tests:

```
src/lib.rs:2:1: 42:2 warning: module `Roman` should have a snake case
name such as `roman`, #[warn(non_snake_case)] on by default
src/lib.rs:2 pub mod Roman {
```


bradleybeddoes:
Apologies, this should be a struct and thus uppercase is valid.

PR #190, opened 2016-08-28
space_age: properly check float delta

### LINKED ISSUE #188: space-age tests aren't checking results properly
This passes and has NO reason to: 

http://exercism.io/submissions/2d8927ee710c403c8c822d9bfba4f565

durka:
Missing parentheses were causing the test to check `expected - abs(actual)` instead of `abs(expected - actual)`.

Fixes #188.

### DIFF
--- a/exercises/space-age/tests/space-age.rs
+++ b/exercises/space-age/tests/space-age.rs
@@ -3,7 +3,7 @@ extern crate space_age;
 use space_age::*;
 
 fn assert_in_delta(expected: f64, actual: f64) {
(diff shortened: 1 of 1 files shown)


petertseng:
Shame I didn't catch it earlier, but yup, this is the right thing to do.

Verified that a solution that always returns `1e100` passes without the change in this PR and fails with the change in this PR.

petertseng:
Thanks.

IanWhitney:
D'oh. Thanks!

PR #195, opened 2016-09-02
WIP: Updating config.json

IanWhitney:
Working on #179

The new exercises collection is in place and I'm populating the topics
values.

### DIFF
--- a/config.json
+++ b/config.json
@@ -67,126 +67,189 @@
       "slug": "hamming",
       "difficulty": ,
       "topics": [
+        "Exception handling",
+        "Result",
+        "Higher-orderfunctions",
+        "Char"
       ]
     },
     {
       "slug": "scrabble-score",
       "difficulty": ,
       "topics": [
+        "Char",
+        "Higher-order functions",
+        "Maps"
       ]
     },
     {
       "slug": "pangram",
       "difficulty": ,
       "topics": [
+        "Char",
+        "Higher-order functions",
+        "ascii"
       ]
     },
     {
       "slug": "nucleotide-count",
       "difficulty": ,
       "topics": [
+        "Maps",
+        "Entry API",
+        "Char"
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -67,126 +67,189 @@
       "slug": "hamming",
       "difficulty": ,
       "topics": [
+        "Exception handling",
+        "Result",
+        "Higher-orderfunctions",
```
petertseng: add space

config.json
```diff
@@ -67,126 +67,189 @@
...
       "topics": [
+        "Enumerations",
+        "Immutability"
       ]
     },
     {
       "slug": "bracket-push",
       "difficulty": ,
       "topics": [
+        "From trait"
```
petertseng: comma

config.json
```diff
@@ -3,6 +3,322 @@
...
+      "slug": "roman-numerals",
+      "difficulty": ,
+      "topics": [
+        "From trait",
+        "Display trait",
+        "Maps"
+      ]
+    },
+    {
+      "slug": "hexadecimal",
```
petertseng: (just happened) we kicked out hexadecimal for all-your-base, replace here

petertseng:
There has not been much activity here for a while.

What would you say if I were to take over it?

My plan is to just use problems.md exactly for the topics, and a difficulty of 1 for everything. We can refine both items later.

petertseng:
Just as #179 states, we should take it in steps. Therefore, my proposal.

IanWhitney:
The problem I ran in to is that I find the difficulty rating impossible. What does 'difficult' mean? Is it hard because it introduces Lifetimes? Because it's algorithmically hard? And guessing at the difficulty of exercises I haven't done was even harder.

Starting everything with a difficulty of 1 is fine.

IanWhitney:
Replaced by #229. closing.

PR #199, opened 2016-09-11
move tournament between wordy and custom-set

### LINKED ISSUE #174: Large jump in difficulty when tournament challenge is reached
Exercises previous to tournament only really include one problem, while this one includes quite a few and includes a test suite that doesn't break the challenge down into the component parts.

This results in a rather sudden and uncomfortable jump in difficulty. I feel that the tournament challenge should be perhaps moved later in the series or simplified.

petertseng:
Reviewers: I decided I'd go and take responsibility for this since I was the one who moved it earlier in the first place.

Take my word with a grain of salt since I haven't done this exercise myself yet.

Commit message follows.

---

In the beginning, tournament was between parallel-letter-frequency and
rectangles.

Then, in #152, File I/O was removed so it was deemed to be easier. So we
moved it to between grade-school and robot-simulator.

However, #174 rightly points out that tournament is still a bit involved
simply becaue there are multiple moving parts to it and the path to a
solution that even passes the first test is a bit long.

This placement is a general feeling from how much work I think one would
need to do on the input before being able to get the expected output:

In tournament, you have to look at each line, parse it, look at the
result, increment two win/draw/loss counts (one for each team that
played). then at the end calculate scores and sort.

wordy also has string parsing, so let's try right after wordy.

Closes #174


IanWhitney:
Ok

PR #201, opened 2016-09-11
Implement Pascals Triangle

IanWhitney:
Following the [official tests](https://github.com/exercism/x-common/blob/f04888ed72009304844f0eb4d63a3c958dc39a07/exercises/pascals-triangle/canonical-data.json)

We can't test negative or no rows because the function declares u32 in its signature.


IanWhitney:
Ok. I'm happy with the implementation, so it's time to hear about all the dumb mistakes I made in it.

As for position, I think this can go fairly early. The weirdest part of it is returning a Vec of Vecs, and that's not _that_ weird. The meat of this problem is just math, all of which I cribbed from Wikipedia.

Other implementations of this I've seen use the values from the previous row to calculate the current row, which means handling out-of-bounds vec indexes. We'll want to take that into account when determining position.

IanWhitney:
Oh, and all of the number casting. That is kinda weird, if you're used to a language that is more forgiving with numeric types.

IanWhitney:
API question:

```
PascalsTriangle::new(3).rows()
```

or 

```
PascalsTriangle::with_rows(3)
```

or

```
PascalsTriangle::rows(3)
```

exercises/pascals-triangle/example.rs
```diff
@@ -0,0 +1,25 @@
...
+    pub fn rows(&self) -> Vec<Vec<u32>> {
+        (0..self.row_count).map(|row| PascalsTriangle::row(row)).collect()
+    }
+
+    pub fn row(number: u32) -> Vec<u32> {
+        let mut r = vec![1];
+
+        for p in 1..(number + 1) {
+            if let Some(last) = r.pop() {
+                let next = last as f32 * ((number as f32 + 1f32 - p as f32) / p as f32);
```
petertseng: Since you thought all the numeric casting was weird, perhaps you would like to remove it, if you were able? After all, the entries in Pascal's triangle are all integers, and the formula for doing this calculation mentions no rounding of any sort, so that implies it should be possible without resorting to floating point numbers.

then someone may rightfully complain "But written as-is, we do have to cast to floats, because `(number + 1 - p) / p` may be fractional!"

True, but a * (b / c) is equal to (a * b) / c.

petertseng: Someone may then righfully say  "(a * b) / c leaves us more susceptible to integer overflow" though, which I can't deny! Even using u64 will only stave that off for so long.

IanWhitney: I am missing some math concept here. 

I had to introduce the floats so that it would properly determine the 3rd element in row 2. With ints, the formula looked like

```rust
let next = 2 * ((2 + 1 - 2) / 2);
```

or

```rust
let next = 2 * (1 / 2);
```

With ints, `1/2` returns 0. So my 2nd row became `[1,2,0]`

What am I missing here?

petertseng: `let next = (2 * 1) / 2` is 1.

petertseng:
I don't think I would mind any of the three proposed APIs.

I would place no later than Roman numerals. Probably no earlier than, say... Hamming.

exercises/pascals-triangle/example.rs
```diff
@@ -0,0 +1,25 @@
...
+        (0..self.row_count).map(|row| PascalsTriangle::row(row)).collect()
+    }
+
+    pub fn row(number: u32) -> Vec<u32> {
+        let mut r = vec![1];
+
+        for p in 1..(number + 1) {
+            if let Some(last) = r.pop() {
+                let next = (last * (number + 1 - p)) / p;
+                r.extend(&[last, next])
```
petertseng: you could use `r.last()` so that you can just `push(next)` instead of having to pop the last element and put it back on with extend.

IanWhitney: In my testing, `last` returns a borrow, so I can't mutate the vector inside the block.

Example: https://play.rust-lang.org/?gist=4420b3f3b8fa09fbeb78b8eb2b3ca99c&version=stable&backtrace=0

petertseng: Hmm. I wonder. I thought that was supposed to work, and then it didn't. I quickly tested and found that with the addition of a strategically-placed `&`, it does work:

```rust
    if let Some(&y) = x.last() {
        x.push(y + 5);
    }
```

https://play.rust-lang.org/?gist=a3ace6f29c85b05e9f5027d31d0b4072&version=stable&backtrace=0

I may need a moment to try to puzzle out why one works and not the other. If someone has assistance to offer in this regard, I sure would appreciate it.

petertseng: Though obviously we need not let that hold up this PR.

IanWhitney: Just me mis-reading the docs. `last()` returns `Option<&T>`, not `Option<T>`.

PR #203, opened 2016-09-14
react: example: Check all dependents to avoid partial write

### LINKED ISSUE #202: react: example solution adds dependency incorrectly if given valid then invalid dependency ID
https://github.com/exercism/xrust/pull/191/files#r78686049

petertseng:
Fixes #202


petertseng:
I could write a test to deal with this, but this is quite an implementation detail. I will add the test and see what reviewers think.

petertseng:
Test added. Example solution fails the attached test unless it has the attached code change. It's not guaranteed that this will catch every single implementation that incorrectly does partial writes, but it's the best I can do, I think.

In case you wonder, this test doesn't seem necesary to add to https://github.com/exercism/x-common/pull/358 because most other tracks don't have to deal with cell IDs like we do; they can just have the cells reference each other (we didn't do that in Rust because mutual references are tricky).

PR #204, opened 2016-09-14
Implement Acronym

gregcline:
I implemented the tests based on the canonical suite and wrote an example solution. I haven't added it to config.json or problems.md yet since I wasn't sure where you would want it.

### DIFF
--- a/exercises/acronym/example.rs
+++ b/exercises/acronym/example.rs
@@ -1,26 +1,19 @@
-///
-/// Techies love their TLA (Three Letter Acronyms)!
-///
-/// Help generate some jargon by writing a program that converts a long name
-/// like Portable Network Graphics to its acronym (PNG).
-#[allow(unused_variables)]
 pub fn abbreviate(phrase: &str) -> String {
     phrase.split(|c: char| c.is_whitespace() || !c.is_alphanumeric())
           .flat_map(|word| split_camel(word))
           .filter(|word| !word.is_empty())
           .map(|word| word.split_at(1).0.to_uppercase())
-          .collect::<String>()
+          .collect()
 }
 
 fn split_camel(phrase: &str) -> Vec<String> {
     let chars: Vec<char> = phrase.chars().collect();
     let mut words: Vec<String> = Vec::new();
-    let mut prev_ind: usize = 0;
+    let mut word_start: usize = 0;
     for (i, c) in chars.iter().enumerate() {
-        if i > 0 && i == chars.len() - 1 ||
-           i > 0 && c.is_lowercase() && chars[i + 1].is_uppercase() {
-            words.push(chars[prev_ind..i + 1].iter().cloned().collect());
-            prev_ind = i + 1;
+        if i == chars.len() - 1 || c.is_lowercase() && chars[i + 1].is_uppercase() {
+            words.push(chars[word_start..i + 1].iter().cloned().collect());
+            word_start = i + 1;
         }
     }
     words
--- a/exercises/acronym/example.rs
+++ b/exercises/acronym/example.rs
@@ -1,9 +1,9 @@
 pub fn abbreviate(phrase: &str) -> String {
     phrase.split(|c: char| c.is_whitespace() || !c.is_alphanumeric())
           .flat_map(|word| split_camel(word))
-          .filter(|word| !word.is_empty())
-          .map(|word| word.split_at(1).0.to_uppercase())
-          .collect()
+          .filter_map(|word| word.chars().next())
+          .collect::<String>()
+          .to_uppercase()
 }
 
 fn split_camel(phrase: &str) -> Vec<String> {
--- a/exercises/acronym/example.rs
+++ b/exercises/acronym/example.rs
@@ -17,7 +17,8 @@ fn split_camel(phrase: &str) -> Vec<String> {
     let mut words: Vec<String> = Vec::new();
     let mut prev_ind: usize = 0;
     for (i, c) in chars.iter().enumerate() {
-        if i > 0 && i == chars.len() - 1 || i > 0 && c.is_lowercase() && chars[i + 1].is_uppercase() {
+        if i > 0 && i == chars.len() - 1 ||
+           i > 0 && c.is_lowercase() && chars[i + 1].is_uppercase() {
             words.push(chars[prev_ind..i + 1].iter().cloned().collect());
             prev_ind = i + 1;
         }
--- a/exercises/acronym/example.rs
+++ b/exercises/acronym/example.rs
@@ -1,0 +1,28 @@
+///
+/// Techies love their TLA (Three Letter Acronyms)!
+///
+/// Help generate some jargon by writing a program that converts a long name
+/// like Portable Network Graphics to its acronym (PNG).
+#[allow(unused_variables)]
+pub fn abbreviate(phrase: &str) -> String {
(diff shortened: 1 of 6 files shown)


exercises/acronym/example.rs
```diff
@@ -0,0 +1,27 @@
+///
+/// Techies love their TLA (Three Letter Acronyms)!
+///
+/// Help generate some jargon by writing a program that converts a long name
+/// like Portable Network Graphics to its acronym (PNG).
+#[allow(unused_variables)]
```
petertseng: this can be removed from the example, right? The example probably does not have any remaining unused_variables.

exercises/acronym/example.rs
```diff
@@ -0,0 +1,27 @@
+///
+/// Techies love their TLA (Three Letter Acronyms)!
```
petertseng: I think it is possible this can get removed from the example, since it won't get shown to students and this information is in the README already.

exercises/acronym/example.rs
```diff
@@ -0,0 +1,27 @@
...
+///
+/// Help generate some jargon by writing a program that converts a long name
+/// like Portable Network Graphics to its acronym (PNG).
+#[allow(unused_variables)]
+pub fn abbreviate(phrase: &str) -> String {
+    phrase.split(|c: char| c.is_whitespace() || !c.is_alphanumeric())
+          .flat_map(|word| split_camel(word))
+          .filter(|word| !word.is_empty())
+          .map(|word| word.split_at(1).0.to_uppercase())
+          .collect::<String>()
```
petertseng: I think the `::<String>`` is not necessary since the compiler knows by using the fact tha `abbreviate` is `-> String`.

That said I dont mind if it stays.

exercises/acronym/tests/acronym.rs
```diff
@@ -0,0 +1,37 @@
+extern crate acronym;
+
+#[test]
+fn basic() {
+  assert_eq!(acronym::abbreviate("Portable Network Graphics"), "PNG");
+}
+
+#[test]
+// #[ignore]
```
petertseng: don't forget to uncomment

exercises/acronym/example.rs
```diff
@@ -0,0 +1,27 @@
...
+}
+
+fn split_camel(phrase: &str) -> Vec<String> {
+    let chars: Vec<char> = phrase.chars().collect();
+    let mut words: Vec<String> = Vec::new();
+    let mut prev_ind: usize = 0;
```
petertseng: Can this be named for clarity? I imagine a future reader of this code may ask "index of *what*"?

exercises/acronym/example.rs
```diff
@@ -0,0 +1,27 @@
...
+    for (i, c) in chars.iter().enumerate() {
+        if i > 0 && i == chars.len() - 1 ||
```
petertseng: Would it be better to deduplicate the `i > 0` and only write it once, or would that make the boolean expression too hard to read, do you think? I tried it out and it doesn't look too bad, but what is your opinion on it?

Can you explain why the `i > 0` needs to be there? It looks like `i == 0` only affects words such as `eMail`, right? Does any word cause this function to have unexpected behaviour without the check for `i > 0` ?

petertseng:
Placement: I would definitely place no earlier than difference-of-squares, and definitely no later than all-your-base. I don't have anything more specific yet (which means we can just pick something in that range), unless someone wants to convince me otherwise.

I thought placing right after or right before hamming seemed right since they both iterate over the characters in a string. However, it also seems that it's clear from reading the README how to do Hamming, whereas it's not so clear for acronym. Acronym definitely has more cases to think about than hamming. So I might actually say to go for the later end of that range, like maybe between etl and sieve. I can always be convinced otherwise.

exercises/acronym/example.rs
gregcline: You're right. This and the type hint on the `collect()` were left over from previous implementations I tried and no longer need to be there.

exercises/acronym/example.rs
```diff
@@ -0,0 +1,20 @@
+pub fn abbreviate(phrase: &str) -> String {
+    phrase.split(|c: char| c.is_whitespace() || !c.is_alphanumeric())
+          .flat_map(|word| split_camel(word))
+          .filter(|word| !word.is_empty())
```
petertseng: I wondered whether any element in the vec returned by `split_camel` can ever be empty, but I don't think it can, because that implies `word_start == i + 1`, and that can never be because `word_start` is only ever set to a previous iteration's `i + 1` (and previous iteration's `i` is less than this iteration's `i`) or zero. Can you come up with a situation where it would?

You may ask "is it a concern that I had to think about this", and maybe it isn't - knowing that no word is empty will prevent panics on the `split_at(1)`, so maybe this is to make abundantly clear that it won't panic. So I appreciate that it prevents the reader form having to worry about that.

For these reasons, I don't mind if this line stays, I leave it up to you.

If you want to remove it but are concerned that now the reader needs to expend effort to ensure that we don't have an empty word, I can suggest that:

* You only want the first character of each word
* https://stackoverflow.com/questions/37635978/how-do-i-access-the-first-character-of-a-string-in-rust tells us a way to do that that returns an `Option`, so the reader can be sure that the case is covered.

gregcline: You're totally right. I actually included that to deal with issues like `PHP: Hypertext Preprocessor`  where `.split()` would result in empty strings from having two characters I split on next to each other before I added `split_camel`. But since `split_camel` is in a flat map now and will return an empty list in the case of an empty string it effectively filters out those empty strings.

I think adding that using `.chars().next()` would be a good way to make it clear that we are handling the empty string case since the fact that `split_camel` dealt with it wasn't clear even to me. And I would like including `Option` handling just to reinforce the concept since it's so different from the `try catch` pattern.

gregcline: Do you think this line is unclear as an alternative?
```rust
.filter_map(|word| word.chars().next().map(|c| c.to_uppercase().collect::<String>()))
```
It deals with the possibility of an empty word, but it ends up being kind of verbose. I didn't realize that `char.to_uppercase()` returns an iterator rather than just a `char`.

petertseng: Interesting that `char.to_uppercase()` does that. I assume that this line replaces the `filter` and `map` lines, but not the `collect` line. So now there are two `collect`! What if we tried to avoid that by taking advantage of `String#to_uppercase`(https://doc.rust-lang.org/std/string/struct.String.html#method.to_uppercase) ? (You probably have to put the type hint back, that's fine).

Ultimately I'm OK with that `filter_map` as-is but let's see if we can `collect` fewer times, eh?

IanWhitney:
Sorry that it took me a while to get to this PR. Thanks for all the work, @gregcline!

gregcline:
Thank you @petertseng for the guidance. You made me first real PR a very good experience!

petertseng:
Nicely done! Congratulations on your first PR!

:tada:

IanWhitney:
It can take a few days for new problems to show up in the site, by the way. I usually keep an eye on it and let people know when new problems are deployed.

kytrinyx:
> It can take a few days for new problems to show up in the site, by the way.

Shoot—that's my fault. I think I need to figure out if someone is willing to help out with this, or if we can find a way to auto-deploy.

PR #206, opened 2016-09-17
An equilateral triangle is a special case of an isosceles triangle.

elahn:
An isosceles triangle is a triangle with (at least) two equal sides. A triangle with all sides equal is called an equilateral triangle, and a triangle with no sides equal is called a scalene triangle. An equilateral triangle is therefore a special case of an isosceles triangle having not just two, but all three sides and angles equal.

http://mathworld.wolfram.com/IsoscelesTriangle.html

### DIFF
--- a/exercises/triangle/tests/triangle.rs
+++ b/exercises/triangle/tests/triangle.rs
@@ -23,7 +23,7 @@ fn equilateral_triangles_have_equal_sides() {
     let sides = [2, 2, 2];
     let triangle = Triangle::build(sides).unwrap();
     assert!(triangle.is_equilateral());
-    assert!(!triangle.is_isosceles());
+    assert!(triangle.is_isosceles());
     assert!(!triangle.is_scalene());
 }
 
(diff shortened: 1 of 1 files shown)


IanWhitney:
Thanks for the PR! Much appreciated.

The Rust implementation of the triangle exercise follows the [current canonical test suite](https://github.com/exercism/x-common/blob/9a0cc1b96f2f9c8f7614f1337d8f58bd05a939b9/exercises/triangle/canonical-data.json), which doesn't do these sorts of tests. 

Instead of changing only our tests, I suggest a PR against the standard test suite. If that PR is merged, then all tracks that implement the Triangles can improve their tests.

IanWhitney:
@elahn I started a conversation on the topic here: https://github.com/exercism/x-common/issues/379

elahn:
This should be reopened now, right?

IanWhitney:
I'd like to wait for the PR in x-common to get merged first.

PR #209, opened 2016-09-20
Bug fix for the last test of `Allergies`

k0pernicus:
The last test must not contains false positives, but 509 is greater than 255 so `expected` must contain `Allergen::Peanuts`.

### DIFF
--- a/exercises/allergies/tests/allergies.rs
+++ b/exercises/allergies/tests/allergies.rs
@@ -128,6 +128,7 @@ fn allergic_to_everything() {
 #[ignore]
 fn scores_over_255_do_not_trigger_false_positives() {
     let expected = vec![Allergen::Eggs,
+                        Allergen::Peanuts,
                         Allergen::Shellfish,
                         Allergen::Strawberries,
                         Allergen::Tomatoes,


petertseng:
I think the test is correct as-is - 509 & 255 is 252, so the result for 509 should be the same as the result for 253, and the result for 253 does not contain Peanuts

petertseng:
I think the test is correct as-is - 509 & 255 is 253, so the result for 509 should be the same as the result for 253, and the result for 253 does not contain Peanuts

petertseng:
This test is present in the Rust track because it is present in the common test data at https://github.com/exercism/x-common/blob/master/exercises/allergies/canonical-data.json - if we wish to change it, it is best to change it there so that all tracks can update their tests

A discussion for numbers > 255 is at https://github.com/exercism/x-common/issues/224

k0pernicus:
Ok so write directly in the README file that if the numbers is greater than 255, you have to make a `&` with it?

I think it exists a better solution to this end, returning a `Result` type creating the structure `Allergies`.
This type can return the `Self` object for Allergies if the `uid` given as parameter is correct - else an `Error` type and we can't continue after that?

petertseng:
Yes, we should discuss at https://github.com/exercism/x-common/issues/224 whether the correct solution is to `& 255` or say that numbers greater than 255 are errors.

IanWhitney:
I suggest we close this until https://github.com/exercism/x-common/issues/224 is resolved

Issue #194, opened 2016-09-01
[question] why `extern crate hamming as dna;`


ashleygwilliams:
i am interested in the reasoning behind the line `extern crate hamming as dna;`.

IanWhitney:
It was called `point_mutations` and the original implementer aliased the crate name as `dna`, probably just for brevity

https://github.com/exercism/xrust/commit/6895c352c5f8deb2c503410546f9db94fb2e54ec

Exercism then changed the name of this exercise to `hamming` and when I updated the tests I just did a find & replace.

In short, no real good reason.

petertseng:
In addition to what's said, it seems that we may not want to talk about DNA in Hamming exercise https://github.com/exercism/x-common/issues/323

So, it is probably appropriate to remove the `as dna`

Henning-K:
I'd like to take this up as it seems like a relatively easy fix and good exercise for my first pull request.

petertseng:
Nicely done!

ashleygwilliams:
cool! :tada:

PR #211, opened 2016-09-30
Fixes issue #194.

Henning-K:
This fixes issue #194 as the unnecessary alias
`extern crate hamming as dna`
is shortened to just the extern crate statement along with the then necessary changes to the used module name in the asserts.


IanWhitney:
Thanks! Travis was [failing with Rust builds for a bit](https://github.com/travis-ci/travis-ci/issues/6654#issuecomment-250739509), but may be fixed now. I'm going to re-run tests.

IanWhitney:
Looking good now. Thanks!

PR #212, opened 2016-09-30
Added .gitignore files to fix #192

### LINKED ISSUE #192: include a gitignore file in exercises

Henning-K:
This PR attempts to fix issue #192 by adding .gitignore files (following the standard Rust way as generated by cargo for every new project) to all the exercises' subdirectories.

These only contain one line, namely:
`target`
Maybe by creating future subdirectories of exercises directly with cargo, #192 won't occur again.


Henning-K:
Oh dear, I just saw that there is already a fix for this issue which I apparently read right over when starting this PR. It also seems the other PR is more complete in they realized and kept in mind that what users/students write as a solution is a library and not an executable AND that those two (executable, lib) have different standardized .gitignore files. As such I'll close this and hope the other PR gets merged soon so that the referenced issue ( #192 ) can be closed.

PR #213, opened 2016-10-02
Implement Bowling

IanWhitney:
Implement the bowling exercise, following the canonical test suite (along with https://github.com/exercism/x-common/pull/418 which has not been merged at the time of this PR)

- [X] Tests -- details in 03ae9ec
- [x] Example Implementation
- [x] Placement in config.json
- [x] Placement in problems.md
- [x] Ignore all but first test


exercises/bowling/tests/bowling.rs
```diff
@@ -0,0 +1,262 @@
...
+}
+
+fn you_can_not_roll_more_than_ten_pins_in_a_single_roll() {
+    let game = BowlingGame::new();
+
+    for x in (0..19) {
+        game.roll(0);
+    }
+
+    game.roll(11);
```
petertseng: how about checking that this `is_err()`?

exercises/bowling/tests/bowling.rs
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    assert!(game.score().is_err());
+}
+
+fn you_can_not_score_a_game_with_more_than_ten_frames() {
+    let game = BowlingGame::new();
+
+    for x in (0..21) {
+        game.roll(0);
```
petertseng: I would have thought that the first twenty would be `is_ok` and the last one would be `is_err`

exercises/bowling/tests/bowling.rs
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    assert!(game.score().is_err());
```
petertseng: I would have thought that this game *can* be scored (its score is zero) as soon as the twentieth ball is rolled. The twenty-first ball should not be allowed and should not invalidate the game from being scored.

exercises/bowling/tests/bowling.rs
```diff
@@ -0,0 +1,262 @@
...
+}
+
+fn if_the_last_frame_is_a_spare_you_can_not_create_a_score_before_extra_roll_is_taken() {
+    let game = BowlingGame::new();
+
+    for x in (0..17) {
```
petertseng: given that this is 17, the last frame isn't a spare. The ninth frame gets a 5, the tenth frame gets a 5.

exercises/bowling/tests/bowling.rs
```diff
@@ -0,0 +1,262 @@
...
+    game.roll(5);
+    game.roll(5);
+
+    assert!(game.score().is_err());
+}
+
+fn if_the_last_frame_is_a_strike_you_can_not_create_a_score_before_extra_rolls_are_taken() {
+    let game = BowlingGame::new();
+
+    for x in (0..17) {
```
petertseng: given that this is 17, rolling a 10 gives us a spare on the ninth frame instead of a strike on the tenth

exercises/bowling/tests/bowling.rs
```diff
@@ -0,0 +1,262 @@
...
+        game.roll(0);
+    }
+
+    game.roll(5);
+    game.roll(5);
+
+    assert!(game.score().is_err());
```
petertseng: below, you test that `game.roll(10)` will afterward allow `game.score().is_ok()`. Should a similar thing also be tested here?

exercises/bowling/tests/bowling.rs
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    assert_eq!(game.score().unwrap(), 31);
+}
+
+#[test]
+fn if_the_last_frame_is_a_spare_you_get_one_extra_roll_that_is_scored_once() {
+    let game = BowlingGame::new();
+
+    for x in (0..17) {
```
petertseng: 17 zeroes followed by two 5's makes a 5 in the ninth frame and a 5 in the tenth frame

exercises/bowling/tests/bowling.rs
```diff
@@ -0,0 +1,262 @@
...
+    assert_eq!(game.score().unwrap(), 17);
+}
+
+#[test]
+fn a_strike_earns_ten_points_in_a_frame_with_a_single_roll() {
+    let game = BowlingGame::new();
+
+    game.roll(10);
+
+    for x in (0..17) {
```
petertseng: with 17 balls following the strike, the game is not yet finished

exercises/bowling/tests/bowling.rs
```diff
@@ -0,0 +1,262 @@
...
+fn points_scored_in_the_two_rolls_after_a_strike_are_counted_twice_with_consecutive_strikes() {
+    let game = BowlingGame::new();
+
+    game.roll(10);
+    game.roll(10);
+    game.roll(10);
+    game.roll(5);
+    game.roll(3);
+
+    for x in (0..11) {
```
petertseng: there are 6 frames left at this point, so 11 rolls isn't enough for them (0..11 is exclusive of 11 so this rolls 11 balls not 12)

exercises/bowling/tests/bowling.rs
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    assert_eq!(game.score().unwrap(), 81);
+}
+
+#[test]
+fn if_the_last_frame_is_a_strike_you_get_two_extra_rolls() {
+    let game = BowlingGame::new();
+
+    for x in (0..17) {
```
petertseng: 17 doesn't make the last frame a strike, this makes frame 9 a strike

exercises/bowling/tests/bowling.rs
```diff
@@ -0,0 +1,262 @@
...
+    game.roll(1);
+
+    assert_eq!(game.score().unwrap(), 18);
+}
+
+#[test]
+fn strikes_in_extra_rolls_after_a_strike_in_the_final_frame_do_not_get_the_bonus() {
+    let game = BowlingGame::new();
+
+    for x in (0..17) {
```
petertseng: 17 followed by spare in 9th, strike in 10th, one extra roll - second extra roll hasn't been rolled yet so game wouldn't be complete

exercises/bowling/tests/bowling.rs
```diff
@@ -0,0 +1,262 @@
...
+    game.roll(10);
+    game.roll(10);
+
+    assert_eq!(game.score().unwrap(), 30);
+}
+
+fn all_strikes_is_a_perfect_score_of_300() {
+    let game = BowlingGame::new();
+
+    for x in (0..11) {
```
petertseng: a perfect game is 12 strikes not 11

exercises/bowling/tests/bowling.rs
```diff
@@ -0,0 +1,262 @@
...
+
+fn you_can_not_roll_more_than_ten_pins_in_a_single_frame() {
+    let game = BowlingGame::new();
+
+    for x in (0..18) {
+        game.roll(0);
+    }
+
+    game.roll(5);
+    game.roll(6);
```
petertseng: how about checking that this `is_err()`?

IanWhitney:
In regards to error handling, there seem to be a few options.

1. `score` returns a `result`.

This is the approach I started with. Most of the boundaries tested deal with the game as a whole, not individual rolls, so I figured it made sense to put the `result` in the `score` function.

2. Have `roll` return a result

As suggested [here](https://github.com/exercism/xrust/pull/213#discussion_r81464170). I'm not sure what the idioms of Rust say here, but in my mind `roll` is a command, so having it return something seems to violate [command-query separation](https://en.wikipedia.org/wiki/Command–query_separation)

A way we could have `roll` return something would be to follow the approach we took in [robot simulator](https://github.com/exercism/xrust/blob/master/exercises/robot-simulator/example.rs) and make the game immutable. Each call to `roll` would return a new `Result<BowlingGame, Err>`.

petertseng:
I hope the above options aren't presented as mutually exclusive. It could be the case that both should return Result!

But before we judge whether that should be true, let me state my reasoning and motivations, then we can better decide.

So! Can we fail to take a score? Yes that can fail if the game is not complete yet. So `score` could return `Option<uint>`, or `Result<uint, E>` where the `E` explains why we can't take the score (the `something` is likely not checked in the tests; we would likely just test `is_err`).

Okay, what about individual rolls. Can individual rolls be invalid? It seems so.

* If I try to roll 11
* If I roll a 7 in the same frame as a 5 (there aren't that many pins there)
* If I roll anything after twenty rolls (and there are no extra rolls)

Well, given that there are some invalid individual rolls, how do we deal with them? How is that reflected in the API of this bowling game? Let's say that `roll` returns nothing (and therefore returns no indication of whether the last roll was valid). So let's suppose that I call `roll(5)`, `roll(6)`, then `roll(0)` 18 times. Okay, that's 20 rolls, that's a full game, so I can take the score now, right? Now I try to take the `score` and find that I get an error. Argh! In the absolute best case, the error result from `score` would be so kind to tell me that my `roll(6)` way back there was invalid, but in the worst case it might not, and I would be left to my own devices to figure out what happened, since there was no feedback on any of the individual rolls.

I think the general principle I'm trying to discuss is that it is less confusing if errors are reported as soon as possible, rather at some unspecified time in the future. This makes it easier for me to figure out what went wrong, instead of having to figure out the source of a seemingly-unrelated failure to `score`.

So what does this say about what `roll` should return? To me I did think this meant it should return `Result<(), E>` (again, the contents of the `E` would not be checked during testing).

But nothing I've said about `score` or `roll` needs to be set in stone - anything that meets the goals described here would do!

IanWhitney:
If `roll` returned a `Result`, what would the test look like? In my mind, they look like:

```rust
fn you_can_not_roll_more_than_ten_pins_in_a_single_roll() {
 let game = BowlingGame::new();
 assert!(game.roll(11).is_err());
}
```

But maybe you have something else in mind. 

The advantage, as you mentioned, is that we'd know right away that we have an error.

But, I think that if we went this way then the first time students would see this behavior is in the exception tests. I don't think our earlier tests would have code like:

```rust
 for x in (0..20) {
  game.roll(0).expect("Invalid Roll"); //or some other way of handling Result.
}
```

Or do you think they should?

This approach would also combine two behaviors into one method call: mutating state and asserting if the mutation was successful. There definitely are methods in Rust that do this ([write](https://doc.rust-lang.org/std/io/trait.Write.html#tymethod.write) for example), so it's certainly idiomatic. But I think it's also idiomatic to consistently wrap those method calls in something that handles the Result, so we'd want to do it throughout the tests.

petertseng:
> ```rust
> assert!(game.roll(11).is_err());`
> ```

Yeah, that's what I think as well.

> But, I think that if we went this way then the first time students would see this behavior is in the exception tests.

Ah, this is a bit unfortunate. And we can't even write plain `game.roll(0)` because `Result` is `must_use`, so the least intrusive thing we might conceivably do is `let _ = game.roll(0)`

I remember we had similar conversations in #89.

> Is there a way in Rust to have students avoid using a Result value until they reach the test that introduces error handling? Because that's what I think would be best in terms of learning curve. I'd love students' initial implementations to be simple,

That from https://github.com/exercism/xrust/pull/89#issuecomment-202462374 seems very cromulent here. Perhaps it's worth revisiting that and seeing whether there's anything that's applicable to this exercise. I don't have the answers, but I will have a think when I can.

petertseng:
So I guess what happened back then was a similar thing - after deciding that it was possible for queen creation to fail, but queen creation will be used in all of the other tests, it made sense then to reorder the tests to test queen creation first.

> A way we could have `roll` return something would be to follow the approach we took in [robot simulator](https://github.com/exercism/xrust/blob/master/exercises/robot-simulator/example.rs) and make the game immutable. Each call to `roll` would return a new `Result<BowlingGame, Err>`

This would be interesting too! A fun part is the act of "roll zero 18 times", taking the previous `Result` and turning into the next. We could make [`foldM`](http://hackage.haskell.org/package/base-4.9.0.0/docs/Control-Monad.html#v:foldM) ! (If you ask me whether I'm serious about that, my answer is always "it's as serious as you'd like it to be")

IanWhitney:
In a hectic pre-release period for a project, so I don't have a bunch of time to think about Bowling right now. I'm thinking the Queen Attack approach makes the most sense here.

IanWhitney:
A couple of thoughts on how to write the tests, now that `roll` will return a result

Reading the [Must Be Used](https://doc.rust-lang.org/std/result/#results-must-be-used) doc that was linked earlier, the tests could read

```rust
for x in (0..20) {
  game.roll(0).expect("roll failed");
}
```

or

```rust
for x in (0..20) {
  game.roll(0).is_ok(); //when I tried this in playground it was fine, even though the docs suggest putting this in an `assert!`
}
```

The `let _ = game.roll(0)` approach also works, but I don't think it's the way I want to go. I'd rather use the standard Result API.

petertseng:
I imagine (but have not verified) that if using `game.roll(0).is_ok()` and the `roll` function unexpectedly returns an `err`, it will go by and not panic, since `is_ok` only convert it a `bool`. Whether we are all right with that can depend, but there are probably a few tests for which we will care.

IanWhitney:
> I imagine (but have not verified) that if using game.roll(0).is_ok() and the roll function unexpectedly returns an err, it will go by and not panic, since is_ok only convert it a bool. Whether we are all right with that can depend, but there are probably a few tests for which we will care.

This [is correct](https://play.rust-lang.org/?gist=497060e05160bbd379dc5ddeda4fa3c2&version=stable&backtrace=0) 

I think I can get this finished up today, now that the new canonical tests are merged into master.

IanWhitney:
Ok. If those tests look good I can maybe finally get around to implementing this dang thing.

exercises/bowling/tests/bowling.rs
```diff
@@ -0,0 +1,320 @@
...
+
+    assert!(game.score().is_err());
+}
+
+#[test]
+fn a_game_score_is_err_if_more_than_ten_frames_have_been_rolled() {
+    let game = BowlingGame::new();
+
+    for _ in 0..11 {
+        game.roll(0).is_ok();
```
petertseng: I suggest asserting `is_err` on ones rolled after the tenth frame

exercises/bowling/tests/bowling.rs
```diff
@@ -0,0 +1,320 @@
...
+
+    assert!(game.roll(11).is_err());
+}
+
+#[test]
+fn a_game_score_is_ok_if_ten_frames_have_been_rolled() {
+    let game = BowlingGame::new();
+
+    for _ in 0..10 {
+        game.roll(0).is_ok();
```
petertseng: for this and others - the bool from `is_ok` is being discarded, and I think this is confusing because it makes the intent unclear. If we're not going to check whether it's OK or not, it seems clearer to say `let _ = game.roll(0);` so that there is no confusion about intent.

exercises/bowling/tests/bowling.rs
```diff
@@ -0,0 +1,320 @@
...
+
+    game.roll(10).is_ok();
+    game.roll(10).is_ok();
+    game.roll(10).is_ok();
+
+    assert_eq!(game.score().unwrap(), 30);
+}
+
+#[test]
+fn a_strike_with_the_one_roll_bonus_after_a_spare_in_the_last_fram_does_not_get_a_bonus() {
```
petertseng: s/fram/frame

exercises/bowling/tests/bowling.rs
```diff
@@ -47,20 +51,22 @@ fn a_game_score_is_err_if_fewer_than_ten_frames_have_been_rolled() {
 }
 
 #[test]
+#[ignore]
 fn a_game_score_is_err_if_more_than_ten_frames_have_been_rolled() {
```
IanWhitney: I thought about that, but decided it was probably unnecessary. But I could be convinced otherwise.

exercises/bowling/example.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+impl BowlingGame {
+    pub fn new() -> Self {
+        BowlingGame { frames: vec![Frame::new()] }
+    }
+
+    pub fn roll(&mut self, pins: u16) -> Result<(), &'static str> {
+        if pins > 10 {
+            Err("Greate than 10")
```
petertseng: optionally:

in the below section, you take the approach of using `return Err(...)` to avoid having to wrap everything in an `else`. Is that also appropriate to do here?

IanWhitney:
I wouldn't worry about the current example code; I only pushed it up so that it would be easier to talk about problem topics & placement.

IanWhitney:
Example refactored to something I'm ok with. That `bonus_valid()` code, though. Yikes.

I tried a couple of different approaches. Enum. Traits. Didn't like either of them. So I'm sticking with the single `Frame` struct.

The added tests cover the weird logic of fill balls after a final strike. There's a pull-request to update the common test suite. https://github.com/exercism/x-common/pull/418

petertseng:
This makes me realize a curious behaviour that we don't test - If an `Err` roll is made, can a subsequent `Ok` roll be made?

I believe the answer for this example solution is "no", and if I wanted to be principled about it I would say that the answer should be "yes" - it's less desirable to have an erroneous roll get us into a state from which we can't recover (except by making a new game). I tested a similar thing in `react` in #203. Maybe we don't care to test that in this problem.

petertseng:
> but I think it's the frame's responsibility to know what to do
> with a roll, not the game's responsibility to decide which frame to add
> a roll to.

Oh I agree with that. I simply choose to optimise it by saying that a frame knows when it will no longer need to accept any rolls, and that the game can take advantage of that information.

But this is not something that need go in this example, more something for the site.

> Yeah, it's always going to be the current frame & previous
> frame, I realize.

Well, sometimes the previous two, even!

IanWhitney:
> Well, sometimes the previous two, even!

Exactly why I don't want to think about it. I will get it wrong. Doing a filter on `!frame.is_complete()` and then adding the roll the remaining frames makes sense, though.

IanWhitney:
> I believe the answer for this example solution is "no", and if I wanted to be principled about it I would say that the answer should be "yes" - it's less desirable to have an erroneous roll get us into a state from which we can't recover (except by making a new game). I tested a similar thing in react in #203. Maybe we don't care to test that in this problem.

I think this gets into requirements beyond the problem. If we had an `undo` function, for example, we'd want to test that it worked.

IanWhitney:
Weird. I just pushed a bunch of stuff and Travis isn't picking it up.

IanWhitney:
I assume it's this http://time.com/4540265/internet-outages-ddos-dns/?xid=tcoshare

petertseng:
> I think this gets into requirements beyond the problem. If we had an `undo` function, for example, we'd want to test that it worked. 

Ah, you're right. `undo` certainly would increase the scope.

The proposed test, then, without `undo`:

```rust
assert!(game.roll(11).is_err());
assert!(game.roll(0).is_ok());
```

IanWhitney:
Is this good to go? There's the one example function I could clean up, but I can also live with it being in the example code.

exercises/bowling/example.rs
```diff
@@ -0,0 +1,136 @@
...
+                    if *first == 10 {
+                        self.bonus_score() <= 20
+                    } else {
+                        self.bonus_score() <= 10
+                    }
+                } else {
+                    true
+                }
+            } else {
+                true
```
petertseng: I expect this line (corresponding to "bonus done, but is neither a spare nor a strike") to be unreachable. Correspondingly, I'd prefer `unreachable!` to express that intention. What say you?

exercises/bowling/example.rs
```diff
@@ -0,0 +1,136 @@
...
+    fn rolls_valid(&self) -> bool {
+        self.roll_score() <= 10
+    }
+
+    fn bonus_valid(&self) -> bool {
+        if self.bonus_done() {
+            if self.is_spare() {
+                self.bonus_score() <= 10
+            } else if self.is_strike() {
+                if let Some(first) = self.bonus.iter().rev().last() {
```
petertseng: can you explain why `.rev().last()` instead of `.next()`?

petertseng:
I was hoping work could be done on `bonus_valid`, but unfortunately the only improvement I could make was to move the `bonus_done` to own line:

```rust
        if !self.bonus_done() {
            return true;
        }   

        if self.is_spare() {
            self.bonus_score() <= 10
        } else if self.is_strike() {
            ...
        } else {
            unreachable!("Neither a spare nor a strike?");
        }  
```

I'm not sure that's even an improvement anyway so maybe we should let it be. I'll leave that one up to you.

I do think that `unreachable` should go there though.

Other than that I have nothing else to change. I look forward to working on this one.

exercises/bowling/example.rs
IanWhitney: Because I thought the function was called `first`, which it's not. Thanks.

PR #214, opened 2016-10-02
attribute the rust icon (moved here from exercism.io)

catb0t:
a step in fixing https://github.com/exercism/exercism.io/issues/3112.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,3 +13,7 @@ Please see the [contributing guide](https://github.com/exercism/x-api/blob/maste
 The MIT License (MIT)
 
 Copyright (c) 2014 Katrina Owen, _@kytrinyx.com
+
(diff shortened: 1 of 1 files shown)


petertseng:
Better late than never (should have asked for it in #144)

Verified that:

* This is the same statement as https://github.com/exercism/exercism.io/blob/master/ATTRIBUTION.md
* The statement is true according to what is seen at https://www.rust-lang.org/en-US/legal.html
* It's in keeping with https://github.com/exercism/exercism.io/issues/3112

So let's go.

petertseng:
Oh, an afterthought... I believe the license stipulates that we should link to it. I'll do it after I deal with some other business. Please feel free to beat me to it.

PR #215, opened 2016-10-02
Follow the current canonical test set. Fixes #196

### LINKED ISSUE #196: final test in word counts exercise does not seem like a good idea
is there a class of symbols that this test is referring to? this doesn't seem like a good test. it requires that the student hardcode splitting on spaces AND underscores AND commas... and potentially Some Other Undefined "symbols". is there a reason you think this test is useful? this adds a lot of complexity that seems not terrible useful. i.e. this breaks if the sentence is "my twitter handle is @ag_dubs"

```rust
#[test]
#[ignore]
fn test_symbols_are_separators() {
    check_word_count(
        "hey,my_spacebar_is_broken.",
        vec![("hey", 1),
             ("my", 1),
             ("spacebar", 1),
             ("is", 1),
             ("broken", 1)]);
}
```

ijanos:


petertseng:
Yep, thanks!

Amended commit message to prefix with "word-count".

PR #217, opened 2016-10-05
Implemented atbash cipher

cynicaldevil:
cc @IanWhitney @petertseng 
I have lumped the whole thing into a single commit, but if you want smaller commits for each step, I can do that too.

### DIFF
--- a/config.json
+++ b/config.json
@@ -17,6 +17,7 @@
     "pangram",
     "nucleotide-count",
     "word-count",
+    "atbash-cipher",
     "etl",
     "acronym",
     "sieve",
--- a/exercises/atbash-cipher/example.rs
+++ b/exercises/atbash-cipher/example.rs
@@ -1,0 +1,36 @@
+use std::ascii::AsciiExt;
+
+fn ascii(ch: char) -> u8 {
+    ch as u8
+}
+
+fn get_transpose(ch: char) -> char {
+    if ch.is_digit(10) {
+        ch
+    } else {
+        (ascii('z') - ascii(ch) + ascii('a')) as char
+    }
+}
+
+pub fn encode(plaintext: &str) -> String {
+    plaintext
+    .to_lowercase()
+    .chars()
+    .filter(|&ch| ch.is_ascii())
+    .filter(|&ch| ch.is_alphanumeric())
+    .map(|ch| get_transpose(ch))
+    .collect::<Vec<char>>()
+    .chunks(5)
+    .map(|slice| slice.iter().cloned().collect::<String>())
+    .collect::<Vec<String>>()
+    .join(" ")
+}
+
+pub fn decode(ciphertext: &str) -> String {
+    ciphertext
+    .split::<char>(' ')
+    .collect::<String>()
+    .chars()
+    .map(|ch| get_transpose(ch))
+    .collect::<String>()
+}
--- a/exercises/atbash-cipher/tests/atbash-cipher.rs
+++ b/exercises/atbash-cipher/tests/atbash-cipher.rs
@@ -1,0 +1,82 @@
+extern crate atbash_cipher as cipher;
+
+#[test]
+fn test_encode_yes() {
+    assert_eq!("bvh", cipher::encode("yes"));
+}
+
+#[test]
+#[ignore]
+fn test_encode_no() {
+    assert_eq!("ml", cipher::encode("no"));
+}
+
+#[test]
+#[ignore]
+fn test_encode_omg() {
+    assert_eq!("lnt", cipher::encode("OMG"));
+}
+
+#[test]
+#[ignore]
+fn test_encode_spaces() {
+    assert_eq!("lnt", cipher::encode("O M G"));
+}
+
+#[test]
+#[ignore]
+fn test_encode_mindblowingly() {
+    assert_eq!("nrmwy oldrm tob", cipher::encode("mindblowingly"));
+}
+
+#[test]
+#[ignore]
+fn test_encode_numbers() {
+    assert_eq!("gvhgr mt123 gvhgr mt", cipher::encode("Testing,1 2 3, testing."));
+}
+
+#[test]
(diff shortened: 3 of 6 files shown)


IanWhitney:
Great! Thanks so much for the work. I have some suggestions that I'll add as line notes.

config.json
```diff
@@ -46,7 +46,8 @@
     "rectangles",
     "forth",
     "circular-buffer",
-    "react"
+    "react",
+    "atbash-cipher"
```
IanWhitney: We'll want to move this earlier in the problem track. Around word-count, I think.

exercises/atbash-cipher/example.rs
```diff
@@ -0,0 +1,39 @@
...
+        (ascii('z') - ascii(ch) + ascii('a')) as char
+    }
+}
+
+pub fn encode(plaintext: &str) -> String {
+    plaintext
+    .to_lowercase()
+    .chars()
+    .filter(|&ch| ch.is_alphanumeric())
+    .filter(|&ch| (ch as u8) < 128)
```
IanWhitney: `is_ascii` may simplify this: https://doc.rust-lang.org/std/primitive.str.html#method.is_ascii

exercises/atbash-cipher/example.rs
```diff
@@ -0,0 +1,39 @@
...
+    .map(|ch| get_transpose(ch))
+    .fold("".to_string(),|encoded_result, ch| {
```
IanWhitney: This works as a way to split the encoded string into groups of 5. Another way would be to build the string without spaces and then use `chunks` to split into 5-character chunks.

https://doc.rust-lang.org/std/primitive.slice.html#method.chunks

You may also be able to do the same thing with `split`

exercises/atbash-cipher/example.rs
```diff
@@ -0,0 +1,39 @@
...
+        } else {
+            encoded_result + &(ch.to_string())
+        }
+    })
+}
+
+pub fn decode(ciphertext: &str) -> String {
+    ciphertext
+    .chars()
+    .fold("".to_string(), |decoded_result, ch| {
```
IanWhitney: I think doing a filter to remove spaces & then a map to transpose would be more idiomatic. Saves you the `if`, at least.

exercises/atbash-cipher/tests/atbash-cipher.rs
```diff
@@ -0,0 +1,70 @@
...
+    assert_eq!("bvh", cipher::encode("yes"));
+}
+
+#[test]
+fn test_encode_no() {
+    assert_eq!("ml", cipher::encode("no"));
+}
+
+#[test]
+fn test_encode_OMG() {
```
IanWhitney: Rust gives a warning about this name. It would prefer `test_encode_omg`, which is fine.

problems.md
```diff
@@ -55,6 +55,7 @@ phone-number |  option, format, unwrap_or, iters, match
 wordy | Result, string parsing, operators (optional)
 tournament |  enum, sorting, hashmap, structs
 custom-set |  generic over type, vector, equality, struct
+atbash-cipher | str vs string, primitive types, iterators
```
IanWhitney: I'd add chars and ascii to this

cynicaldevil:
@IanWhitney I have implemented the changes you requested, and also added the `ignore` attribute to all the required tests.

IanWhitney:
As mentioned in another PR I'm a little time-crunched right now. I'd like to see what @petertseng says, but my (very cursory) look at this was OK.

exercises/atbash-cipher/example.rs
```diff
@@ -0,0 +1,36 @@
...
+    .collect::<Vec<String>>()
+    .join(" ")
+}
+
+pub fn decode(ciphertext: &str) -> String {
+    ciphertext
+    .chars()
+    .filter(|&ch| !ch.is_whitespace())
+    .fold("".to_string(), |decoded_result, ch| {
+            decoded_result + &(get_transpose(ch).to_string())
```
petertseng: given that the interior of this `fold` has been simplified greatly since the first iteration of this PR, does it now make sense to make this a `map` then `collect<String>` instead of a `fold`?

problems.md
```diff
@@ -55,6 +55,7 @@ phone-number |  option, format, unwrap_or, iters, match
 wordy | Result, string parsing, operators (optional)
 tournament |  enum, sorting, hashmap, structs
 custom-set |  generic over type, vector, equality, struct
+atbash-cipher | str vs string, primitive types, iterators, chars, ascii
```
petertseng: don't forget to move this too, we'd like it to be in the same place relative to other problems as it is in `config.json`

petertseng:
I think I just have those two things to say, and then this will be good to go.

cynicaldevil:
@petertseng Done!

exercises/atbash-cipher/example.rs
```diff
@@ -0,0 +1,36 @@
...
+    .chunks(5)
+    .map(|slice| slice.iter().cloned().collect::<String>())
+    .collect::<Vec<String>>()
+    .join(" ")
+}
+
+pub fn decode(ciphertext: &str) -> String {
+    ciphertext
+    .split::<char>(' ')
+    .map(|slice| slice.iter.cloned().collect::<String>())
```
petertseng: I think that needs to be an `iter()` instead of `iter`, otherwise it won't compile. almost there!

cynicaldevil:
Ah, my bad. I have removed the `map` and replaced it with more idiomatic code now.

petertseng:
Ah, even better! All right, everything is looking good so we are merging it! Thank you!

IanWhitney:
Yes! Thanks a lot @cynicaldevil! Next time the site is deployed your problem will be available. I'll publicize your hard work then. Much appreciated.

PR #218, opened 2016-10-15
replace 'problems' key with 'exercises' in config.json

cgrayson:
see #179


cgrayson:
Whoops, didn't see #195, never mind. 😃

PR #219, opened 2016-10-22
Implement grains

IanWhitney:
Follows the [standard test suite](https://github.com/exercism/x-common/blob/ca0c64ffd489ff53844699bc2d566bfa478b5592/exercises/grains/canonical-data.json)

I did not include the test of square -1 since that would require the
function signature to be `i32`.

Placement is at the end of the "Introduction" section. It introduces a
couple of concepts that may be new,

- declaring number types
- panic & should_panic

I thought it would be nice to get some exposure to `panic` before
introducing `result` in the Hamming exercise.


petertseng:
Seems pretty uncontroversial. Conscious decision to show panics .

PR #220, opened 2016-11-04
Added into_rust to list of learning materials for Rust

tbelahi:


docs/LEARNING.md
```diff
@@ -2,6 +2,7 @@
 
 * [The Rust Programming Language](http://doc.rust-lang.org/stable/book/) is a great resource for getting started with Rust as well as diving deeper into specific features of Rust.
 * [Rust by Example](http://rustbyexample.com) shows you examples of the most common things you will be writing in Rust.
+* [Into_rust()](http://intorust.com/) "Screencasts for learning Rust."
```
petertseng: I was about to say what's with the extra parentheses after rust, but it's part of the name at http://intorust.com/about/ so okay.

Are the quotes necessary?

(Then again, someone may ask why the quotes are there for Rustlings and #rust-beginners... maybe they should all go)

petertseng: This seems appropriate, then. It looks like the below were also direct quotes: https://www.rust-lang.org/en-US/community.html and https://github.com/carols10cents/rustlings also contain the exact quotes listed for the below entries. Since the used text is also a quote, we are now being consistent.

petertseng:
Thanks!

Note that I had to squash the two commits into one, since it would not have been understandable to have two commits that simply say "Update LEARNING.md"

PR #225, opened 2016-11-16
Add Alphametics exercise. Fixes #182

### LINKED ISSUE #182: Implement Alphametics
The problem is described in [its readme](https://github.com/exercism/x-common/blob/master/exercises/alphametics/description.md)
- [ ] Use the [standard test definition](https://github.com/exercism/x-common/blob/master/exercises/alphametics/canonical-data.json) as the basis for test/alphametics.rs file
- [ ] Create an `example.rs` solution
- [ ] Place the problem in `config.json`
- [ ] Place the problem and describe its topics in `problems.md`

ijanos:
I'm not sure where should we put this in problems.md. I think it belongs towards the end of "Getting Rusty". What do you think?


ijanos:
Hmm, it looks like travis fails on compiling example.rs due to missing dependencies. How should I fix that? Why is it even tries to compile that file?

IanWhitney:
Gigasecond is a good example of an exercise that requires crates. You need to list them in your Cargo.toml file

https://github.com/exercism/xrust/blob/3c48af8880211d3da2284f549bf45e13099bcab9/exercises/gigasecond/Cargo.toml

Sorry that it took me so long to get to this. Vacation + News = Me Avoiding My Computer.

ijanos:
The exercise is not about any specific crate. I just used two convenient creates for the example code. I feel providing a pre-filled [dependencies] would make people think they should use those and I want to avoid that.

IanWhitney:
Sure, but since your example code uses external crates it will not pass CI until those crates are declared as dependencies in the Cargo.toml file. 

When the CI runs it compiles your example.rs file and ensures that it passes all of the test cases. And that compilation can't happen unless the dependencies are declared.

ijanos:
I see that, but what should I do if I don't want those dependencies to show up in `Cargo.toml` when other people download the exercise? Because when you `exercism fetch` it will not download `example.rs` but it will download `Cargo.toml` and I don't see why should this file contain references to something that is not even downloaded, I fear that it will misguide people into the same solution. 

Maybe there should be a Cargo-example.toml file? I feel like this pull request is not the right place to discuss that, do you want me to open an issue somewhere for this? Or do You think this is a non-issue and those dependencies for example code should be there for everyone?

petertseng:
Funnily enough, I too have limited access so expect delays in responses from me.

> I feel providing a pre-filled [dependencies] would make people think they should use those and I want to avoid that.

I would prefer presenting a clean Cargo.toml to students, because I agree with your reasoning.

If there is a strong desire to have this PR merged quickly, I would find acceptable a Cargo.toml with the dependencies for now if an issue is filed to discuss removing it and we at least come up with a plan.

> I feel like this pull request is not the right place to discuss that, do you want me to open an issue somewhere for this?

It does seem to make sense to open a separate issue in this repo. Despite the fact that this PR is the first place where this problem comes up, it can be applicable to others as well so makes sense to have the separate issue.

I would wait to post the following info in that issue, but since I have limited availability, I'll just say it now:

In the Haskell track, package.yaml plays a similar role to Cargo.toml: You can declare dependencies there. I have some history to share.

In https://github.com/exercism/xhaskell/pull/189#r70165604 , I noted that an example was depending on a package, and this package would show up in the delivered package.yaml to students.

Eventually we have https://github.com/exercism/xhaskell/issues/420 and presented a clean package.yaml to students, without example's dependencies (example uses a separate package.yaml file)

We don't need to use the exact same solution, it only suffices that the example can use a separate Cargo.toml and that https://github.com/exercism/xrust/blob/master/_test/check-exercises.sh understands how to use it - I would suggest that if Cargo-example.toml exists, move it to Cargo.toml, otherwise do nothing - that way we don't have to maintain a whole bunch of unnecessary Cargo.toml files; we only have to maintain the ones that actually need a separate one.

exercises/alphametics/example.rs
```diff
@@ -0,0 +1,70 @@
...
+fn test_equation(puzzle: &str, substitutions: &HashMap<char, u8>) -> bool {
+    // Create a new String with characters changed to numbers
+    let puzzle: String = puzzle.chars().map(|c| {
+        if let Some(&n) = substitutions.get(&c) {
+            // If the character is in the substitutions, get the number and
+            // convert it to a char
+            char::from_digit(n as u32,10).unwrap()
+        } else {
+            // Otherwise just copy over the character
+            c
```
petertseng: do you ever expect this to happen if we call `test_equation`? If it does happen, then the `parse::<u32>().unwrap()` later on would fail, right? If so, I might consider just `unwrap()` here as well, or have an `unreachable!` in here.

exercises/alphametics/example.rs
```diff
@@ -0,0 +1,70 @@
...
+
+pub fn solve(puzzle: &str) -> Option<HashMap<char, u8>> {
+    // Get unique letters from the puzzle
+    let letters: HashSet<char> = puzzle.chars().filter(|&c| c.is_alphabetic() && c.is_uppercase()).collect();
+    let letters: Vec<char> = letters.into_iter().collect();
+
+    // All available numbers for substitution
+    let numbers: &[u8] = &[0, 1, 2, 3, 4, 5, 6, 7, 8, 9];
+
+    // Iterate every combination with the length of uniqe letters in the puzzle
```
petertseng: minor: spelling (uniqe -> unique)

petertseng:
> I think it belongs towards the end of "Getting Rusty"

I'll agree.

I can't think of many topics to add to https://github.com/exercism/xrust/blob/master/problems.md , I only have "string parsing", "math", "combinations" or something like that.

petertseng:
I'll be able to take a closer look after the 28th.

How are we doing on runtime? I only see in the test output that the last test ran for more than 60 seconds; is there someting that tell us how long does the entire thing take?

exercises/alphametics/example.rs
ijanos: The else branch will run every time an equation is parsed. This will convert a String like `"A + B == C"` to another String which looks like this `"1 + 2 == 3"`. The the function chops up this string later and parses the individual numbers.

petertseng: Oh, so it is for the `+` and `=` and spaces. Got it.

ijanos:
As mentioned in the comment on the top of the file, this is a rather naive brute force example which will run for a long time on large inputs, like the last test. On top of that I did not do any tricks to make this solution  faster in the hope of better readability.

It is not that bad on my machine though, definitely under 10 seconds. Running `cargo test --release` helps, I'm not sure which if travis runs tests in debug or release mode.

exercises/alphametics/tests/alphametics.rs
```diff
@@ -0,0 +1,64 @@
+extern crate alphametics;
+use std::collections::HashMap;
+
+fn test_helper(puzzle: &str, solution:  &[(char, u8)])  {
```
petertseng: perhaps just a single space between the `:` and `&`?

petertseng: is there (should there) be a better name for `test_helper`?

Then again, I don't have a particularly better one. `assert_solution_matches`?

exercises/alphametics/example.rs
```diff
@@ -0,0 +1,70 @@
...
+            let substitution: HashMap<char, u8> = letters.iter().zip(p).map(|(&c, &n)| (c, n)).collect();
+            if test_equation(puzzle, &substitution) {
+                // We found a good substitution
+                return Some(substitution);
+            }
+        }
+    }
+    // If we tested every combination and did not found a solution then return None
+    None
+}
```
petertseng: newline after?

exercises/alphametics/example.rs
```diff
@@ -0,0 +1,70 @@
...
+use std::collections::HashSet;
+use std::char;
+
+fn test_equation(puzzle: &str, substitutions: &HashMap<char, u8>) -> bool {
+    // Create a new String with characters changed to numbers
+    let puzzle: String = puzzle.chars().map(|c| {
+        if let Some(&n) = substitutions.get(&c) {
+            // If the character is in the substitutions, get the number and
+            // convert it to a char
+            char::from_digit(n as u32,10).unwrap()
```
petertseng: space after the comma?

exercises/alphametics/Cargo-example.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "alphametics"
+version = "0.0.0"
+
+[dependencies]
+itertools = "0.5"
+permutohedron = "0.2"
```
petertseng: newline after?

petertseng:
OK, since Ian's good and everything I said is dealt with, let's merge. I'll squash.

petertseng:
Thanks!

IanWhitney:
Thanks, @ijanos?

PR #227, opened 2016-11-20
Use example.rs specific Cargo.toml file during CI

ijanos:
Add support for optional `Cargo-example.toml` files to satisfy the CI and do not pollute the empty project config file. See discussion in #225

One thing I could not figure out though: where is it defined what files the CLI tool will download? Do I have to modify anything else to prevent the download of `Cargo-examle.toml`?


IanWhitney:
Nice. Thanks much, @ijanos!

PR #231, opened 2016-12-12
Implement Largest Series Product

IanWhitney:
Follows the canonical test suite as [it currently stands](https://github.com/exercism/x-common/blob/107c4416d3674d1ae10ba0a912aad9d1c4f3e909/exercises/largest-series-product/canonical-data.json)

Though I have started with a rust-specific test that shows students that `lsp` returns a Result.

Placement is early in the Getting Rusty section. Results have been introduced by this point, as have some Higher Order Functions. The `windows` and `product` iterators will be new. I've pointed students to them in the HINTS.md file.


exercises/largest-series-product/HINTS.md
```diff
@@ -0,0 +1,6 @@
+# Largest Series Product in Rust
+
+These iterators may be useful, depending on your approach
+
+- [Windows](https://doc.rust-lang.org/std/slice/struct.Windows.html)
+- [Product](https://doc.rust-lang.org/std/iter/trait.Product.html)
```
ijanos: I remember getting lost in the Rust documentation when I started learning the language, it is not trivial from the linked pages to get to the one with the examples. I would rather link the slice and iterator pages of the documentation here because it contains the descriptions of what these functions do and provides code examples.

  * https://doc.rust-lang.org/std/primitive.slice.html#method.windows 
  * https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.product

exercises/largest-series-product/example.rs
```diff
@@ -0,0 +1,24 @@
+use std::str::FromStr;
+
+pub fn lsp(string_digits: &str, span: usize) -> Result<u64, String> {
+    if span == 0 {
+        return Ok(1);
+    }
+
+    if string_digits.chars().any(|c| !c.is_numeric()) {
```
ijanos: `char::is_numeric` accepts all kinds of Unicode characters (see [doc](https://doc.rust-lang.org/std/primitive.char.html#method.is_numeric)). I propose using `c.is_digit(10)` here.

exercises/largest-series-product/example.rs
```diff
@@ -0,0 +1,24 @@
...
+        return Err(String::from("All characters must be numbers"));
+    }
+
+    let products: Vec<u64> = string_digits.chars()
+        .map(|c| u64::from_str(&c.to_string()).unwrap())
```
ijanos: This will allocate a string on the heap just to throw it away. A more memory friendly version would be:

`.map(|c| c.to_digit(10).unwrap())`

exercises/largest-series-product/example.rs
```diff
@@ -0,0 +1,24 @@
...
+        .collect::<Vec<u64>>()
+        .windows(span)
+        .map(|w| w.into_iter().product())
+        .collect();
+
+    if let Some(x) = products.iter().max() {
+        Ok(x.clone())
```
ijanos: Is `clone` necessary here? `u64` is a Copy type, it should create a clone on move automatically.

exercises/largest-series-product/tests/largest-series-product.rs
```diff
@@ -0,0 +1,105 @@
...
+
+#[test]
+#[ignore]
+fn empty_string_and_non_zero_span_is_an_error() {
+    assert!(lsp("", 1).is_err());
+}
+
+#[test]
+#[ignore]
+fn panics_if_string_contains_non_digits() {
```
petertseng: To me, `panics` implies the function under test would use `panic!`, but it is just returning an `Err`.

exercises/largest-series-product/example.rs
petertseng: It's because x is a `&u64` here. I would change the `let Some(x)` to a `let Some(&x)`, how about it?

exercises/largest-series-product/tests/largest-series-product.rs
IanWhitney: Old test name that reflected my first panic-based approach. Good catch.

IanWhitney:
I think we're good. Let me know if you see anything else.

PR #235, opened 2016-12-18
Implement sum-of-multiples

### LINKED ISSUE #234: Implement Sum of Multiples
Standard test suite is here: https://github.com/exercism/x-common/blob/aa12f2e403270854e709b20a2aa62cf24e60d171/exercises/sum-of-multiples/canonical-data.json

IanWhitney:
Closes #234

Tests follow the current standard. I have some problems with the current
standard in that almost none of the tests introduce new behaviors or
have useful descriptions. But I'll address that with a PR against
the current standard test suite, not here.

My example implementation is likely poor, but it works.

I'm placing this early in the track because it doesn't really introduce
much Rust-specific behavior beyond the `Vec` being borrowed. The solution
is an algorithmic problem, not a language fluency one.


exercises/sum-of-multiples/example.rs
```diff
@@ -0,0 +1,17 @@
+use std::collections::BTreeSet;
+
+pub fn sum_of_multiples(limit: u32, factors: &Vec<u32>) -> u32 {
```
petertseng: by no means do I demand this, but I look at https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#slicing and see "In Rust, it's more common to pass slices as arguments rather than vectors when you just want to provide a read access." - so I suggest `&[u32]` here.

(Note that doing this should not require the test code to change in any way - the `vec!` in the test code can still stay)

exercises/sum-of-multiples/example.rs
```diff
@@ -0,0 +1,17 @@
...
+    let mut multiples: BTreeSet<u32> = BTreeSet::new();
+
+    for f in factors {
+        let mut multiplier = 2;
+        let mut x = f.clone();
```
petertseng: Trying to think about what looks best here, given https://github.com/exercism/xrust/pull/231#discussion_r91913638 . I was thinking about `*f` instead, or a `for &f in factors` followed by a plain `let mut x = f` here. In the end, I don't really care, just pick.

IanWhitney:
Changes made and I rebased off of master to get the updated `config.json` file.

exercises/sum-of-multiples/example.rs
petertseng: Not to say that I completely understand everything that goes on here, but at least for Vectors I believe it's the combination of the facts:

* [Vector](https://doc.rust-lang.org/std/vec/struct.Vec.html) implements Deref with Target `[T]`, so that `v.deref()` gives `&[T]`.
* https://doc.rust-lang.org/book/deref-coercions.html

> If you have a type `U`, and it implements `Deref<Target=T>`, values of `&U` will automatically coerce to a `&T`

Let `U = Vec<E>` and `T = [E]`.

* the condition is met. We have a type `Vec<E>`, and it implements `Deref<Target=[E]>`
* Thus, values of `&Vec<E>` coerce to `&[E]`
* And that's what we are doing in our tests. This function takes a `&[u32]`, but we pass a `&Vec<u32>`, but it works.

Issue #236, opened 2016-12-19
config.json duplicates information from problems.md


petertseng:
After the merging of #229, problems.md now duplicates config.json in both problem order (it was already duplicating it in this regard) and topics (added). So, what should we do about this duplication?

There are two pieces of information in problems.md that are not present in config.json

1. **Background** including links to posts explaining why we order the problems as we do.
1. **Section headers**

```
## Introduction

The first section contains the sort of stuff you expect when learning any programming languages: conditionals, booleans, looping and some higher-order functions.
```

Option 1: We could do something really silly like "the **first** (not ALL, that is duplication) problem of a section gets a `section_name` and a `section_description`", like this:

```json
    {
      "slug": "hello-world",
      "difficulty": 1,
      "section_name": "Introduction",
      "section_description": "The first section contains the sort of stuff you expect when learning any programming languages: conditionals, booleans, looping and some higher-order functions.",
      "topics": [
        "Some/None",
        "println!"
      ]
    },
   {
      "slug": "gigasecond",
      "difficulty": 1,
      "topics": [
        "crates",
        "math"
      ]
    }
```

(and so on)

Option 2: Keep the problem order duplicated in problems.md along with keeping the section headers there, but remove all topics from it, pointing to config.json as the authoritative source for the topics?

Option 3: Remove problems.md completely, losing our section headers and background.

Option 4: Keep problems.md as-is, accepting the duplication of topics.

Please also provide alternative options, if you have any.

I find option 4 completely unacceptable, but can be convinced of the other three options. My leading option is 1. This means I might execute option 1 at a time convenient to me if nobody provides a better alternative before that time (since there has been time to ponder over this while considering #229)

ijanos:
Here another idea: what if we create another markup file (eg: sections.json) that contains headers, descriptions and add a "section" key to every problem in config.json and then create problems.md automatically based on the two files if we want a human readable version. 

Otherwise I would choose option 2. Option 1 is way too hackish for me and I think it would be hard to maintain that file to someone who does not read this issue.

petertseng:
Ah, automatically generating based on config.json and the other file seems OK.

Would you propose that we keep the generated problems.md in the repo (as a matter of convenience), or not (as a matter of not committing generated files)?

(If yes, I might even suggest a CI check that the generated file matches the inputs, in case someone changes the inputs then forgets to regen)

[...]

PR #238, opened 2017-01-04
nucleotide-count: Error on invalid nucleotide

### LINKED ISSUE #149: test cases for nucleotide count should include handing of invalid sequences

petertseng:
`count` and `nucleotide_counts` both return a Result now.

Closes #149

---

[Too many other language implementations](http://x.exercism.io/problems/nucleotide-count) to look through them all and see what they did. There is no canonical-data.json for this problem yet and I do not have the time to make one.

### DIFF
--- a/exercises/nucleotide-count/example.rs
+++ b/exercises/nucleotide-count/example.rs
@@ -1,13 +1,26 @@
 use std::collections::HashMap;
+use std::collections::hash_map::Entry;
 
-pub fn count(nucleotide: char, input: &str) -> usize {
-    input.chars().filter(|&c| c == nucleotide).count()
+static VALID_NUCLEOTIDES: &'static str = "ACGT";
+
+pub fn count(nucleotide: char, input: &str) -> Result<usize, char> {
+    let valid = |x: char| { VALID_NUCLEOTIDES.contains(x) };
+    if valid(nucleotide) && input.chars().all(valid) {
+        Ok(input.chars().filter(|&c| c == nucleotide).count())
+    } else {
+        Err(nucleotide)
+    }
 }
 
-pub fn nucleotide_counts(input: &str) -> HashMap<char, usize> {
-    let mut map: HashMap<char, usize> = "ACGT".chars().map(|c| (c, 0)).collect();
+pub fn nucleotide_counts(input: &str) -> Result<HashMap<char, usize>, char> {
+    let mut map: HashMap<char, usize> = VALID_NUCLEOTIDES.chars().map(|c| (c, 0)).collect();
     for nucleotide in input.chars() {
-        *map.entry(nucleotide).or_insert(0) += 1;
+        match map.entry(nucleotide) {
+            Entry::Vacant(_) => return Err(nucleotide),
+            Entry::Occupied(mut view) => {
+                *view.get_mut() += 1;
+            }
+        }
     }
-    map
+    Ok(map)
 }
--- a/exercises/nucleotide-count/example.rs
+++ b/exercises/nucleotide-count/example.rs
@@ -1,5 +1,4 @@
 use std::collections::HashMap;
-use std::collections::hash_map::Entry;
 
 static VALID_NUCLEOTIDES: &'static str = "ACGT";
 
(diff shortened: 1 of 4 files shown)


exercises/nucleotide-count/example.rs
```diff
@@ -1,13 +1,26 @@
...
+    }
 }
 
-pub fn nucleotide_counts(input: &str) -> HashMap<char, usize> {
-    let mut map: HashMap<char, usize> = "ACGT".chars().map(|c| (c, 0)).collect();
+pub fn nucleotide_counts(input: &str) -> Result<HashMap<char, usize>, char> {
+    let mut map: HashMap<char, usize> = VALID_NUCLEOTIDES.chars().map(|c| (c, 0)).collect();
     for nucleotide in input.chars() {
-        *map.entry(nucleotide).or_insert(0) += 1;
+        match map.entry(nucleotide) {
```
petertseng: you know, maybe instead of all this I could just check `valid(nucleotide)` (I would make it a top-level function), so that I wouldn't have to deal with `Entry::` cases

exercises/nucleotide-count/tests/nucleotide-count.rs
```diff
@@ -17,19 +17,31 @@ fn check_dna(s: &str, pairs: &[(char, usize)]) {
 
 #[test]
 fn test_count_empty() {
-    assert_eq!(dna::count('A', ""), 0);
+    assert_eq!(dna::count('A', "").unwrap(), 0);
```
petertseng: should we do anything specific here to make abundandly clear that it's a result? see https://github.com/exercism/xrust/blob/master/exercises/largest-series-product/tests/largest-series-product.rs#L6

problems.md
```diff
@@ -36,7 +36,7 @@ hamming |  Result
 pascals-triangle | Math, Vec, Index (optional)
 scrabble-score |  chaining higher-order functions, HashMap (optional)
 pangram | filter, ascii (optional)
-nucleotide-count |  filter, entry api, mutablity, match
+nucleotide-count |  Result, filter, entry api, mutablity, match
```
petertseng: duped, but to be solved by #236

exercises/nucleotide-count/example.rs
```diff
@@ -1,13 +1,26 @@
...
-pub fn count(nucleotide: char, input: &str) -> usize {
-    input.chars().filter(|&c| c == nucleotide).count()
+static VALID_NUCLEOTIDES: &'static str = "ACGT";
+
+pub fn count(nucleotide: char, input: &str) -> Result<usize, char> {
+    let valid = |x: char| { VALID_NUCLEOTIDES.contains(x) };
+    if valid(nucleotide) && input.chars().all(valid) {
+        Ok(input.chars().filter(|&c| c == nucleotide).count())
+    } else {
+        Err(nucleotide)
```
petertseng: for `count('A', "AX")` this returns `A` instead of `X`, but we don't check it in the tests so I'm not motivated.

exercises/nucleotide-count/example.rs
ijanos: I feel that you are not supposed to deal with entry cases directly. I suggest using if let here: 

```rust
    if let Some(n) = map.get_mut(nucleotide) {
      *n += 1;
    } else {
       return Err(nucleotide);
    }
```

exercises/nucleotide-count/tests/nucleotide-count.rs
ijanos: I think using `is_err` is enough.

exercises/nucleotide-count/example.rs
petertseng: Fantastic. Did exactly that (though `&nucleotide` needed)

petertseng:
Squashing is recommended. I'll do it soon, just giving y'all a last chance to look.

exercises/nucleotide-count/tests/nucleotide-count.rs
IanWhitney: Agreed. For implementations that expect Result or Option, I usually start with some `is_err` and `is_ok` tests just to make it clear.

petertseng: Added. One for each function.

PR #239, opened 2017-01-04
parallel-letter-frequency: Use &str with lifetime, not String

petertseng:
The previous version had a comment that explained a need for copying the
`&str` into `String`s, but this can be avoided.

The disadvantage is now that this function can *only* be invoked with
`&'static str`, which means that it can't (for example) accept arbitrary
user input as part of a command-line program. So it works for our style
of testing (all our tests do indeed use `&'static str`) but is
unrealistic code.

---

Reviewers should consider the disadvantage carefully before deciding whether to merge. Of course, this is only the example and doesn't affect the test code. On the other hand, it could be better to keep the example code realistic.

In the extreme case, we might even choose to add a test that creates non-`'static` strings, so that using `&'static str` is not possible (I do not advocate for this, but y'all can if you like)

### DIFF
--- a/exercises/parallel-letter-frequency/example.rs
+++ b/exercises/parallel-letter-frequency/example.rs
@@ -5,20 +5,18 @@ use std::sync::mpsc::channel;
 
 /// Compute the frequency of each letter (technically of each unicode codepoint) using the given
 /// number of worker threads.
-pub fn frequency(texts: &[&str], num_workers: usize) -> HashMap<char, usize> {
+pub fn frequency(texts: &[&'static str], num_workers: usize) -> HashMap<char, usize> {
     assert!(num_workers > 0);
     let part_size_floor = texts.len() / num_workers;
     let rem = texts.len() % num_workers;
     let part_size = if rem > 0 { part_size_floor + 1 } else { part_size_floor };
-    let mut parts: Vec<Vec<String>> = Vec::with_capacity(part_size);
+    let mut parts: Vec<Vec<&str>> = Vec::with_capacity(part_size);
     for _ in 0 .. num_workers {
         parts.push(Vec::with_capacity(part_size));
     }
     let mut i = 0;
-    for line in texts.iter() {
-        // We'll need to clone those strings in order to satisfy some lifetime guarantees. Basically
-        // it's hard for the system to be sure that the threads spawned don't outlive the strings.
-        parts[i].push(line.to_string());
+    for &line in texts.iter() {
+        parts[i].push(line);
         i = (i + 1) % num_workers;
     }
 
@@ -41,7 +39,7 @@ pub fn frequency(texts: &[&str], num_workers: usize) -> HashMap<char, usize> {
     results
 }
 
-fn count(lines: Vec<String>) -> HashMap<char, usize> {
+fn count(lines: Vec<&str>) -> HashMap<char, usize> {
     let mut results: HashMap<char, usize> = HashMap::new();
     for line in lines.iter() {
         for c in line.chars() {
--- a/exercises/parallel-letter-frequency/tests/parallel-letter-frequency.rs
+++ b/exercises/parallel-letter-frequency/tests/parallel-letter-frequency.rs
@@ -42,6 +42,13 @@ fn test_no_texts() {
 
 #[test]
 #[ignore]
+fn test_hello_world() {
+    let hi = format!("Hello, {}!", "world");
+    frequency::frequency(&[&hi], 4);
+}
+
+#[test]
+#[ignore]
 fn test_one_letter() {
     let mut hm = HashMap::new();
     hm.insert('a', 1);


ijanos:
I don't like this. Who will look at examples? I guess some people who want to simply "cheat" and those who are stuck and want to peek at a solution. It may look like a function accepting only `'static str` is a good solution to these people, while in the real world it is rarely useful nor a good pattern to use.

If we merge this at least put a warning comment before the function... and by doing that we are back to the original problem with the comment explaining things.

IanWhitney:
@ijanos Do you have more details on why it's not a good pattern? I would like to know more!

From my experience, I use the examples a lot when learning a new language. I usually know how to solve the problem, but I don't know the language's syntax or idioms. So I'll look at the example code to figure out what I need to use. I don't know if that's common, but this use case is one of the reasons we settled on the goal of having our example solutions be, well, exemplary. We want them to use reasonably idiomatic, well-written Rust.

petertseng:
The problem is explained shortly with: It is now impossible for this function to take a `&str` with non-`static` lifetime, and this is a *very* harsh restriction.

I can try to come up with example code soon, but not now.

petertseng:
No intent to merge - better close now to avoid any mistakes. I do still plan to get back to you with a reason why it's not going to get merged.

ijanos:
@IanWhitney Accepting a static string slice basically means accepting only strings available at compile time. Imagine using a library that has a restriction like that, you cannot call it on anything you don't know during compile, like data from a file, data downloaded from somewhere or user input.

petertseng:
Reopening to show Travis. Please do not merge.

petertseng:
The commit https://github.com/exercism/xrust/pull/239/commits/5070848fcb065b5cbb2053f0558af9212f8bac41 should show the restrictiveness of `&'static str`. I couldn't even pass the result of a `format!` into `frequency`.

petertseng:
Oops, type error. Have to make Travis do it again.

petertseng:
The commit https://github.com/exercism/xrust/pull/239/commits/8fe6c5bcb988557cb41fe58970001ef9380ffc93 should show the restrictiveness of `&'static str`. I couldn't even pass the result of a `format!` into `frequency`. You will see that the build fails: https://travis-ci.org/exercism/xrust/builds/189277410

IanWhitney:
Ok, I think I see what's going on here now. My understanding was that all `&str` were really `&'static str` under the hood. That was my understanding from [the book](https://doc.rust-lang.org/book/strings.html). 

So, based on that belief, I thought these two lines are identical

```
pub fn frequency(texts: &[&str], num_workers: usize) -> HashMap<char, usize> {
```
```
pub fn frequency(texts: &[&'static str], num_workers: usize) -> HashMap<char, usize> {
```

But the non-lifetime version is allowing Rust to do some implicit conversion of 

``` rust
let hi = format!("Hello, {}!", "world");
&[&hi]
```

into something it can think of as a str. Right?

But one you say "I only accept static str" then the compiler can't do the conversion since it can't satisfy the lifetime requirements.

🦆

petertseng:
> But the non-lifetime version is allowing Rust to do some implicit conversion of
> ```rust
> let hi = format!("Hello, {}!", "world");
> &[&hi]
> ```
> into something it can think of as a str. Right?

[`format!`](https://doc.rust-lang.org/std/macro.format.html) creates a value of type `String`. If we take a reference to it, [deref coercion](https://doc.rust-lang.org/book/deref-coercions.html) allows it to be used wherever a `&str` is required, because [String implements Deref<str>](https://doc.rust-lang.org/std/string/struct.String.html#deref)

---

> But one you say "I only accept static str" then the compiler can't do the conversion since it can't satisfy the lifetime requirements.

My understanding. Please correct.

> the non-lifetime version

The version without lifetime annotations follows the rules of [lifetime elision](https://doc.rust-lang.org/book/lifetimes.html#lifetime-elision) which should mean that it is equivalent to:

```rust
pub fn frequency<'a, 'b>(texts: &'a [&'b str], num_workers: usize) -> HashMap<char, usize>
```

The only requirement on `texts` is that `frequency` may not outlive the slice nor the strings contained within it. Put another way, the minimum lifetime of the slice and strings is the scope of `frequency`. This requirement is the [standard rule for borrowing](https://doc.rust-lang.org/book/references-and-borrowing.html#the-rules).

At this point, let's see why we have to do `parts[i].push(line.to_string())` instead of just parts[i].push(line)`. We have to examine the type of `thread::spawn`(https://doc.rust-lang.org/std/thread/fn.spawn.html) - its argument (normally a closure, but anything implementing `FnOnce`) must at least have `'static` lifetime (as long as the program). This is because the threads can outlive their spawner.

The strings in `texts` are not guaranteed to have `'static` lifetime. The only constraint imposed on them is that they must live as long as `frequency`. Since they are not guaranteed to have `'static` lifetime, they cannot be closed over by the closure.

*Aside*: If we wanted to have the type system prove that the threads terminate before their spawner, that seems a bit tricky! But [it's been done](https://aturon.github.io/crossbeam-doc/crossbeam/struct.Scope.html)! I have not investigated this create to see what it is doing.

---

What happens in this PR? If I understand lifetime elision properly, what I wrote should be equivalent to:

```rust
pub fn frequency<'a>(texts: &'a [&'static str], num_workers: usize) -> HashMap<char, usize>
```

For the `texts` slice, the slice is still required to only live for as long as `frequency` does. But now the strings are required to live for lifetime `&'static`, as long as the program. So, because `frequency` now imposes this requirement on its callers, `frequency` is now at liberty to allow closures to close over the strings in `texts`. But imposing the requirement on the callers leads to unrealistic code, and would not make a good example for our students. As I showed, you can no longer take references of arbitrary `String`s, such as those created by `fmt!` or any other function, because those `String`s do not live for `&'static` time.

petertseng:
> But the non-lifetime version is allowing Rust to do some implicit conversion of
> ```rust
> let hi = format!("Hello, {}!", "world");
> &[&hi]
> ```
> into something it can think of as a str. Right?

[`format!`](https://doc.rust-lang.org/std/macro.format.html) creates a value of type `String`. If we take a reference to it, [deref coercion](https://doc.rust-lang.org/book/deref-coercions.html) allows it to be used wherever a `&str` is required, because [String implements Deref<str>](https://doc.rust-lang.org/std/string/struct.String.html#deref)

---

> But one you say "I only accept static str" then the compiler can't do the conversion since it can't satisfy the lifetime requirements.

My understanding. Please correct.

> the non-lifetime version

The version without lifetime annotations follows the rules of [lifetime elision](https://doc.rust-lang.org/book/lifetimes.html#lifetime-elision) which should mean that it is equivalent to:

```rust
pub fn frequency<'a, 'b>(texts: &'a [&'b str], num_workers: usize) -> HashMap<char, usize>
```

The only requirement on `texts` is that `frequency` may not outlive the slice nor the strings contained within it. Put another way, the minimum lifetime of the slice and strings is the scope of `frequency`. This requirement is the [standard rule for borrowing](https://doc.rust-lang.org/book/references-and-borrowing.html#the-rules).

At this point, let's see why we have to do `parts[i].push(line.to_string())` instead of just `parts[i].push(line)`. We have to examine the type of [`thread::spawn`](https://doc.rust-lang.org/std/thread/fn.spawn.html) - its argument (normally a closure, but anything implementing `FnOnce`) must at least have `'static` lifetime (as long as the program). This is because the threads can outlive their spawner.

The strings in `texts` are not guaranteed to have `'static` lifetime. The only constraint imposed on them is that they must live as long as `frequency`. Since they are not guaranteed to have `'static` lifetime, they cannot be closed over by the closure passed to `thread::spawn`.

*Aside*: If we wanted to have the type system prove that the threads terminate before their spawner, that seems a bit tricky! But [it's been done](https://aturon.github.io/crossbeam-doc/crossbeam/struct.Scope.html)! I have not investigated this crate to see what it is doing.

---

What happens in this PR? If I understand lifetime elision properly, what I wrote should be equivalent to:

```rust
pub fn frequency<'a>(texts: &'a [&'static str], num_workers: usize) -> HashMap<char, usize>
```

For the `texts` slice, the slice is still required to only live for as long as `frequency` does. But now the strings are required to live for lifetime `&'static`, as long as the program. So, because `frequency` now imposes this requirement on its callers, `frequency` is now at liberty to allow closures passed to `thread::spawn` to close over the strings in `texts`.

But imposing the requirement on the callers leads to unrealistic code, and would not make a good example for our students. As I showed, references of arbitrary `String`s (such as those created by `fmt!` or any other function) are no longer acceptable as arguments of `frequency`, because those `String`s do not live for `&'static` time.

petertseng:
EBWODP: Clarified that 

> But one you say "I only accept static str" then the compiler can't do the conversion since it can't satisfy the lifetime requirements.

is I believe a reasonable summary (previous text was unclear on whether it is)

petertseng:
One piece of the picture I realized isn't clear:

So why is it that we can allow the closure passed to `thread::spawn` to close over `part`? I believe the answer is:

`parts` is `Vec<Vec<String>>` - everything there is owned. Inside the for loop, using the `move` keyword, we move ownership of `part` into the closure, so the closure owns a `Vec<String>`. Since it's owned by the closure, it satisfies the requirement of `&'static` lifetime.

petertseng:
EBWODP: No, it's because a `Vec<String>` contains no references. The closure can close over `Vec<String>`, but not over `Vec<&str>`.

PR #241, opened 2017-01-05
travis: Test in release mode

petertseng:
Alphametics is unbearably slow in debug mode.

---

Reviewers should verify that there is a runtime improvement before merging. I was personally worried about whether `--release` would increase compile times for all exercises while only benefitting alphametics.

### DIFF
--- a/_test/check-exercises.sh
+++ b/_test/check-exercises.sh
@@ -49,7 +49,7 @@ for exercise in exercises/*/tests; do
             cargo test --quiet --no-run
         else
             # Run the test and get the status
-            cargo test
+            cargo test --release
         fi
     )
 


petertseng:
* With this PR: https://travis-ci.org/exercism/xrust/builds/189352439
* Without this PR: https://travis-ci.org/exercism/xrust/builds/189267725

Looks much better.

IanWhitney:
I can't think of a reason to not use release mode.

ijanos:
I intentionally created an idiomatic but naive (hence slow) example for that exercise. If it is causing problems I'm fine with replacing it with something faster. 

A reason to not use release mode: it will skip overflow checks. Try compiling [this](https://play.rust-lang.org/?code=fn%20main()%20%7B%0A%20%20%20%20let%20x%20%3D%200u32%3B%0A%20%20%20%20println!(%22%7B%7D%22%2C%20x-1)%3B%0A%7D&version=stable&backtrace=0) in release and debug mode and to the difference.

petertseng:
Usually I don't notice the slowness. Yesterday at #239 I created sufficiently many Travis builds that they got backed up waiting for each other. Even created two for every commit (that's weird, maybe because I was closing and reopening. But I canceled the ones I didn't want).

Well, if a faster implementation can get decent times and still in debug mode so that we keep overflow checks, I'm all for it. I will make a column by column one at some point in the future, but don't be afraid to beat me to it.

Another option then can be to only run in --release for selected exercises, and the only such selected exercise being alphametics currently. Don't have any great ideas on how to make the "selected" bit, maybe a `.meta/test-in-release-mode` (`.meta` so it doesn't get served to students).

I'll currently close since the problem is only felt if we queue up multiple builds in rapid succession, and there are other ways we want to look at.

PR #242, opened 2017-01-07
problems.md: Remove topics

### LINKED ISSUE #236: config.json duplicates information from problems.md

petertseng:
History:

Thanks to the hard work of #167 and #127, we have topics in problems.md
that tell why we ordered problems a certain way.

After #229 (motivated by #179), we copied the topics into config.json,
since now all tracks can take advantage of the "topics" concept.

To reduce duplication, we can remove topics from problems.md now.

What's left in problems.md is a bit of prose explaining why we order our
problems the way they do, and the division of the problems into
sections.

Closes #236


problems.md
```diff
@@ -14,75 +14,67 @@ The actual source of truth of problem order is [config.json](config.json), but t
...
-hello-world |  Some/None. println!
-gigasecond |  crates, math
-leap |  math, booleans, conditionals
-raindrops |  case (or `format`). Mutable string
-bob |  chars, string functions
-beer-song |  case, string concatenation, vector (optional), loop
-difference-of-squares |  fold & map
-sum-of-multiples | algorithm, borrowing
-grains | math, panic
+* hello-world
```
petertseng: Or alternatively, would "This section starts at `hello-world` and ends at `grains`" be acceptable? Might be less maintenance/duplication that way?

petertseng: Added that as a separate commit, see if we like it.

IanWhitney:
The "This section begins with..." seem like they'll go quickly out of date. How about something more abstract, that states our values without going into the exact implementation.

https://gist.github.com/IanWhitney/3dc56aef520afe3146924b3552abdc52 is an idea

petertseng:
Your proposal seems good to me, since it removes the last traces of duplication. So now, I have to ask: Do you mind if I start with that?

I'll see if I have any changes to suggest.

problem_ordering.md
```diff
@@ -0,0 +1,41 @@
...
+
+If a problem solution requires students to use both `Result` and `HashMap`, then at least one of these concepts should be introduced by an earlier problem.
+
+If the track has several problems that rely on a concept, group them together to reinforce learning.
+
+### Progress Through Rust
+
+#### Early Problems
+
+Early problems should focus on Rust syntax and concepts that are present in many programming languages: conditionals, looping, higher-order-functions.
```
petertseng: it gives me a bit of pause to see higher-order functions in the same sentence. But why is that? I realized:

* It seems "weird" to put this in the section before `Result` and `Option`... because I usually associate higher-order functions, `Result`, and `Option` all with functional programming.
* But on further reflection, it is true in my experience that a language can  have higher-order functions without either `Result` or `Option` (the latter two need static typing to really make sense)
* It does seem that higher-order functions take more effort to understand, but maybe my estimates of the effort required to understand any concept in this sentence are [wildly wrong](https://en.wikipedia.org/wiki/Curse_of_knowledge).

So despite my "weird" feeling, the sentence is correct and I've no change to suggest. I just wanted to share the feeling.

problem_ordering.md
```diff
@@ -0,0 +1,41 @@
...
+Problems in the middle of the of the track should introduce concepts that are less common. `Result` and `Option` and Traits, for example. Or syntax such as `while let`.
+
+Through the middle section of the track students should gain familiarity with a wide range of Rust concepts.
+
+#### Later Problems
+
+Problems late in the track should do one of two things:
+
+1. Introduce advanced concepts -- Lifetimes, Concurrency, Unsafety
+2. Require a complex solution that synthesizes all the knowlegde students have gained to date
```
petertseng: spelling of "knowledge"

petertseng:
spelling error in commit message

s/hrad/hard/

problem_ordering.md
IanWhitney: I think HOF will depend entirely on the students' background. Those coming from Ruby/JS/Python/Java 8 will be familiar. I think they are becoming pretty standard, though that might just be my years of Ruby speaking.

PR #245, opened 2017-01-12
Benchmark parallel-letter-frequency. Closes #244

### LINKED ISSUE #244: Add benchmarks.
I've been creating benchmarks for some exercises in the `benches` dir and testing with `cargo bench`.

Using test::Bencher, they would only work on nightly, but that could be documented easily enough and `cargo test` would still work fine on stable.

This Go exercise is a pretty compelling reason to have that. Is your parallel implementation even faster than a simple sequential implementation? Find out:

https://github.com/exercism/xgo/blob/master/exercises/parallel-letter-frequency/parallel_letter_frequency_test.go#L48-L58

Would a PR adding a benchmark to that example to start be accepted?

glennpratt:
Addresses #244.

This runs benchmarks in a separate Travis-CI job. This way they run in parallel, only on nightly and benchmark failures aren't mixed with test failures.

`cargo bench` output:

```
running 2 tests
test bench_parallel   ... bench:     114,524 ns/iter (+/- 33,720)
test bench_sequential ... bench:      43,880 ns/iter (+/- 613)
test result: ok. 0 passed; 0 failed; 0 ignored; 2 measured
```

You can see why I made this benchmark. 😄 

Visible in the job: https://travis-ci.org/exercism/xrust/jobs/191194852

### DIFF
--- a/_test/check-exercises.sh
+++ b/_test/check-exercises.sh
@@ -6,6 +6,12 @@ if [ -z "$DENYWARNINGS" ]; then
     set -e
 fi
 
+if [ -n "$BENCHMARK" ]; then
+    files=exercises/*/benches
+else
+    files=exercises/*/tests
+fi
+
 tmp=${TMPDIR:-/tmp/}
 mkdir "${tmp}exercises"
 
@@ -13,7 +19,7 @@ exitcode=0
 
 # An exercise worth testing is defined here as any top level directory with
 # a 'tests' directory
-for exercise in exercises/*/tests; do
+for exercise in $files; do
     # This assumes that exercises are only one directory deep
     # and that the primary module is named the same as the directory
     directory=$(dirname "${exercise}");
@@ -38,7 +44,10 @@ for exercise in exercises/*/tests; do
             sed -i '/\[ignore\]/d' $test
         done
 
-        if [ -n "$DENYWARNINGS" ]; then
+        # Run benchmarks instead of tests when enabled.
+        if [ -n "$BENCHMARK" ]; then
+            cargo bench
+        elif [ -n "$DENYWARNINGS" ]; then
             sed -i -e '1i #![deny(warnings)]' src/lib.rs
 
             # No-run mode so we see no test output.
--- a/exercises/parallel-letter-frequency/HINTS.md
+++ b/exercises/parallel-letter-frequency/HINTS.md
@@ -1,0 +1,32 @@
+# Parallel Letter Frequency in Rust
+
+Learn more about concurrency in Rust here:
+
+- [Concurrency](https://doc.rust-lang.org/book/concurrency.html)
+
+## Bonus
+
+This exercise also includes a benchmark, with a sequential implementation as a
+baseline. You can compare your solution to the benchmark. Observe the
+effect different size inputs have on the performance of each. Can you
+surpass the benchmark using concurrent programming techniques?
+
+As of this writing, test::Bencher is unstable and only available on
+*nightly* Rust. Run the benchmarks with Cargo:
+
+```
+cargo bench
+```
+
+If you are using rustup.rs:
+
+```
+rustup run nightly cargo bench
+```
+
+- [Benchmark tests](https://doc.rust-lang.org/book/benchmark-tests.html)
+
+Learn more about nightly Rust:
+
+- [Nightly Rust](https://doc.rust-lang.org/book/nightly-rust.html)
+- [Rustup: Working with nightly](https://github.com/rust-lang-nursery/rustup.rs#working-with-nightly-rust)
--- a/exercises/parallel-letter-frequency/benches/benchmark.rs
+++ b/exercises/parallel-letter-frequency/benches/benchmark.rs
@@ -1,0 +1,99 @@
+#![feature(test)]
+extern crate parallel_letter_frequency;
+extern crate test;
+
+use std::collections::HashMap;
+use test::Bencher;
+
+#[bench]
+fn bench_tiny_parallel(b: &mut Bencher) {
+    let tiny = &["a"];
+    b.iter(|| parallel_letter_frequency::frequency(tiny, 3));
+}
+
+#[bench]
+fn bench_tiny_sequential(b: &mut Bencher) {
+    let tiny = &["a"];
+    b.iter(|| frequency(tiny));
+}
+
+#[bench]
+fn bench_small_parallel(b: &mut Bencher) {
+    let texts = all_texts(1);
+    b.iter(|| parallel_letter_frequency::frequency(&texts, 3));
+}
+
+#[bench]
+fn bench_small_sequential(b: &mut Bencher) {
+    let texts = all_texts(1);
+    b.iter(|| frequency(&texts));
+}
+
+#[bench]
+fn bench_large_parallel(b: &mut Bencher) {
+    let texts = all_texts(30);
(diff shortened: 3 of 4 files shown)


glennpratt:
Oh, I need to update the readme.

exercises/parallel-letter-frequency/sequential-letter-frequency/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+            if let Some(c) = chr.to_lowercase().next() {
+                (*map.entry(c).or_insert(0)) += 1;
+            }
+        }
+    }
+
+    map
+}
+
+// Poem by Friedrich Schiller. The corresponding music is the European Anthem.
```
glennpratt: This might not be the ideal place to put test fixtures. 

I could them to an rs file outside `src`, `benches` and `tests` and pull it in with [`include!("path/to/a/file")`](https://doc.rust-lang.org/std/macro.include.html).

Or use some other fixture pattern I'm not aware of.

glennpratt:
On second thought, I don't know where I would put documentation about this. Probably wouldn't want it in every exercise, but I'm not sure what other options there are.

ijanos:
I'm wondering do we need to run the benchmarks in travis? Is there any andvantage of doing that? I think only checking that the code compiles is enough. I assume `cargo bench --no-run` would do that.

ijanos:
And of course thank you for you contribution! It looks good overall, just a few things about the organisation of files: I see you added a whole new create to the exercise with a full implementation so everyone who would start working on the exercise will see the sequential implementation. I think that is your intention, right? 

Let's see what the others think about this, but personally I think it would be better omit the new crate, add the sequential code to `example.rs` and document that the benches are there if students wants to test their implementations they should be encouraged to write their own sequential implementation.

glennpratt:
The pattern was stolen from:

https://github.com/exercism/xgo/blob/master/exercises/parallel-letter-frequency/parallel_letter_frequency_test.go

Though that goes farther and suggests you write your code to use it.

I imagine in most exercises, having an example implementation wouldn't make sense, but in this one I don't think it gives much away.

glennpratt:
The sequential implementation is in a separate crate so it could be called in the test and the benchmark. I could drop calling it in the test and just put it in the benchmark, then I just need to figure out how to share the texts.

ijanos:
Hmmm sharing the text seems tricky, I'm not fond of putting it in `lib.rs`. I think it is okay to duplicate it in the bench and in the test. I can also imagine that the tests can use a different text than the bench.

glennpratt:
@ijanos updated to duplicate the text and put the sequential implementation in the benchmark.

exercises/parallel-letter-frequency/benches/benchmark.rs
```diff
@@ -0,0 +1,73 @@
...
+}
+
+#[bench]
+fn bench_parallel(b: &mut Bencher) {
+    let texts = all_texts();
+    b.iter(|| parallel_letter_frequency::frequency(&texts, 3));
+}
+
+/// Simple sequential char frequency. Can it be beat?
+pub fn frequency(texts: &[&str]) -> HashMap<char, usize> {
```
glennpratt: I'm in favor of leaving this visible. I personally don't think this exercise should focus too much on the details of getting the count right and more on concurrency techniques.

In other words, reading this doesn't give away anything about threads or whatever tools you'd use to solve this.

ijanos: I see your point and I tend to agree, let's see what other members say. We are setting up a pattern here that we will use if we add benchmarks to other exercises.

_test/check-exercises.sh
```diff
@@ -38,7 +44,10 @@ for exercise in exercises/*/tests; do
             sed -i '/\[ignore\]/d' $test
         done
 
-        if [ -n "$DENYWARNINGS" ]; then
+        # Run benchmarks in instead when enabled.
```
petertseng: I think the "in" can be deleted?

exercises/parallel-letter-frequency/benches/benchmark.rs
petertseng: I can agree with that. I can't speak for every single track, but the Go track also provides students with a frequency function: https://github.com/exercism/xgo/blob/master/exercises/parallel-letter-frequency/frequency.go

Edit: Oh... you [already mentioned](https://github.com/exercism/xrust/pull/245#issuecomment-272192375) that you took this pattern from the Go track. I should really read all of the comments on this PR before jumping in to review...

exercises/parallel-letter-frequency/HINTS.md
```diff
@@ -0,0 +1,22 @@
+# Parallel Letter Frequency in Rust
+
+Learn more about concurrency in Rust here:
+
+- [Concurrency](https://doc.rust-lang.org/book/concurrency.html)
+
+## Bonus
+
+This exercise also includes a benchmark, with a sequential implementation
+as a baseline. As of this writing, Bencher is unstable and only available on
```
IanWhitney: I think this Bonus section could describe more what we want the students to do. Right now this only mentions that the benchmark exists, but not why or what students should do with it.

> This exercise also includes a benchmark, with a sequential implementation as a baseline. You can compare your solution to the benchmark, or you can try to surpass the benchmark using concurrent programming techniques.

Or something like that.

glennpratt:
@IanWhitney @petertseng thank you both for the feedback. I've incorporated it into the latest diff.

I also included larger input sizes where the parallel solution does surpass sequential. I think this is now more informative for the student, if even if they don't try to pre-create a thread pool or something 😄 .

```
test bench_large_parallel   ... bench:     821,598 ns/iter (+/- 308,897)
test bench_large_sequential ... bench:   1,218,690 ns/iter (+/- 120,466)
test bench_small_parallel   ... bench:     125,682 ns/iter (+/- 105,233)
test bench_small_sequential ... bench:      42,321 ns/iter (+/- 6,397)
test bench_tiny_parallel    ... bench:     112,742 ns/iter (+/- 94,979)
test bench_tiny_sequential  ... bench:         100 ns/iter (+/- 0)
```

ijanos:
I like this much more with the inclusion of the parallel-faster example.

petertseng:
Currently thinking about:

> Performant code (usually) goes in the opposite direction, right? That's usually the way. To get the best performance you frequently sacrifice aspects of design/idiom/etc.
>
> So this is a mindshift for students, right? Instead of being encouraged to write one kind of Rust, we're now suggesting they do a different kind.
>
> How do we tell students that they need to change their thinking? That they need to start considering tradeoffs that we haven't explored up until now?

It seems to me that even without explicitly optimising for performance (and possibly sacrificing style while doing so), the parallel implementation will be demonstrated to be faster on large inputs. It seems possible that a student may be satisfied with this. For small inputs it will probably be the case that the overhead exceeds the benefit from parallel computation, and I hope this is what students learn from the provided benchmark. It is possible that some students will try to interpret it as "performance at all costs", in which case we might try to adjust some text to explain.

Soon after 24 hours have passed since the latest iteration was pushed, I think I'll merge.

petertseng:
Very nice, thank you!

IanWhitney:
Thanks!

glennpratt:
Woohoo. @ijanos @petertseng @IanWhitney thanks for all the feedback.

PR #247, opened 2017-01-14
Implement Luhn

### LINKED ISSUE #237: Implement Luhn
Following the [current test suite](https://github.com/exercism/x-common/blob/aa12f2e403270854e709b20a2aa62cf24e60d171/exercises/luhn/canonical-data.json) implement the [luhn exercise](https://github.com/exercism/x-common/tree/aa12f2e403270854e709b20a2aa62cf24e60d171/exercises/luhn)

IanWhitney:
Closes #237

This mostly follows the current test suite
(https://github.com/exercism/x-common/issues/491), with one change.

The final canonical test adds an alphabetical character to a known
invalid string and then asserts that it's still invalid.

Instead what the test should do is alter a known-valid string and show
that it is now invalid.

That's what I've done here. I'll also propose the change to canonical
test suite.


IanWhitney:
I haven't placed the problem yet, so this can not be merged.

exercises/luhn/example.rs
```diff
@@ -0,0 +1,14 @@
+pub fn is_valid(candidate: &str) -> bool {
+    if candidate.chars().any(|c| c.is_alphabetic()) || candidate.chars().count() == 1 {
+        return false;
+    }
+
+    candidate.chars()
+        .filter(|c| c.is_numeric())
```
petertseng: can I suggest `filter_map` here? It seems perfect for this use.

IanWhitney: yep.

IanWhitney:
Thinking about placement. I don't think my solution is that far from what people will do. The [existing crate](https://lunemec.github.io/rust-luhn/src/luhn2/src/lib.rs.html) uses another crate to handle digits > 10 after doubling. If people go down that path their solutions will vary a bit from mine, but not by a bunch.

So the solution mostly requires a good handle on 

- Converting a string to digits
- Working with iteration

I would put it between Sum of Multiples and Scrabble Score.

IanWhitney:
This works, but I've been thinking about some other, more Rusty ways of approaching it.

### Option 1: Use `From` 

This way we could validate strings and different types of unsized ints. The test API would look like:

```rust
let luhn = Luhn::from("012345");
assert!(!luhn.is_valid());
let luhn = Luhn::from(12345);
assert(!luhn.is_valid());
```

### Option 2: Use Traits

Students would create a `LuhnValidator` trait and then implement it for different types. Test API would look like:

```
let string_luhn = String::from("012345")
assert!(!string_luhn.valid_luhn());

let u8_luhn: u8 = 8; 
assert!(!u8_luhn.valid_luhn());
//and so on
```

IanWhitney:
### Option 3

Do all of them. Have 3 Luhn exercises that show different ways of solving the problem, so that students can see each approach.

petertseng:
I know you already placed, but I'm going to make my placement comment without looking at where you placed to avoid bias, and if our ideas match up then great.

Place no later than robot-simulator.
Place no earlier than hamming.

Fantastic, we agree.

exercises/luhn/example.rs
```diff
@@ -0,0 +1,13 @@
...
+    if candidate.chars().any(|c| c.is_alphabetic()) || candidate.chars().count() == 1 {
+        return false;
+    }
+
+    candidate.chars()
+        .filter_map(|c| c.to_digit(10))
+        .rev()
+        .enumerate()
+        .map(|(index, digit)| if index % 2 == 0 { digit } else { digit * 2 })
+        .map(|digit| if digit > 10 { digit - 9 } else { digit })
```
petertseng: sorry, but this code says `!is_valid("59")` and `is_valid("50")`, when instead it should say `is_valid("59")` and `!is_valid("50")`.

exercises/luhn/tests/luhn.rs
```diff
@@ -0,0 +1,38 @@
...
+
+#[test]
+#[ignore]
+fn single_zero_string_is_invalid() {
+    assert!(!is_valid("0"));
+}
+
+#[test]
+#[ignore]
+fn valid_canadian_sin_is_valid() {
```
petertseng: long cat is long.

(No change necessary)

petertseng:
EBWODP: 

> Yeah, interesting way to add functions to existing types. Do we have any exercise that already does this?

Looking at our git grep "impl.*for", I guess not.

exercises/luhn/example.rs
IanWhitney: Ah, good catch. Should be `digit > 9`.

IanWhitney:
Just this week I was trying to figure out how to monkey-patch Ruby to add methods to the number `2`. Not that this is a _good_ idea, but it was interesting. For the same reason I think show traits on primitives could be interesting, though the documentation of the exercise should say that it is not recommended.

I'm going to experiment with Option 3 and will either update this PR or create a new one.

PR #248, opened 2017-01-16
Example of Luhn parts 2 & 3, From Trait and Custom Trait

IanWhitney:
This is an implementation of the idea I discussed here

https://github.com/exercism/xrust/pull/247#issuecomment-272669098

After students do the original Luhn exercise, they can learn how to use
the From trait to allow validation of multiple types. Then they can implement it again with a custom type.


exercises/luhn-from/HINTS.md
```diff
@@ -0,0 +1,9 @@
+# Luhn with From Trait
+
+Before doing this exercise you should probably do the original Luhn exercise.
+
+`exercism fetch rust luhn`
+
+In the original Luhn exercise you only validated strings, but the Luhn algorithm can be applied integers as well.
```
petertseng: "applied to"

exercises/luhn-trait/HINTS.md
```diff
@@ -0,0 +1,16 @@
+# Luhn Validation with a Trait
+
+Before doing this exercise you should probably do the original Luhn exercise and its successor, Luhn From Trait 
+
+- `exercism fetch rust luhn`
+- `exercism fetch rust luhn-from`
+
+In the original Luhn exercise you only validated strings, but the Luhn algorithm can be applied integers as well.
```
petertseng: "applied to"

exercises/luhn-trait/HINTS.md
```diff
@@ -0,0 +1,16 @@
...
+
+In Luhn From you implemented a From trait, which also required you to create a Luhn struct. 
+
+Instead of creating a Struct just to perform the validation, what if you you validated the primitives (i.e, String, u8, etc.) themselves?
+
+In this exercise you'll create and implement a custom [trait](https://doc.rust-lang.org/book/traits.html) that performs the validation
```
petertseng: End sentence with a period, I think.

petertseng:
EBWODP: I modified my diff to include files of the new problem, as we have done. Still same error.

IanWhitney:
I am unfamiliar with Trackler. Here's what I see after setting it up and pointing it's version of the Rust track to this branch.

```ruby
rust_track = Trackler.tracks["rust"]
implementations = rust_track.implementations
luhn_from = implementations.detect {|impl| impl.problem.slug == 'luhn-from'}
# => #<Trackler::Implementation:0x007fa6101799f0 @track_id="rust",...>
luhn_from.files
# => NoMethodError: undefined method `[]' for nil:NilClass
```

The error is thrown [here](https://github.com/exercism/trackler/blob/master/lib/trackler/problem.rb#L75), because [the metadata method](https://github.com/exercism/trackler/blob/master/lib/trackler/problem.rb#L108) returns nil.

I think that's a fixable problem in Trackler. What it's trying to do here is pull together the language-independent files for an exercise. But if those files don't exist there should be fall-back behavior. 

I think. I mean, I've only looked at that code for 5 minutes so I'm sure there's complexity I'm missing. I wanna dig into it more. Will have to wait until Wednesday at the earliest, though.

IanWhitney:
Not sure when it'll be deployed, but track-specific exercises should be possible soon.

https://github.com/exercism/trackler/pull/24

config.json
```diff
@@ -132,6 +132,26 @@
       ]
     },
     {
+      "slug": "luhn-from",
+      "difficulty": 1,
```
petertseng: now that I gave Luhn a 4, perhaps this should be as well

config.json
```diff
@@ -132,6 +132,26 @@
...
+      "topics": [
+        "from trait",
+        "str to digits",
+        "iterators",
+        "higher-order functions"
+      ]
+    },
+    {
+      "slug": "luhn-trait",
+      "difficulty": 1,
```
petertseng: > now that I gave Luhn a 4, perhaps this should be as well

same

exercises/luhn-from/example.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+impl From<String> for Luhn {
+    fn from(s: String) -> Self {
+        Luhn { digits: s.chars().collect() }
+    }
+}
+
+impl From<&'static str> for Luhn {
+    fn from(s: &'static str) -> Self {
```
petertseng: I understand it's only the example, but I'm not clear on why this would have to be `&'static str` specifically. What would happen if this were just `&str`? It seems unnecessarily restrictive to me to only accept `&'static str`.

exercises/luhn-trait/HINTS.md
```diff
@@ -0,0 +1,16 @@
+# Luhn Validation with a Trait
+
+Before doing this exercise you should probably do the original Luhn exercise and its successor, Luhn From Trait 
+
+- `exercism fetch rust luhn`
+- `exercism fetch rust luhn-from`
+
+In the original Luhn exercise you only validated strings, but the Luhn algorithm can be applied to integers as well.
+
+In Luhn From you implemented a From trait, which also required you to create a Luhn struct.
```
petertseng: here, it says "Luhn From", whereas above it says "Luhn From Trait". In the header of the `HINTS.md` of the `luhn-from`, it says "Luhn with From Trait". Intentional? Seems OK I guess since they should all be distinguishable.

exercises/luhn-trait/example.rs
```diff
@@ -0,0 +1,55 @@
...
+            .filter_map(|c| c.to_digit(10))
+            .rev()
+            .enumerate()
+            .map(|(index, digit)| if index % 2 == 0 { digit } else { digit * 2 })
+            .map(|digit| if digit > 9 { digit - 9 } else { digit })
+            .sum::<u32>() % 10 == 0
+    }
+}
+
+impl Luhn for &'static str {
```
petertseng: same - I'd ask if there is any requirement that these be `&'static str` rather than just `&str`.

IanWhitney:
Rebased on master and made @petertseng's requested changes. 

I'm not sure if the updated Trackler is deployed yet, so I don't know what will happen if this gets deployed.

IanWhitney:
Nearly done. After rebasing down to a single commit I realized that I had the Luhn followups in the wrong part of the track. 

"Luhn: Using the From Trait" now moves to just after Bracket Push (which is our example that uses `From`). In the Luhn implementation students will implement `From` for multiple types, so it makes sense to put it after Bracket Push where students only implement it for one type.

"Luhn: Using a Custom Trait" moves to after Space Age. Space Age solutions use custom traits and `From`, so pairing these exercises made sense.

IanWhitney:
Checked with Katrina. Trackler is updated, so merging this is fine.

kytrinyx:
@IanWhitney trackler assumes that there's a `description.md` file either in `x-common/exercises/:slug/description.md` or in `x:track_id/exercises/:slug/description.md`. I don't see that here.

PR #249, opened 2017-01-19
Tests for scrabble-score's optional extensions

juleskers:
So I set myself the challenge of doing scrabble-score's extensions for `:double`- and `:triple`-scoring letters.
In the course of doing so, I wrote tests. I thought these may be useful to the source, so here is my offer to include them upstream.

Comments, suggestions, nits and other (constructive) criticism welcome; this is my first contribution to exercism.io, so I may need some guidance ;-)

### DIFF
--- a/exercises/scrabble-score/example.rs
+++ b/exercises/scrabble-score/example.rs
@@ -1,40 +1,35 @@
-use std::collections::HashMap;
-
 pub fn score(word: &str) -> u16 {
-    let values = dictionary();
-    word.to_lowercase()
-        .chars()
-        .map(|c| values.get(&c).clone())
-        .fold(0, |score, v| score + v.unwrap_or(&0))
+    let mut score: u16 = 0;
+    let mut previous_char_value: u16 = 0;
+
+    // normalise the word to lowercase, for case-insensitivity, also, for easier iteration,
+    // reduce the multi-character value multipliers :double and :triple to a single char.
+    let tokenised_word = word.to_lowercase().replace(":double", "2").replace(":triple", "3");
+
+    for a_char in tokenised_word.chars() {
+        let value = char_value(a_char, previous_char_value);
+        score += value;
+        previous_char_value = value;
+    }
+
+    score
 }
 
-fn dictionary() -> HashMap<char, u16> {
-    let mut values = HashMap::new();
-    values.insert('a', 1);
-    values.insert('b', 3);
-    values.insert('c', 3);
-    values.insert('d', 2);
-    values.insert('e', 1);
-    values.insert('f', 4);
-    values.insert('g', 2);
-    values.insert('h', 4);
-    values.insert('i', 1);
-    values.insert('j', 8);
-    values.insert('k', 5);
-    values.insert('l', 1);
-    values.insert('m', 3);
-    values.insert('n', 1);
-    values.insert('o', 1);
-    values.insert('p', 3);
-    values.insert('q', 10);
-    values.insert('r', 1);
-    values.insert('s', 1);
-    values.insert('t', 1);
-    values.insert('u', 1);
-    values.insert('v', 4);
-    values.insert('w', 4);
-    values.insert('x', 8);
-    values.insert('y', 4);
-    values.insert('z', 10);
-    values
+/// Calculates the value of the current character.
+///
+/// The previous characters' value is required in case the current
+/// token is a multiplier, which affects the score of the previous char.
+fn char_value(the_char: char, prev_value: u16) -> u16 {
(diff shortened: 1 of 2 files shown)


juleskers:
Aha, the Travis tests fail because my new tests fail.
It seems the reference implementation doesn't actually do the double and triple letter scoring.

petertseng:
Oh yeah! and before I forget, I have to ask to ignore all but the first test, as I did in the other PR

IanWhitney:
I think this is a good start. Thanks.

juleskers:
Thanks for the feedback! I'll try to incorporate it!
The #[ignore] will be added, but probably only after I've passed the Travis round at least once.
(Or is travis smart enough to un-ignore the tests?)

I may be slow to respond, but that won't mean I've abandoned this, it's just that I'm doing this in the brief lost minutes during my commute :-)

I'm undecided on `:garbage`;
I was considering that `:doubleSomeword` would be the syntax for a double-scoring word; The spec leaves open how this would be expected.

petertseng:
> (Or is travis smart enough to un-ignore the tests?)

Yes, it does. I should have explained it, my bad for not doing so!
You don't necessarily need to look at this script, but if you are curious, our Travis CI runs https://github.com/exercism/xrust/blob/master/_test/check-exercises.sh#L42, thus unignoring all tests.

> I was considering that `:doubleSomeword` would be the syntax for a double-scoring word; The spec leaves open how this would be expected.

Ah! This could be a fine idea. You are right about the spec.

As in the other PR, this could be a thing all languages can benefit from, so it could be worth opening an issue in x-common to discuss how the spec should be interpreted. And if we reach an agreement, a PR can be sent to x-common.

If agreement can't be reached in x-common, we could possibly clarify the specification for our track by adding a exercises/scrabble-score/HINTS.md - this is a special file that gets appended to the README.

juleskers:
update!
* I've #[ignored] all newly added tests
* example code updated to pass all tests (locally ran check_exercises.sh, passed!)
* slight tweak of one test name

feel free to nitpick over variable names and comments! It's the example after all, so it should be clear!

I'm opening a discussion issue on x-common next.

exercises/scrabble-score/example.rs
```diff
@@ -1,40 +1,35 @@
...
-    values.insert('w', 4);
-    values.insert('x', 8);
-    values.insert('y', 4);
-    values.insert('z', 10);
-    values
+/// Calculates the value of the current character.
+///
+/// The previous characters' value is required in case the current
+/// token is a multiplier, which affects the score of the previous char.
+fn char_value(the_char: char, prev_value: u16) -> u16 {
```
petertseng: I might also consider `c` here.

IanWhitney:
a delimiter like `[]` works for me

- double `[]`
- triple `<>`

- aw[e]some
- aw<e>some
- [awesome]
- <awesome>
- <aw[e]some>
- etc.

IanWhitney:
And I'm fine having tests in place to handle the "extra credit" problems, even if those tests aren't in canonical.

juleskers:
Ah, delimiters! I like how simple the implementation could become:
(I wonder if that compiles...? Mutating the multiplyer inside the match AND reading it in very the same expression looks maybe a bit too compact to work. But maybe Rust lets us have nice things!)

`````rust
score += Match c {
   '[' => multiplyer *= 2; 0,
   ']' => multiplyer /= 2; 0,
   '<' => multiplyer *= 3; 0,
   '>' => multiplyer /= 3; 0,
   'a'|'e'|'u'.... => 1, 
   ... => 2,
   _ => 0 
} * multiplyer
`````

IanWhitney:
If I'm reading this PR correctly, there are just a few small changes @petertseng mentioned in comments. Are those blocking merge? Or can we move on this PR without those changes?

petertseng:
I had a different reading of this PR: I thought all of my comments are unimportant, and what we would want to discuss before merge is:

> In-line notation seem super awkward to me. Parsing seems fraught with problems, as you mention. I think a more robust structure is necessary.

And that whatever notation we use should be added to HINTS.md

IanWhitney:
Ah, yes. I saw the comment approving the ideas of delimiters, but hadn't checked the code to see if delimiters had been implemented.

juleskers:
After the discussion on the common track, I understood that the optional exercise was intentionally left vague. Figuring out a syntax was explicitly said to be part of the exercise.
That's why I haven't pursued this further.

IanWhitney:
Ok. Sounds like we can close, then. Thanks for your work on this, @juleskers!

PR #250, opened 2017-01-19
scrabble-score: add a test for the entire alphabet

juleskers:
After passing all tests for scrabble-score, I noticed I still didn't have all letters.
Adding them to my implementation at that point of course was trivial, but since there wasn't a test, I wrote one.
Here it is offered for inclusion at the source!

so comments, suggestions, nits and other (constructive) criticism welcome;
This is only my second[1] exercism contribution, so I may need some guidance ;-)

[1] [first contribution](https://github.com/exercism/xrust/pull/249)

### DIFF
--- a/exercises/scrabble-score/tests/scrabble-score.rs
+++ b/exercises/scrabble-score/tests/scrabble-score.rs
@@ -61,3 +61,8 @@ fn non_english_scrabble_letters_do_not_score() {
 fn empty_words_are_worth_zero() {
     assert_eq!(score(""), 0);
 }
+
+#[test]
+fn all_letters_work() {
+    assert_eq!(score("abcdefghijklmnopqrstuvwxyz"), 87);
+}
--- a/exercises/scrabble-score/tests/scrabble-score.rs
+++ b/exercises/scrabble-score/tests/scrabble-score.rs
@@ -63,6 +63,7 @@ fn empty_words_are_worth_zero() {
 }
 
 #[test]
+#[ignore]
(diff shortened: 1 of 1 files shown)


IanWhitney:
I know that in the [canonical tests](https://github.com/exercism/x-common/blob/c79801789c796ee140caa1bbde5a6439eb63ef49/exercises/scrabble-score/canonical-data.json) we've pulled out words that couldn't be valid scrabble words. I'm guessing the same argument will apply here.

Before we update the Rust tests, I think the change should happen in the canonical tests.

juleskers:
Thanks for the feedback!
I'll make sure to add an #[ignore].

As for the canonical tests, that makes a lot of sense, I wasn't aware there was so much coordination going on between tracks!

As for the alphabet, you are correct that it's not a scrabble word. I would argue that the alphabet itself is the clearest test-case to prove that the entire alphabet is present.
If words are a hard requirement, I could try "the quick brown fox jumps over the lazy dog" (more a sentence than a word, I admit, but maybe in 8 asserts?)

As I mentioned in the other Pull-request, I won't be the fastest to reply, as I do this in my lost minutes on the train.

petertseng:
It did get added in x-common - don't forget to add the `ignore` here.

juleskers:
Added! I had it lying around already, but didn't have time/internet to push it.

petertseng:
Thanks!

(I squashed the two commits before merging since these two commits make the most sense as a single one)

PR #252, opened 2017-01-29
config.json: Add difficulties

### LINKED ISSUE #179: Update config.json to match new specification
For the past three years, the ordering of exercises has been done based on gut feelings and wild guesses. As a result, the progression of the exercises has been somewhat haphazard.

In the past few months maintainers of several tracks have invested a great deal of time in analyzing what concepts various exercises require, and then reordering the tracks as a result of that analysis.

It would be useful to bake this data into the track configuration so that we can adjust it over time as we learn more about each exercise.

To this end, we've decided to add a new key _exercises_ in the config.json file, and deprecate the `problems` key.

See exercism/discussions#60 for details about this decision.

Note that we will **not** be removing the `problems` key at this time, as this would break the website and a number of tools.

The process for deprecating the old `problems` array will be:
- Update all of the track configs to contain the new _exercises_ key, with whatever data we have.
- Simultaneously change the website and tools to support both formats.
- Once all of the tracks have added the _exercises_ key, remove support for the old key in the site and tools.
- Remove the old key from all of the track configs.

In the new format, each exercise is a JSON object with three properties:
- _slug_: the identifier of the exercise
- _difficulty_: a number from 1 to 10 where 1 is the easiest and 10 is the most difficult
- _topics_: an array of strings describing topics relevant to the exercise. We maintain
  a list of common topics at https://github.com/exercism/x-common/blob/master/TOPICS.txt. Do not feel like you need to restrict yourself to this list;
  it's only there so that we don't end up with 20 variations on the same topic. Each
  language is different, and there will likely be topics specific to each language that will
  not make it onto the list.

The _difficulty_ rating can be a very rough estimate.

The _topics_ array can be empty if this analysis has not yet been done.

Example:

```
"exercises": [
  {
    "slug": "hello-world" ,
    "difficulty": 1,
    "topics": [
        "control-flow (if-statements)",
        "optional values",
        "text formatting"
    ]
  },
  {
    "difficulty": 3,
    "slug": "anagram",
    "topics": [
        "strings",
        "filtering"
    ]
  },
  {
    "difficulty": 10,
    "slug": "forth",
    "topics": [
        "parsing",
        "transforming",
        "stacks"
    ]
  }
]
```

It may be worth making the change in several passes:
1. Add the _exercises_ key with the array of objects, where _difficulty_ is 1 and _topics_ is empty.
2. Update the difficulty settings to reflect a more accurate guess.
3. Add topics (perhaps one-by-one, in separate pull requests, in order to have useful discussions about each exercise).

petertseng:
As noted in #195, difficulty rating is hard.

This commit proposes a difficulty rating scale that should be completely
unambiguous, requiring no judgment:

In the last version of problems.md before it was rewritten in #242:

* Everything in "Introduction" stays at 1.
* Everything in "Getting Rusty" gets a 4.
* Everything in "Rust Gets Strange" gets a 7.
* Everything in "Putting it all Together" gets a 10.

It is acknowledged that this is not a perfect difficulty scheme. See,
for example, how many exercises are at 4.
However, it is more accurate than leaving them all at 1 because it is
generally true that those exercises in "Getting Rusty" are more
difficult than those in "Introduction".
It is a starting point from which further adjustments can be made.
I posit that there is little motivation to adjust further until the
difficulty ratings actually show up on the website, since until then
they are just arbitrary numbers with no meaning.

Closes #179.

### DIFF
--- a/config.json
+++ b/config.json
@@ -81,14 +81,14 @@
     },
     {
       "slug": "hamming",
-      "difficulty": 1,
+      "difficulty": 4,
       "topics": [
         "Result"
       ]
     },
     {
       "slug": "pascals-triangle",
-      "difficulty": 1,
+      "difficulty": 4,
       "topics": [
         "Math",
         "Vec",
@@ -97,7 +97,7 @@
     },
     {
       "slug": "scrabble-score",
-      "difficulty": 1,
+      "difficulty": 4,
       "topics": [
         "chaining higher-order functions",
         "HashMap (optional)"
@@ -105,7 +105,7 @@
     },
     {
       "slug": "pangram",
-      "difficulty": 1,
+      "difficulty": 4,
       "topics": [
         "filter",
         "ascii (optional)"
@@ -113,7 +113,7 @@
     },
     {
       "slug": "nucleotide-count",
-      "difficulty": 1,
+      "difficulty": 4,
       "topics": [
         "Result",
         "filter",
@@ -124,7 +124,7 @@
     },
     {
       "slug": "luhn",
-      "difficulty": 1,
+      "difficulty": 4,
       "topics": [
         "str to digits",
         "iterators",
@@ -133,7 +133,7 @@
     },
     {
       "slug": "largest-series-product",
-      "difficulty": 1,
+      "difficulty": 4,
       "topics": [
         "Result",
         "windows",
@@ -143,7 +143,7 @@
     },
     {
       "slug": "word-count",
-      "difficulty": 1,
+      "difficulty": 4,
       "topics": [
         "hashmap",
         "str vs string",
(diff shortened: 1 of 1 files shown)


IanWhitney:
> it is better than leaving them all at 1

Why?

petertseng:
Commit message updated.

> However, it is more accurate than leaving them all at 1 because it is
> generally true that those exercises in "Getting Rusty" are more
> difficult than those in "Introduction".

This of course only answers the "better" question if it is true that we care about "accuracy". If we don't care (one could rightly argue that *we have no reason to care* since it doesn't show up anywhere), then this argument falls flat.

petertseng:
> if it is true that we care about "accuracy". If we don't care

Or the opposite - if we care enough that this inaccurate measure is **worse** than making no attempt.

I think I've made clear that I actually do not care, so I'm not the right person to elaborate on the above point - I'll leave that to others.

IanWhitney:
A guess is a fine place to start. Maybe later we'll learn something new about difficulty that will improve our guesses. LGTM.

PR #253, opened 2017-01-29
luhn: Update example and add more tests

gibfahn:
The luhn test suite didn't cover a lot of cases, including:
 - Strings of zeros
 - Non-alphabetic characters
 - Numbers which are only valid if you reverse the string

This adds those tests and updates the example to properly reject symbols (and check that the sum isn't zero).

Let me know if I'm doing something wrong!

**_EDIT:_** FYI I was previously using `!c.is_whitespace()` rather than `c != ' '`. I switched because technically other whitespace like `\t` and `\n` isn't allowed by the letter of the rules.

### DIFF
--- a/exercises/luhn/example.rs
+++ b/exercises/luhn/example.rs
@@ -1,5 +1,6 @@
 pub fn is_valid(candidate: &str) -> bool {
-    if candidate.chars().any(|c| c.is_alphabetic()) || candidate.chars().count() == 1 {
+    if candidate.chars().filter(|c| c.is_digit(10)).take(2).count() <= 1 ||
+       candidate.chars().any(|c| !c.is_digit(10) && c != ' ') {
         return false;
     }
 
--- a/exercises/luhn/tests/luhn.rs
+++ b/exercises/luhn/tests/luhn.rs
@@ -15,6 +15,12 @@ fn single_zero_string_is_invalid() {
 
 #[test]
 #[ignore]
+fn simple_valid_sin() {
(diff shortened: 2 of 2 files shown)


IanWhitney:
Thanks for the contribution! For most of our exercises we use the [canonical test data](https://github.com/exercism/x-common/blob/9b580e86170475cf571a1b2b7c8e2c5433e67b41/exercises/luhn/canonical-data.json) defined in the [x-common repo](https://github.com/exercism/x-common).

If we're going to improve Luhn's test suite, we should start by changing the canonical test file. Then all the tracks can implement the improved tests.

gibfahn:
https://github.com/exercism/x-common/pull/522 has now landed, I've updated this, PTAL.

exercises/luhn/example.rs
```diff
@@ -1,5 +1,6 @@
 pub fn is_valid(candidate: &str) -> bool {
-    if candidate.chars().any(|c| c.is_alphabetic()) || candidate.chars().count() == 1 {
+    if input.chars().filter(|c| c.is_digit(10)).take(2).count() <= 1 ||
```
petertseng: you seem to have intended to change the input's name to `input`, but didn't make the same change in the argument. I believe if you were to do that, the tests could pass.

gibfahn: My bad, copying from my solution (which used input) to example, I forgot to change the name back. Hopefully fixed.

PR #255, opened 2017-02-01
hello-world: simplify to unconditional "Hello, World!"

petertseng:
The introductory exercise should require as little Rust knowledge as
necessary. We'd like to introduce `cargo test` here since it will be
used for all exercises, but introducing Option is too much.

https://github.com/exercism/x-common/issues/520

---

I wouldn't recommend merging this until the corresponding change is made in x-common.
Let's get it ready to go before then.

### DIFF
--- a/config.json
+++ b/config.json
@@ -8,7 +8,6 @@
       "slug": "hello-world",
       "difficulty": 1,
       "topics": [
-        "Some/None",
         "println!"
       ]
     },
--- a/exercises/hello-world/example.rs
+++ b/exercises/hello-world/example.rs
@@ -1,6 +1,3 @@
-pub fn hello(name: Option<&str>) -> String {
-    match name {
-        Some(n) => format!("Hello, {}!", n),
-        None => "Hello, World!".to_string(),
-    }
+pub fn hello() -> &'static str {
+    "Hello, World!"
 }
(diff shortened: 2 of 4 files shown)


ijanos:
I'm in favour of simplification but looking at this got me thinking. Strings in rust are not simple, while the classic tutorial work with a simple `println!("Hello, world!");` here we have to call `.to_string()` or `.into()` or `String::from()` to satisfy the test. 

If the purpose of this exercise is just to test that everything works then I'm starting to lean in the direction to use a primitive type here, a bool could work with something like this: 

```
pub fn everything_works() -> bool {
    true
}
```

At this point however we have no connection to the common description. So another approach could be accepting a static string slice in the test. No conversion to `String` required but the type signature could be confusing or hard to guess if we don't provide a stub. 

```
pub fn hello() -> &'static str {
    "Hello, World!"
}
```

Thoughts?

petertseng:
Hmm, good point. I would say there are various measures we can take to lighten the load. We might combine some of these measures:

* static str, so we don't have to do to_string
* provide a stub
* provide a failing implementation that says "Goodbye, world" - then the student only has to change the Goodbye to Hello, should give the idea.

petertseng:
Let's take all three. Note that using `&'static str` did not require any change to the test.

petertseng:
OK, take a look at that comment. Feel free to suggest changes to the wording.

ijanos:
Looks good to me, but keep in mind that I am not a native English speaker :)

petertseng:
Also removed `Some/None` from the config.json for hello-world.

PR #256, opened 2017-02-02
grade-school: Expect `Option<Vec<String>>` from grade; add stub file

### LINKED ISSUE #254: Could the grade-school tests be less prescriptive?
I'm working through the Rust exercisms and have so far enjoyed the journey. grade-school led to some frustration though. At first I thought that was just down to me trying to program on the train home after having had a few beers, but when I decided (which I hadn't done on any other of the exercises so far) to just submit an empty solution in order to see what others have done I couldn't help but notice I wasn't alone in my struggles, others commented on/criticised the tests as follows:

> [...], but that requires rewriting the tests to take ownership.
http://exercism.io/submissions/0e24aaf3d672454eb536acdba0dd7073

> // test forces me to use `Vec` instead of `BTreeSet`
--> http://exercism.io/submissions/604fd3bd27614f98a939eeb40428f70d

> The test cases here made it such that one was expected to solve this problem in one way. [...]
--> http://exercism.io/submissions/3f9fc61277c94385be7338fb2c8e04e4

> I gave up...
--> http://exercism.io/submissions/c11bd73b8f3c48dcb23c4dba66cc0d41

> I hated this example. The tests have been written in such a way that they assume very specific function definitions, [...]
--> http://exercism.io/submissions/948e0931d14d42bbb11dcc4df01f09c8

> // I would prefer to use a BTreeSet here, or at least
> // BinaryHeap, but the test definition seems to make it
> // impossible [...]
--> http://exercism.io/submissions/845cdbef93374155ad7632c9f339aabf

petertseng:
Previously, grade was expected to return an `Option<&Vec<String>>`.
Because there's a reference inside the lifetime, any implementing School
would have been required to maintain ownership of some vectors so that
they can be lended out. This unnecessarily restricts the implementation.
For example, some implementations may instead have preferred to use
BTreeSet.

If we expect grade to return `Option<Vec<String>>`, the fact that
ownership is given means that a new vector can be made from whatever
internal representation is used.

Closes #254.

### DIFF
--- a/exercises/grade-school/src/lib.rs
+++ b/exercises/grade-school/src/lib.rs
@@ -1,0 +1,26 @@
+#[allow(unused_variables)]
+
+pub struct School {
+}
+
+impl School {
+    pub fn new() -> School {
+        unimplemented!()
+    }
+
+    pub fn add(&mut self, grade: u32, student: &str) {
+        unimplemented!()
+    }
+
+    pub fn grades(&self) -> Vec<u32> {
+        unimplemented!()
+    }
+
+    // If grade returned an `Option<&Vec<String>>`,
+    // the internal implementation would be forced to keep a `Vec<String>` to lend out.
+    // By returning an owned vector instead,
+    // the internal implementation is free to use whatever it chooses.
+    pub fn grade(&self, grade: u32) -> Option<Vec<String>> {
+        unimplemented!()
+    }
+}
--- a/exercises/grade-school/example.rs
+++ b/exercises/grade-school/example.rs
@@ -21,7 +21,7 @@ impl School {
         s
     }
 
-    pub fn grade(&self, grade: u32) -> Option<&Vec<String>> {
-        self.grades.get(&grade)
+    pub fn grade(&self, grade: u32) -> Option<Vec<String>> {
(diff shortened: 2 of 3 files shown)


pviecelli:
Thanks! That makes more sense now! :)

exercises/grade-school/src/lib.rs
```diff
@@ -0,0 +1,22 @@
...
+
+    pub fn add(&mut self, grade: u32, student: &str) {
+        unimplemented!()
+    }
+
+    pub fn grades(&self) -> Vec<u32> {
+        unimplemented!()
+    }
+
+    pub fn grade(&self, grade: u32) -> Option<Vec<String>> {
```
petertseng: https://github.com/exercism/xrust/issues/254#issuecomment-276057035

> It would be nice if, when a constraining choice like this is made, the README or stub file could be explicit about why that interface was chosen.

I failed to do this. It could be something like "If grade returned an `Option<&Vec<String>>`, the internal implementation would be forced to keep a `Vec<String>` to lend out. By returning an owned vector instead, the internal implementation is free to use whatever it chooses."

petertseng:
Ten days, one definite vote that this improves things, and no objections. I'm merging it.

PR #257, opened 2017-02-03
added rotational-cipher exercise

LogoiLab:
This cipher is the same as atbash except it is implemented in a way that allows for 27 possible keys. Some call this the Caesar cipher but the official cryptography term is rotational cipher. For more info see the readme in this pull request.

This exercise should come right after the atbash cipher in terms of ordering and difficulty.

### DIFF
--- a/exercises/rotational-cipher/Cargo.toml
+++ b/exercises/rotational-cipher/Cargo.toml
@@ -1,0 +1,4 @@
+[package]
+name = "rotational-cipher"
+version = "0.0.0"
+authors = ["Chad Baxter <cbaxter@mail.umw.edu>"]
--- a/exercises/rotational-cipher/example.rs
+++ b/exercises/rotational-cipher/example.rs
@@ -1,0 +1,16 @@
+pub fn rotate(text: &str, shift_key: u8) -> String {
+
+    text.chars().map(|c| {
+        let case = if c.is_uppercase() {
+            'A'
+        } else {
+            'a'
+        } as u8;
+        if c.is_alphabetic() {
+            (((c as u8 - case + shift_key) % 26) + case) as char
+        } else {
+            c
+        }
+    }).collect::<String>()
+
+}
--- a/exercises/rotational-cipher/tests/rotational-cipher.rs
+++ b/exercises/rotational-cipher/tests/rotational-cipher.rs
@@ -1,0 +1,61 @@
+extern crate rotational_cipher as cipher;
+
+#[test]
+fn rotate_a_1() {
+    assert_eq!("b", cipher::rotate("a", 1));
+}
+
+#[test]
+#[ignore]
+fn rotate_a_26() {
+    assert_eq!("a", cipher::rotate("a", 26));
+}
+
+#[test]
+#[ignore]
+fn rotate_a_0() {
+    assert_eq!("a", cipher::rotate("a", 0));
+}
+
+#[test]
+#[ignore]
+fn rotate_m_13() {
+    assert_eq!("z", cipher::rotate("m", 13));
+}
+
+#[test]
+#[ignore]
+fn rotate_n_13_with_wrap() {
+    assert_eq!("a", cipher::rotate("n", 13));
+}
+
+#[test]
+#[ignore]
+fn rotate_caps() {
+    assert_eq!("TRL", cipher::rotate("OMG", 5));
+}
+
+#[test]
+#[ignore]
+fn rotate_spaces() {
+    assert_eq!("T R L", cipher::rotate("O M G", 5));
+}
+
+#[test]
+#[ignore]
+fn rotate_numbers() {
+    assert_eq!("Xiwxmrk 1 2 3 xiwxmrk", cipher::rotate("Testing 1 2 3 testing", 4));
+}
+
+#[test]
+#[ignore]
+fn rotate_punctuation() {
+    assert_eq!("Gzo\'n zvo, Bmviyhv!", cipher::rotate("Let\'s eat, Grandma!", 21));
+}
+
+#[test]
+#[ignore]
+fn rotate_all_the_letters() {
+    assert_eq!("Gur dhvpx oebja sbk whzcf bire gur ynml qbt.",
+        cipher::rotate("The quick brown fox jumps over the lazy dog.", 13));
+}
--- a/config.json
+++ b/config.json
@@ -163,6 +163,17 @@
       ]
     },
     {
+      "slug": "rotational-cipher",
+      "difficulty": 4,
+      "topics": [
+        "str vs string",
+        "primitive types",
+        "iterators",
+        "chars",
+        "ascii"
+      ]
+    },
+    {
       "slug": "etl",
       "difficulty": 4,
       "topics": [
--- a/exercises/rotational-cipher/Cargo.lock
+++ b/exercises/rotational-cipher/Cargo.lock
@@ -1,0 +1,4 @@
+[root]
+name = "rotational-cipher"
+version = "0.0.0"
+


LogoiLab:
I'v only ever coded on nightly and am fairly new to Rust, I might need some help getting this working for beta and stable.

exercises/rotational-cipher/Cargo.toml
```diff
@@ -0,0 +1,4 @@
+[package]
+name = "rotational-cipher"
+version = "0.0.0"
+authors = ["Chad Baxter <cbaxter@mail.umw.edu>"]
```
petertseng: note - the second exercise with an `authors` line (the first is nucleotide-codons).

exercises/rotational-cipher/README.txt
```diff
@@ -0,0 +1,71 @@
+# Rotational Cipher
+
+Create an implementation of the rotational cipher, also sometimes called the Caesar cipher.
+
+The Caesar cipher is a simple shift cipher that relies on
+transposing all the letters in the alphabet using an integer key
+between `0` and `26`. Using a key of `0` or `26` will always yield
```
petertseng: "Using a key of `0` or `26`" implies that the range is inclusive on both ends. This is ok, but careful when combined with a below line:

exercises/rotational-cipher/README.txt
```diff
@@ -0,0 +1,71 @@
...
+The most commonly used rotational cipher is `ROT13`.
+
+A `ROT13` on the Latin alphabet would be as follows:
+
+```plain
+Plain:  abcdefghijklmnopqrstuvwxyz
+Cipher: nopqrstuvwxyzabcdefghijklm
+```
+
+It is stronger than the Atbash cipher because it has 26 possible keys, and 24 usable keys.
```
petertseng: 0 through 26 inclusive on both ends has 27 elements, rather than 26... but I know you mean that there are 26 *unique* keys since 0 is congruent to 26. It would be good to get some clarity on this point though, in case someone else other than me objects "there are 27 elements in the range!"

There are 25 usable keys - the alignments of (every letter of the alphabet except "a") with "a", rather than 24.

exercises/rotational-cipher/README.txt
```diff
@@ -0,0 +1,71 @@
...
+- ROT5  `omg` gives `trl`
+- ROT0  `c` gives `c`
+- ROT26 `Cool` gives `Cool`
+- ROT13 `The quick brown fox jumps over the lazy dog.` gives `Gur dhvpx oebja sbk whzcf bire gur ynml qbt.`
+- ROT13 `Gur dhvpx oebja sbk whzcf bire gur ynml qbt.` gives `The quick brown fox jumps over the lazy dog.`
+
+## BONUS!
+non-ASCII characters are downgraded this means that an `é` would become an `e`.
+
+## Rust Installation
```
petertseng: you will not want to include these. They are provided by https://github.com/exercism/xrust/blob/master/exercises/TRACK_HINTS.md, and including in the README duplicates them.

exercises/rotational-cipher/README.txt
```diff
@@ -0,0 +1,71 @@
...
+The [exercism/xrust](https://github.com/exercism/xrust) repository on GitHub is the home for all of the Rust exercises. If you have feedback about an exercise, or want to help implement new exercises, head over there and create an issue. Members of the [rust track team](https://github.com/orgs/exercism/teams/rust) are happy to help!
+
+If you want to know more about Exercism, take a look at the [contribution guide](https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md).
+
+[help-page]: http://exercism.io/languages/rust
+[crates-and-modules]: http://doc.rust-lang.org/stable/book/crates-and-modules.html
+
+## Source
+
+Wikipedia [http://en.wikipedia.org/wiki/Atbash](http://en.wikipedia.org/wiki/Atbash)
```
petertseng: still atbash?

exercises/rotational-cipher/README.txt
```diff
@@ -0,0 +1,71 @@
...
+
+## Submitting Incomplete Problems
```
petertseng: you will not want to include this. This is included by https://github.com/exercism/trackler/blob/27da0508be0138338c6d583705c60e72231c9701/lib/trackler/implementation.rb#L99 and including it in the readme will cause it to be duplicated

exercises/rotational-cipher/tests/rotational-cipher.rs
```diff
@@ -0,0 +1,80 @@
...
+#[test]
+#[ignore]
+fn test_rotate_numbers() {
+    assert_eq!("Xiwxmrk 1 2 3 xiwxmrk", cipher::rotate("Testing 1 2 3 testing", 4));
+}
+
+#[test]
+#[ignore]
+fn test_rotate_punctuation() {
+    assert_eq!("Gzo\'n zvo, Bmviyhv!", cipher::rotate("Let\'s eat, Grandma!", 21));
```
petertseng: https://twitter.com/neilhimself/status/822312562809110529

exercises/rotational-cipher/README.txt
LogoiLab: yea i tried to reuse as much from atbash as possible i'll fix that

exercises/rotational-cipher/README.txt
LogoiLab: ill edit that out

exercises/rotational-cipher/README.txt
LogoiLab: ill edit that

exercises/rotational-cipher/README.txt
LogoiLab: yes, ill try to redefine that better.

exercises/rotational-cipher/README.txt
LogoiLab: the below line will be changed to reflect

exercises/rotational-cipher/Cargo.toml
LogoiLab: do you recommend no attribution to an exercise designer?

petertseng: Since there is prior precedent, I don't object, though I won't speak for anyone else who may feel more strongly.

I will note that you will probably want to put yourself as the source when adding the exercise to x-common, as there is a source and source_url in metadata - example at https://github.com/exercism/x-common/blob/master/exercises/paasio/metadata.yml#L3

Unless you think that should be a wikipedia link

LogoiLab:
just made all updates and submitted a pull request to x-common

exercises/rotational-cipher/example.rs
```diff
@@ -0,0 +1,16 @@
+pub fn rotate(text: & 'static str, shift_key: u8) -> String {
+
+    return text.chars().map(|c| {
```
petertseng: how about eliding the `return` and semicolon? I don't know if there is some Rust style guide that demands it, but at least https://github.com/Manishearth/rust-clippy/wiki#needless_return asks for it.

exercises/rotational-cipher/example.rs
```diff
@@ -0,0 +1,16 @@
+pub fn rotate(text: & 'static str, shift_key: u8) -> String {
```
petertseng: This might be OK since this is the example and won't be served to students, but I wouldn't want to restrict this function to only being able to take `&'static str`. I would expect it to be able to just take `&str`. So, if you remove `'static`, does the function still work? If not, why not?

exercises/rotational-cipher/example.rs
LogoiLab: Sorry that's my Java creeping in.

exercises/rotational-cipher/example.rs
LogoiLab: I'll look at it and play with it there was a reason I added static I think cause nightly was complaining about something.

exercises/rotational-cipher/example.rs
```diff
@@ -0,0 +1,16 @@
+pub fn rotate(text: str, shift_key: u8) -> String {
```
petertseng: we'll want to have `&str`, rather than `str`, so that it compiles.

petertseng: with `str`, it wouldn't compile: https://play.rust-lang.org/?gist=5ca113a1aed9629f2fc22114f3b55015&version=stable&backtrace=0

with `&str`, it would: https://play.rust-lang.org/?gist=3ac4315e213c2b756bdfba34ebea828c&version=stable&backtrace=0

LogoiLab:
OK I've gotten everything fixed up so far. Anything else?

petertseng:
Very good, thank you!

PR #258, opened 2017-02-03
replace nucleotide-codons with protein-translation

petertseng:
As noted in https://github.com/exercism/x-common/issues/268 these two
exercises are the same. In this transitionary period, we deprecate
nucleotide-codons (but keep its files available).

If we wish to replace nucleotide-codons with protein-translation
completely, then a database surgery will be needed to transfer all
submissions from nucleotide-codons to protein-translation.

### DIFF
--- a/exercises/protein-translation/tests/proteins.rs
+++ b/exercises/protein-translation/tests/proteins.rs
@@ -1,22 +1,22 @@
-extern crate nucleotide_codons as codons;
+extern crate protein_translation as proteins;
 
 #[test]
 fn test_methionine() {
-    let info = codons::parse(make_pairs());
+    let info = proteins::parse(make_pairs());
     assert_eq!(info.name_for("ATG"), Ok("methionine"));
 }
 
 #[test]
 #[ignore]
 fn test_cysteine_tgt() {
-    let info = codons::parse(make_pairs());
+    let info = proteins::parse(make_pairs());
     assert_eq!(info.name_for("TGT"), Ok("cysteine"));
 }
 
 #[test]
 #[ignore]
 fn test_cysteine_tgy() { // "compressed" name for TGT and TGC
-    let info = codons::parse(make_pairs());
+    let info = proteins::parse(make_pairs());
     assert_eq!(info.name_for("TGT"), info.name_for("TGY"));
     assert_eq!(info.name_for("TGC"), info.name_for("TGY"));
 }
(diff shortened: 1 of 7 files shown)


petertseng:
Because https://github.com/exercism/x-common/pull/532 is now merged, it's now official: The `nucleotide-codons` exercise is now globally deprecated. It behooves this track to follow suit and deprecate it.

exercises/protein-translation/tests/proteins.rs
```diff
@@ -0,0 +1,97 @@
...
+    assert_eq!(info.name_for("CGN"), Ok("arginine"));
+    assert_eq!(info.name_for("MGR"), Ok("arginine"));
+}
+
+#[test]
+#[ignore]
+fn test_invalid() {
+    let info = proteins::parse(make_pairs());
+    assert!(info.name_for("").is_err());
+    assert!(info.name_for("VWX").is_err()); // X is not a shorthand
```
IanWhitney: If we broke these out into separate tests could the test name describe what we're testing? It would remove the need for the comments.

exercises/protein-translation/tests/proteins.rs
```diff
@@ -0,0 +1,97 @@
...
+#[test]
+#[ignore]
+fn test_cysteine_tgt() {
+    let info = proteins::parse(make_pairs());
+    assert_eq!(info.name_for("TGT"), Ok("cysteine"));
+}
+
+#[test]
+#[ignore]
+fn test_cysteine_tgy() { // "compressed" name for TGT and TGC
```
IanWhitney: This comment references TGC but the test does not include it.

exercises/protein-translation/tests/proteins.rs
```diff
@@ -0,0 +1,97 @@
...
+    let info = proteins::parse(make_pairs());
+    assert_eq!(info.name_for("TGY"), Ok("cysteine"));
+}
+
+#[test]
+#[ignore]
+fn test_stop() {
+    let info = proteins::parse(make_pairs());
+    assert_eq!(info.name_for("TAA"), Ok("stop codon"));
```
IanWhitney: This is probably just me not understanding the example code, but I don't see anything in the example that could return `Ok("stop codon")`

petertseng: > nything in the example that could return `Ok("stop codon")`

`parse(make_pairs())`, and one k/v pair in `make_pairs` is:

```rust
      ("stop codon", vec!["TAA", "TAG", "TGA"])
```

Issue #200, opened 2016-09-11
Are stub `lib.rs` files a good idea?


IanWhitney:
Initially suggested in #117, we've been adding these stub `lib.rs` files to most of our recent exercises. I'm still not sold on them, and I wanted to see what other people think.

In the current [Maintaining a Language Track](https://github.com/exercism/exercism.io/blob/8841b3cdb1632fb56dd9ed0223a898fa4227d78e/docs/maintaining-a-track.md) document it is suggested that we:

> improve the test suites to avoid pushing people towards specific solutions

Which I fully agree with. And I think that stub files also push people towards specific solutions. 

Take, for example, Space Age. The test API is generic enough

```rust
let duration = Duration::from(1_000_000_000);
Earth::years_during(&duration)
```

There's more than one way to solve this.

But then the [stub file](https://github.com/exercism/xrust/blob/40924af5adf32a4c2df3e4da76160dc76ed2904b/exercises/space-age/src/lib.rs) tells students to use Traits. Which they do. I don't think I've seen a solution that removed the Trait.

In [Triangle](https://github.com/exercism/xrust/pull/197), a possible stub file would be:

```rust
pub struct Triangle;

impl Triangle {
    pub fn build(sides: [u16; 3]) -> Result<Self, ()> {
        unimplemented!()
    }

    pub fn is_equilateral(&self) -> bool {
        unimplemented!()
    }

    pub fn is_isosceles(&self) -> bool {
        unimplemented!()
    }

    pub fn is_scalene(&self) -> bool {
        unimplemented!()
    }
}
```

Which leaves the student with very little to do. The design of the code has already been done. Students are not likely to take this stub and write a solution that uses Enums, Traits or Boxes. Solutions using those approaches are interesting to write and discuss.

IanWhitney:
Summoning @kytrinyx to see if she has thoughts.

petertseng:
Looking for that exercise where you consciously decided not to add a stub file... ah, it was #178. It was not for the exact same reason as described in here (here: "it restricts the solution" and in #178 "it reduces learning since coming up with the right signature should be part of the learning process").

In React at #191 I chose to include the stub file since:

* there are many methods to implement
* you have to satisfy the compiler on all of the tests before you can get the first one passing!
* specifying the interface IMO does not restrict the implementation (notice that the `Cell` from the example solution is not present in the interface at all! Only `CellID`, and what the reactor does with a `CellID` when it is passed one for `value` or `set_value` is entirely its own business)

[...]

PR #266, opened 2017-03-09
Update hello-world GETTING_STARTED.md to match simplified exercise

### LINKED ISSUE #265: Hello World - mismatch
After https://github.com/exercism/xrust/commit/3275c180e172c9d9d286bfec22ed9984b1cd1afe
GETTING_STARTED.md isn't correct for current test.
@petertseng

jgallagher:
Closes #265.

### DIFF
--- a/exercises/hello-world/GETTING_STARTED.md
+++ b/exercises/hello-world/GETTING_STARTED.md
@@ -18,96 +18,62 @@ Run the test suite. It can be run with `cargo`, which is installed with rust.
 $ cargo test
 ```
 
-This will fail, complaining that `hello-world` could not compile.
-
-To fix this, create a new directory called `src`.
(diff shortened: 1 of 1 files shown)


jgallagher:
One thing I noticed looking at the next several exercises is that most of them don't include a `src/lib.rs` at all, but a few have one with all the necessary public functions defined (with `unimplemented!()` bodies). The latter seems a lot smoother to pick up and would reduce the need for extra process instructions compared to the former.

petertseng:
What I see from #117 and #200 is definite agreement there should be a `src/lib.rs`. There wasn't agreement on whether it should be empty or contain the function signatures.

I think we can get wider agreement on signatures for the initial exercises, to help people start. On later exercises, even if we can't agree, an empty src/lib.rs would be good.

PR #268, opened 2017-04-02
Include standard .gitignore file in all exercises

mkantor:
Most of the exercises already had one, but a few were missing.


IanWhitney:
Thanks!

Issue #269, opened 2017-04-02
Should there be a placeholder src/lib.rs for all exercises?


mkantor:
I noticed that several of the exercises have placeholder implementations with various levels of verbosity (e.g. [all-your-base](https://github.com/exercism/xrust/blob/master/exercises/all-your-base/src/lib.rs), [grains](https://github.com/exercism/xrust/blob/master/exercises/grains/src/lib.rs), [react](https://github.com/exercism/xrust/blob/master/exercises/react/src/lib.rs), [space-age](https://github.com/exercism/xrust/blob/master/exercises/space-age/src/lib.rs)). However this is absent from most of the exercises. I was unable to find documentation to suggest whether these files should or should not exist.

Anyway, it could probably be considered a bug one way or the other. I'm willing to help standardize things.

petertseng:
What I see in #117 and #200: I'm not sure what to make of discussions on what they should contain, but everyone seemed to agree that a lib/src.rs should at least exist, even if it is empty.

> documentation

Our track's README is anemic and does not contain most of the starter content in https://github.com/exercism/request-new-language-track/blob/master/README.md , as it seems it was based off of https://github.com/exercism/request-new-language-track/blob/53b185fdc67f0e5e7bc2b0891c2a1cb30e4b5f05/README.md instead. Even though not every single piece of starter content needs to be included, I recommend at least a review of each item in that starter README to evaluate whether it should be included.

mkantor:
@petertseng thanks, I neglected to search closed issues but I see there's a lot of good discussion already there. After reading everyone's arguments I'm on the fence about providing stub signatures, but agree that at the very least empty src/lib.rs files in every exercise would be nice. I'll send a PR to that effect in the near future.

IanWhitney:
Here's my thoughts.

When we decide if there should be stub files we should think about who our students are. 

If our student is someone new to Rust, then I think there is a lot of value in having them write the entire function themselves, including the signature.  Not every exercise will have a interesting signature to write, but enough of them do. And thinking about "I want to accept type `x` and return type `y`" is the core of writing Rust (or at least it is for me, your mileage may vary).

If our student is someone familiar with Rust but trying to write better-designed implementations, then having the method signature makes more sense. They already understand that part of Rust and want to work on 'cleaner' code.

Personally, I'm a Rust learner, so I favor the first approach.

jcdyer:
> The second approach is basically reading the signature

[...]

PR #270, opened 2017-04-03
Add src/lib.rs files to all exercises

mkantor:
See #117, #200, and #269.

Since the jury is sill out on whether/how to provide stub functions, I opted to simply add empty files in this commit. This makes for a smoother user experience, as students will no longer have to invoke `mkdir src; touch src/lib.rs` for every exercise.


petertseng:
(I corrected a typo sill -> still, but of course that didn't stop us)

Issue #228, opened 2016-11-24
Triangle: equilateral triangles are isosceles


clemensw:
I agree the coding is more interesting, but I have a hard time accepting
    assert!(triangle.is_equilateral());
    assert!(!triangle.is_isosceles());

petertseng:
Given that https://github.com/exercism/x-common/issues/379 has been discussed and https://github.com/exercism/x-common/pull/424 has been merged as a result, we should comply with the suite.

Re-visiting https://github.com/exercism/xrust/pull/206 would be a quick win in this area, though it may be worth it to rewrite the test suite follow the structure of the new JSON file.

IanWhitney:
Agreeing with @petertseng here. I think if you have an approach that you want to suggest, updating the official JSON file is the way to go.

PR #272, opened 2017-04-05
triangle: Allow equilateral triangles to also be isosceles

mkantor:
[The triangle problem description](https://github.com/exercism/x-common/blob/d7a3c09b6758a8a5d59a2e140e6684031316280a/exercises/triangle/description.md) says the following:

> An isosceles triangle has at least two sides the same length. (It is sometimes specified as having exactly two sides the same length, but for the purposes of this exercise we'll say at least two.)

However, previously the tests for this exercise only allowed isosceles triangles with exactly two sides of the same length.

Removing the expectation that equilateral triangles must return false for `.is_isosceles()` allows students the freedom to decide which definition of "isosceles" they want to implement.

### DIFF
--- a/exercises/triangle/tests/triangle.rs
+++ b/exercises/triangle/tests/triangle.rs
@@ -23,7 +23,6 @@ fn equilateral_triangles_have_equal_sides() {
     let sides = [2, 2, 2];
     let triangle = Triangle::build(sides).unwrap();
     assert!(triangle.is_equilateral());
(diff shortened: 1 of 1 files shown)


petertseng:
Closes #228 (well, it was already closed as of this writing, but the recommended action had not been taken) and supersedes #206

IanWhitney:
Thanks!

PR #273, opened 2017-04-06
triangle: Fix spelling of "isosceles" in tests

mkantor:
This word is hard to spell! See exercism/x-common#749 for another misspelling.


IanWhitney:
God I hate that word. I got it wrong all the time. Thanks for fixing it!

PR #274, opened 2017-04-16
PascalsTriangle: two more test cases!

niklasad1:
Hi,
Added two more test cases 
/Niklas A

### DIFF
--- a/exercises/pascals-triangle/tests/pascals-triangle.rs
+++ b/exercises/pascals-triangle/tests/pascals-triangle.rs
@@ -51,7 +51,7 @@ fn last_of_seven_rows() {
 
 #[test]
 #[ignore]
-fn last_of_twenty_rows() {
+fn last_of_twentyone_rows() {
     let pt = PascalsTriangle::new(21);
     let expected: Vec<u32> = vec![1,20,190,1140,4845,15504,38760,77520,125970,167960,184756,167960,125970,77520,38760,15504,4845,1140,190,20,1];
     assert_eq!(expected, pt.rows().pop().unwrap());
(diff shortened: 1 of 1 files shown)


IanWhitney:
Thanks! 

A couple of questions.

- What edge cases do these tests cover that the existing tests don't?
- Are these tests only useful for Rust, or should we add them to the [canonical test suite](https://github.com/exercism/x-common/blob/master/exercises/pascals-triangle/canonical-data.json) that all language tracks follow?

niklasad1:
Hmm, they are not testing any edge case what I'm aware but I thought it that people can get lucky or hardcode in the answer which is much harder with 21 rows ;) 

Use them if you want to but the existing tests may be sufficient.

petertseng:
I have no opinion on whether this track should have these tests, so I'm acknowledging that I'm being weak, evading the question, and passing the buck by deferring to x-common. If the same test cases are added to x-common, the Rust track will take them.

It looks like from https://github.com/exercism/x-common/pull/362 that some tracks test 20 but it was not made standard since once it works for four rows it probably works for subsequent rows.

Perhaps going up to five or more helps us because the fifth row is the first time a value appears that does not have 1 as one of its addends. That can be an argument for adding five (or some higher number) to x-common.

I will be happy to post the same opinion in x-common.

petertseng:
We look forward to seeing these cases in x-common at https://github.com/exercism/x-common/blob/master/exercises/pascals-triangle/canonical-data.json and would like to see it merged there before merging here.

petertseng:
Let's reopen this PR when it is merged in x-common!

PR #277, opened 2017-04-24
add run-length-encoding

zombiefungus:


exercises/run-length-encoding/Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[package]
+name = "run-length-encoding"
+version = "0.1.0"
+authors = ["zombiefungus <divmermarlav@gmail.com>"]
+
+[dependencies]
```
petertseng: just realised not all of our Cargo.toml have this line! Guess it's optional. The only exercise with a `[dependencies]` but no dependencies is acronym.

This line **is** included by `cargo new` though. So it makes sense to have.

exercises/run-length-encoding/example.rs
```diff
@@ -0,0 +1,45 @@
+use std::cmp;
+
+pub fn encode(input: &str) -> String {
+    input
+        .chars()
+        .fold((String::new(),' ',0,1), |(mut acc, last, last_n, pos), c| {
```
petertseng: spaces after the commas in the tuple? To be consistent with the fact that there are spaces after commas in other places

exercises/run-length-encoding/example.rs
```diff
@@ -0,0 +1,45 @@
...
+                (acc, c, 1, pos + 1)
+            }
+        }).0
+}
+
+pub fn decode(input: &str) -> String {
+    input
+        .chars()
+        .fold((String::new(), 0), |(mut acc, last_n), c| {
+            if c.is_digit(10) { // keep reading.. keeeep reading
```
petertseng: is this comment meant to tell the **reader** of the code that it's necessary to keep reading (I don't see the value of this) or that the **code** should keep reading? Can the comment be clearer about what the code will read, in that case?

exercises/run-length-encoding/example.rs
```diff
@@ -0,0 +1,45 @@
...
+                (acc, 10*last_n + c.to_digit(10).unwrap())
```
petertseng: Since there's an `is_digit` followed by a `to_digit.unwrap`, what about using [`if let`](https://doc.rust-lang.org/book/if-let.html) above? Less need to trust that the unwrap will succeed.

exercises/run-length-encoding/tests/run-length-encoding.rs
```diff
@@ -0,0 +1,87 @@
+extern crate run_length_encoding as cipher;
+
+// encoding tests
+
+#[test]
+// #[ignore]
```
petertseng: this one can probably be removed

exercises/run-length-encoding/tests/run-length-encoding.rs
```diff
@@ -0,0 +1,87 @@
...
+fn test_encode_empty_string() {
+    assert_eq!("", cipher::encode(""));
+}
+
+#[test]
+// #[ignore]
```
petertseng: whereas these should probably be uncommented, right?

I wanted to look for documentation where we say why we have tests ignored, but it is not present. Our documentation is not sufficient in this area. I will have to mention this in #279 .

exercises/run-length-encoding/tests/run-length-encoding.rs
```diff
@@ -0,0 +1,87 @@
...
+fn test_encode_lowercase_characters() {
+    assert_eq!("2a3b4c", cipher::encode("aabbbcccc"));
+}
+
+// decoding tests
+
+#[test]
+// #[ignore]
+fn test_decode_empty_string() {
+    assert_eq!("" , cipher::decode(""));
```
petertseng: there's a space before the comma here, but not in most other cases. Should it be removed?

exercises/run-length-encoding/tests/run-length-encoding.rs
```diff
@@ -0,0 +1,87 @@
...
+#[test]
+// #[ignore]
+fn test_decode_single_characters_only() {
+    assert_eq!("XYZ", cipher::decode("XYZ"));
+}
+
+#[test]
+// #[ignore]
+fn test_decode_string_with_no_single_characters() {
+    assert_eq!("AABBBCCCC" , cipher::decode("2A3B4C"));
```
petertseng: there's a space before the comma here, but not in most other cases. Should it be removed?

exercises/run-length-encoding/tests/run-length-encoding.rs
```diff
@@ -0,0 +1,87 @@
...
+}
+
+#[test]
+// #[ignore]
+fn test_decode_single_characters_with_repeated_characters() {
+    assert_eq!("WWWWWWWWWWWWBWWWWWWWWWWWWBBBWWWWWWWWWWWWWWWWWWWWWWWWB" ,
```
petertseng: there's a space before the comma here, but not in most other cases. Should it be removed?

exercises/run-length-encoding/tests/run-length-encoding.rs
```diff
@@ -0,0 +1,87 @@
+extern crate run_length_encoding as cipher;
```
petertseng: I'm unsure about it because I wouldn't consider run-length encoding as a means to obscure a message to prevent it from being read. That's what I think the definition of a cipher, but I apologise if I was mistaken since English isn't my first language. I'm not sure I have a better suggestion though - simply `encoding`, perhaps? Is shortening it to `rle` acceptable in Rust?

exercises/run-length-encoding/example.rs
zombiefungus: haha sorry, the idea is that you can't conclude that the number of digits to place is c because there may be other digits latter. This is just a personal comment, i'll remove it

exercises/run-length-encoding/example.rs
zombiefungus: well, I am learning rust my self and I haven't used it. Maybe this should be implemented by someone with more experience

exercises/run-length-encoding/tests/run-length-encoding.rs
zombiefungus: victim of copy paste

exercises/run-length-encoding/example.rs
petertseng: Then my challenge is to try it.

zombiefungus: done

zombiefungus:
nice, thanks

petertseng:
There were no objections in 48 hours (and a few more). It is time to merge. Thank you for your contribution!

PR #284, opened 2017-05-09
config: Ignore description.md and metadata.yml

petertseng:
They were created for #248, but since they're not in the ignore_pattern
they are being served.

See for proof:
http://x.exercism.io/v2/exercises/rust/luhn-from
http://x.exercism.io/v2/exercises/rust/luhn-trait

And note how description.md and metadata.yml appear inside the response.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,6 +1,7 @@
 {
   "slug": "rust",
   "language": "Rust",
+  "ignore_pattern": "example|description.md|metadata.yml",
   "repository": "https://github.com/exercism/xrust",
   "active": true,
   "exercises": [


config.json
```diff
@@ -1,6 +1,7 @@
 {
   "slug": "rust",
   "language": "Rust",
+  "ignore_pattern": "example|description.md|metadata.yml",
```
petertseng: fun fact: this will match a file name descriptionxmd, but we have no files of that name, so I don't care. it will go away if files move to .meta.

Insti: another fun fact: When you need to put a backslash in json string you need two, so the correct regex still looks odd.
`"ignore_pattern": "example|description\\.md|metadata\\.yml"`

Insti: another fun fact: the ignore_pattern is not anchored so you'll also match `the_description.md_file_contains_the_description`

PR #285, opened 2017-05-09
copy description.md and metadata.yml to .meta directory

petertseng:
If, as proposed in https://github.com/exercism/trackler/issues/45, the
proper location of these files moves to the .meta directory, this allows
Trackler to freely change its expectations without having to wait for
this track to move and/or support multiple locations at once.

If the proposal is accepted and implemented, the original files may be
removed. Of course, until then, the original files must remain, since
Trackler still expects them at the original location.


Insti:
I will submit a PR to remove the old files once the trackler changes are live.

kytrinyx:
The trackler changes have been merged. The changes will go live in the next deploy (today).

Issue #280, opened 2017-05-04
Test generation suite?


zombiefungus:
I really don't know how would this be done in rust, but I have seen that other tracks generate their tests from the canonical cases. What do you guys think about that?

petertseng:
It would make following x-common changes easier.

The people making the generators would want to think about two major questions.

* How to read the JSON files, whether that be serde, json-rust, or other.
* Once the files are read, how they turn into tests/*.rs. Templating engine, simple string interpolation, convert to using table-based tests, other.

Additional considerations when writing the generators: The more work each generator is to write, the less likely someone will want to write it.
To reduce the amount of work per generator, common pieces should be factored out. Of course, it may be hard to figure out what the common pieces are before there are at least two generators. That's why this one might not be clear until writing generators. Allow for it in the design if possible.

ijanos:
I don't think it is feasible to generate full tests because of the lack of types in the JSON files. 

A tool that creates some kind of skeleton test file could work or a tool that checks and compares the number of tests in xrust and xcommon. That would make easier to follow changes.

petertseng:
> A tool that creates some kind of skeleton test file could work or a tool that checks and compares the number of tests in xrust and xcommon. That would make easier to follow changes.

If we would include the x-common test version in a standard comment somewhere, we could compare the xrust vs x-common test versions too.

> lack of types in the JSON files.

What of the other languages that have test generators such as Ruby, Go, OCaml, Scala? Surely they dealt with the problem of not having types in the JSON files.

ijanos:
> What of the other languages that have test generators such as Ruby, Go, OCaml, Scala? Surely they dealt with the problem of not having types in the JSON files.

I did a quick check and I found that Scala and OCaml generators both provide a hand-made templates for each exercise. They basically create the types by hand and then the generator builds all the test cases for exercises with templates. This could work with Rust as well, but it is a lot of work and their generators seems language specific (I did not investigate this too much) so I don't think we can use theirs. 

> If we would include the x-common test version in a standard comment somewhere in xrust, we could compare the xrust vs x-common test versions too.

This is much less work than the other approach and it would help keeping up with the changes in x-common. Implementing this should not be hard, maybe we could create an issue and mark it "good first patch".

petertseng:
> If we would include the x-common test version in a standard comment somewhere in xrust, we could compare the xrust vs x-common test versions too.
> 
[...]

PR #288, opened 2017-05-13
tournament 1.3.0: Add simple start cases, remove invalid lines

petertseng:
https://github.com/exercism/x-common/pull/773

### DIFF
--- a/exercises/tournament/Cargo.lock
+++ b/exercises/tournament/Cargo.lock
@@ -1,4 +1,4 @@
 [root]
 name = "tournament"
-version = "0.0.0"
+version = "1.3.0"
 
--- a/exercises/tournament/Cargo.toml
+++ b/exercises/tournament/Cargo.toml
@@ -1,3 +1,3 @@
 [package]
 name = "tournament"
-version = "0.0.0"
+version = "1.3.0"
--- a/exercises/tournament/example.rs
+++ b/exercises/tournament/example.rs
@@ -65,12 +65,12 @@ fn write_tally(results: &HashMap<String, TeamResult>) -> String {
                   Equal => a.1.cmp(&(b.1)).reverse(),
                   other => other
               });
-    let header = format!("{:30} | MP |  W |  D |  L |  P\n", "Team");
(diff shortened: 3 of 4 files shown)


petertseng:
Note that this test file was created with a test generator, as suggested by #280. The generator follows:

```ruby
require 'json'

data = JSON.parse(File.read(File.join(__dir__, 'canonical-data.json')))

def quote(s)
  ?" + s + ?"
end

def fmt(s)
  return '""' if s.empty?
  return quote(s.first) if s.size == 1
  indent = ' ' * 8
  "#{quote(s.first + '\n')}.to_string() +\n" + s[1...-1].map { |l| "#{indent}#{quote(l + '\n')} +\n" }.join + indent + quote(s.last)
end

puts 'extern crate tournament;'

data['cases'].each_with_index { |c, i|
  puts
  puts '#[test]'
  puts '#[ignore]' if i != 0
  puts "fn #{c['description'].tr(' ', ?_).tr(',()', '').downcase}() {"
  puts "    let input = #{fmt(c['input'])};"
  puts "    let expected = #{fmt(c['expected'])};"
  puts
  puts '    assert_eq!(tournament::tally(&input), expected);'
  puts ?}
}
```

petertseng:
Unfortunately the generator seems unsuitable for submitting to this repository, as it was not written in Rust.

petertseng:
That of course does not affect the validity of this PR.

petertseng:
> I have no problem adding a test generator written in another language to the repo. Do we have some rule against this?

There is no rule; the only consideration is the concern that it will hinder contributions. If someone comes to xrust thinking the only necessary language to know is Rust, but suddenly, surprise, you also need to know a different language to be able to contribute to the generators, that's not ideal.

I'm not going to have any qualms about writing and posting quick scripts until a generator written in Rust comes along and I could maybe be convinced that adding them to the repo is a good idea, but either way they'll be availble until then.

petertseng:
> but suddenly, surprise, you also need to know a different language to be able to contribute to the generators

Of course, my comment conveniently ignored the fact that we have shell scripts in this repository too

PR #289, opened 2017-05-18
difference-of-squares 1.1.0: Add tests for 1

petertseng:
exercism/x-common#791

---

Supporting commit: Test one function at a time

Instead of testing all three functions for 5 and then all three
functions for 100, let's test one function at a time, so that the
student can make sure one function is solid before moving onto the next.

This also better matches the order the tests are listed in in x-common
(though we are free to change the order if we really want).

Honestly, I don't feel strongly about this, so this is just arbitrary.

### DIFF
--- a/exercises/difference-of-squares/Cargo.lock
+++ b/exercises/difference-of-squares/Cargo.lock
@@ -1,4 +1,4 @@
 [root]
 name = "difference-of-squares"
-version = "0.0.0"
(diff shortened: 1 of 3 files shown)


petertseng:
Reviewers, let me know if you feel strongly about testing one function at a time to decide whether to keep the commit or not.

PR #292, opened 2017-05-19
rna-transcription 1.0.0: reject invalid inputs

petertseng:
The invalid input cases originally came from
exercism/x-common#119

Today's JSON file can be found at
https://github.com/exercism/x-common/blob/master/exercises/rna-transcription/canonical-data.json

This track was already using all the valid inputs and simply needed to
add these invalid inputs.

### DIFF
--- a/exercises/rna-transcription/tests/rna-transcription.rs
+++ b/exercises/rna-transcription/tests/rna-transcription.rs
@@ -35,3 +35,21 @@ fn test_transcribes_thymine_to_adenine() {
 fn test_transcribes_all_dna_to_rna() {
     assert_eq!(Ok(dna::RibonucleicAcid::new("UGCACCAGAAUU")), dna::DeoxyribonucleicAcid::new("ACGTGGTCTTAA").to_rna())
 }
+
+#[test]
+#[ignore]
+fn handles_invalid_input() {
+    assert!(dna::DeoxyribonucleicAcid::new("U").to_rna().is_err());
+}
+
+#[test]
+#[ignore]
+fn handles_completely_invalid_input() {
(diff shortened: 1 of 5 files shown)


exercises/rna-transcription/tests/rna-transcription.rs
```diff
@@ -9,29 +9,47 @@ fn test_acid_equals_acid() {
 #[test]
 #[ignore]
 fn test_transcribes_cytosine_guanine() {
-    assert_eq!(dna::RibonucleicAcid::new("G"), dna::DeoxyribonucleicAcid::new("C").to_rna());
+    assert_eq!(Ok(dna::RibonucleicAcid::new("G")), dna::DeoxyribonucleicAcid::new("C").to_rna());
```
petertseng: Well, this is a different approach from, say, #238 which used `unwrap`. Note that https://github.com/exercism/xrust/blob/master/exercises/nucleotide-codons/tests/codons.rs does use `assert_eq` with `Ok` though. I think I like this better. The difference is what happens when I return an `Err` when I should return `Ok`. With `assert_eq!(Ok(...), ...)`:

```
---- test_transcribes_cytosine_guanine stdout ----
	thread 'test_transcribes_cytosine_guanine' panicked at 'assertion failed: `(left == right)` (left: `Ok(RibonucleicAcid { nucleotides: "G" })`, right: `Err(())`)', tests/rna-transcription.rs:12
```

With `assert_eq!(..., ....unwrap())`;

```
---- test_transcribes_cytosine_guanine stdout ----
	thread 'test_transcribes_cytosine_guanine' panicked at 'called `Result::unwrap()` on an `Err` value: ()', src/libcore/result.rs:859
```

So why don't I go ahead and say now that I like `assert_eq(Ok(...), ...)` better since it shows me both sides.

PR #294, opened 2017-05-19
forth: return `Vec<Value>`, not `String`, from `stack`

petertseng:
Questioning the assumptions: Why must the output be a string?

Reasons I thought of:

* It was this way in the original version of the exercise:
  https://github.com/exercism/exercism.io/pull/1188
    * But, we have no obligation to follow that.
* It matches the fact that we have strings as inputs.
    * But, it is probably more work for the student to create a string.
    * It can be argued that it's easier to check the vector.

Previous discussions:
https://github.com/exercism/x-common/pull/394#discussion_r81705078
https://github.com/exercism/xhaskell/pull/412

This doesn't get this track all the way to forth 1.0.0, but is a useful
step there.

### DIFF
--- a/exercises/forth/example.rs
+++ b/exercises/forth/example.rs
@@ -58,12 +58,8 @@ impl Forth {
         }
     }
 
-    pub fn format_stack(&self) -> String {
-        let mut s = self.stack.iter().fold(String::new(), |s, v| {
-            s + &v.to_string() + " "
-        });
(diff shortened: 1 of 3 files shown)


petertseng:
will need rebase for #296, will do that later.

petertseng:
I did that badly. Better now.

PR #296, opened 2017-05-28
assert that eval() returns ok where it should

gsauthof:
This detects failed attempts where the stack looks as expected but
eval still returns the wrong result.

Also, this should point out some implementation errors more clearly
because you see a specific error result and not just the mismatching
stack content. Especially when there are multiple eval() calls in a row.

### DIFF
--- a/exercises/forth/tests/forth.rs
+++ b/exercises/forth/tests/forth.rs
@@ -1,5 +1,3 @@
-#![allow(unused_must_use)]
-
 extern crate forth;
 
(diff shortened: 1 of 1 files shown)


petertseng:
I think I'm for this. I would like to see the ` #![allow(unused_must_use)]` on line1 removed too if this is going to happen, please

gsauthof:
Good point. The first line with that warning suppressing attribute is gone now.

(And, as expected, because of the new asserts no warnings are emitted.)

PR #299, opened 2017-06-09
Rename the global exercise hints file

kytrinyx:
We've had some difficulty coming up with a good name for the file
that gets included in all of the exercise READMEs for a given track.

These are global hints, like how to run the test suite, which are
relevant to all the exercises on a track.

We started with SETUP.md in the root of the repository, then renamed
that to exercises/TRACK_HINTS.md because SETUP.md was misleading and
confusing, but then we realized that TRACK_HINTS.md was a bit ambiguous
and confusing as well.

Finally we settled on putting the file in the docs directory, since
this is user-facing documentation, and calling the file
EXERCISE_README_INSERT.md

See https://github.com/exercism/meta/issues/5 for context.


petertseng:
I anticipate that the Rust Beta build will fail, but it should be fixed separately and this PR can still be merged.

petertseng:
Useful information for all-repos PRs, see how the org's job limit takes it:

PR submitted to xbash 16:48

xrust Travis CI start running about 19:23

petertseng:
> Travis CI start running about 19:23

start time is probably not accurate, but I care about the finish times: finish at about 19:25 and 19:34

Issue #267, opened 2017-03-15
Tests for protein-translation don't match its README


jgallagher:
The [README](http://exercism.io/exercises/rust/protein-translation/readme) for this exercise describe the problem as (roughly) "given a sequence of codons, output the sequence of proteins". However, the tests in `protein-translation` (a) only seem to expect a single protein name from a single codon and (b) talk about codon "compression" which is not mentioned in the README; e.g.,

```rust
fn test_cysteine_tgy() { // "compressed" name for TGT and TGC
    let info = proteins::parse(make_pairs());
    assert_eq!(info.name_for("TGT"), info.name_for("TGY"));
    assert_eq!(info.name_for("TGC"), info.name_for("TGY"));
}

...

fn test_arginine_name() {
    // In arginine CGA can be "compresed" both as CGN and as MGR
    let info = proteins::parse(make_pairs());
    assert_eq!(info.name_for("CGA"), Ok("arginine"));
    assert_eq!(info.name_for("CGN"), Ok("arginine"));
    assert_eq!(info.name_for("MGR"), Ok("arginine"));
}
```

I glanced at a couple other languages' versions of this exercise and saw tests that look much more like the README and don't talk about compressed names ([go](https://github.com/exercism/xgo/blob/master/exercises/protein-translation/protein_translation_test.go), [elixir](https://github.com/exercism/xelixir/blob/master/exercises/protein-translation/protein_translation_test.exs)).

petertseng:
The questions arising from this:

* So does that mean https://github.com/exercism/x-common/issues/268 was incorrect? the nucleotide-codons and protein-translation exercises are not actually the same?
* Should #258 be reverted? Should the nucleotide-codons exercise be un-deprecated in the Rust track, if it is in fact different from protein-translation?
* Should https://github.com/exercism/x-common/pull/532 be reverted? Should the nucleotide-codons exercise be un-deprecated globally, if it is in fact different from protein-translation?
* Should the tests of protein-translation of the Rust track be changed to match the README?
* How can we to decide what the tests should be? We need someone to work on https://github.com/exercism/x-common/issues/577
* Should the [protein-translation readme](https://github.com/exercism/x-common/blob/master/exercises/protein-translation/description.md) be changed to include anything about compression?

jgallagher:
It appears to me that exercism/x-common#268 is correct at a high level:

> both these exercises do the exact same thing: they translate codons to their correct names.

but the two exercises do so quite differently:

[...]

PR #304, opened 2017-06-15
Remove tests for compression & added tests for of_rna

go717franciswang:
This resolves https://github.com/exercism/xrust/issues/267
* removed test cases that tested for compression because README did not mention about it
* added test cases for function `of_rna`, which are copied from [xpython](https://github.com/exercism/xpython/blob/master/exercises/protein-translation/protein_translation_test.py)

### DIFF
--- a/exercises/protein-translation/example.rs
+++ b/exercises/protein-translation/example.rs
@@ -12,29 +12,18 @@ pub fn parse<'a>(pairs: Vec<(&'a str, &'a str)>) -> CodonInfo<'a> {
 
 impl<'a> CodonInfo<'a> {
     pub fn name_for(&self, codon: &str) -> Result<&'a str, &'static str> {
-        if codon.len() != 3 {
-            return Err("invalid length")
+        match self.actual_codons.get(&codon) {
+            Some(name) => Ok(name),
+            None => Err("Invalid")
         }
+    }
 
-        let mut valid = true;
-        let lookup: String = codon.chars().map(|l| {
-            // Get an example of a "letter" represented by the possibly encoded letter.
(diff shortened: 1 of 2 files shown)


petertseng:
Hey there, thanks for the contribution.

I hope you can excuse me as my allowed time to use the computer running out soon and I could not finish reviewing it, but I am able to ask a few questions:

* After this PR is merged, would you consider https://github.com/exercism/xrust/issues/267 to be complete? As in, everything raised in that issue is addressed?
    * If the answer is yes, could you add one of the keywords listed in https://help.github.com/articles/closing-issues-via-commit-messages/ to your commit message?
    * If the answer is no, could you explain here and/or in #267 what is left to do after this PR is merged?
* I see an & in this commit message. When I see this, it leads me to wonder if this commit should be split up in to two commits. One commit would be "Remove test cases for compression" and the second commit would be "add test cases for `of_rna`".
    * This would make the PR easier to review. If the two changes are combined into one commit, the danger is that we will look at a change and must spend time thinking whether it is "Remove test cases for compression" or "add test cases for `of_rna`" to understand whether the change makes sense. If they are two separate commits, this becomes easy to understand.
    * If it does make sense to keep the two changes as one commit (it is impossible to have one change without the other), all you have to do is say so.

go717franciswang:
Hi Peter,
I can split the commit into 2. Let me create another pull request, and try again.
It will resolve https://github.com/exercism/xrust/issues/267 by updating the tests to match description in README.

petertseng:
Nice. You could feel free to reuse the same PR by force pushing the branch and no need to create a new one. It is an option you can take, but you are not required to do it. No need to change anything now that we have chosen a new PR this time.

PR #306, opened 2017-06-17
Add hints to gigasecond lib.rs

leodasvacas:
The signature of the function is not at all obvious.

EDIT: It seems this is being discussed in #269, to be it's obvious we want the signatures, the exercises  baffling to a begginer otherwise. I want to use exercism to teach rust at a workshop, I'll gladly add a bunch of stubs if the PRs will be accepted.

### DIFF
--- a/exercises/gigasecond/src/lib.rs
+++ b/exercises/gigasecond/src/lib.rs
@@ -1,0 +1,6 @@
+extern crate chrono;
+use chrono::*;
+
+pub fn after(start: DateTime<UTC>) -> DateTime<UTC> {
+    // Return a UTC DateTime one billion seconds after start.
+}
--- a/exercises/gigasecond/src/lib.rs
+++ b/exercises/gigasecond/src/lib.rs
@@ -1,6 +1,7 @@
 extern crate chrono;
 use chrono::*;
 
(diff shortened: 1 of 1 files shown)


IanWhitney:
I lean to having the comment above, but it's not a blocker.

ijanos:
I stand corrected :)

petertseng:
One might also think that the comment, being written as an imperative (rather than third person), was a code comment saying what the code will do, rather than a documentation comment.

I'll move it up to match our existing ones and add unimplemented I suppose, then merge. Wouldn't want to spend too much time thinking about which of the two to use, otherwise the value of the time spent exceeds the value gained from either alternative.

petertseng:
All right, very good, thank you.

> being discussed in #269, to be it's obvious we want the signatures, the exercises baffling to a begginer otherwise

reporting experiences of bafflement there will be helpful! whether they be personal experience or working with someone and observing that person's experience

> I'll gladly add a bunch of stubs if the PRs will be accepted.

I personally would. It would be useful to check at #269 to make sure people are on the same page. If it is true that figuring out how to infer the signature from reading the tests or error messages is a skill to learn (an argument against having stubs), perhaps just the first few exercise?

PR #307, opened 2017-06-18
Remove leading 'x' from repository name

kytrinyx:
The leading 'x' is kind of arbitrary. Especially now that we can set
topics on the repositories, we don't need a pattern to distinguish what
is a track or not.

The repository itself has already been renamed. GitHub redirects from
the old name to the new name, so we do not have to rush to fix links to
the old repository name, though we should update them for the sake of
clarity.

See exercism/meta#1

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,7 +1,7 @@
 {
   "slug": "rust",
   "language": "Rust",
-  "repository": "https://github.com/exercism/xrust",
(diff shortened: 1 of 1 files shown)


IanWhitney:
I especially approve because the `x` key on my new MacBook Pro sucks.

kytrinyx:
LOL. That's an excellent reason to approve.

PR #308, opened 2017-06-18
docs: remove leading 'x' from repository name

petertseng:
The leading 'x' is kind of arbitrary. Especially now that we can set
topics on the repositories, we don't need a pattern to distinguish what
is a track or not.

The repository itself has already been renamed.

See https://github.com/exercism/meta/issues/1

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
-# xRust
+# Exercism Rust Track
 
-[![Build Status](https://travis-ci.org/exercism/xrust.svg?branch=master)](https://travis-ci.org/exercism/xrust)
-[![Join the chat at https://gitter.im/exercism/xrust](https://badges.gitter.im/exercism/xrust.svg)](https://gitter.im/exercism/xrust?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
+[![Build Status](https://travis-ci.org/exercism/rust.svg?branch=master)](https://travis-ci.org/exercism/rust)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,7 +1,7 @@
 # xRust
 
-[![Build Status](https://travis-ci.org/exercism/xrust.svg?branch=master)](https://travis-ci.org/exercism/xrust)
-[![Join the chat at https://gitter.im/exercism/xrust](https://badges.gitter.im/exercism/xrust.svg)](https://gitter.im/exercism/xrust?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
+[![Build Status](https://travis-ci.org/exercism/rust.svg?branch=master)](https://travis-ci.org/exercism/rust)
+[![Join the chat at https://gitter.im/exercism/rust](https://badges.gitter.im/exercism/rust.svg)](https://gitter.im/exercism/rust?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
```
petertseng: so... we should rename the gitter channel, eh?

kytrinyx: Yeah, I think that would make sense.

petertseng: Created rust. xrust can be deleted whenever, I suppose

petertseng:
The EXERCISE_README_INSERT change moved into https://github.com/exercism/rust/commit/5efdea27d40ff632b5767c1c84941ea74a0e6e62 of #314 because nobody reviewed this.

PR #309, opened 2017-06-20
config: rectangles topics: replace lifetimes with algorithm

petertseng:
This has been a long time coming.
Ever since https://github.com/exercism/rust/pull/167#r70376973 lifetimes
should not have been a topic for rectangles.
The reasons were discussed there already:

* The "Putting it All Together" problems do not emphasise Rust language
  features in general.
* The specific feature "lifetimes" is not implied by this problem. A
  sampling of ten submitted solutions shows two solutions with
  lifetimes.

I used the https://github.com/exercism/x-common/issues/834 command to
inspect other tracks' topics to see if we can glean any insight here.

C#/F#: parsing, transforming
Lua: strings, performance, control-flow (if/else, loops), algorithms

Based on these, my recommendation:

A possible topic is string parsing, because we are reading in a string
and trying to make sense of it. However this is not parsing in the usual
sense that one would use a parser for, which usually deals with text.

If we were to declare "string parsing" a topic of rectangles, we would
probably use it for ocr-numbers too (but we do not; the topics for
ocr-numbers are lines, slices, chunks).

"algorithm" is chosen because significant work will likely go into
devising an algorithm for determining the rectangles, since that is not
immediately obvious from the problem description.  Solutions will likely
display a diverse set of approaches.

One might surmise that the alphametics exercise could have this topic as
well, but this commit deals with rectangles only.

### DIFF
--- a/config.json
+++ b/config.json
@@ -470,7 +470,7 @@
         "Enum",
         "structs",
         "traits",
-        "Lifetimes"
+        "algorithm"
       ]
     },
     {


petertseng:
It is a difficult one for me, which is why I had not come up with one before this point either. Diverse approaches means it's hard to come up with something, I spupose?

I suppose one can have combinatorics, if one takes the approach of checking every pair of corners, but not all take this approach.

I'll think more about it before merging. Topics are not used, so no rush.

IanWhitney:
Is there a record for most discussion over a one-line change?

petertseng:
failed to find a better topic. further thought about it would cause the time spent to exceed the value of the change (the value of the change is almost zero because topics are not used). so it has to be merged.

PR #311, opened 2017-06-25
Add more tests to the `bob` exercise

vvv:
Tests are copied from the Haskell language track:
https://github.com/exercism/haskell/blob/master/exercises/bob/test/Tests.hs

### DIFF
--- a/exercises/bob/example.rs
+++ b/exercises/bob/example.rs
@@ -6,13 +6,14 @@ pub fn reply(message: &str) -> &str {
 }
 
 fn is_silence(message: &str) -> bool {
-    message == ""
+    message.trim().is_empty()
 }
 
 fn is_yelling(message: &str) -> bool {
-    message.chars().all(|char| !char.is_lowercase())
+    let s = message.trim_matches(|c: char| !c.is_alphabetic());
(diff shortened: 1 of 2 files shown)


IanWhitney:
Looks like these tests are also in the the official suite

https://github.com/exercism/problem-specifications/blob/3833e0554f890fa40c774e1bfd79023542638d4f/exercises/bob/canonical-data.json

vvv:
@IanWhitney: Nice! I was unaware of “problem-specifications” repository.

Should I rewrite the tests (bob/tests/bob.rs)?

IanWhitney:
We do like it if our test suite matches the official suite, so if you want to tackle that then I say 👍

petertseng:
If doing that, then https://github.com/exercism/rust/blob/master/exercises/bob/Cargo.toml#L3 may be updated to match the problem-specifications version. That version is used to track what version of problem-specifications we are compliant with. It's currently 0.0.0 so we are not in compliance with any problem-specifications version.

vvv:
@petertseng Got it, thank you.

vvv:
Abandoned in favour of #312, which is rebased.

PR #312, opened 2017-06-28
bob: Update tests to conform with the problem spec

vvv:
* Make unit tests comply with the [problem specification](https://github.com/exercism/problem-specifications/blob/master/exercises/bob/canonical-data.json) v1.0.0.
* Update Cargo.toml version correspondingly.
* Update example.rs.
* Delete Cargo.lock.

### DIFF
--- a/exercises/bob/Cargo.lock
+++ b/exercises/bob/Cargo.lock
@@ -1,4 +1,0 @@
-[root]
-name = "bob"
-version = "0.0.0"
-
--- a/exercises/bob/example.rs
+++ b/exercises/bob/example.rs
@@ -6,13 +6,14 @@ pub fn reply(message: &str) -> &str {
 }
 
 fn is_silence(message: &str) -> bool {
-    message == ""
+    message.trim().is_empty()
 }
 
 fn is_yelling(message: &str) -> bool {
-    message.chars().all(|char| !char.is_lowercase())
+    let s = message.trim_matches(|c: char| !c.is_alphabetic());
+    !s.is_empty() && s.to_uppercase() == s
 }
 
 fn is_question(message: &str) -> bool {
-    message.ends_with("?")
+    message.trim_right().ends_with("?")
 }
--- a/exercises/bob/Cargo.toml
+++ b/exercises/bob/Cargo.toml
@@ -1,3 +1,3 @@
 [package]
 name = "bob"
-version = "0.0.0"
+version = "1.0.0"
--- a/exercises/bob/tests/bob.rs
+++ b/exercises/bob/tests/bob.rs
@@ -1,7 +1,7 @@
 extern crate bob;
 
 #[test]
-fn test_statement() {
+fn test_stating_something() {
     assert_eq!("Whatever.", bob::reply("Tom-ay-to, tom-aaaah-to."));
 }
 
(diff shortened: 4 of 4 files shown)


exercises/bob/Cargo.lock
```diff
@@ -1,4 +0,0 @@
-[root]
```
petertseng: should all Cargo.lock be deleted?

exercises/bob/example.rs
```diff
@@ -6,13 +6,14 @@ pub fn reply(message: &str) -> &str {
 }
 
 fn is_silence(message: &str) -> bool {
-    message == ""
+    message.trim().is_empty()
 }
 
 fn is_yelling(message: &str) -> bool {
-    message.chars().all(|char| !char.is_lowercase())
```
petertseng: don't really care since the PR as written passes the tests, but that means it was possible to keep `!is_lowercase`:


```rust
message.chars().any(char::is_alphabetic) && message.chars().all(|c| !c.is_lowercase())
```

Or its equivalent

```rust
message.chars().any(char::is_alphabetic) && !message.chars().any(char::is_lowercase)
```

exercises/bob/Cargo.lock
petertseng: Or maybe just all Cargo.lock for exercises that do not declare any dependencies (all exercises except gigasecond and robot-name)???

vvv:
> why was the original using !char.is_lowercase() instead of char.is_uppercase()

@ijanos They are not the same. `!char.is_lowercase()` returns true for uppercase chars, digits, punctuation... anything but lowercase chars. `char.is_uppercase()` returns true for uppercase chars only.

ijanos:
@vvv Yeah, I know, I should have been more verbose, so my thought was to filter characters first and then use is_uppercase to have a more readable line. Like this: 

```rust
message.chars().filter(char::is_alphabetic).all(char::is_uppercase)
```

exercises/bob/Cargo.lock
ijanos: I think we should delete all Cargo.lock files, I don't see any advantage keeping them around.

vvv:
@ijanos The code you proposed
```rust
message.chars().filter(char::is_alphabetic).all(char::is_uppercase)
```
does not compile:
```
error[E0281]: type mismatch: the type `fn(char) -> bool {std::char::<impl char>::is_alphabetic}` implements the trait `std::ops::FnOnce<(char,)>`, but the trait `for<'r> std::ops::FnOnce<(&'r char,)>` is required (expected &char, found char)
  --> src/lib.rs:13:21
   |
13 |     message.chars().filter(char::is_alphabetic).all(char::is_uppercase)
   |                     ^^^^^^
```

```rust
message.chars().filter(|c| c.is_alphabetic()).all(char::is_uppercase)
```
does compile but fails `test_non_letters_with_question`, `test_only_numbers`, and `test_question_with_only_numbers`:
```
---- test_non_letters_with_question stdout ----                                 
        thread 'test_non_letters_with_question' panicked at 'assertion failed: `(left == right)` (left: `"Sure."`, right: `"Whoa, chill out!"`)', tests/bob.rs:84                   

---- test_only_numbers stdout ----      
        thread 'test_only_numbers' panicked at 'assertion failed: `(left == right)` (left: `"Whatever."`, right: `"Whoa, chill out!"`)', tests/bob.rs:58        

---- test_question_with_only_numbers stdout ----                                
        thread 'test_question_with_only_numbers' panicked at 'assertion failed: `(left == right)` (left: `"Sure."`, right: `"Whoa, chill out!"`)', tests/bob.rs:63                  
```

The problem with the proposed code is that
```rust
assert!("".chars().all(char::is_uppercase));
```
passes.

ijanos:
You caught me typing up rust without checking it first :) You are right, changing that line will break the example.

vvv:
@ijanos This wasn't my intention. :) I liked your solution better and wanted to use it. Could not.
Oh well.

ijanos:
Oh no problem :) the example.rs is not an important part of an exercise. For a complete solution that uses is_uppercase check my first solution. It is more than a year old, it may fail on some new test cases though. 

http://exercism.io/submissions/857e8780b9cc40caa717d5ccd2992be8

PR #315, opened 2017-07-14
Add configuration defaults for nextercism

kytrinyx:
We will be moving to a tree-shaped track rather than a linear one, as described in the [progression & learning in Exercism](https://github.com/exercism/docs/blob/master/about/conception/progression.md) design document.

In order to support this, we need to expand the metadata that exercises are configured with.

Note that 'core' exercises are never unlocked by any other exercises. Core exercises appear in the track in the order that they are listed in the array.

Non-core exercises depend on only one exercise (unlocked_by: ). At the moment we are assuming that this is a core exercise, but there is no reason that it needs to be.

active: false is the equivalent of _deprecated_, which was stored in a separate array. This array is being removed in this change.

With these defaults the track in nextercism will have no core exercises, and all the exercises will be available as 'extras' from the start.

See https://github.com/exercism/meta/issues/16

### DIFF
--- a/config.json
+++ b/config.json
@@ -5,14 +5,20 @@
   "active": true,
   "exercises": [
     {
+      "uuid": "13ec1ebe-d71b-436f-ab12-25305e814171",
       "slug": "hello-world",
+      "core": false,
+      "unlocked_by": null,
       "difficulty": 1,
       "topics": [
         "println!"
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


petertseng:
At https://github.com/exercism/discussions/issues/159#issuecomment-315466131 I am discussing whether certain keys (definitely `difficulty` and `topics`, and perhaps `core` and `depends_on`) can be removed for the inactive exercises.

PR #317, opened 2017-07-16
Update chrono to 0.4, for UTC -> Utc change.

sacherjj:
When I tried to figure out issues on Playground, I noticed UTC -> Utc.  This looks like a change between 0.3 and 0.4 chrono.   I've updated to use 0.4 and updated the tests and lib.rs.  

Edit: And example.rs after second commit.


petertseng:
the commits will be squashed since they do not build separately.

petertseng:
Thanks.

PR #318, opened 2017-07-17
Adding pig-latin exercise

sacherjj:
This is adding an implementation of the pig-latin exercise.

I added an entry in json, but I do not know if uuid is supposed to tie to anything.  I generated a fresh one. 

I did not see any regex topics.  Also used lazy_static, which I thought was interesting for people to learn.  

I can't test full setup on windows, and will look into implementing a version of check-exercises.sh that works for windows.  I'm currently using Travis-CI on my fork as testing, but this causes more commits.

I'm not sure why my commits from the previous pull request are still in this one (Chrono -> 0.4).  Pull requests in Git are still new to me.  So any pointers are appreciated.

### DIFF
--- a/exercises/gigasecond/Cargo.lock
+++ b/exercises/gigasecond/Cargo.lock
@@ -2,12 +2,12 @@
 name = "gigasecond"
 version = "1.0.0"
 dependencies = [
- "chrono 0.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "chrono 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
 name = "chrono"
-version = "0.3.1"
+version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "num 0.1.39 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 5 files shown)


petertseng:
Thanks for another contribution. While I can't review the code yet, it looks like there are enough questions that it is still valuable for me to write a comment.

> I added an entry in json, but I do not know if uuid is supposed to tie to anything. I generated a fresh one.

generating a fresh one is correct.

> I did not see any regex topics. Also used lazy_static, which I thought was interesting for people to learn.

Be warned that there is no guarantee that students will use `lazy_static`. I do not know if that means anything in this PR needs to change, just be aware.

> I can't test full setup on windows, and will look into implementing a version of check-exercises.sh that works for windows.

Thanks.

> I'm currently using Travis-CI on my fork as testing, but this causes more commits.

Makes sense. The commits will likely be squashed when merging this PR, unless there is some specific reason why it makes sense to keep separate commits.

> I'm not sure why my commits from the previous pull request are still in this one (Chrono -> 0.4). Pull requests in Git are still new to me. So any pointers are appreciated.

At the top of this PR it says "sacherjj  wants to merge 7 commits into exercism:master from sacherjj:master"

The PR thus lists every commit that is on sacherjj:master but is not on exercism:master. Since you have commits https://github.com/sacherjj/rust/commit/a76558a58b6fca8f7a7fa129954d4ecec6ae9794 and https://github.com/sacherjj/rust/commit/83c1b86761e087b7b0eff54cfe2a0e4e2dc51398 on `sacherjj:master` but they were squashed and merged as https://github.com/exercism/rust/commit/aeead677a7ff715d02537ffb07a106a490086e35 , they are are not on `exercism:master` and thus they show up in the PR.

I would suggest to rebase on `exercism:master` to remove those two commits, and the "Merge remote-tracking branch 'upstream/master'" commit.

sacherjj:
Did not think about lazy_static being not probable for user, but regex being so.   I will clean that up and rebase to see if I can make this cleaner before merge.

PR #321, opened 2017-07-20
Adding say exercise.

sacherjj:
I tried to get a balance between following the instructions of the exercise and being more rustic.

I see the encode_u64 as all that would be needed in Rust.  But also wrapped it to more closely follow the exercise.   Trying to get the solvers to not just get to a solution, but think how the strong typing could eliminate work and increase safety if used properly.


exercises/say/example.rs
```diff
@@ -0,0 +1,66 @@
...
+    } else if n < 100 {
+        let small = n % 10;
+        let mut out = String::from(TENS[n as usize / 10]);
+        if small > 0 {
+            out.push('-');
+            out.push_str(SMALL[small as usize]);
+        }
+        out
+    } else if n < 1000 {
+		let mut out = String::from(SMALL[n as usize / 100]);
```
petertseng: indentation seems unusual on these lines, as if spaces and tabs were mixed. can you check?

exercises/say/tests/say.rs
```diff
@@ -0,0 +1,153 @@
...
+//  Testing bounds given in problem, which requires a signed data type.
+//
+//  How could data type selection use Rust compiler to eliminate
+//  issues with receiving negative number at compile time, 
+//  rather than returning error?
+*/
+#[test]
+#[ignore]
+fn test_negative() {
+    assert!(if let Err(_) = say::encode(-1) { true } else { false });
```
petertseng: I suggest `is_err()` here.

exercises/say/tests/say.rs
```diff
@@ -0,0 +1,153 @@
...
+}
+
+#[test]
+#[ignore]
+fn test_too_big() {
+    assert!(if let Err(_) = say::encode(1_000_000_000_000) { true } else { false });
```
petertseng: I suggest `is_err()` here.

exercises/say/tests/say.rs
petertseng: I would say the function should just take in `u64` in the first place, if only the README didn't ask for `-1` as a test case. Maybe delete it as a suggested case for the README, but that means we have to have a custom README, that's a pain.

sacherjj: Sounds good.  I struggled with needs of exercise vs needs of language.  However, the goal is to teach the language.  And I agree that this means using it for the best conditions.

exercises/say/example.rs
sacherjj: You are correct.  Fixing.

sacherjj:
Since we need a custom README.md, I moved much of what was in the test tests/say.rs as comments into the README.md.   

By implementing with u64, we don't need any Result() return, just String as all cases can be covered.

(By the way, figured out what was going wrong with my git --amend.  This is my second amend for this commit.  Thanks for the mentoring.)

sacherjj:
I also am looking at [error-handling](https://github.com/exercism/problem-specifications/tree/master/exercises/error-handling), which also requires a custom text.  It seems like this would be just exercises around Result and Option.   Also looking at flatten-array (which might not be possible, unless it is flatten-tuple.  

Having RUST_SPECIFIC.md in directories if it makes sense.  A link is then included at the top of the rust section (or it is included in, not sure how the build system works.)   Rust has enough really new and different features, that there will often be these edge cases to describe.

Another alternative is a big block comment at the top of src/lib.rs.   I've seen this on one or two exercises.

sacherjj:
Made modifications and amended the commit.

petertseng:
oh and

* tests look good (well, that is covered in "we do comply with 1.0.0" kind of)

petertseng:
Thanks for another exercise!

PR #324, opened 2017-07-22
Adding empty src/lib.rs for pig-latin

sacherjj:
When I tested pulling down pig-latin using exercism tool, there was no src/lib.rs.  I assume this was uploaded with new code that now requires it to exist.


petertseng:
Yup, thanks.

PR #326, opened 2017-07-23
check-exercises: Require presence of src/lib.rs

petertseng:
As done in https://github.com/exercism/rust/pull/270

### DIFF
--- a/_test/check-exercises.sh
+++ b/_test/check-exercises.sh
@@ -31,7 +31,13 @@ for exercise in $files; do
     # Run in subshell to change workdir without affecting current workdir.
     (
         cd $workdir
-        [ -d src ] || mkdir src
+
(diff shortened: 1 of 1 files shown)


petertseng:
Please excuse me for a moment. Going to make sure that when a src/lib.rs is deleted that the build fails. Please do not merge while the testing is in progress.

petertseng:
It has failed as it should. I will leave that commit **in the PR** as evidence and to make an example failure more visible in the historical records. They will be squashed when merged.

petertseng:
There are other files that should be checked. Such as whether there is a tests file.

But I suggest we leave that for configlet to check in https://github.com/exercism/configlet/issues/8 since all tracks have tests files (even if their names are different)
We only need to specially check files that are only present in the Rust track.

PR #327, opened 2017-07-23
error-handling: Add exercise

sacherjj:
Option and Result were a little weird for me learning, so I think forcing conversions between them like this should help new programmers get the ideas.  Especially if someone is new to monads, like I was.

ErrorKind is new to me.  I'd like to know if there are other types of errors or conditions I should handle.  Having it here as mainly an "it exists" thing for the user.  

I struggled with if there should be a shell for one or more of the functions in src/lib.rs.  I could see people maybe not getting the generics <> syntax, etc. early in the track.

### DIFF
--- a/exercises/error-handling/.meta/hints.md
+++ b/exercises/error-handling/.meta/hints.md
@@ -1,0 +1,11 @@
+## Rust Specific Track Information
+
+In Rust, it is expected (and enforced by the compiler) that all possible outcomes are handled.  There is no throw and catch error system that is present in other languages.
+
+There is panic!(), when the program gets into an unknown and dangerous state.  In this situation, we want the program to shutdown down cleanly.  This is not a state you would catch and continue, it is more of a fatal and crashing condition.
+
+This exercise will deal with [Result](https://doc.rust-lang.org/std/result/enum.Result.html), which is the way Rust allows of dealing with an error condition.
+
+It will also work with [Option](https://doc.rust-lang.org/std/option/), which handles conditions where things may not exist.  By using an Option instead of a null or nil type, there is no way to try to use an object that doesn't exist in Rust.
+
+In this exercise, you have a partial coded Ice Cream Truck implementation.  Complete the two existing methods so they work and implement the missing method.  You will be returning and handling Option and Result types.  If you have not run across this before, Enums can have function implementations.
--- a/exercises/error-handling/Cargo.lock
+++ b/exercises/error-handling/Cargo.lock
@@ -1,0 +1,4 @@
+[root]
+name = "error-handling"
+version = "1.0.0"
+
--- a/exercises/error-handling/Cargo.toml
+++ b/exercises/error-handling/Cargo.toml
@@ -1,0 +1,6 @@
+[package]
+name = "error-handling"
+version = "1.0.0"
+authors = ["sacherjj <sacherjj@gmail.com>"]
+
+[dependencies]
--- a/exercises/error-handling/example.rs
+++ b/exercises/error-handling/example.rs
@@ -1,0 +1,81 @@
+use std::cmp;
+
+#[derive(Eq,PartialEq,Debug)]
+pub enum IceCream {
+    Chocolate,
+    Vanilla,
+    Strawberry,
+}
+
+impl IceCream {
+    /// Convert various string versions of ice cream flavors to a
+    /// proper IceCream enum or return Error
+    pub fn get_flavor(request: &str) -> Result<IceCream, &'static str> {
+        match request {
+            "Chocolate" | "chocolate" | "choc" | "c" | "C" => Ok(Chocolate),
+            "Strawberry" | "strawberry" | "straw" | "s" | "S" => Ok(Strawberry),
+            "Vanilla" | "vanilla" | "v" | "V" | "plain" => Ok(Vanilla),
+            _ => Err("Unknown Flavor"),
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct IceCreamTruck {
+    // These fields are only pub, because tests are not in the same file
+    // and we need to verify inventory after actions.
+    pub chocolate: u32,
+    pub vanilla: u32,
+    pub strawberry: u32,
+}
+
+use IceCream::*;
+
+impl IceCreamTruck {
+    pub fn new(chocolate: u32, vanilla: u32, strawberry: u32) -> Self {
+        IceCreamTruck { chocolate: chocolate,
+                        vanilla: vanilla,
+                        strawberry: strawberry }
+    }
+
+    /// Buy ice cream and update inventory.
+    /// Return quantity purchased or indication that you could not purchase.
+    pub fn buy(&mut self, qty: u32, flavor: IceCream) -> Option<u32> {
+        let pos_qty: u32;
+        match flavor {
+            Chocolate => {
+                pos_qty = cmp::min(qty, self.chocolate);
+                self.chocolate -= pos_qty;
+            },
+            Vanilla => {
+                pos_qty = cmp::min(qty, self.vanilla);
+                self.vanilla -= pos_qty;
+            },
+            Strawberry => {
+                pos_qty = cmp::min(qty, self.strawberry);
+                self.strawberry -= pos_qty;
+            },
+        }
+        if pos_qty > 0 {
+            Some(pos_qty)
+        } else {
+            None
+        }
+    }
+
+    /// Receives a Vector of tuples with string name of flavor and quantity.
+    /// Converts string name into flavor and buy ice cream.
+    ///
+    /// Returns a vector of quantity_ordered if flavor is valid and
+    /// ice cream inventory exists or None.
+    pub fn process_order(&mut self, orders: Vec<(&str, u32)>) -> Vec<Option<u32>> {
+        let mut out: Vec<Option<u32>> = vec![];
+        for (name, qty) in orders {
+            match IceCream::get_flavor(name) {
+                Ok(flavor) => out.push(self.buy(qty, flavor)),
+                Err(_) => out.push(None),
+            }
+        } 
+        out
+    }
+}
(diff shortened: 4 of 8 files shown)


sacherjj:
I forgot to #[ignore] all tests past the first.  Corrected this with an amend.

sacherjj:
With including the src/lib.rs file, we can include a section that is `don't touch`.  Then use that as something that the user needs to call into.  

Option and Result are confusing starting out.  Perhaps this is a simple exercise of dealing with an Option and Result return type from existing functions, that comes very early in the track.  These two functions are already in lib.rs.  Then it becomes a very early exercise to get people comfortable with them.   I'll do a little more thinking around this one.

sacherjj:
Here is what I'm thinking about, leaving the shell here and letting them implement.  This is for making Result and Option.  Trying to figure out how to have them use something to handle Result and Option.  

```
use std::cmp;

pub enum IceCream {
    Chocolate,
    Vanilla,
    Strawberry,
}

pub struct IceCreamTruck {
    chocolate: u32,
    vanilla: u32,
    strawberry: u32,
}

use IceCream::*;

impl IceCreamTruck {
    pub fn new(chocolate: u32, vanilla: u32, strawberry: u32) -> Self {
        IceCreamTruck { chocolate: chocolate,
                        vanilla: vanilla,
                        strawberry: strawberry }
    }

    pub fn buy(&mut self, qty: u32, flavor: IceCream) -> Option<u32> {
        let mut pos_qty: u32 = 0;
        match flavor {
            Chocolate => {
                pos_qty = cmp::min(qty, self.chocolate);
                self.chocolate -= pos_qty;
            },
            Vanilla => {
                pos_qty = cmp::min(qty, self.vanilla);
                self.vanilla -= pos_qty;
            },
            Strawberry => {
                pos_qty = cmp::min(qty, self.strawberry);
                self.strawberry -= pos_qty;
            },
        }
        if pos_qty > 0 {
            Some(pos_qty)
        } else {
            None
        }
    }

    pub fn get_flavor(request: &str) -> Result<IceCream, String> {
        match request {
            "Chocolate" | "chocolate" | "choc" | "c" | "C" => Ok(Chocolate),
            "Strawberry" | "strawberry" | "straw" | "s" | "S" => Ok(Strawberry),
            "Vanilla" | "vanilla" | "v" | "V" => Ok(Vanilla),
            _ => Err("Unknown Flavor".to_string()),
        }
    }
}

```

sacherjj:
I think this is fully flushed out.  While we don't need to specifically process Option types with the process_order function, I believe this included enough of both to get the user familiar with them.

I decided to remove the impl for the Enum, as this may be new.  I mention it in the hints.md, so they don't get stuck.

We don't really show using various Err value as conditions, but I think any more and this will get too complex.

exercises/error-handling/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+    Chocolate,
+    Vanilla,
+    Strawberry,
+}
+
+#[derive(Debug)]
+pub struct IceCreamTruck {
+    // These fields are only pub, because tests are not in the same file
+    // and we need to verify inventory after actions.
+    pub chocolate: u32,
```
petertseng: this representation is a little unfortunate when considering the fact that `IceCream` exists above. One might expect we would have map of `IceCream` to quantity, but this is no longer possible given the struct must be of this form.

as an alternative, might imagine a function that queries the ice cream truck for how many of each flavour it has left, perhaps?

Then the fields would not need to be `pub`, and in fact the student may choose any internal representation.

exercises/error-handling/example.rs
```diff
@@ -0,0 +1,81 @@
...
+impl IceCreamTruck {
+    pub fn new(chocolate: u32, vanilla: u32, strawberry: u32) -> Self {
+        IceCreamTruck { chocolate: chocolate,
+                        vanilla: vanilla,
+                        strawberry: strawberry }
+    }
+
+    /// Buy ice cream and update inventory.
+    /// Return quantity purchased or indication that you could not purchase.
+    pub fn buy(&mut self, qty: u32, flavor: IceCream) -> Option<u32> {
```
petertseng: I know this is an exercise that is supposed to show off options, but I can't help but notice that since we currently never have `Some(0)` (only have `None`) that we might as well just use a straight `u32` here and no option. There needs to be a situation where `Some(0)` could be a valid answer. Perhaps if you try to buy 0 ice creams you will always get a `Some(0)`, that could work. In that case we would just want tests for buying 0 ice creams of a flavour and that will be good!

exercises/error-handling/example.rs
```diff
@@ -0,0 +1,81 @@
...
+pub enum IceCream {
+    Chocolate,
+    Vanilla,
+    Strawberry,
+}
+
+impl IceCream {
+    /// Convert various string versions of ice cream flavors to a
+    /// proper IceCream enum or return Error
+    pub fn get_flavor(request: &str) -> Result<IceCream, &'static str> {
```
petertseng: makes me wish https://doc.rust-lang.org/std/convert/trait.TryFrom.html were stable so we could use it here.

exercises/error-handling/example.rs
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /// Receives a Vector of tuples with string name of flavor and quantity.
+    /// Converts string name into flavor and buy ice cream.
+    ///
+    /// Returns a vector of quantity_ordered if flavor is valid and
+    /// ice cream inventory exists or None.
+    pub fn process_order(&mut self, orders: Vec<(&str, u32)>) -> Vec<Option<u32>> {
+        let mut out: Vec<Option<u32>> = vec![];
+        for (name, qty) in orders {
```
petertseng: this is a `map` and I think it makes sense to have the example solution express it as such.

exercises/error-handling/example.rs
petertseng: any reason why this function is not present in the stub, even though the stub defines all other functions that the test would expect?

exercises/error-handling/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+impl IceCreamTruck {
+    pub fn new(chocolate: u32, vanilla: u32, strawberry: u32) -> Self {
+        IceCreamTruck { chocolate: chocolate,
+                        vanilla: vanilla,
+                        strawberry: strawberry }
+    }
+
+    /// Buy ice cream and update inventory.
+    /// Return quantity purchased or indication that you could not purchase.
+    pub fn buy() {
```
petertseng: I suppose I understand why there are no args here or return type, because the student should write it to understand option. OK. I guess that means this needs a `.meta/ALLOWED_TO_NOT_COMPILE`.

exercises/error-handling/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    /// Receives a Vector of tuples with string name of flavor and quantity.
+    /// Converts string name into flavor and buy ice cream.
+    ///
+    /// Returns a vector of quantity_ordered if flavor is valid and
+    /// ice cream inventory exists or None.
+    pub fn process_order() {
```
petertseng: I suppose I understand why there are no args here or return type, because the student should write it to understand option. OK. I guess that means this needs a `.meta/ALLOWED_TO_NOT_COMPILE`.

exercises/error-handling/.meta/hints.md
```diff
@@ -0,0 +1,11 @@
+## Rust Specific Track Information
```
petertseng: the README must be regenerated so that these instructions appear inside it.

exercises/error-handling/.meta/hints.md
```diff
@@ -0,0 +1,11 @@
...
+
+In Rust, it is expected (and enforced by the compiler) that all possible outcomes are handled.  There is no throw and catch error system that is present in other languages.
+
+There is panic!(), when the program gets into an unknown and dangerous state.  In this situation, we want the program to shutdown down cleanly.  This is not a state you would catch and continue, it is more of a fatal and crashing condition.
```
petertseng: > shutdown down

down is repeated.

exercises/error-handling/Cargo.lock
```diff
@@ -0,0 +1,4 @@
+[root]
+name = "error-handling"
+version = "1.0.0"
```
petertseng: there is no 1.0.0 in https://github.com/exercism/problem-specifications/tree/master/exercises/error-handling therefore this must be `0.0.0` or `0.1.0` your choice.

exercises/error-handling/Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[package]
+name = "error-handling"
+version = "1.0.0"
```
petertseng: there is no 1.0.0 in https://github.com/exercism/problem-specifications/tree/master/exercises/error-handling therefore this must be `0.0.0` or `0.1.0` your choice.

exercises/error-handling/example.rs
petertseng: Never mind, this is explained.

> I decided to remove the impl for the Enum, as this may be new functionality for a user who has only seen impl for structs. I mention it in the hints.md, so they don't get stuck, but want them to research a little if this is new to them.

I will think about whether I agree with it.

coriolinus:
@sacherjj Are you still working on this?

sacherjj:
I will be able to in a month or so.  We had a death in the family and I'm trying to help settle things.

coriolinus:
I'm sorry for your loss. Please take the time that you need.

coriolinus:
@sacherjj Are you still interested in working on this issue? It's fine if you're not, but if you are, I'd like to see some progress. My goal is to clean up the open pull requests. It's been six months since you projected you'd resume work, and 10 since the PR was opened, so we should either drive this forward to completion or close it.

sacherjj:
Close this. Too many other things have been getting in the way on completing this.

PR #328, opened 2017-07-24
clock: Add exercise

sacherjj:
I think we found the winner of the most test cases.   Test Driven Drudgery.

Is there a not equal assert?   I changed to assert!( 1 != 2 ).

Thought about using Add trait for adding int to Clock as minutes add, but thought it would get too confusing.  I figured the Display trait was enough.   Thoughts?

### DIFF
--- a/config.json
+++ b/config.json
@@ -167,6 +167,18 @@
       ]
     },
     {
+      "uuid": "543a3ca2-fb9b-4f20-a873-ff23595d41df",
+      "slug": "clock",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 4,
+      "topics": [
+        "traits",
+        "derive",
+        "struct"
+      ]
+    },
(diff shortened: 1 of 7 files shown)


sacherjj:
Another option would be to not mention the Trait.  Then it would be assumed that the user would implement it as `fn to_string()`.   Using the `.meta/hints.md`, put in an extended exercise for trying this with the Trait.  This would allow clean md. links to the docs as well.  

I think I like that option better.  Comments?

ijanos:
I like this option too.

sacherjj:
Code is updated to use the .hints and eliminate mention of Display trait until then.   Also fixed the assert_ne!.

sacherjj:
I have been using hints.md as more of a rust_specific.md.   So I have been writing them with expectation that they exist directly below the common exercise text.

PR #329, opened 2017-07-24
nth-prime: Add exercise

sacherjj:
Feel like this is fairly simple as it comes to Rust specific.  Just looping and math.  I have it early in the difficulty.


ijanos:
As a fun fact, iterators also have a method called [nth](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.nth) and you can solve this with an iterator that creates the primes on demand. 

```rust
let mut primes = (2..).filter(|&n| (2..(n as f32).sqrt() as u32 + 1).all(|i| n % i != 0));
primes.nth(10_000).unwrap()
```

Sadly, returning an iterator from a function is kind of a type nightmare without `impl Trait` so I would not advise using it in any Exercism exercise.

sacherjj:
I was playing with the sqrt ending algorithm (in non-iter looping) and managed to infinite loop VSCode into requiring a kill.  So I simplified mine.  

Iterators in rust allow very compact statements.  I worry about taking it too far and playing code golf.  :)

exercises/nth-prime/tests/nth-prime.rs
```diff
@@ -0,0 +1,30 @@
...
+#[test]
+#[ignore]
+fn test_big_prime() {
+    assert_eq!(np::nth(10001), Ok(104743));
+}
+
+#[test]
+#[ignore]
+fn test_zeroth_prime() {
+    assert!(np::nth(0).is_err());
```
petertseng: Well, I made this change and now I want to ask what y'all think about it. I don't think I would want to encourage students to have `Result<T, E>` where `E` is `()` since then we might as well just have `Option`, right? So, it seems better to just use `is_err` here instead of `assert_eq!(np::nth(0), Err(()))` . Agreed?

sacherjj: This makes sense to me.

PR #330, opened 2017-07-24
proverb: Add exercise

sacherjj:
Fairly simple one.


exercises/proverb/tests/proverb.rs
```diff
@@ -0,0 +1,52 @@
...
+                        "For want of a shoe the horse was lost.",
+                        "And all for the want of a horseshoe nail.").join("\n");
+    assert_eq!(build_proverb(input), expected);
+}
+
+#[test]
+#[ignore]
+fn test_one_piece() {
+    let input = vec!("nail");
+    let expected = vec!("And all for the want of a nail.").join("\n");
```
ijanos: This could be just a simple `"And all for the want of a nail.\n"` no need to allocate a Vector.

sacherjj:
I don't know where I picked up `vec!()`.  Looking at the official docs it is `[]` and that makes more sense for an array styled type.  I'll do a little clean up on this one.

exercises/proverb/Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[package]
+name = "proverb"
+version = "0.0.0"
```
petertseng: I had to call it version 0 since we have no https://github.com/exercism/problem-specifications/blob/master/exercises/proverb/canonical-data.json existing. That means if we call it 1.0.0 and then a 1.0.0 comes into existence that we do not comply with, we may falsely believe ourselves to be compliant, which defeats the purpose of tracking compliance

exercises/proverb/README.md
```diff
@@ -0,0 +1,50 @@
...
+The [exercism/rust](https://github.com/exercism/rust) repository on GitHub is the home for all of the Rust exercises. If you have feedback about an exercise, or want to help implement new exercises, head over there and create an issue. Members of the [rust track team](https://github.com/orgs/exercism/teams/rust) are happy to help!
+
+If you want to know more about Exercism, take a look at the [contribution guide](https://github.com/exercism/docs/blob/master/contributing-to-language-tracks/README.md).
+
+[help-page]: http://exercism.io/languages/rust
+[crates-and-modules]: http://doc.rust-lang.org/stable/book/crates-and-modules.html
+
+## Source
+
+Wikipedia [http://en.wikipedia.org/wiki/For_Want_of_a_Nail](http://en.wikipedia.org/wiki/For_Want_of_a_Nail)
```
petertseng: while I do have suspicions that `[text](url)` is better than `text [url](url)` this is the way the generator does it so until that changes I'm going to make the READMEs that way so that no READMEs break the pattern

PR #332, opened 2017-07-25
accumulate: Add exercise

sacherjj:
I learned something new from this one.  However, I'm not sure that function pointers is the best way for people to solve this or if it is generally the correct thing to do in Rust.   

I'm looking for some feedback from more experienced on this.  I have not seen function pointers much in Rust docs.  But this may just be because map takes care of them for most cases.

Do you think .meta\hints.md is required?  I only found function pointers by searching with that text and hitting on a Stack Overflow answer.  I worry if the user doesn't understand that phrase, they could get stuck.

### DIFF
--- a/exercises/accumulate/README.md
+++ b/exercises/accumulate/README.md
@@ -28,6 +28,18 @@ Solve this one yourself using other basic tools instead.
 Lisp specific: it's perfectly fine to use `MAPCAR` or the equivalent,
 as this is idiomatic Lisp, not a library function.
 
+## Hints
+
+We are dealing with two types of situations.  One is a function pointer and
+the other is a closure.
+
+It may help to look at the [Fn trait](https://doc.rust-lang.org/std/ops/trait.Fn.html).
+
+Help with passing a closure into a function may be found in
+the ["closures as input parameters" section](https://rustbyexample.com/fn/closures/input_parameters.html) of
+[Rust by Example](https://rustbyexample.com/).
+
+
 ## Rust Installation
 
 Refer to the [exercism help page][help-page] for Rust installation and learning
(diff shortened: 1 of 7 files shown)


exercises/accumulate/README.md
```diff
@@ -0,0 +1,67 @@
...
+The [exercism/rust](https://github.com/exercism/rust) repository on GitHub is the home for all of the Rust exercises. If you have feedback about an exercise, or want to help implement new exercises, head over there and create an issue. Members of the [rust track team](https://github.com/orgs/exercism/teams/rust) are happy to help!
+
+If you want to know more about Exercism, take a look at the [contribution guide](https://github.com/exercism/docs/blob/master/contributing-to-language-tracks/README.md).
+
+[help-page]: http://exercism.io/languages/rust
+[crates-and-modules]: http://doc.rust-lang.org/stable/book/crates-and-modules.html
+
+## Source
+
+The Pig Latin exercise at Test First Teaching by Ultrasaurus [https://github.com/ultrasaurus/test-first-teaching/blob/master/learn_ruby/pig_latin/](https://github.com/ultrasaurus/test-first-teaching/blob/master/learn_ruby/pig_latin/)
```
ijanos: I think this link is not the right one.

sacherjj: Thought I had fixed all my bad sources...  Hopefully correct now.

sacherjj:
I thought researching and learning how to properly make the type signatures was a decent challenge for someone who has not done it before.  If you know that, it is a simple for loop.  

I haven't fully wrapped my head around where I would use FnMut or FnOnce.

petertseng:
Thanks.

You may notice that among the things I do is to run the generator on the README. Not sure if I should be asking everyone to install configlet and doing that. Probably something I should discuss with other maintainers.

Everything else is just small changes and I just did it so that we can get things merged faster.

PR #333, opened 2017-07-25
pythagorean-triplet: Add Exercise

sacherjj:
Another simple one.   Decided to include option, rather than have a return 0 at the bottom of the method.  Make it a little Rustic.


sacherjj:
I agree.  Euler is not about learning programming and usually about showing my how much higher level math I've forgotten since finishing my EE degree.  :)

exercises/pythagorean-triplet/Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[package]
+name = "pythagorean_triplet"
+version = "0.0.0"
```
petertseng: there is no canonical JSON in https://github.com/exercism/problem-specifications/tree/master/exercises/pythagorean-triplet , so it couldn't have been called 1.0.0, so I put it back at 0.

petertseng:
If you could do one of the two choices:

* Say whether this problem should come before or after prime-factors
* Rebase on master (in order to place pythagorean-triplet in relation to prime-factors in config.json)

PR #337, opened 2017-07-31
use assert_ne!(a, b) instead of assert!(a != b)

petertseng:
This gives a better error message when the assertion is violated.

This was added in August 2016, after each of the changed tests were
added.


IanWhitney:
Still wish they had `refute`, though.

PR #341, opened 2017-08-13
Implement exercise crypto-square

coriolinus:
The [crypto-square](https://github.com/exercism/problem-specifications/blob/master/exercises/crypto-square/description.md) problem looks interesting, and isn't implemented yet for Rust, so I figured I'll take a shot at implementing it.


coriolinus:
Travis build fails because the example implementation depends on `itertools` as commented [here](https://github.com/coriolinus/rust/blob/crypto-square/exercises/crypto-square/example.rs#L3-L4), but that line's not in the default `Crates.toml` because I didn't want to constrain the way people approached the problem.

Maintainers: how is this problem usually solved? Do I need to rewrite the example to avoid the external dependency?

coriolinus:
With travis passing, I think this is just waiting for a maintainer to squash-rebase-merge this in.

exercises/crypto-square/Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[package]
+name = "crypto-square"
+version = "0.1.0"
```
petertseng: given that https://github.com/exercism/problem-specifications/blob/master/exercises/crypto-square/canonical-data.json exists, normally I would say this should be 2.0.0 and using the test cases from this file. However, since https://github.com/exercism/problem-specifications/issues/750 exists, I wish to exempt crypto-square from following problem-specifications in its current form.

coriolinus:
Thanks! 

I hadn't realized that [`canonical-data.json`](https://github.com/exercism/problem-specifications/blob/master/exercises/crypto-square/canonical-data.json) existed. Still, I do think it makes more sense to only test the end result, not the intermediate functions.

config.json
```diff
@@ -289,6 +289,20 @@
       ]
     },
     {
+      "uuid": "0cc485e9-43ba-4d97-a622-ee4cb8b9f1f7",
+      "slug": "crypto-square",
```
petertseng: this is, I believe, the only exercise that doesn't define `"core"` or `"unlocked_by"`

Interesting that configlet doesn't care. Does Nextercism care? Does this exercise show up on Nextercism, and if so, in what capacity?

petertseng: Far as I can tell it is a non-core already-unlocked (unlocked_by null) exercise. I did not read any code to verify this statement so I'm prepared to be wrong.

coriolinus: It is interesting, but I'm not sure what to do with the information. I presume that if Nextercism cares, we'll discover it and fix it upon adoption. For now my basic attitude is that configlet is the current canonical judge of what is correct; as it doesn't care, the file is valid.

PR #343, opened 2017-08-22
Add default introductory code example

kytrinyx:
Instead of having logic for a fallback in the backend, we're choosing a
default for the snippet file.

For tracks that have core exercises, we pick the first core exercise.
For tracks without these, we pick the first exercise listed in the config.

See https://github.com/exercism/meta/issues/89

### DIFF
--- a/docs/SNIPPET.txt
+++ b/docs/SNIPPET.txt
@@ -1,0 +1,3 @@
+pub fn hello() -> &'static str {
+    "Hello, World!"
+}


petertseng:
I have a mind to ask whether the code located at https://www.rust-lang.org/en-US/ would be suitable, but we would have to change the "Editable and runnable" portion

kytrinyx:
It's too long and too wide for our example snippet.

PR #344, opened 2017-08-22
config: replaced "math" topic with "mathematics"

tleen:
See exercism/problem-specifications#884

### DIFF
--- a/config.json
+++ b/config.json
@@ -21,7 +21,7 @@
       "difficulty": 1,
       "topics": [
         "crates",
-        "math"
+        "mathematics"
       ]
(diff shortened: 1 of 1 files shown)


tleen:
Nothing to do with the topic. Just to keep things [consistent with the other tracks](https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt), some used "math" some "mathematics". Now that topics will be showing up as people do tracks wanted to bring them in-line cross-track.

ijanos:
Oh, I missed that. Great, thank you :)

PR #347, opened 2017-08-24
Implement exercise Poker

coriolinus:
Staking a claim on implementing the [poker](https://github.com/exercism/problem-specifications/tree/master/exercises/poker) exercise.


coriolinus:
Travis passes, so I think this should be all set for someone with merge authority. 

Note that the command `configlet generate . --only poker` eliminated all problem-specific text from the README, so either I was using it wrong or we'll want to override its output.

exercises/poker/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+/// Given a list of poker hands, return a list of those hands which win.
+///
+/// Note the type signature: this function should return _the same_ reference to
+/// the winning hand(s) as were passed in, not reconstructed strings which happen to be equal.
+pub fn winning_hands<'a>(_: &[&'a str]) -> Option<Vec<&'a str>> {
```
ijanos: I would prefer using a variable name instead of `_`

coriolinus: 1. Any variable name we use will need to begin with an underscore anyway, to prevent compiler warnings about unused variables.
2. Using `_` encourages the implementer choose their own variable name.
3. The only point of including the function stub is so that users don't have to guess and check the expected signature by inspecting the tests. While the name used is technically part of the public signature of the function, I figured it best to minimize the degree to which we constrain the individual implementers.

If you feel very strongly about it, I can change it, but I think there are good reasons to leave it as is.

ijanos: Ok, I'm convinced.

ijanos:
Here is what I'm thinking about: I think the rust track would really benefit from an exercise that teaches implementing `PartialOrd` as implementing traits is a really important concept. I see two options:

- modify this exercise to focus on the `PartialOrd` implementation. 
- leave this one as is and add a simple track specific exercise about `PartialOrd` somewhere before this one and modify the poker readme nudging people to use that knowledge to create the implementation for this one. 

I think the second option would be better, what do you think?

coriolinus:
I'd agree that traits are an important concept to learn, and that this exercise is a pretty natural fit for using that trait implementation. After all, that's the path I went in my own solution. That said, I tend to prefer exercises which allow a broad range of implementations over those which railroad the implementer into a particular design choice, and the required function signature already requires that people know about lifetimes in function signatures.

In short, I agree that the second path is better. I'm going to adjust the readme now to suggest the use of `PartialOrd` without actually requiring it from the tests.

ijanos:
This is awesome, thank you! I will wait a few days with the merge to give a chance to other maintainers to chime in, if you don't mind.

coriolinus:
@ijanos It's been two weeks with no further comments; perhaps you could merge this now?

exercises/poker/README.md
```diff
@@ -0,0 +1,47 @@
+# Poker
+
+Pick the best hand(s) from a list of poker hands.
+
+See [wikipedia](https://en.wikipedia.org/wiki/List_of_poker_hands) for an overview of poker hands.
+
+## Hints
+
+- Ranking a list of poker hands can be considered a sorting problem.
```
petertseng: these hints are Rust-specific, therefore for the README generator to not remove them, they need to be placed in .meta/hints.md . I did so in a commit.

petertseng:
Question: README generator writes the description wrapped to some number of columns, because those are how the files appear in https://github.com/exercism/problem-specifications/blob/master/exercises/poker/description.md / https://raw.githubusercontent.com/exercism/problem-specifications/master/exercises/poker/description.md and https://github.com/exercism/rust/blob/master/config/exercise-readme-insert.md / https://raw.githubusercontent.com/exercism/rust/master/config/exercise-readme-insert.md .

So, regenerating the README results in this diff:

```
diff --git i/exercises/poker/README.md w/exercises/poker/README.md
index 75e1554..2f978d0 100644
--- i/exercises/poker/README.md
+++ w/exercises/poker/README.md
@@ -2,7 +2,8 @@
 
 Pick the best hand(s) from a list of poker hands.
 
-See [wikipedia](https://en.wikipedia.org/wiki/List_of_poker_hands) for an overview of poker hands.
+See [wikipedia](https://en.wikipedia.org/wiki/List_of_poker_hands) for an
+overview of poker hands.
 
 ## Hints
 
@@ -13,9 +14,11 @@ See [wikipedia](https://en.wikipedia.org/wiki/List_of_poker_hands) for an overvi
 - Rust provides the [`PartialOrd` trait](https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html) to handle the case of sortable things which do not have a total order. However, it doesn't provide a standard `sort` method for `Vec<T> where T: PartialOrd`. The standard idiom to sort a vector in this case is `your_vec.sort_by(|a, b| a.partial_cmp(b).unwrap_or(Ordering::{Less|Equal|Greater}));`, depending on your needs. `
 - You might consider implementing a type representing a poker hand which implements `PartialOrd`.
 
+
 ## Rust Installation
 
-Refer to the [exercism help page][help-page] for Rust installation and learning resources.
+Refer to the [exercism help page][help-page] for Rust installation and learning
+resources.
 
 ## Writing the Code
 
@@ -25,9 +28,14 @@ Execute the tests with:
 $ cargo test
 ```
 
-All but the first test have been ignored. After you get the first test to pass, remove the ignore flag (`#[ignore]`) from the next test and get the tests to pass again. The test file is located in the `tests` directory. You can also remove the ignore flag from all the tests to get them to run all at once if you wish.
+All but the first test have been ignored.  After you get the first test to
+pass, remove the ignore flag (`#[ignore]`) from the next test and get the tests
+to pass again.  The test file is located in the `tests` directory.   You can
+also remove the ignore flag from all the tests to get them to run all at once
+if you wish.
 
-Make sure to read the [Crates and Modules](https://doc.rust-lang.org/stable/book/crates-and-modules.html) chapter if you haven't already, it will help you with organizing your files.
+Make sure to read the [Crates and Modules](https://doc.rust-lang.org/stable/book/crates-and-modules.html) chapter if you
+haven't already, it will help you with organizing your files.
 
 ## Feedback, Issues, Pull Requests
 
@@ -35,13 +43,12 @@ The [exercism/rust](https://github.com/exercism/rust) repository on GitHub is th
 
 If you want to know more about Exercism, take a look at the [contribution guide](https://github.com/exercism/docs/blob/master/contributing-to-language-tracks/README.md).
 
+
 ## Source
 
-J Dalbey's Programming Practice problems <http://users.csc.calpoly.edu/~jdalbey/103/Projects/ProgrammingPractice.html>
+Inspired by the training course from Udacity. [https://www.udacity.com/course/viewer#!/c-cs212/](https://www.udacity.com/course/viewer#!/c-cs212/)
 
 ## Submitting Incomplete Solutions
-
 It's possible to submit an incomplete solution so you can see how others have completed the exercise.
-
-[crates-and-modules]: http://doc.rust-lang.org/stable/book/crates-and-modules.html
-[help-page]: http://exercism.io/languages/rust
```

I'm wondering whether the hints should get wrapped too, to match.

exercises/poker/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Feedback, Issues, Pull Requests
+
+The [exercism/rust](https://github.com/exercism/rust) repository on GitHub is the home for all of the Rust exercises. If you have feedback about an exercise, or want to help implement new exercises, head over there and create an issue. Members of the [rust track team](https://github.com/orgs/exercism/teams/rust) are happy to help!
+
+If you want to know more about Exercism, take a look at the [contribution guide](https://github.com/exercism/docs/blob/master/contributing-to-language-tracks/README.md).
+
+## Source
+
+J Dalbey's Programming Practice problems <http://users.csc.calpoly.edu/~jdalbey/103/Projects/ProgrammingPractice.html>
```
petertseng: this is not the source listed in https://github.com/exercism/problem-specifications/blob/master/exercises/poker/metadata.yml.

petertseng:
Well, my constraints are that I don't want to maintain any README file that *some* generator does not generate (even if it isn't configlet's generator), so I would prefer to use the README as it was appears in https://github.com/exercism/rust/pull/347/commits/f4b2ae5c2dfbfe3b02d8a8151baa982d5ed0131a , hoping nobody has objections to that (the fact that some lines are wrapped and some aren't). Assuming nobody has complaints, then there is nothing more to be said

petertseng:
Thanks for implementing this exercise!

PR #353, opened 2017-09-26
Add missing .gitigore files for the exercises

attilahorvath:
I like to keep my Exercism solutions backed up on GitHub (e.g. https://github.com/attilahorvath/exercism-rust) and I've seen others do the same.

However, while solving the Clock exercise I've noticed that it was missing the `.gitignore` file so all kinds of build artifacts got added to the commit accidentally.

This PR is intended to add these missing `.gitignore` files to all current exercises in the Rust track to avoid such issues in the future.


ijanos:
Thank you for adding these! Cheers.

petertseng:
> to avoid such issues in the future.

Perhaps CI should ensure it!

Alternatively, it would be great if it would go in `$EXERCISM/rust/.gitignore` instead of `$EXERCISM/rust/$EXERCISE_SLUG/.gitignore`. Perhaps https://github.com/exercism/cli/issues/418 would help us, wonder how to use it.

PR #354, opened 2017-10-04
READMEs: Import problem-specifications commits

petertseng:


petertseng:
If you ask, the import was done by hand. I'll think about automating it later.

petertseng:
It is intended that all commit messages link to the original problem-specifications PR, and that the author of each commit is the author of the problem-specifications commit. Let me know if I got any of that wrong.

petertseng:
In case someone ask: No, it's not too bad if this gets squashed. My *preference* is to keep them separate commits so the rationale for each change is maintained, but I wouldn't complain if they get squashed into one commit with "regenerate READMEs from problem-specifications" since this is basically what this is doing (minus space-age). The strength of my preference is 4/10, so if someone has any stronger preference, you pick which way you like.

ijanos:
I would prefer squashing these, the repo history would be cleaner in my opinion.

petertseng:
OK, I defer. And besides, if someone wants to see the individual commits, they at least have this PR to refer to, as long as GitHub continues to exist.

petertseng:
The effect of Squash and Merge button on this PR seems to be (based on what I observed in https://github.com/exercism/haskell/pull/603) that the resulting commit will have author of me (I submitted the PR) and committer of the one who merges. I'm fine with that.

petertseng:
I will not be squashing the commits before merging because I prefer to be able to see the history of the individual changes, and since they are going to be squashed on merge, I need to have this PR as a fallback as where to see it.

I plan to use something like the following for the commit message of the squashed commit.

```
regenerate READMEs

Relevant problem-specifications issues:

https://github.com/exercism/problem-specifications/issues/810
https://github.com/exercism/problem-specifications/issues/889
https://github.com/exercism/problem-specifications/pull/892
https://github.com/exercism/problem-specifications/pull/909
https://github.com/exercism/problem-specifications/pull/913
https://github.com/exercism/problem-specifications/pull/915
https://github.com/exercism/problem-specifications/pull/916
```

For this PR I see no need to violate my policy, so until it receives Approval (in the GitHub sense), I'm just going to keep lumping on problem-specifications changes into this same PR.

petertseng:
>  going to be squashed on merge

(Squash and Merge button, of course)

petertseng:
The kindergarten-garden commit was a mistake; that exercise is not in this track.

petertseng:
Proposed commit message of squashed comit is updated.

```
regenerate READMEs

Relevant problem-specifications issues:

https://github.com/exercism/problem-specifications/issues/356
https://github.com/exercism/problem-specifications/issues/810
https://github.com/exercism/problem-specifications/pull/880
https://github.com/exercism/problem-specifications/issues/889
https://github.com/exercism/problem-specifications/pull/892
https://github.com/exercism/problem-specifications/pull/909
https://github.com/exercism/problem-specifications/pull/913
https://github.com/exercism/problem-specifications/pull/915
https://github.com/exercism/problem-specifications/pull/916
https://github.com/exercism/problem-specifications/pull/929
https://github.com/exercism/problem-specifications/pull/945
https://github.com/exercism/problem-specifications/pull/946
```

petertseng:
Better to list the issues along with the PRs they are associated with.

```
regenerate READMEs

Corresponding problem-specifications PRs (and issues, if appropriate):

* https://github.com/exercism/problem-specifications/pull/892
    * and https://github.com/exercism/problem-specifications/issues/889
* https://github.com/exercism/problem-specifications/pull/909
* https://github.com/exercism/problem-specifications/pull/913
* https://github.com/exercism/problem-specifications/pull/915
* https://github.com/exercism/problem-specifications/pull/916
    * and https://github.com/exercism/problem-specifications/issues/810
* https://github.com/exercism/problem-specifications/pull/929
    * and https://github.com/exercism/problem-specifications/issues/356
* https://github.com/exercism/problem-specifications/pull/946
    * and https://github.com/exercism/problem-specifications/pull/945
    * and https://github.com/exercism/problem-specifications/pull/880
```

exercises/proverb/README.md
```diff
@@ -1,16 +1,18 @@
 # Proverb
```
coriolinus: As soon as exercism/problem-specifications#994 is merged, this will no longer be in sync.

coriolinus:
@petertseng Given the amount of activity in the problem specifications, and the amount of work you do checking that various readmes match the generated version, is there a strong reason to put them in the repo at all? It's easy to imagine a system in which they're generated dynamically by merging relevant info from the problem-specifications repo and .meta; this seems superior to having to manually track and update them.

petertseng:
> is there a strong reason to put them in the repo at all

There is a reason. I'll leave it up to you to determine whether https://github.com/exercism/meta/issues/15 is considered strong.

> It's easy to imagine a system in which they're generated dynamically by merging relevant info from the problem-specifications repo and .meta

https://github.com/exercism/trackler/blob/master/lib/trackler/implementation.rb#L51
https://github.com/exercism/trackler/blob/master/lib/trackler/implementation.rb#L92-L125
https://github.com/exercism/prototype/blob/4d093ec7766b292f014205d1747015b90b6468c5/app/services/git/syncs_track.rb#L94

> this seems superior to having to manually track and update them.

I agree.

coriolinus:
Ok, it looks like it's exercism policy to have READMEs which are stored within the repo but dynamically generated. I still think it's a bad idea for humans to have to manage this stuff manually.

I propose the following:

1. Someone (maybe me) writes a server capable of handling a GitHub [PullRequestEvent](https://developer.github.com/v3/activity/events/types/#pullrequestevent) webhook.
2. Someone (not me) runs this server somewhere permanently online.
3. This server monitors PRs: on merge, it calls the appropriate already-existing scripts to generate the README.
4. The server uploads/updates the exercise README using the [API](https://developer.github.com/v3/repos/contents/#update-a-file)
5. We add a note in the Rust track README telling people not to bother with manually writing READMEs, but instead to add entries to `.meta` as appropriate for their exercise. 

Sound plausible?

petertseng:
The only thing I would advise before potentially starting this work is just making sure that Katrina did not already do the same, given the statement of intent in https://github.com/exercism/discussions/issues/200#issuecomment-338470291 - I will assume that it's just a statement of intent and that nothing has been done yet, but still good to check.

Here are some things you might take into account when making such a thing:

I want the resulting commit to:

* contain the original commit message (so we can see the rationale, not just "automatically regenerate")
* contain a link to the problem-specifications PR (so that we can find any discussion that may have happened there)
* have its Author set to be the author of the original commit (it's only fair to give credit where credit is due)
* have its AuthorDate set to be the date of the original commit (this generally seems like the right thing to do, but I don't really have a reason why this needs to be important)
* (obviously those above two points don't apply to Committer and CommitterDate; those get to be the bot or whatever we want)

Feel free to convince me that I should not want any of those things, though.

coriolinus:
Good point. Digging around, I discovered exercism/meta#96, which seems to cover the scope of this work. However, there's still some discussion as to how much it should actually do; the thrust of the comments seems to be that it's simpler and generally more correct to trigger CI fails when something goes wrong than it is to attempt to generate things automatically. 

There are good points there, so let's reformulate again: what if we simply added a test which runs `configlet generate` and fails the CI if the output is not the same as the existing `README.md`? I'm not familiar with how Travis is configured; could it be as simple as simply adding a script to [`/_test`](https://github.com/exercism/rust/tree/master/_test)?

petertseng:
> what if we simply added a test which runs `configlet generate` and fails the CI if the output is not the same as the existing `README.md`

I will find this useful, especially for new exercises. We have seen today that we have not applied some README changes to new exercises.

We have to be careful of situations such as: I make a change to exercise X, but in problem-specifications a change to exercise Y is made so my change to exercise X can't proceed until we match the change to exercise Y.

> I'm not familiar with how Travis is configured; could it be as simple as simply adding a script to /_test?

It could be as simple as adding a script; the location does not matter, only that it is listed in https://github.com/exercism/rust/blob/master/.travis.yml#L3

coriolinus:
Is that actually a problem, though? If the default behavior of `configlet generate` is to regenerate the READMEs for every exercise in the track, we can add an instruction for submitters which says "if the build fails because of READMEs, just run `configlet generate` and update your PR; that should cause the build to succeed". 

Pros: easy for contributors; keeps all exercise READMEs updated.
Cons: Multiple README updates get packed into a single commit; theoretical chance that the README and exercise implementation could diverge.

petertseng:
> we can add an instruction for submitters which says "if the build fails because of READMEs, just run `configlet generate` and update your PR; that should cause the build to succeed".

I can't say for sure that the following will be a contributor's reaction to this, but I would think "why do I have to make these irrelevant changes in order to get the build to succeed? Why do I have to download this random tool that I don't even know I can trust, and who knows what will happen if I run it?" I would find it a bad experience for contributors.

It may make life slightly harder for maintainers too, since if we intend to merge (say) a PR that adds an exercise and it has a commit for unrelated README change, we can no longer use Squash and Merge button since it doesn't make sense to make the two unrelated things into one commit, whereas we might otherwise be able to (and want to).

coriolinus:
Those are good points. In that case, what we want is for Travis to be able to detect which exercises are affected by the current PR, and only generate and compare READMEs for those. Something like

```bash
for exercise in $(git diff --name-only master..$(git rev-parse --abbrev-ref HEAD) | grep exercises/ | cut -d'/' -f2 -s | sort -fu); do
    # `$exercise` is a track name here, so Travis needs to run 
    # `configlet generate --only $exercise` and compare its output to the existing README
done
```

Using a loop like that, we can fail if and only if a README already affected by the PR is now out of sync.

petertseng:
The approach is acceptable because it doesn't start requiring PRs to have unrelated changes.

Note that with this approach, our READMEs still may remain out of date for an unbounded amount of time, simply if there are no PRs affecting a README that goes out of date. Someone interested in that might try to devise a scheme that can help with that.

Issue #279, opened 2017-04-25
Document why we ignore all tests but the first


petertseng:
We should document this somewhere in this track. We may incorporate by reference any other documentation elsewhere in Exercism such as https://github.com/exercism/exercism.io/issues/856, but it is convenient to have it on hand somewhere in this track, so we can explain in what way the Exercism-wide policy affects this track.

This documentation should go in the README (or some suitable place) so it is visible to contributors who may wish to contribute a new exercise.

It would probably also be good to show this in a place visible to students, so they understand why their tests have `#[ignore]` everywhere. Consider placing it in docs/TESTS.md I suppose? https://github.com/exercism/xrust/blob/master/docs/TESTS.md - it shows up on http://exercism.io/languages/rust/tests.

Note that before #266, GETTING_STARTED.md used to say that tests were ignored but **still did not explain why**. Now there is only one test (which is not ignored) in hello-world, so it it is no longer the right place to put this explanation, just providing a bit of history.

I am unsure of whether to nominate this for good first patch. On the one hand, I figure this is a decision with historical context behind it, and the explanation might be best written by someone with that context. On the other hand, the explanation might instead be best written by a newcomer who benefitted from one-test-at-a-time and so can explain it firsthand.

So, feel free to chime in if you have experienced the benefits and want to explain it.

petertseng:
It would probably also be good to show this in a place visible to students. Consider placing it in docs/TESTS.md I suppose? https://github.com/exercism/xrust/blob/master/docs/TESTS.md - it shows up on http://exercism.io/languages/rust/tests.

Note that before #266, GETTING_STARTED.md used to say that tests were ignored but **still did not explain why**. Now there is only one test (which is not ignored) in hello-world, so it it is no longer the right place to put this explanation, just providing a bit of history.

I am unsure of whether to nominate this for good first patch. On the one hand, I figure this is a decision with historical context behind it, and the explanation might be best written by someone with that context. On the other hand, the explanation might instead be best written by a newcomer who benefitted from one-test-at-a-time and so can explain it firsthand.

So, feel free to chime in if you have experienced the benefits and want to explain it.

petertseng:
Hello. There is a paragraph in https://github.com/exercism/docs/blob/master/contributing-to-language-tracks/exercise-test-suites.md added by https://github.com/exercism/docs/pull/34 that explains that some tracks do this, and gives some explanation.

Your contribution might simply be to link to that document in this track's README (or other suitable place).

[...]

PR #359, opened 2017-10-08
Fixes #310 Conform phone number unit test to NANP and updated track to 1.2.0

### LINKED ISSUE #310: Phone number unit test not conforming to README (and neither to NANP)
The README states (correctly according to the [NANP format](https://en.wikipedia.org/wiki/North_American_Numbering_Plan#Modern_plan) that the first digit of the Area and Exchange codes should be in the range 2-9 (so excluding 0 and 1).

However, the unit tests do not conform to this: invalid numbers (e.g. 1234567890) are supposed to be recognized as valid to pass the test. The example solution also does not conform.

Should I create a pull request with a correction or am I getting confused?

vinmaster:


vinmaster:
This is for issue: https://github.com/exercism/rust/issues/310. Please review 😄

petertseng:
> This is for issue: #310

Can you confirm whether the issue is resolved if this PR is merged?

If it is, we need you to use one of the words in https://help.github.com/articles/closing-issues-using-keywords/ in your commit message or your PR description so that the issue will automatically be closed.

If it is not, could you describe briefly in #310 what work will still need to be done after this is merged?

Thank you.

exercises/phone-number/tests/phone-number.rs
```diff
@@ -5,90 +5,70 @@ fn to_some_string(s: &str) -> Option<String> {
...
-    assert_eq!(phone::number("(123) 456-7890"), to_some_string("1234567890"));
+fn test_cleans_the_number() {
+    assert_eq!(
+        phone::number("(223) 456-7890"),
+        to_some_string("2234567890")
+    );
 }
 
 #[test]
-#[ignore]
```
petertseng: please leave the `ignore`s in place, as discussed in https://github.com/exercism/rust/issues/279 and https://github.com/exercism/docs/blob/master/language-tracks/exercises/anatomy/test-suites.md (where it says "In order to provide a better user experience, some language tracks only leave the first test in an exercise test suite active.")

exercises/phone-number/example.rs
```diff
@@ -18,10 +25,14 @@ pub fn area_code(s: &str) -> Option<String> {
 }
 
 pub fn pretty_print(s: &str) -> String {
```
petertseng: since `pretty_print` is no longer being tested, you can and should now remove this function

You also can/should remove area_code

vinmaster:
Thanks for reviewing! I've made the changes mentioned

PR #361, opened 2017-10-10
count ignores

### LINKED ISSUE #352: I don't like having to manually verify that second+later tests are ignored. Can it be automated?
As noted in #279, we ignore all tests except the first.

I don't like having to check by hand whether every new exercise submission has done this.

~~Maybe we can automate it~~ We are automating it in https://github.com/exercism/rust/pull/361 by counting the number of `#[ignore]` lines and ensuring that it is exactly one smaller than the number of `#[test]` lines.

~~By suggestion of a contributor (not myself) another possible way to do it is to run `cargo test` and make sure it says `1 passed; 0 failed`. One disadvantage is it involves having to actually compile and run, whereas counting `ignore` and `test` does not.~~

All current exercises have been fixed so they have exactly one fewer `ignore` line than `test` line.

The remaining task is to add it to Travis CI so that future exercises also receive its benefit.

petertseng:
I don't like having to compare the number of tests and ignores manually,
so this script helps with that.

Closes #352.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@
 language: rust
 script:
 - "./_test/check-exercises.sh"
+- "sh ./_test/count-ignores.sh"
 - "./bin/fetch-configlet"
 - "./bin/configlet lint ."
 sudo: false
--- a/_test/count-ignores.sh
+++ b/_test/count-ignores.sh
@@ -1,0 +1,16 @@
+#!/bin/sh
+
+dir=$(dirname $(dirname $0))
+exitcode=0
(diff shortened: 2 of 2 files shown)


ijanos:
I would love to see this in the CI

petertseng:
I hate to decide to act in contravention of my own policy on the spot by merging without Approval (in the GitHub sense), but given that I need this to review other PRs, my ability to maintain this track is impaired without this merged.

I request faster Approvals (in the GitHub sense).

PR #362, opened 2017-10-10
Fixed inaccurate test/ignores.

f3rn0s:
After creating a script to count tests/ignore amounts, this is a fix for the incorrect exercises.

#352

### DIFF
--- a/exercises/say/tests/say.rs
+++ b/exercises/say/tests/say.rs
@@ -14,7 +14,7 @@ fn test_zero() {
 //
 /*
 #[test]
-#[ignoe]
+#[ignore]
 fn test_negative() {
     assert_eq!(say::encode(-1), String::from("won't compile"));
 }
--- a/exercises/say/tests/say.rs
+++ b/exercises/say/tests/say.rs
@@ -13,10 +13,14 @@ fn test_zero() {
 // If the below test is uncommented, it should not compile.
 //
 /*
-#[test]
-fn test_negative() {
-    assert_eq!(say::encode(-1), String::from("won't compile"));
-}
+
+ #[test]
+ #[ignore]
+ #[should_panic]
+ fn test_negative() {
+     assert_eq!(say::encode(-1), String::from("won't compile"));
+ }
+
 */
 
 #[test]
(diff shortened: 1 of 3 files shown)


exercises/say/tests/say.rs
```diff
@@ -13,13 +13,18 @@ fn test_zero() {
 // If the below test is uncommented, it should not compile.
 //
 /*
-#[test]
-fn test_negative() {
-    assert_eq!(say::encode(-1), String::from("won't compile"));
-}
+
+ #[test]
```
petertseng: this is changing more lines than I expect it to change. A space got added before each line. That makes it hard to understand what changed, since now each line of this test is marked as a changed line. If this change (add space to each line) needs to be made for some reason, it would want to be a separate commit

exercises/say/tests/say.rs
```diff
@@ -13,13 +13,18 @@ fn test_zero() {
...
+ #[ignore]
+ #[should_panic]
```
petertseng: I'm unsure `should_panic` makes sense, since who can say whether the test should panic if it doesn't even compile? I would not add such a line.

I kind of wish Rust would have a `#[should_not_compile]` or something, but that is something to bring up with the Rust project, Is uppose

petertseng: > I kind of wish Rust would have a `#[should_not_compile]`

That is https://github.com/laumann/compiletest-rs apparently. a project for another day (and probably not something we can make students download)

exercises/say/tests/say.rs
```diff
@@ -15,8 +15,6 @@ fn test_zero() {
 /*
 
  #[test]
- #[ignore]
```
petertseng: the `#[ignore]` should still be added, right? should not be removed

exercises/say/tests/say.rs
```diff
@@ -15,6 +15,7 @@ fn test_zero() {
 /*
 
  #[test]
+ #[ignoe]
```
petertseng: spelled wrong

petertseng:
Thanks!

PR #363, opened 2017-10-10
Added ignore/tests counter

f3rn0s:
Created a bash script that when run will find invalid test files.

Automates according to #352

### DIFF
--- a/count_ignores.sh
+++ b/count_ignores.sh
@@ -1,0 +1,40 @@
+current_dir=$(pwd)
+
+cd exercises
+
+for exercise_name in *
+do
+
+#Get filename
+file_name="$exercise_name/tests/*.rs"
+
+#Commented tests
+comment_test_num=$(cat $file_name | grep "// \#\[test\]" | wc -l)
+
+#Number of tests
+test_num=$(cat $file_name | grep "\#\[test\]" | wc -l)
+test_num=$(($test_num - $comment_test_num))
+
+#Commented ignores
+comment_ignore_num=$(cat $file_name | grep "// \#\[ignore\]" | wc -l)
+
+#Number of ignores
+ignore_num=$(cat $file_name | grep "\#\[ignore\]" | wc -l)
+ignore_num=$(($ignore_num - $comment_ignore_num))
+
+#Number of tests minus 1 should be the number of ignores
+compare=$(($test_num - 1))
+
+#Print logic
+if [ "$compare" -eq "$ignore_num" ]
+then
+    #Valid
+    echo -n
+else
+    echo "$file_name : "
+    echo "Tests: $test_num"
+    echo "Ignores: $ignore_num"
+fi
+done
+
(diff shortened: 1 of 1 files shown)


count_ignores.sh
```diff
@@ -0,0 +1,40 @@
+current_dir=$(pwd)
+
+cd exercises
```
petertseng: because this always does `cd exercises`, the caller of this script must be at the root of https://github.com/exercism/rust when running it, which is more disadvantageous compared to being able to run it anywhere

count_ignores.sh
```diff
@@ -0,0 +1,40 @@
...
+
+#Get filename
+file_name="$exercise_name/tests/*.rs"
+
+#Commented tests
+comment_test_num=$(cat $file_name | grep "// \#\[test\]" | wc -l)
+
+#Number of tests
+test_num=$(cat $file_name | grep "\#\[test\]" | wc -l)
+test_num=$(($test_num - $comment_test_num))
```
petertseng: could consider counting those number of lines for which `#[test]` is the first thing on the line (possibly preceded by some number of spaces), rather than taking all lines with `#[test]` anywhere on them and subtracting the comments.

count_ignores.sh
```diff
@@ -0,0 +1,40 @@
...
+
+#Commented ignores
+comment_ignore_num=$(cat $file_name | grep "// \#\[ignore\]" | wc -l)
+
+#Number of ignores
+ignore_num=$(cat $file_name | grep "\#\[ignore\]" | wc -l)
+ignore_num=$(($ignore_num - $comment_ignore_num))
```
petertseng: could consider counting those number of lines for which `#[ignore]` is the first thing on the line (possibly preceded by some number of spaces), rather than taking all lines with `#[ignore]` anywhere on them and subtracting the comments.

count_ignores.sh
```diff
@@ -0,0 +1,40 @@
...
+cd exercises
+
+for exercise_name in *
+do
+
+#Get filename
+file_name="$exercise_name/tests/*.rs"
+
+#Commented tests
+comment_test_num=$(cat $file_name | grep "// \#\[test\]" | wc -l)
```
petertseng: It is probably preferable to `grep PATTERN FILE` rather than `cat FILE | grep PATTERN`, right? Is there a difference between the two?

count_ignores.sh
```diff
@@ -0,0 +1,40 @@
...
+
+#Print logic
+if [ "$compare" -eq "$ignore_num" ]
+then
+    #Valid
+    echo -n
+else
+    echo "$file_name : "
+    echo "Tests: $test_num"
+    echo "Ignores: $ignore_num"
```
petertseng: it would be good if the script would exit nonzero if any file were found to have an unexpected number of `ignore` lines. That will make it usable in CI.

PR #364, opened 2017-10-12
Implement exercise collatz-conjecture

jgilray:


jgilray:
My first ported exercise.  I think everything went fine (thanks to @NobbZ for the help with rebase), but I am confused about the "WIP commit" being automatically removed.  That didn't seem to happen.

jgilray:
In this doc: https://github.com/exercism/docs/blob/master/you-can-help/implement-an-exercise-from-specification.md it seems to indicate that the empty commit would be removed, but maybe I needed to squash my two commits?  Still learning!

petertseng:
Thanks. While my allowed time is up and thus I won't be able to review the code, I can at least answer the question of removing the commit.

As I understand it, a regular rebase (a rebase without any other flags) does not remove any commits (well, okay, it removes merge commits, but those aren't under discussion here). In contrast, an interactive rebase by default will remove empty commits (since it will comment them out when it asks you to edit the commit list).

jgilray:
OK, I squashed the empty commit and the real commit into one commit.  One confusion: there is a lot written about using configlet generate to update the README.md.  I assumed that since the description in problem-speficiations was correct I didn't need to do that, is that correct?

petertseng:
> update the README.md. I assumed that since the description in problem-speficiations was correct I didn't need to do that, is that correct?

README.md for collatz-conjecture needs to be present in this repository. Two reasons:

* The student will only receive files that are in this repository, and we would like students to receive a README.md so they may refer to it
* (using hello-world as an example) the page http://exercism.io/exercises/rust/hello-world/readme uses the exact contents of README.md in this repository, and we would like that page to have the correct content

jgilray:
I created ```README.md``` with ```configlet generate . --only collatz-conjecture```

jgilray:
I **believe** that I have done everything needed to port the problem, looking forward to reviews.

jgilray:
Did another squash to make it look like a single commit

jgilray:
OK, now I think I made a mistake, I did a ```git merge upstream/master``` while I was in the collatz-conjecture branch... (I forgot to switch to master).  Oops... what to do now??

jgilray:
Did some reading... can I just do a git reset on my branch and push again?

petertseng:
> Did some reading... can I just do a git reset on my branch and push again

Sure, `git reset --hard cd2cd2da5bf87b009646023747901dd59aefc079`, followed by a push, will do.

jgilray:
OK thanks @petertseng, Done.  I hope everything is copacetic now

exercises/collatz-conjecture/example.rs
```diff
@@ -0,0 +1,13 @@
+// return Ok(x) where x is the number of steps required to reach 1
+pub fn collatz(n: u64) -> Result<u64, &'static str> {
+    if n == 0 {
+        return Err("Only positive numbers are allowed");
```
petertseng: For the purposes of consistency with the `else` case, the `return` and semicolon should be removed here.

exercises/collatz-conjecture/example.rs
```diff
@@ -0,0 +1,13 @@
...
+    } else if n == 1 {
+        return Ok(0);
```
petertseng: For the purposes of consistency with the `else` case, the `return` and semicolon should be removed here.

exercises/collatz-conjecture/example.rs
```diff
@@ -0,0 +1,13 @@
...
+    } else {
+        Ok(1 + match n % 2 {
+            0 => collatz(n / 2).unwrap(),
```
petertseng: Optional: it is a little unfortunate that these `unwrap` are needed. Since we know that non-zero inputs won't result in non-zero outputs, it would be possible to take advantage of it with a non-`pub` function that is `u64 -> u64`.

I will leave it to you whether you wish to actually do that.

config.json
```diff
@@ -68,6 +68,16 @@
       ]
     },
     {
+      "uuid": "f9afd650-8103-4373-a284-fa4ecfee7207",
+      "slug": "collatz-conjecture",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 1,
+      "topics": [
+        "error handling with Result<T, E>"
```
petertseng: We only used the topic `Result` in most[1] other exercises that use `Result`. It seems it would be better to be consistent. But I can't point to any documentation that says "topics must be consistent within a track" (can anyone find any such documentation), so I have no ground to stand on with such a request.

[1]: I say "most" rather than "all" because roman-numerals uses `results`

petertseng: My question about placing an exercise with `result` in this position: In https://github.com/exercism/rust/pull/219 we made an explicit choice to put an exercises that uses `panic` before any exercise that uses `result`:

> I thought it would be nice to get some exposure to `panic` before
introducing `result` in the Hamming exercise.

So now the question is: If this desire is still valid, it should be that `collatz-conjecture` should be placed at least after `grains`.
If the desire is no longer valid for some reason, all we need is to record that decision here, about why it is no longer valid.

exercises/collatz-conjecture/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+// return Ok(x) where x is the number of steps required to reach 1
+pub fn collatz(n: u64) -> Result<u64, &'static str> {
+    if n == 0 {
+        return Err("Only positive numbers are allowed");
```
petertseng: Is it OK to provide this code to the student, rather than simply providing only the function signature and letting the student fill in the entirety of the body? If the code is provided, a disadvantage is that the student doesn't need to do any work (and therefore doesn't get the experience of learning-by-doing) to get the `test_0` to pass.

exercises/collatz-conjecture/tests/collatz-conjecture.rs
```diff
@@ -0,0 +1,32 @@
...
+#[test]
+#[ignore]
+fn test_1000000() {
+    assert_eq!(Ok(152), collatz(1000000));
+}
+
+#[test]
+#[ignore]
+fn test_0() {
+    assert_eq!(Err("Only positive numbers are allowed"), collatz(0));
```
petertseng: Depending on the answer to whether the `return Err(...)` should be provided, I would think we would only want to test `is_err` on the returned error, to allow the freedom to choose what the `Err` value is.

config.json
jgilray: Until others chime in, I left config.json alone.  I like having this exercise fairly early as it can help students start to look at Result and Option as well.  Of course I will move the exercise if that is the concensus.

exercises/collatz-conjecture/src/lib.rs
jgilray: As an early exercise, I think leaving this in is the easiest way to show how Result is used.  The student still needs to figure out how to return Ok().

exercises/collatz-conjecture/tests/collatz-conjecture.rs
jgilray: Agreed, but see my comment above.

jgilray:
Thanks @petertseng for your thought-provoking review.  I appreciate the time and effort you put in here!

config.json
petertseng: > Of course I will move the exercise if that is the concensus.

Very well, then my contribution to it will be to suggest placing it between grains and hamming. It seems to require fewer mental switches to go from panic -> result -> result than result -> panic -> back to result.

exercises/collatz-conjecture/example.rs
```diff
@@ -0,0 +1,21 @@
...
+            _ => collatz_positive(n * 3 + 1)
+        }
+    }
+}
+
+// return Ok(x) where x is the number of steps required to reach 1
+pub fn collatz(n: u64) -> Result<u64, &'static str> {
+    if n < 1 {
+        Err("Only positive numbers are allowed")
+    } else if n == 1 {
```
petertseng: I have high confidence this `else if n == 1` is not necessary. Let me know if I am wrong on that.

exercises/collatz-conjecture/src/lib.rs
petertseng: Since I have no knowledge of how to best teach students something, I will simply accept this since I don't know any better. That means my only constraint is: Since this is provided, it must be placed before any other exercise with `Result` (since they won't provide it). If it were removed, there would be more freedom of placement.

exercises/collatz-conjecture/tests/collatz-conjecture.rs
petertseng: I changed my mind. Even though we are leaving `return Err("Only positive...")` in, I still say this test should only check `is_err`, since I am considering the habits it may instill, and I don't want to instill the habit of checking exact string error messages.

exercises/collatz-conjecture/example.rs
jgilray: If I remove the else, then I must use "return Err(..." in the clause above.

config.json
jgilray: Done!

jgilray:
I think I've made all the changes that @petertseng requested.  Thanks again for the review.  Next I'm going to port this exercise to the Racket track.

exercises/collatz-conjecture/example.rs
petertseng: I agree that what you say is true, but I am still thinking lines 16 and 17 are not necessary.

jgilray:
final (?) squash down to one commit

petertseng:
Thank you for implementing this exercise for our track!

jgilray:
Thanks @petertseng, I very much appreciate your assistance.  Let me know if there is anything I can do to help in the future.

PR #365, opened 2017-10-15
Changes test assertions to compare wrapped values

### LINKED ISSUE #338: Prefer to assert_eq!(a, Ok(b)) rather than assert_eq!(a.unwrap(), b)
In https://github.com/exercism/rust/pull/292#discussion_r117407845 we decided that asserting equality to an `Ok` gives better error messages than attempting to unwrap (which panics if the observed value is an Err instead of Ok).

We should try to adhere to this.

Find instances of unwrap in asserts by using `git grep 'assert.*unwrap'` (or  `git grep unwrap exercises/*/tests` for a larger hammer) and remove the unwraps by changing the other argument to an `Ok`).

nfiles:
Resolves #338 

Test assertions should compare `Option` or `Result` values directly instead of unwrapping them.

### DIFF
--- a/exercises/bowling/tests/bowling.rs
+++ b/exercises/bowling/tests/bowling.rs
@@ -72,7 +72,7 @@ fn twenty_zero_pin_rolls_scores_zero() {
         let _ = game.roll(0);
     }
 
-    assert_eq!(game.score().unwrap(), 0);
+    assert_eq!(game.score(), Ok(0));
 }
 
 #[test]
@@ -85,7 +85,7 @@ fn ten_frames_without_a_strike_or_spare() {
         let _ = game.roll(6);
     }
 
-    assert_eq!(game.score().unwrap(), 90);
+    assert_eq!(game.score(), Ok(90));
 }
 
 #[test]
@@ -100,7 +100,7 @@ fn spare_in_the_first_frame_followed_by_zeros() {
         let _ = game.roll(0);
     }
 
-    assert_eq!(game.score().unwrap(), 10);
+    assert_eq!(game.score(), Ok(10));
 }
 
 #[test]
@@ -116,7 +116,7 @@ fn points_scored_in_the_roll_after_a_spare_are_counted_twice_as_a_bonus() {
         let _ = game.roll(0);
     }
 
-    assert_eq!(game.score().unwrap(), 16);
+    assert_eq!(game.score(), Ok(16));
 }
 
 #[test]
@@ -134,7 +134,7 @@ fn consecutive_spares_each_get_a_one_roll_bonus() {
         let _ = game.roll(0);
     }
 
-    assert_eq!(game.score().unwrap(), 31);
+    assert_eq!(game.score(), Ok(31));
 }
 
 #[test]
@@ -150,7 +150,7 @@ fn if_the_last_frame_is_a_spare_you_get_one_extra_roll_that_is_scored_once() {
     let _ = game.roll(5);
     let _ = game.roll(7);
 
-    assert_eq!(game.score().unwrap(), 17);
+    assert_eq!(game.score(), Ok(17));
 }
 
 #[test]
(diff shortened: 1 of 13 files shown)


exercises/bowling/tests/bowling.rs
```diff
@@ -347,7 +347,7 @@ fn the_two_balls_after_a_final_strike_can_not_be_a_non_strike_followed_by_a_stri
 
 #[test]
 #[ignore]
-fn second_bonus_ball_after_a_final_strike_can_not_score_an_invalid_number_of_pins_even_if_first_is_strike() {
+fn second_bonus_ball_after_a_final_strike_can_not_score_an_invalid_number_of_pins_even_if_first_is_strike(){
```
petertseng: accidental removal of space?

exercises/nucleotide-count/tests/nucleotide-count.rs
```diff
@@ -9,10 +9,10 @@ fn check_dna(s: &str, pairs: &[(char, usize)]) {
     // check for the presence and value of each key in the given pairs vector.
     let mut m: HashMap<char, usize> = dna::nucleotide_counts(s).unwrap();
     for &(k, v) in pairs.iter() {
-        assert_eq!((k, m.remove(&k).unwrap()), (k, v));
+        assert_eq!((k, m.remove(&k)), (k, Some(v)));
     }
     // may fail with a message that clearly shows all extra pairs in the map
-    assert_eq!(m.iter().collect::<Vec<(&char,&usize)>>(), vec!());
+    assert_eq!(m.iter().collect::<Vec<(&char, &usize)>>(), vec!());
```
petertseng: the addition of a space is good, but isn't related to removing unwrap, so it should go in a separate commit

nfiles:
I have committed a revised changeset that reverts the spacing issues. :)

nfiles:
No problem!

> After this is merged, are there any remaining instances of unwrap that should be changed?

I think the remaining `unwrap` calls are okay, but I could certainly be wrong. I'm still pretty new to Rust. None of them are being directly used in a test assertion. If the worry is that the `unwrap` call can cause a `panic` at any point in the test (not just the assertion), it might make sense to `assert!(wrapped.is_ok())` before calling unwrap. I'm not sure if that makes sense.

`unwrap` is still called the most in `queen-attack`, `react`, and `triangle`, but none of them are within a test assertion.

1. `queen-attack` - `ChessPosition::new` is unwrapped and passed directly into `Queen::new`.
1. `react` - `reactor.create_compute` is unwrapped because it is passed into other methods to set up the test assertions.
1. `triangle` - `Triangle::build` is unwrapped before the test assertions. An `is_ok` and `is_err` result are both tested.
1. `circular-buffer` - `buffer.read().unwrap().len()` is called within the test assertion, but the unwrapped value is not being directly compared, so I thought it was okay.
1. `nucleotide-count` - `dna::nucleotide_counts(s)` is unwrapped, but the unwrapped `HashMap` is tested with a couple of different iterators, and I don't know how to replace that with a test assertion on a wrapped object.

That's a good point about the `as_slice`/`vec!` instances (`all-your-base` and `variable-length-quantity`). The `example.rs` file was my reference for those and they explicitly returned `Vec` in the optional type.

This pull request resolves #338

petertseng:
Thanks again.

When I am able, I will also check to see whether I agree with your assessment, and assuming I do then it will be time to merge it.

> This pull request resolves #338

That needs to be in the PR description (the one that currently says "Related to Issue #338") OR any commit message in the PR.

nfiles:
Sorry, I understand now. I have updated the PR description.

petertseng:
OK. I agree with all your assessments. And they are definitely not unwraps that are compared, so objective achieved. I may consider a new issue for providing better error messages when the unwraps fail, if there is demand for it (or someone else may open it, that will be sufficient demand). As for this PR, it is time to merge, and close the referenced issue.

I will be making a change to the commit messages. They should be written in imperative form to match with the messages that `git` emits (e.g. on `git merge` and `git revert`), so at a minimum the "Changes" should be "Change"

petertseng:
Thanks (yet) again, great stuff.

PR #367, opened 2017-10-18
Implement exercise isogram

hekrause:


exercises/isogram/.gitignore
```diff
@@ -0,0 +1,9 @@
+# Generated by Cargo
+# will have compiled files and executables
+/target/
+/.idea/
```
petertseng: I think .idea should not be here. It is not generated by Cargo, so the comment is inaccurate.

```
$ cargo new testproj; cat testproj/.gitignore
     Created library `testproj` project
/target/
**/*.rs.bk
Cargo.lock
```

I think things other than those three entries should not be present.

exercises/isogram/Cargo.toml
```diff
@@ -0,0 +1,3 @@
+[package]
+name = "isogram"
+version = "0.1.0"
```
petertseng: https://github.com/exercism/problem-specifications/blob/master/exercises/isogram/canonical-data.json#L6 is at 1.2.0. So, the version here should also be at 1.2.0 and use the cases listed in https://github.com/exercism/problem-specifications/blob/master/exercises/isogram/canonical-data.json , OR write in this PR a comment explaining why those cases are not appropriate for use in the Rust track.

exercises/isogram/example.rs
```diff
@@ -0,0 +1,11 @@
+pub fn check(word: &str) -> bool {
+
+    // Filter all non-Alphabetic character out and collect them in a new String
+    let normalized_string: String = word.to_lowercase().chars().filter(|c| c.is_alphabetic()).collect();
+
+    // Find the char element from back and front and compare the index
+    let closure = |x: char, word: &str| word.find(x).unwrap() == word.rfind(x).unwrap();
```
petertseng: this should be named meaningfully saying what the function does. It determines whether a letter is unique.

exercises/isogram/tests/isogram.rs
```diff
@@ -0,0 +1,62 @@
...
+#[test]
+fn only_lower_case_characters() {
+    assert_eq!(check("isogram"), true, "\"isogram\" is a isogram.")
+}
+
+#[test]
+fn one_duplicated_character() {
+    assert_eq!(check("eleven"),
+               false,
+               "\"eleven\" has duplicated \'e\' isogram.")
```
petertseng: I don't understand what `duplicated \'e\' isogram.` means. I think it should be `duplicated \'e\' so is not an isogram`, or `duplicated \'e\'`.

exercises/isogram/tests/isogram.rs
```diff
@@ -0,0 +1,62 @@
...
+    assert_eq!(check("subdermatoglyphic"),
+               true,
+               "\"subdermatoglyphic\" is a isogram.")
+}
+
+#[test]
+fn one_duplicated_character_mixed_case() {
+    assert_eq!(check("Alphabet"),
+               false,
+               "\"Alphabet\" has duplicated \'a\' isogram.")
```
petertseng: > I don't understand what `duplicated \'e\' isogram.` means. I think it should be `duplicated \'e\' so is not an isogram`, or `duplicated \'e\'`.

same

exercises/isogram/tests/isogram.rs
```diff
@@ -0,0 +1,62 @@
+extern crate isogram;
+
+use isogram::check;
+
+#[test]
+fn empty_string() {
+    assert_eq!(check(""), true, "An empty string is a isogram.")
```
petertseng: `a isogram` should be `an isogram`.

exercises/isogram/tests/isogram.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+#[test]
+fn only_lower_case_characters() {
+    assert_eq!(check("isogram"), true, "\"isogram\" is a isogram.")
```
petertseng: `a isogram` should be `an isogram`.

exercises/isogram/tests/isogram.rs
```diff
@@ -0,0 +1,62 @@
...
+    assert_eq!(check("eleven"),
+               false,
+               "\"eleven\" has duplicated \'e\' isogram.")
+}
+
+#[test]
+fn longest_reported_english_isogram() {
+    assert_eq!(check("subdermatoglyphic"),
+               true,
+               "\"subdermatoglyphic\" is a isogram.")
```
petertseng: `a isogram` should be `an isogram`.

exercises/isogram/tests/isogram.rs
```diff
@@ -0,0 +1,62 @@
...
+    assert_eq!(check("Alphabet"),
+               false,
+               "\"Alphabet\" has duplicated \'a\' isogram.")
+}
+
+#[test]
+fn hypothetical_isogramic_word_with_hyphen() {
+    assert_eq!(check("thumbscrew-japingly"),
+               true,
+               "\"thumbscrew-japingly\" is a isogram.")
```
petertseng: `a isogram` should be `an isogram`.

exercises/isogram/tests/isogram.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+#[test]
+fn isogram_with_duplicated_hyphen() {
+    assert_eq!(check("six-year-old"),
+               true,
+               "\"six-year-old\" is a isogram.")
```
petertseng: `a isogram` should be `an isogram`.

exercises/isogram/tests/isogram.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+#[test]
+fn made_up_name_that_is_an_isogram() {
+    assert_eq!(check("Emily Jung Schwartzkopf"),
+               true,
+               "\"Emily Jung Schwartzkopf\" is a isogram.")
```
petertseng: `a isogram` should be `an isogram`.

exercises/isogram/tests/isogram.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+#[test]
+fn duplicated_character_in_the_middle() {
+    assert_eq!(check("accentor"),
+               false,
+               "\"accentor\" has duplicated \'c\' isogram.")
```
petertseng: > I don't understand what `duplicated \'e\' isogram.` means. I think it should be `duplicated \'e\' so is not an isogram`, or `duplicated \'e\'`.

same

exercises/isogram/tests/isogram.rs
petertseng: can you confirm whether it is better to use `assert_eq!(something, true, message)` rather than `assert!(something, message)` ? Which is better, and why?

hekrause: I would prefer the ```assert_eq!``` because the output when the assertion fails is more informative than with the ```assert!``` macro.

With ```assert_eq!```:
grafik


With ```assert!```:
grafik

Is there an advantage i missed to prefer one over the other?

hekrause:
@petertseng 
Seems like the CI build is passed but not updated github?!
Can you rerun the CI?

config.json
```diff
@@ -818,3 +830,4 @@
     "trinary"
   ]
 }
+
```
petertseng: should not add this newline

config.json
```diff
@@ -810,6 +810,18 @@
       "uuid": "496fd79f-1678-4aa2-8110-c32c6aaf545e",
       "slug": "hexadecimal",
       "deprecated": true
+    },
+    {
+      "uuid": "357d63b9-002e-2680-561c-8aa540f6a4ea7181b64",
+      "slug": "isogram",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 3,
```
petertseng: given this exercise is of difficulty 3, I would think it should be placed between the difficulty 1 exercises and the difficulty 4 exercises, right? It doesn't seem right to have it be the last exercise of the track; someone may find it a strange drop in difficulty to finish React and then have Isogram come up next.

config/maintainers.json
```diff
@@ -47,3 +47,4 @@
     }
   ]
 }
+
```
petertseng: should not add this newline

exercises/isogram/tests/isogram.rs
petertseng: > I would prefer the `assert_eq!` because the output when the assertion fails is more informative than with the `assert!` macro.

> with `assert!`
> `panicked at '"eleven" has more than one 'e''`

It's true that it's unclear, but perhaps it could be mdae clear by changing the message to `"eleven has more than one 'e' and thus is not a pangram`.

I have decided that it doesn't matter to me what choice you make, only that I see the reasoning for it. Therefore, if I see information that may alter that reasoning, I am obligated to say it.

config.json
hekrause: Ah okay I did not realize that this was structured that way.

config.json
```diff
@@ -154,6 +154,18 @@
...
+      "topics": [
+        "chars",
+        "iterators",
+        "strings"
+      ]
+  },
```
petertseng: indentation consistency

config.json
```diff
@@ -154,6 +154,18 @@
       ]
     },
     {
+      "uuid": "357d63b9-002e-2680-561c-8aa540f6a4ea7181b64",
```
petertseng: this is too long to be a UUID, assuming that UUID is https://en.wikipedia.org/wiki/Universally_unique_identifier. The last segment should be 12 characters, not 19 as it currently is.

hekrause: I generated this uuid with the ```configlet uuid``` command. But this numbers are always that long.
Is this the right comand?

petertseng: then `configlet uuid` is generating wrong UUIDs.

petertseng: therefore, by definition, it is not the right command

petertseng: and that warrants an issue for configlet, since I presume `configlet uuid` is advertised to generate UUIDs (and thus is failing its advertisement)

exercises/isogram/tests/isogram.rs
petertseng: I don't find the `assert_eq` much more helpful because there is no indication of which of the two (true/false) are the observed/expected values, unless there is some convention going on that I'm not aware of (such as "right" being the "right"/"correct" answer and/or the explanation being next to the "right" line).

If I'm working with booleans, I think I will be able to tell without assistance that there are only two values and that my test must be failing because I'm producing the wrong value and the correct value is the only other possibility. However, maybe we would like to contend this is not as easy for a newcomer?

hekrause: I think it depends on the level of the person working at this exercise. For people which are new in rust a kind of "verbose" output with more information is more helpfull. If I could choose how somebody should implement test for me being new to rust i would prefer the ```assert_eq!```.

petertseng:
Reminder to myself to regenerate the README. Since I ran out of the allotted time, I cannot do it for some hours.

Possible inconsistencies:

* All sentences of "X is an isogram." end with period. All sentences of "X has more than one 'x'" do not end with period
* All sentences of "X has more than one 'x'" do not state the consequence of having more than one 'x', namely, that it makes X not an isogram.

I don't really care, but those could possibly be fixed if desired.

exercises/isogram/tests/isogram.rs
```diff
@@ -10,51 +10,63 @@ fn empty_string() {
 #[test]
 #[ignore]
 fn only_lower_case_characters() {
-    assert_eq!(check("isogram"), true, "An \"isogram\" should be an isogram.")
+    assert_eq!(check("isogram"),
+               true,
+               "An \"isogram\" should be an isogram.")
```
petertseng: I would say this "An" should be removed.

petertseng:
Thank you for implementing this exercise!

PR #369, opened 2017-10-23
Normalize the config file to minimize noise later

kytrinyx:
We will be scripting a number of changes to the track configuration.
This runs configlet fmt, which ensures that any further changes can be minimized.

See https://github.com/exercism/meta/issues/95

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,820 +1,818 @@
 {
-  "language": "Rust",
-  "blurb": "Rust is a compiled programming language designed for speed, concurrency, and memory safety. Rust programs can run almost anywhere, from low-power embedded devices to web servers.",
   "active": true,
+  "blurb": "Rust is a compiled programming language designed for speed, concurrency, and memory safety. Rust programs can run almost anywhere, from low-power embedded devices to web servers.",
   "exercises": [
     {
-      "uuid": "13ec1ebe-d71b-436f-ab12-25305e814171",
-      "slug": "hello-world",
       "core": false,
-      "unlocked_by": null,
       "difficulty": 1,
+      "slug": "hello-world",
       "topics": [
-        "println!"
-      ]
+        "println"
+      ],
+      "unlocked_by": null,
+      "uuid": "13ec1ebe-d71b-436f-ab12-25305e814171"
     },
     {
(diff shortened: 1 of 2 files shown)


config.json
```diff
@@ -1,820 +1,818 @@
 {
-  "language": "Rust",
```
petertseng: This seems too important to shunt to the bottom of the file. I have evidence in https://github.com/exercism/bash/commit/6fabd6f26c76e799335ee43434826f9fac4ea3a9 and https://github.com/exercism/bash/pull/91 that this caused a mistaken assumption that the `language` attribute was absent. If this should indeed move, I need to hear a way to prevent others from making the same mistake.

config.json
```diff
@@ -1,820 +1,818 @@
...
       "core": false,
-      "unlocked_by": null,
       "difficulty": 1,
+      "slug": "hello-world",
       "topics": [
-        "println!"
-      ]
+        "println"
+      ],
+      "unlocked_by": null,
```
petertseng: shouldn't it be the case that unlocked_by and core are next to each other? They seem too closely related to separate in this manner.

config.json
```diff
@@ -1,820 +1,818 @@
...
   "active": true,
+  "blurb": "Rust is a compiled programming language designed for speed, concurrency, and memory safety. Rust programs can run almost anywhere, from low-power embedded devices to web servers.",
   "exercises": [
     {
-      "uuid": "13ec1ebe-d71b-436f-ab12-25305e814171",
-      "slug": "hello-world",
       "core": false,
-      "unlocked_by": null,
       "difficulty": 1,
+      "slug": "hello-world",
```
petertseng: slug seems important. it seems like it should come first.

petertseng: slug seems important. it seems like it should come first or second (behind uuid)

kytrinyx:
There's a discussion happening in exercism/meta#95 about how this normalization degrades the readability of the configs.

That discussion has some good suggestions for improvements to the serializer.
I'd appreciate it if you would weigh in on that discussion with any suggestions for improving how we handle configs, consistency, and normalization.

Meanwhile, I'm going to close this and wait for that discussion to play out.

I'm sorry about the noise, and any extra work that I've caused you.

PR #370, opened 2017-10-25
Add more stubs

elbaro:


elbaro:
Oops, there is `example.rs` but `fetch rust` doesn't download the file.
#269 is related

petertseng:
Thanks.

> there is `example.rs` but `fetch rust` doesn't download the file.

Correct, it does not.

> Update lib.rs

I'd like to ask the commit message to be amended because "Update" is too vague. A function stub was added, so the commit message should say that.

exercises/perfect-numbers/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+pub enum Classification {
+
```
petertseng: I request that the three possible variants of `Classification` be included in the stub, since there is no point in keeping them from the student; the test expects them and the description lists them already.

The other reason is that the stub will not compile without it.

The precedent for this request is in https://github.com/exercism/rust/blob/master/exercises/forth/src/lib.rs#L7-L12 and https://github.com/exercism/rust/blob/master/exercises/robot-simulator/src/lib.rs#L4-L10.

exercises/perfect-numbers/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+pub enum Classification {
+
+}
+
+pub fn classify(num: u64) -> Result<Classification, &'static str> {
+
```
petertseng: there should be `unimplemented!()` here, otherwise the stub will not compile because the compiler thinks `()` is being returned

The reason for this request is so that `cargo test -- --no-run` can be run on this stub, for the purpose of automatically ensuring that the signature provided in the stub is correct (I am unable to manually ensure this, because of time).

exercises/pythagorean-triplet/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+pub fn find() -> Option<u32> {
+
```
petertseng: there should be `unimplemented!()` here, otherwise the stub will not compile because the compiler thinks `()` is being returned

The reason for this request is so that `cargo test -- --no-run` can be run on this stub, for the purpose of automatically ensuring that the signature provided in the stub is correct (I am unable to manually ensure this, because of time).

petertseng:
Very good, thank you! Will you do us the favour of combining the commits such that the final state is two commits:

* one commit that adds the perfect-numbers stub in its final form
* one commit that adds the pythagorean-triplet stub in its final form

We wouldn't want this repo to have intermediate commits where there are stubs that are not runnable

(Ask if you need advice on how to achieve it. One might think of various ways like an interactive rebase or reset, add, commit, add, commit)

petertseng:
Thank you for contributing these stubs!

Issue #371, opened 2017-10-25
If function stubs are provided, we'd like to prevent accidentally adding an incorrect signature


petertseng:
It was an issue in https://github.com/exercism/rust/pull/339#issuecomment-319645344.

One possible way this might be accomplished: CI would check that if function stubs are provided, they compile.

PR #339, opened 2017-08-01
Add stubs to some exercises

leodasvacas:
Add stubs to leap, raindrops, bob, beer song, proverb difference of squares and sum of multiples.

PR #372, opened 2017-10-27
travis: Enforce that stubs compile, supporting commit (add unimplemented)

petertseng:
travis: Enforce that stubs compile

Adding a stub with an incorrect signature would be confusing to
students. We will prevent that with CI.

Was desired in https://github.com/exercism/rust/pull/339#issuecomment-319645344

Closes https://github.com/exercism/rust/issues/371.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: rust
 script:
 - "./_test/check-exercises.sh"
 - "sh ./_test/ensure-lib-src-rs-exist.sh"
+- "sh ./_test/ensure-stubs-compile.sh"
 - "sh ./_test/count-ignores.sh"
 - "./bin/fetch-configlet"
 - "./bin/configlet lint ."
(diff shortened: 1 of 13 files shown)


petertseng:
I had to expand this to have more. Individual commit messages explain.

petertseng:
Sorry, but I am unable to effectively do my duties as a track maintainer (review others' PRs) without this being merged, so if I can't get a review in 24 hours I have to just test it myself and merge if it passes my own tests.

petertseng:
OK, split up the other changes into other PRs.

coriolinus:
Looks generally good to me!

I think we'll need to add a `.meta/ALLOWED_TO_NOT_COMPILE` exception to the `decimal` exercise and possibly others before merging; I don't think that we want to have this new check cause existing exercises to fail.

petertseng:
`decimal` is the only exercise that needs it.

PR #374, opened 2017-10-28
travis: Move lib/src.rs check to separate script

petertseng:
The reasons to do this:

1. Easier to run (and thus test) it separately
2. Previous scheme wasn't propagating the exit codes correctly. the
   `exitcode` is a new variable local to the subshell (as I understand),
   so a missing stub file would not actually cause an exit code of 1.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@
 language: rust
 script:
 - "./_test/check-exercises.sh"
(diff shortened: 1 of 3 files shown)


petertseng:
Sort of important for continued effectiveness, but not that bad. I'm going to need to merge it if there come many more PRs for new exercises and that I may forget, but otherwise no rush I suppose.

coriolinus:
Looks good to me.

PR #375, opened 2017-10-28
Add two-bucket test

ktomsic:
This test is listed in the problem-specifications repo ([1]), but hadn't been
added to the rust track yet. This commit adds the description, test
cases, and a sample solution based on breadth-first search.

[1] https://github.com/exercism/problem-specifications/blob/master/exercises/two-bucket

### DIFF
--- a/config.json
+++ b/config.json
@@ -647,6 +647,18 @@
       ]
     },
     {
+      "uuid": "1850fb3f-9dad-449a-90b6-9d90038cf34d",
+      "slug": "two-bucket",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 4,
+      "topics": [
+        "algorithm",
+        "loops",
+        "mathematics"
+      ]
+    },
+    {
       "uuid": "c21c379b-fb23-449b-809a-3c6ef1c31221",
       "slug": "pig-latin",
       "core": false,
--- a/exercises/two-bucket/README.md
+++ b/exercises/two-bucket/README.md
@@ -1,0 +1,68 @@
+# Two Bucket
+
+Given two buckets of different size, demonstrate how to measure an exact number of liters by strategically transferring liters of fluid between the buckets.
+
+Since this mathematical problem is fairly subject to interpretation / individual approach, the tests have been written specifically to expect one overarching solution.
+
+To help, the tests provide you with which bucket to fill first. That means, when starting with the larger bucket full, you are NOT allowed at any point to have the smaller bucket full and the larger bucket empty (aka, the opposite starting point); that would defeat the purpose of comparing both approaches!
+
+Your program will take as input:
+- the size of bucket one, passed as a numeric value
+- the size of bucket two, passed as a numeric value
+- the desired number of liters to reach, passed as a numeric value
+- which bucket to fill first, passed as a Bucket (either Bucket::One or Bucket::Two)
+
+Your program should determine:
+- the total number of "moves" it should take to reach the desired number of liters, including the first fill - expects a numeric value
+- which bucket should end up with the desired number of liters (let's say this is bucket A) - expects a Bucket (either Bucket::One or Bucket::Two)
+- how many liters are left in the other bucket (bucket B) - expects a numeric value
+
+Note: any time a change is made to either or both buckets counts as one (1) move.
+
+Example:
+Bucket one can hold up to 7 liters, and bucket two can hold up to 11 liters. Let's say bucket one, at a given step, is holding 7 liters, and bucket two is holding 8 liters (7,8). If you empty bucket one and make no change to bucket two, leaving you with 0 liters and 8 liters respectively (0,8), that counts as one "move". Instead, if you had poured from bucket one into bucket two until bucket two was full, leaving you with 4 liters in bucket one and 11 liters in bucket two (4,11), that would count as only one "move" as well.
+
+To conclude, the only valid moves are:
+- pouring from one bucket to another
+- emptying one bucket and doing nothing to the other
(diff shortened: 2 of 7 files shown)


exercises/two-bucket/src/lib.rs
```diff
@@ -0,0 +1,24 @@
...
+    pub other_bucket: u8,
+}
+
+/// Solve the bucket problem
+pub fn solve(_capacity_1: u8,
+             _capacity_2: u8,
+             _goal: u8,
+             _start_bucket: &str) -> BucketStats
+{
+    return BucketStats {
```
coriolinus: It's better form to have the body of the stub function be simply `{ unimplemented!() }`, to make it easy for people to see where they should begin work.

exercises/two-bucket/src/lib.rs
```diff
@@ -0,0 +1,24 @@
+/// A struct to hold your results in.
+#[derive(PartialEq, Eq, Debug)]
+pub struct BucketStats<'a> {
+    /// The total number of "moves" it should take to reach the desired number of liters, including
+    /// the first fill.
+    pub moves: u8,
+    /// Which bucket should end up with the desired number of liters? (Either "one" or "two")
+    pub goal_bucket: &'a str,
```
coriolinus: Let's leverage Rust's capabilities, and use a two-element `enum` to represent the final bucket. It's more type-safe, and a better example for learners.

ktomsic:
Thanks for the really quick feedback! I should have at least caught the config.json thing from the README.md, but I guess I missed the section on linting somehow. Anyway, everything should be updated now.

exercises/two-bucket/src/lib.rs
```diff
@@ -0,0 +1,26 @@
...
+    pub goal_bucket: Bucket,
+    /// How many liters are left in the other bucket?
+    pub other_bucket: u8,
+}
+
+/// Solve the bucket problem
+pub fn solve(capacity_1: u8,
+             capacity_2: u8,
+             goal: u8,
+             start_bucket: &str) -> BucketStats
```
coriolinus: I think this should be an `&Bucket` also; this prevents the need for a potential panic in your example implementation. It'll mean updating the tests, but that's fine.

exercises/two-bucket/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+Your program will take as input:
+- the size of bucket one, passed as a numeric value
+- the size of bucket two, passed as a numeric value
+- the desired number of liters to reach, passed as a numeric value
+- which bucket to fill first, passed as a String (either 'one' or 'two')
+
+Your program should determine:
+- the total number of "moves" it should take to reach the desired number of liters, including the first fill - expects a numeric value
+- which bucket should end up with the desired number of liters (let's say this is bucket A) - expects a String (either 'one' or 'two')
```
coriolinus: This should also be updated to refer to a `Bucket` object.

coriolinus:
No problem; thanks for making the requested changes quickly. 

I've marked up a few more things; it all amounts to using `Bucket`s consistently instead of using `&str` sometimes. 

Once that change is made, I'll be ready to approve this. I won't merge immediately--I want to give the other maintainers time to take a look and make their own comments, as desired--but if there's no further discussion after that, I'd merge this not later than Friday, 3 Oct.

config.json
```diff
@@ -647,6 +647,18 @@
       ]
     },
     {
+      "uuid": "e2b2654f-0f8c-1580-35d1-06a7173a61b9b66d44d",
```
petertseng: the last portion is too long (should be 12 hexadecimal digits but has 19 by my count), so this is not a valid https://en.wikipedia.org/wiki/Universally_unique_identifier (I am assuming this is what a UUID is)

exercises/two-bucket/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+Since this mathematical problem is fairly subject to interpretation / individual approach, the tests have been written specifically to expect one overarching solution.
+
+To help, the tests provide you with which bucket to fill first. That means, when starting with the larger bucket full, you are NOT allowed at any point to have the smaller bucket full and the larger bucket empty (aka, the opposite starting point); that would defeat the purpose of comparing both approaches!
+
+Your program will take as input:
+- the size of bucket one, passed as a numeric value
+- the size of bucket two, passed as a numeric value
+- the desired number of liters to reach, passed as a numeric value
+- which bucket to fill first, passed as a String (either 'one' or 'two')
```
petertseng: for the Rust track, this is no longer a string.

exercises/two-bucket/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+Your program should determine:
+- the total number of "moves" it should take to reach the desired number of liters, including the first fill - expects a numeric value
+- which bucket should end up with the desired number of liters (let's say this is bucket A) - expects a Bucket (either Bucket::One or Bucket::Two)
```
petertseng: Because this line differs from problem-specifications and cannot simply be fixed by adding HINTS.md, I dislike this aspect of problem-specifications. I am going to deal with it there. Not a blocker for this PR.

config.json
ktomsic: Sorry. I generated this with [configlet](https://github.com/exercism/configlet/), but I hadn't seen that [it might not generate valid UUIDs](https://github.com/exercism/configlet/issues/99). This should be fixed now.

exercises/two-bucket/README.md
ktomsic: Good catch. Updated it to indicate that it takes a Bucket.

PR #381, opened 2017-11-02
Add a utility to assist manually testing an exercise

coriolinus:
Instead of having to manually enable tests, copy the examples, and
clean up, this script automates all that.

### DIFF
--- a/_test/check-exercises.sh
+++ b/_test/check-exercises.sh
@@ -31,7 +31,7 @@ return_code=0
 for exercise in $files; do
    # This assumes that exercises are only one directory deep
    # and that the primary module is named the same as the directory
-   directory=$(dirname "${exercise}");
+   directory=$(dirname "${exercise}")
 
    release=""
    if [ -z "$BENCHMARK" -a -f "$directory/.meta/test-in-release-mode" ]; then
--- a/_test/check-exercises.sh
+++ b/_test/check-exercises.sh
@@ -1,73 +1,55 @@
 #!/bin/bash
 
+# test for existence and executability of the test-exercise script
+# this depends on that
+if [ ! -f "./bin/test-exercise" ]; then
+   echo "bin/test-exercise does not exist"
+   exit 1
+fi
+if [ ! -x "./bin/test-exercise" ]; then
+   echo "bin/test-exercise does not have its executable bit set"
+   exit 1
+fi
+
 # In DENYWARNINGS mode, do not set -e so that we run all tests.
 # This allows us to see all warnings.
 if [ -z "$DENYWARNINGS" ]; then
     set -e
 fi
 
+# can't benchmark with a stable compiler; to bench, use
+# $ BENCHMARK=1 rustup run nightly _test/check-exercises.sh
 if [ -n "$BENCHMARK" ]; then
     files=exercises/*/benches
 else
     files=exercises/*/tests
 fi
 
-tmp=${TMPDIR:-/tmp/}
-mkdir "${tmp}exercises"
-
-exitcode=0
-
+return_code=0
 # An exercise worth testing is defined here as any top level directory with
 # a 'tests' directory
 for exercise in $files; do
-    # This assumes that exercises are only one directory deep
-    # and that the primary module is named the same as the directory
-    directory=$(dirname "${exercise}");
+   # This assumes that exercises are only one directory deep
+   # and that the primary module is named the same as the directory
+   directory=$(dirname "${exercise}");
 
-    workdir=$(mktemp -d "${tmp}${directory}.XXXXXXXXXX")
-
-    cp -R -T $directory $workdir
-
-    # Run in subshell to change workdir without affecting current workdir.
-    (
-        cd $workdir
-
-        cp example.rs src/lib.rs
-
-        # Overwrite empty Cargo.toml if an example specific file exists
-        if [ -f Cargo-example.toml ]; then
-            cp Cargo-example.toml Cargo.toml
-        fi
-
-        # Forcibly strip all "ignore" statements from the testing files
-        for test in tests/*.rs; do
-            sed -i '/\[ignore\]/d' $test
-        done
-
-        # Run benchmarks instead of tests when enabled.
-        if [ -n "$BENCHMARK" ]; then
-            cargo bench
-        elif [ -n "$DENYWARNINGS" ]; then
-            sed -i -e '1i #![deny(warnings)]' src/lib.rs
-
-            # No-run mode so we see no test output.
-            # Quiet mode so we see no compile output
-            # (such as "Compiling"/"Downloading").
-            # Compiler errors will still be shown though.
-            # Both flags are necessary to keep things quiet.
-            cargo test --quiet --no-run
-        else
-            # Run the test and get the status
-            cargo test
-        fi
-    )
-
-    status=$?
-
-    if [ $status -ne 0 ]
-    then
-        exitcode=1
-    fi
+   if [ -n "$DENYWARNINGS" ]; then
+      # No-run mode so we see no test output.
+      # Quiet mode so we see no compile output
+      # (such as "Compiling"/"Downloading").
+      # Compiler errors will still be shown though.
+      # Both flags are necessary to keep things quiet.
+      ./bin/test-exercise $directory --quiet --no-run
+      return_code=$(($return_code | $?))
+   else
+      # Run the test and get the status
+      # We use release mode here because, while it somewhat increases
+      # the compile time for all exercises, it substantially improves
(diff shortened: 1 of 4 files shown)


bin/test-exercise
```diff
@@ -0,0 +1,28 @@
...
+
+# eliminate #[ignore] lines from tests
+for test in "tests/*.rs"; do
+    sed -i '/\[ignore\]/d' $test
+done
+
+# run tests
+cargo test
+
+# reset
```
petertseng: optionally, all this reset business may move into a `trap` so that it will be run even if the script gets Ctrl-C in the middle of the `cargo test` run

I don't require it because perhaps that will be a rare occurrence.

One can see an example at https://github.com/exercism/go/blob/master/bin/test-without-stubs#L24-L39

coriolinus:
Agree with both comments. I'll update in the morning.

coriolinus:
Delaying this because today was crazy. Will come back to it later. Integrating it into `check-exercises.sh` properly is a bigger task than it looked at first.

petertseng:
Yes, it seems to be bigger because check-exercises operates in separate modes: run the tests, compile the tests with warnings, and run benchmarks. Will have to figure that out

coriolinus:
That's interesting:

```shell
$ ./_test/check-exercises.sh
./_test/check-exercises.sh: line 41: bin/test-exercise: Permission denied
The command "./_test/check-exercises.sh" exited with 1.
```

Not sure what that means; there are no crazy permissions on `test-exercise`. Looking into it.

coriolinus:
Test fail, but it's not the fault of this PR:

```shell
$ ./_test/check-exercises.sh
The command "./_test/check-exercises.sh" exited with 0.
0.05s$ sh ./_test/ensure-lib-src-rs-exist.sh
The command "sh ./_test/ensure-lib-src-rs-exist.sh" exited with 0.
7.65s$ sh ./_test/ensure-stubs-compile.sh
error: unused variable: `puzzle`
 --> src/lib.rs:4:14
  |
4 | pub fn solve(puzzle: &str) -> Option<HashMap<char, u8>> {
  |              ^^^^^^
  |
note: lint level defined here
 --> src/lib.rs:1:9
  |
1 | #![deny(warnings)]
  |         ^^^^^^^^
  = note: #[deny(unused_variables)] implied by #[deny(warnings)]
  = note: to disable this warning, consider using `_puzzle` instead
error: unused variable: `puzzle`
 --> src/lib.rs:4:14
  |
4 | pub fn solve(puzzle: &str) -> Option<HashMap<char, u8>> {
  |              ^^^^^^
  |
note: lint level defined here
 --> src/lib.rs:1:9
  |
1 | #![deny(warnings)]
  |         ^^^^^^^^
  = note: #[deny(unused_variables)] implied by #[deny(warnings)]
  = note: to disable this warning, consider using `_puzzle` instead
To learn more, run the command again with --verbose.
alphametics's stub does not compile; please make it compile or remove all non-commented lines
decimal's stub is allowed to not compile
Exercises that don't compile:
alphametics
The command "sh ./_test/ensure-stubs-compile.sh" exited with 1.
```

We should update `alphametics`, but I believe that a different PR should handle that.

_test/check-exercises.sh
```diff
@@ -1,73 +1,56 @@
...
-    if [ $status -ne 0 ]
-    then
-        exitcode=1
-    fi
+   # This assumes that exercises are only one directory deep
+   # and that the primary module is named the same as the directory
+   directory=$(dirname "${exercise}");
+
+   if [ -n "$DENYWARNINGS" ]; then
+      sed -i -e '1i #![deny(warnings)]' "$directory/src/lib.rs"
```
petertseng: This places a `[deny(warnings)]` line in each **stub** file, which promptly gets replaced by each **example** file. Therefore, no `[deny(warnings)]` line is present in any file that gets compiled by `cargo`.

The original script placed a `[deny(warnings)]` in each **example** file instead, and I believe that being in the example file is the desired behaviour.

_test/check-exercises.sh
```diff
@@ -1,73 +1,56 @@
...
 
+# can't benchmark with a stable compiler; to bench, use
+# $ BENCHMARK=1 rustup run nightly _test/check-exercises.sh
 if [ -n "$BENCHMARK" ]; then
     files=exercises/*/benches
+    # cargo bench --release is an invalid command
+    release=""
 else
     files=exercises/*/tests
+    release="--release"
```
petertseng: I proposed this change in https://github.com/exercism/rust/pull/241 and it was rejected, for the reason:

> A reason to not use release mode: it will skip overflow checks

I support this change. Please be aware of the reasons for the previous rejection.

_test/check-exercises.sh
```diff
@@ -1,73 +1,56 @@
...
+   else
+      # Run the test and get the status
+      # We use release mode here because, while it somewhat increases
+      # the compile time for all exercises, it substantially improves
+      # the runtime for certain exercises such as alphametics.
+      # Overall this should be an improvement.
+      ./bin/test-exercise $directory $release
+   fi
+   # as the test-exercise command is the last one run in all cases,
+   # its exit code is preserved automatically
```
petertseng: true statement, but we do not have `-e` in DENYWARNINGS mode, therefore the exit code is the exit code of that of the last exercise, is it not? Thus, we would only get a nonzero exit if the very last exercise has any warnings, and all other exercises' warnings are ignored, I think?

_test/check-exercises.sh
coriolinus: I put that in to support testing of the script locally, but I'm only moderately in favor of it. It makes sense to ensure that nothing overflows in the tests, so I'll need to get rid of that.

_test/check-exercises.sh
coriolinus: Good catch; I'll need to change that.

_test/check-exercises.sh
coriolinus: I see. Just tested, and you're correct. I'll update that.

_test/check-exercises.sh
petertseng: A possible nice change to make would to test in release mode if `.meta/test-in-release-mode` is present, or something along those lines.

coriolinus:
Good idea; that will be quick to add.

_test/check-exercises.sh
```diff
@@ -1,73 +1,58 @@
...
 # This allows us to see all warnings.
 if [ -z "$DENYWARNINGS" ]; then
     set -e
 fi
 
+# can't benchmark with a stable compiler; to bench, use
+# $ BENCHMARK=1 rustup run nightly _test/check-exercises.sh
 if [ -n "$BENCHMARK" ]; then
     files=exercises/*/benches
+    # cargo bench --release is an invalid command
```
petertseng: this comment is no longer makes sense since `--release` is not being added here

_test/check-exercises.sh
```diff
@@ -1,73 +1,58 @@
...
-
-    status=$?
-
-    if [ $status -ne 0 ]
-    then
-        exitcode=1
-    fi
+   # This assumes that exercises are only one directory deep
+   # and that the primary module is named the same as the directory
+   directory=$(dirname "${exercise}");
```
petertseng: this line is not yours (you only unindented it) so I think the change should come later, but I note that the semicolon is unnecessary.

bin/test-exercise
```diff
@@ -0,0 +1,84 @@
...
+fi
+
+# If deny warnings, insert a deny warnings compiler directive in the header
+if [ -n "$DENYWARNINGS" ]; then
+   sed -i -e '1i #![deny(warnings)]' "$exercise/src/lib.rs"
+fi
+
+# eliminate #[ignore] lines from tests
+for test in "$exercise/tests/*.rs"; do
+    sed -i '/\[ignore\]/d' $test
```
petertseng: this line is also not yours (you only moved it from one file to another), so shouldn't get fixed here, but it's worth noting that on BSD seds, this gets a:

```
sed: 1: "exercises/accumulate/te ...": invalid command code e
```

Whereas it would work if a `-e` were added, or the `-i` were changed to `-i ''`.

coriolinus:
Ok, I think that squash takes us most of the way there. There's one deviation from the desired end state: removing the [extraneous comment](https://github.com/exercism/rust/pull/381#discussion_r149550275) is in the second commit instead of the first, because `git rebase -i` can squash commits but cannot move changes between commits. 

I'm going to add one more commit to this, to take care of the nitpicks regarding lines copied from the previous implementation, and then I think this will be ready to merge.

coriolinus:
We've gone all the way there! I'll merge as soon as this passes Travis.

PR #378, opened 2017-10-28
Implement exercise isbn-verifier

ktomsic:

PR #368, opened 2017-10-21
Updates all references from first to second edition of Rust book

nfiles:
References #356 (not sure if it closes it or not)

The only remaining references to the first edition are in these problems:

- decimal
  - [ ] derive traits: topic isn't written yet (Appendix C)
- parallel letter frequency
  - [ ] benchmark tests: I can't find any reference to this topic in the second edition
  - [ ] nightly rust: topic isn't written yet (Appendix D)

For reference, this is the search string I used to find the references: `doc.rust-lang.org`

And this is the script I used to update the most common first-edition references:

```sh
#! /bin/sh

# reference in the content
sed \
    -r 's|\[Crates and Modules\]\(https://doc.rust-lang.org/stable/book/crates-and-modules.html\)|\[Modules\]\(https://doc.rust-lang.org/book/second-edition/ch07-00-modules.html\)|' \
    -i exercises/*/README.md

# footnotes
sed \
    -r 's|\[crates-and-modules\]: http://doc.rust-lang.org/stable/book/crates-and-modules.html|\[modules\]: https://doc.rust-lang.org/book/second-edition/ch07-00-modules.html\n\[crates\]: https://doc.rust-lang.org/book/second-edition/ch14-00-more-about-cargo.html|' \
    -i exercises/*/README.md

```

PR #382, opened 2017-11-04
isogram, two-bucket: use second-edition links

petertseng:
We forgot to apply https://github.com/exercism/rust/pull/368 to
https://github.com/exercism/rust/pull/378 and
https://github.com/exercism/rust/pull/375.

### DIFF
--- a/exercises/isogram/README.md
+++ b/exercises/isogram/README.md
@@ -32,7 +32,7 @@ to pass again.  The test file is located in the `tests` directory.   You can
 also remove the ignore flag from all the tests to get them to run all at once
 if you wish.
 
-Make sure to read the [Crates and Modules](https://doc.rust-lang.org/stable/book/crates-and-modules.html) chapter if you
+Make sure to read the [Modules](https://doc.rust-lang.org/book/second-edition/ch07-00-modules.html) chapter if you
 haven't already, it will help you with organizing your files.
 
 ## Feedback, Issues, Pull Requests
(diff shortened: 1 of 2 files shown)


exercises/isogram/README.md
```diff
@@ -42,7 +42,8 @@ The [exercism/rust](https://github.com/exercism/rust) repository on GitHub is th
 If you want to know more about Exercism, take a look at the [contribution guide](https://github.com/exercism/docs/blob/master/contributing-to-language-tracks/README.md).
 
 [help-page]: http://exercism.io/languages/rust
-[crates-and-modules]: http://doc.rust-lang.org/stable/book/crates-and-modules.html
+[modules]: https://doc.rust-lang.org/book/second-edition/ch07-00-modules.html
```
coriolinus: It looks like the links are expressed both in footnote form and inline. I'm pretty sure this is redundant: if we have an inline link of the form `[link_text](https://...)`, we don't need to include a footnote with the link's value. Contrariwise, if we want to use the footnote style, then the place where the link is actually used should look like `[Modules][modules]`.

exercises/isogram/README.md
```diff
@@ -42,7 +42,8 @@ The [exercism/rust](https://github.com/exercism/rust) repository on GitHub is th
...
+[cargo]: https://doc.rust-lang.org/book/second-edition/ch14-00-more-about-cargo.html
```
coriolinus: This footnote isn't referenced at all, and is [stripped out of the rendered document](https://github.com/petertseng/exercism-rust/blob/ae6f16d2d09434816b027dd5051b086cc4114863/exercises/isogram/README.md).

petertseng:
Correct on both counts, but not applicable to this PR.

coriolinus:
Just saw that these changes were applied by a script. Fair enough; there's no reason to add manual work to generated output. I withdraw my objection.

petertseng:
However, it is a true statement that those can be removed. I would simply remove them. Only question is whether https://doc.rust-lang.org/book/second-edition/ch14-00-more-about-cargo.html is needed. I took a < 5 second glance and made a snap decision of "no", but obviously a more reasoned decision should take priority over the five second decision.

PR #386, opened 2017-11-05
Add CI test to ensure that README.md files are present and correct

coriolinus:
Checking readmes takes up non-trivial maintainer time, and is
prone to error. This script adds a check to the CI which ensures
that PRs are only accepted if their README.md files are
present and correct.

The intent and method come from the discussion which begins [here](https://github.com/exercism/rust/pull/354#issuecomment-341902986). 

It begins by isolating the exercises already impacted by the PR,
so that upstream changes to unrelated exercises don't cause undue
difficulty. It then simply ensures that the README.md within
the PR is identical to the one generated by `configlet generate`.

Outputs lists of exercises with missing or incorrect READMEs,
so people for whom CI fails due to this test not passing should
have an easy time understanding what went wrong and how to fix it.

### DIFF
--- a/_test/ensure-readmes-are-updated.sh
+++ b/_test/ensure-readmes-are-updated.sh
@@ -4,13 +4,10 @@ if [ ! -x bin/configlet ]; then
    exit 1
 fi
 
-<<<<<<< HEAD
 if [ ! -d "problem-specifications" ]; then
    git clone git@github.com:exercism/problem-specifications.git problem-specifications
 fi
 
-=======
->>>>>>> 28d4a17e7c424b47d9c6649c05cdd127c1a2ef59
 newline=$'\n  '
 
 missing_readmes=""
@@ -23,11 +20,7 @@ for exercise in $(git diff --name-only master..$(git rev-parse --abbrev-ref HEAD
    else
       existing_readme_checksum=$(md5sum $readme_path | cut -d' ' -f1)
       # generate the new README
-<<<<<<< HEAD
       bin/configlet generate . --only "$exercise" --spec-path "problem-specifications"
-=======
-      bin/configlet generate . --only "$exercise"
->>>>>>> 28d4a17e7c424b47d9c6649c05cdd127c1a2ef59
       generated_readme_checksum=$(md5sum $readme_path | cut -d' ' -f1)
 
       if [ $existing_readme_checksum != $generated_readme_checksum ]; then
--- a/_test/ensure-readmes-are-updated.sh
+++ b/_test/ensure-readmes-are-updated.sh
@@ -5,7 +5,7 @@ if [ ! -x bin/configlet ]; then
 fi
 
 if [ ! -d "problem-specifications" ]; then
-   git clone git@github.com:exercism/problem-specifications.git problem-specifications
+   git clone https://github.com/exercism/problem-specifications.git problem-specifications
 fi
 
 newline=$'\n  '
--- a/_test/ensure-readmes-are-updated.sh
+++ b/_test/ensure-readmes-are-updated.sh
@@ -4,6 +4,10 @@ if [ ! -x bin/configlet ]; then
    exit 1
 fi
 
+if [ ! -d "problem-specifications" ]; then
+   git clone git@github.com:exercism/problem-specifications.git problem-specifications
+fi
+
 newline=$'\n  '
 
 missing_readmes=""
@@ -16,7 +20,7 @@ for exercise in $(git diff --name-only master..$(git rev-parse --abbrev-ref HEAD
    else
       existing_readme_checksum=$(md5sum $readme_path | cut -d' ' -f1)
       # generate the new README
-      bin/configlet generate . --only "$exercise"
+      bin/configlet generate . --only "$exercise" --spec-path "problem-specifications"
       generated_readme_checksum=$(md5sum $readme_path | cut -d' ' -f1)
 
       if [ $existing_readme_checksum != $generated_readme_checksum ]; then
--- a/_test/ensure-readmes-are-updated.sh
+++ b/_test/ensure-readmes-are-updated.sh
@@ -1,0 +1,36 @@
+if [ ! -x bin/configlet ]; then
+   echo "Improper configuration; configlet should exist in bin/ when this script is run"
+   echo "Ping a Rust track maintainer to fix this"
+   exit 1
+fi
+
+newline=$'\n  '
+
+missing_readmes=""
+wrong_readmes=""
+for exercise in $(git diff --name-only master..$(git rev-parse --abbrev-ref HEAD) | grep exercises/ | cut -d'/' -f2 -s | sort -fu); do
+   echo "Checking readme for $exercise"
+   readme_path="exercises/${exercise}/README.md"
+   if [ ! -f $readme_path ]; then
+      missing_readmes="$missing_readmes$newline$exercise"
+   else
+      existing_readme_checksum=$(md5sum $readme_path | cut -d' ' -f1)
+      # generate the new README
+      bin/configlet generate . --only "$exercise"
+      generated_readme_checksum=$(md5sum $readme_path | cut -d' ' -f1)
+
+      if [ $existing_readme_checksum != $generated_readme_checksum ]; then
+         wrong_readmes="$wrong_readmes$newline$exercise"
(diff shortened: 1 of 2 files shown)


_test/ensure-readmes-are-updated.sh
```diff
@@ -0,0 +1,36 @@
...
+   readme_path="exercises/${exercise}/README.md"
+   if [ ! -f $readme_path ]; then
+      missing_readmes="$missing_readmes$newline$exercise"
+   else
+      existing_readme_checksum=$(md5sum $readme_path | cut -d' ' -f1)
+      # generate the new README
+      bin/configlet generate . --only "$exercise"
+      generated_readme_checksum=$(md5sum $readme_path | cut -d' ' -f1)
+
+      if [ $existing_readme_checksum != $generated_readme_checksum ]; then
```
petertseng: optional: I imagine you can check the exit code of `diff` rather than comparing the checksums against one another, if you find that easier.

petertseng:
Now is the time to mention this.

If ever we come across the situation:

* The README in the Rust track should differ from the current problem-specifications description
* The change is unsuitable for inclusion in problem-specifications, perhaps because it is language-specific
* The change cannot be expressed by using `.meta/hints.md` to append content to the problem-specifications description
* We don't want to maintain a separate `.meta/description.md` because the change is small and we don't want to force ourselves to manually follow every single problem-specifications change from then on

We may wish for an alternative way to express we want a different README. I was thinking a solution based on a `.patch` file. Solution is not fully formed yet because it has not become necessary. Kind of hoping it never does, but you never know.

petertseng:
EBWODP

This should go without saying, but for consistency, since I gave a rationale for every other bullet point:

> The README in the Rust track should differ from the current problem-specifications description

As it needed to in https://github.com/exercism/rust/pull/375.

> The change cannot be expressed by using .meta/hints.md to append content to the problem-specifications description

Because it is either a deletion of a line, an insert into the middle of the document, or an edit of an existing line (this third one can be expressed as a composition of the former two)

coriolinus:
> I advise that before merging this PR you should test that the build does fail when expected to

I'm not sure how to test this on Travis without actually marging it. I've performed local tests by adding a do-not-merge branch with missing or incorrect READMEs; here's a transcript of redoing that:

```shell
coriolinus@robocomp:~/projects/exercism.io/xrust$ git checkout -b do-not-merge-test-ensure-readmes-are-updated
Switched to a new branch 'do-not-merge-test-ensure-readmes-are-updated'
do-not-merge-test-ensure-readmes-are-updated
coriolinus@robocomp:~/projects/exercism.io/xrust$ git rm exercises/decimal/README.md
rm 'exercises/decimal/README.md'
do-not-merge-test-ensure-readmes-are-updated│#1
coriolinus@robocomp:~/projects/exercism.io/xrust$ git rm exercises/all-your-base/README.md
rm 'exercises/all-your-base/README.md'
do-not-merge-test-ensure-readmes-are-updated│#2
coriolinus@robocomp:~/projects/exercism.io/xrust$ echo "bogus line" >> exercises/crypto-square/README.md
do-not-merge-test-ensure-readmes-are-updated│~1#2
coriolinus@robocomp:~/projects/exercism.io/xrust$ echo "another bogus line" >> exercises/hello-world//README.md
do-not-merge-test-ensure-readmes-are-updated│~2#2
coriolinus@robocomp:~/projects/exercism.io/xrust$ git add exercises/
do-not-merge-test-ensure-readmes-are-updated│#4
coriolinus@robocomp:~/projects/exercism.io/xrust$ git commit -m "state needs to be committed for the script to work"
[do-not-merge-test-ensure-readmes-are-updated 6d6fc27] state needs to be committed for the script to work
 4 files changed, 2 insertions(+), 128 deletions(-)
 delete mode 100644 exercises/all-your-base/README.md
 delete mode 100644 exercises/decimal/README.md
do-not-merge-test-ensure-readmes-are-updated
coriolinus@robocomp:~/projects/exercism.io/xrust$ _test/ensure-readmes-are-updated.sh
Checking readme for all-your-base
Checking readme for crypto-square
Checking readme for decimal
Checking readme for hello-world
Exercises missing README.md:
  all-your-base
  decimal
Exercises with out-of-date README.md:
  crypto-square
  hello-world
do-not-merge-test-ensure-readmes-are-updated│~2
coriolinus@robocomp:~/projects/exercism.io/xrust$ echo $?
1
```

_test/ensure-readmes-are-updated.sh
coriolinus: We don't actually have both files in place at the same time; the configlet command to regenerate it overwrites the existing README.md, if any. For this reason, I thought it easier to just keep the md5sum in memory and compare that.

coriolinus:
> If ever we come across the situation: ...

Agreed that there should probably exist some way to handle that. A `.patch` file may work, but I think it's out of scope for this PR; that is more a problem for configlet.

coriolinus:
Here's an example of the existing script correctly accepting a valid README:

```shell
coriolinus@robocomp:~/projects/exercism.io/xrust$ touch exercises/decimal/tests/more_tests.rs
ensure-readmes-are-updated│+1
coriolinus@robocomp:~/projects/exercism.io/xrust$ git checkout -b do-not-merge-test-ensure-readmes-are-updated-success
Switched to a new branch 'do-not-merge-test-ensure-readmes-are-updated-success'
do-not-merge-test-ensure-readmes-are-updated-success│+1
coriolinus@robocomp:~/projects/exercism.io/xrust$ git add exercises/ && git commit -m "ensure that exercises with modifications but correct READMES pass"
[do-not-merge-test-ensure-readmes-are-updated-success 348f527] ensure that exercises with modifications but correct READMES pass
 1 file changed, 0 insertions(+), 0 deletions(-)
 create mode 100644 exercises/decimal/tests/more_tests.rs
do-not-merge-test-ensure-readmes-are-updated-success
coriolinus@robocomp:~/projects/exercism.io/xrust$ _test/ensure-readmes-are-updated.sh
Checking readme for decimal
do-not-merge-test-ensure-readmes-are-updated-success
coriolinus@robocomp:~/projects/exercism.io/xrust$ echo $?
0
```

coriolinus:
> I suspect that problem-specifications needs to be present.

I'm fairly confident that it does not. Reasoning: I have successfully used configlet to generate READMEs without having a local clone of problem-specifications anywhere on my machine. It makes more sense for it to pull directly from a canonical Github location for each exercise anyway, IMO, because it can be sure that Github will be up-to-date; the same cannot be said of a local clone.

Still, I'm not actually 100% confident of this. Perhaps someone familiar with configlet internals could chime in with an answer?

petertseng:
> I'm not sure how to test this on Travis without actually merging it. I've performed local tests by adding a do-not-merge branch with missing or incorrect READMEs

That's good, then you can either put the commit(s) on this PR, ensure that Travis fails to build then, then revert them, or put them either on a separate branch on your fork (configure Travis to build your fork) or make a PR for it.

> I have successfully used configlet to generate READMEs without having a local clone of problem-specifications anywhere on my machine.

whoaaaaaaaaaaa

I have no idea how that can work at all, since according to https://github.com/exercism/configlet/blob/master/cmd/generate.go#L55-L58 if it does not exist, it is an error.

Unfortunately I cannot verify that myself since I am not able to run configlet myself, so I will have to take the word of you + Travis for that.

coriolinus:
>> I have successfully used configlet to generate READMEs without having a local clone of problem-specifications anywhere on my machine.
> whoaaaaaaaaaaa

Looks like I was wrong: I certainly _thought_ I'd done this before ever becoming a maintainer, for the second exercise I submitted, but the command did fail in Travis. I'm going to add a section which clones the problem-specifications repo within the script; it seems the most direct path to solving this problem.

coriolinus:
#387 now fails the CI [in the expected manner](https://github.com/exercism/rust/pull/387#issuecomment-342318016), so I've closed it. It now serves as documentation that this PR works as expected.

coriolinus:
That's caused by differences between `bash` and `sh`. It's not ideal, but it has no functional effect, so I'm not bothered by it. The solution would be to have Travis run this script via `bash` instead of `sh`, which would be inconsistent with the rest of the test scripts. 

I'll wait a day or two in case there are more comments, then merge this.

PR #388, opened 2017-11-07
Remove unused variable from Alphametics stub

coriolinus:
Travis fails in general right now, because we have an unused variable in the Alphametics stub. This is causing #381 to fail right now, and possibly others.

This simply removes the unused name from the Alphametics stub. This changes no interfaces or tests; students may rename the variable to anything they prefer. It simply ensures that the `_test/ensure-stubs-compile.sh` test actually passes.

### DIFF
--- a/exercises/alphametics/src/lib.rs
+++ b/exercises/alphametics/src/lib.rs
@@ -1,5 +1,5 @@
 use std::collections::HashMap;
 
-pub fn solve(puzzle: &str) -> Option<HashMap<char, u8>> {
+pub fn solve(_: &str) -> Option<HashMap<char, u8>> {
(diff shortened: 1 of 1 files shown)


petertseng:
I support this change.

I strongly believe this is not the reason #381 fails, if #381 preserved existing behaviour.

Stubs are never compiled with `[deny(warnings)]` in current tests. If #381 fails, then #381 did not preserve that behaviour and instead represents a change in behaviour.

coriolinus:
I need to update #381. Your comment about the deny warnings insertion
script line makes sense, which means that I did introduce an unintentional
change in behavior there.

PR #389, opened 2017-11-07
Init exercise script

### LINKED ISSUE #280: Test generation suite?

coriolinus:
Add a script which generates an exercise stub. Closes #280.

- Generates all required files and formatting
- Generates README.md 
- Generates test cases if the exercise has a properly configured `canonical-data.json`
- Interactively configures `config.json`

This should make it easier for first-time contributors to jump in, hopefully.


bin/init_exercise.py
```diff
@@ -0,0 +1,411 @@
...
+Why a Python script in the Rust track repo? Distribution.
+A rust program would either need to be precompiled for various
+platforms, and available for download (in which case it wouldn't
+conveniently work in the repository), or would need to be included
+as a sub-crate and compiled locally. A python script can simply
+be a single file and depend on the user's system Python, if desired.
+
+This module requires Python3.5 or newer
+"""
+from __future__ import print_function
```
petertseng: My understanding is that this will have no effect on Python 3, but will allow this script to run and print its error messages on Python 2, when it instead would not be able to.

If this guess was correct, then there is no need to respond to it.

bin/init_exercise.py
```diff
@@ -0,0 +1,411 @@
...
+    from contextlib import contextmanager
+    from datetime import datetime
+    from uuid import uuid4
+except ImportError:
+    print("This script requires Python 3.5 or higher", file=sys.stderr)
+    # exiting like this isn't great for library use, but at least it's a quick fail
+    sys.exit(1)
+
+# check version info
+if sys.version_info[0] != 3 or sys.version_info[1] < 5:
```
petertseng: I would voice some concern here about "But what if Python 4 comes out someday and this script works on it without modification!" but I guess this line can be dealt with at that point in the far future, if it still exists when Python 4 comes.

bin/init_exercise.py
```diff
@@ -0,0 +1,411 @@
...
+    "Return the stdout of the given command"
+    sp = subprocess.run(shlex.split(cmd),
+                        stdout=subprocess.PIPE,
+                        universal_newlines=True)
+    if check_returncode:
+        sp.check_returncode()
+    return sp.stdout.strip()
+
+
+REPO_ROOT = output_of('git rev-parse --show-toplevel')
```
petertseng: Potential worry if the script is run when CWD is outside of the rust repo (including possibly in any other repo)?

bin/init_exercise.py
```diff
@@ -0,0 +1,411 @@
...
+    "Make a new exercise with the specified name"
+    with inside(EXERCISES):
+        if os.path.exists(name):
+            print("{} already exists; aborting".format(name), file=sys.stderr)
+            sys.exit(1)
+        subprocess.run(['cargo', 'new', name])
+    exercise_dir = os.path.join(EXERCISES, name)
+    # blank out the default lib.rs
+    with inside(exercise_dir):
+        with open(os.path.join('src', 'lib.rs'), 'w') as lib_rs:
```
petertseng: `truncate` would also be acceptable.

bin/init_exercise.py
```diff
@@ -0,0 +1,411 @@
...
+            lib_rs.write('')
+        os.mkdir('tests')
+        with inside('tests'):
+            with open('{}.rs'.format(name), 'w') as tests_rs:
+                print("use {}::*;".format(name), file=tests_rs)
+                print(file=tests_rs)
+
+    cd = get_problem_specification(name)
+    if cd is None:
+        print("No problem specification for {} found".format(name))
```
petertseng: ~~someone running this might think "oh no! does that mean it failed and did nothing of use?!" so I would suggest to make it clear that it did in fact do something after not finding the specification.~~

bin/init_exercise.py
```diff
@@ -0,0 +1,411 @@
+#!/usr/bin/env python
+"""
+Script to initialize an exercise for Exercism's Rust track
+
+Why a Python script in the Rust track repo? Distribution.
+A rust program would either need to be precompiled for various
+platforms, and available for download (in which case it wouldn't
+conveniently work in the repository), or would need to be included
+as a sub-crate and compiled locally. A python script can simply
```
petertseng: I would not think it would be so bad to include this code as a crate... except for the fact that I would for personal reasons have chosen a language other than Rust to write this code in. Perhaps the fact that you have written this code is also evidence that you also have similar reasons.

The only potential snag is that perhaps it may confuse contributors to this repository, but as I noted in https://github.com/exercism/rust/pull/288#issuecomment-301280919 we already have many scripts in a non-Rust language (shell scripts!), so we do not necessarily make the world any worse by adding a different non-Rust language.

bin/init_exercise.py
```diff
@@ -0,0 +1,411 @@
...
+    return "https://raw.githubusercontent.com/exercism/problem-specifications/master/exercises/{name}/{file}".format(
+        name=name, file=file,
+    )
+
+
+def get_problem_specification(name):
+    """
+    Try to get problem specifications for the exercise of the given name.
+
+    If the problem specifications repo doesn't exist, returns None.
```
petertseng: I'm not convinced the condition is true. I think it's "if the specification does not exist in the problem specifications repo", rather than whether the repo doesn't exist

bin/init_exercise.py
```diff
@@ -0,0 +1,411 @@
...
+            with open('{}.rs'.format(name), 'w') as tests_rs:
+                print("use {}::*;".format(name), file=tests_rs)
+                print(file=tests_rs)
+
+    cd = get_problem_specification(name)
+    if cd is None:
+        print("No problem specification for {} found".format(name))
+        make_new_exercise(name, exercise_dir)
+        with inside(exercise_dir):
+            generate_readme(name, False)
```
petertseng: I had to look for definition of generate_readme to understand what the boolean argument is. I would suggest giving it a name.

bin/init_exercise.py
```diff
@@ -0,0 +1,411 @@
...
+    else:
+        make_exercise_with_specifications(name, exercise_dir, cd)
+        with inside(exercise_dir):
+            generate_readme(name, True)
```
petertseng: > I had to look for definition of generate_readme to understand what the boolean argument is. I would suggest giving it a name.

same

bin/init_exercise.py
petertseng: > ~~someone running this might think "oh no! does that mean it failed and did nothing of use?!" so I would suggest to make it clear that it did in fact do something after not finding the specification.~~

I revoke my objection. It is handled, but in the respective functions (make_new_exercise prints "Creating new exercise from scratch")

bin/init_exercise.py
```diff
@@ -0,0 +1,411 @@
...
+    with inside(exercise_dir):
+        os.mkdir('.meta')
+        with inside('.meta'):
+            with open('description.md', 'w') as description:
+                print("Describe your exercise here", file=description)
+                print(file=description)
+                print("Don't forget that `README.md` is automatically generated; update this within `.meta\description.md`.", file=description)
+            with open('metadata.yml', 'w') as metadata:
+                print("---", file=metadata)
+                print("blurb: \"{} (created {})\"".format(
```
petertseng: Putting the "created" portion here may encourage contributors to leave it there.

I think we do not want to encourage that, because it doesn't seem to be relevant to http://exercism.io/languages/rust/exercises when the exercise was created. I would remove it.

bin/init_exercise.py
```diff
@@ -0,0 +1,411 @@
...
+                    name, datetime.now().isoformat()), file=metadata)
+                print("source: \"\"", file=metadata)
+                print("source_url: \"\"", file=metadata)
+        with inside('tests'):
+            with open('{}.rs'.format(name), 'a') as tests_rs:
```
petertseng: guess: `a` in case the runner of this script had already created some tests, and we do not want this script to overwrite them.

If this guess was correct, then there is no need to respond to it.

bin/init_exercise.py
```diff
@@ -0,0 +1,411 @@
...
+                print("#[ignore]", file=tests_rs)
+            print("fn test_{}() {".format(to_item_name(case['description'])), file=tests_rs)
+            print("    process_case({}, {});".format(literal(case['input']), literal(case['expected'])),
+                  file=tests_rs)
+            print("}", file=tests_rs)
+            print(file=tests_rs)
+
+        def write_cases(cases):
+            for item in cases:
+                if all(key in item for key in ('description', 'input', 'expected')):
```
petertseng: Note that the input isn't necessarily in `input`, but it would be difficult to come up with something that works for all exercises. Some tracks have created ideas and you may find some of them in https://github.com/exercism/discussions/issues/155#issue-233471715 . I wouldn't mind enforcing in the problem-specifications schema that all inputs must appear under `input` (possibly as an object that contains multiple keys, since some problems have multiple inputs)

bin/init_exercise.py
```diff
@@ -0,0 +1,411 @@
...
+                'generate', '.',
+                '--only', exercise_name,
+            ])
+
+
+if __name__ == '__main__':
+    import argparse
+    parser = argparse.ArgumentParser(description='Create a Rust Track exercise for Exercism')
+    parser.add_argument('name', help='name of the exercise to create')
+    parser.add_argument('--dont-create-exercise', action='store_true',
```
petertseng: I am used to seeing `--create-exercise` and `--no-create-exercise`, but I don't care and it might not be the case that this is an accepted convention.

bin/init_exercise.py
petertseng: > I wouldn't mind enforcing in the problem-specifications schema that all inputs must appear under input (possibly as an object that contains multiple keys, since some problems have multiple inputs)

but obviously that is far out of scope for this PR.

bin/init_exercise.py
coriolinus: Agreed. My take on the utility of including this script is that it's a tool which helps contributors, but it is in no way required, so adding it can't make the world worse for a new contributor if they don't have the ability to run it.

And yes, when initially choosing the language, the fact that Python is (for me, for projects of this size) simply faster to code in was a strong incentive. I could write the script in Python in a day; if I'd started it in Rust, it wouldn't yet be ready.

bin/init_exercise.py
coriolinus: I thought that through when writing the line. My reasoning: Python3 included breaking changes from Python2. If there is ever a Python4 in the future, we can expect it to include some changes which break Python3 scripts anyway. This line simply ensures that, for this script at least, someone takes a look at exactly what the breaking changes are and whether things are affected before someone can run this and have things explode in their face.

bin/init_exercise.py
coriolinus: This script will create a Rust exercise in the `exercises` directory of whatever repo the CWD is currently in when the script is run. If the CWD is not in a git repo, the script breaks. If the CWD is in a repo without an `exercises` directory, the script breaks. No attempts are made to do anything smart in those cases; it just prints the traceback and exits. 

In practice, I can't see anyone ever actually running this from within a non-Rust repo; I think the worry is more theoretical than practical. The bits which insert the entry into `config.json` are interactive anway, meaning that if someone runs this as a script they'll use the `--dont-update-config` flag, which means they can just `cp` or `mv` or `rm` the generated directory appropriately.

bin/init_exercise.py
coriolinus: You're right, that's a better statement. Both are true, however.

bin/init_exercise.py
coriolinus: No; if the exercise directory already exists, `make_exercise()` aborts.

The `'a'` is because `make_exercise()` pre-populates the file with `use {name}::*;`.

bin/init_exercise.py
coriolinus: I'm aware that some specifications include their inputs under other keys, but I think the best way forward for scripts like this is to enforce that all input is in `input` (possibly as an object) and all expected results are in `expected`.  As you say, that's an issue out of scope for this PR.

The closest I've seen to a schema for the `canonical-data.json` format is in the [problem-specifications README](https://github.com/exercism/problem-specifications/blob/master/README.md), and that uses `input` and `expected`, so that's all I feel the need to support for now.

bin/init_exercise.py
coriolinus: I haven't seen any other tools of this type; this was all ab nihilo. 

I feel that a verb in the argument is more expressive than simple negation.

bin/init_exercise.py
coriolinus: You're right that more often than not, `input` isn't actually used, which poses a problem. I may introduce code to do something like what vimscript does, now that you've pointed out the issue.

coriolinus:
Made some improvements. Turns out the [`pov`](https://github.com/exercism/problem-specifications/tree/master/exercises/pov) exercise makes for a really good test case: it's not already in Rust, it has several sections with their own comments, and the cases involve deep nesting and internal comments and descriptions.

Ironically, it's one of the exercises which would probably require the most work to actually bring into the Rust track, even with this script doing much of the gruntwork, just because Rust makes you work harder to build recursive structures.

bin/init_exercise.py
```diff
@@ -0,0 +1,583 @@
...
+            sys.exit(1)
+        subprocess.run(['cargo', 'new', name])
+    exercise_dir = os.path.join(EXERCISES, name)
+    # blank out the default lib.rs
+    with inside(exercise_dir):
+        with open('.gitignore', 'w') as gitignore:
+            print("Cargo.lock  # We're building a library, not a binary", file=gitignore)
+            print("            # http://doc.crates.io/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries", file=gitignore)
+        with open(os.path.join('src', 'lib.rs'), 'w') as lib_rs:
+            lib_rs.truncate()
```
petertseng: excuse me that this is the only comment I give right now instead of a full review, as my allotted time is up:

I meant https://docs.python.org/3/library/os.html#os.truncate such that the file no longer needs to be opened

But I'm not sure that actually makes a difference, so I actually don't care.

coriolinus: Huh, ok. It still works, so I suppose it's just a case where there are a few valid paths to accomplish a goal in the standard library.

bin/init_exercise.py
petertseng: One might consider a solution that is not sensitive to where the CWD is: the script determines its own location, and acts on the config.json and exercises directories at the proper relative location to itself.

then we run into problems if the person running the script has moved it before running it.

I will say that I am far more likely to run this with CWD being outside of the repo than I am to move this script.

coriolinus: Yeah, I'm not too worried about people moving the script. 

However, something like `os.chdir(os.path.dirname(__file__))` before this line would probably fix both issues. I'll look into it.

coriolinus: Just pushed a version which does this; local testing confirms that the script now works correctly when CWD is outside the repo.

coriolinus:
Went through the code one last time, and added a few bugfixes and a version bump. I think that once Travis signs off on this, it'll be ready to merge.

PR #390, opened 2017-11-09
Add exercise book-store

coriolinus:
This was a fun challenge to implement in Rust, but it also serves
as a useful demo of the code generation implemented in #389:
past implementing the example, I had to edit five lines of code
in tests/book-store.rs and three lines to add the stub in src/lib.rs.
Everything else was generated automatically.

Comments are invited pertaining both to this exercise and to the output
of the exercise generator.


coriolinus:
Looks like the generator made a bunch of edits to `config.json` beyond the insert which we actually care about. I'll take those out by hand, for this exercise; they're beyond the desired scope. However, I'm going to add them into #389; due to the fact that the generator parses and re-emits `config.json`, it will always inescapably act as a JSON prettifier, and we don't want to force people to manually take out the prettying edits that it makes.

exercises/book-store/example.rs
```diff
@@ -0,0 +1,189 @@
...
+                    book_groups[idx].0.borrow_mut().insert(*book);
+                    continue 'nextbook;
+                }
+            }
+            // if we're here, we still haven't found a place for the book.
+            // better add it to a new group
+            book_groups.push(Group::new_containing(*book));
+        }
+        book_groups.sort();
+        println!("DecomposeGroups::new() --> {:?}", book_groups);
```
petertseng: still need this?

exercises/book-store/tests/book-store.rs
```diff
@@ -0,0 +1,135 @@
...
+
+extern crate book_store;
+use book_store::*;
+
+/// Process a single test case for the property `total`
+///
+/// All cases for the `total` property are implemented
+/// in terms of this function.
+///
+/// Note that you'll need to both name the expected transform which
```
petertseng: Should it be the case that the student still sees this? They may think the "you" refers to them.

exercises/book-store/tests/book-store.rs
```diff
@@ -0,0 +1,135 @@
...
+/// the student needs to write, and name the types of the inputs and outputs.
+/// While rustc _may_ be able to handle things properly given a working example,
+/// students will face confusing errors if the `I` and `O` types are not concrete.
```
petertseng: Should it be the case that the student still sees this? They may be confused because I and O are nowhere in sight here.

exercises/book-store/tests/book-store.rs
```diff
@@ -0,0 +1,135 @@
...
+///
+/// Expected input format: ('basket', 'targetgrouping')
```
petertseng: Should it be the case that the student still sees this? They may be confused because basket and targetgrouping are nowhere in sight here

exercises/book-store/tests/book-store.rs
```diff
@@ -0,0 +1,135 @@
...
+fn process_total_case(input: (Vec<usize>, Vec<Vec<usize>>), expected: f64) {
```
petertseng: I note that `input.1` is never used, but I'm not particularly concerned. Might make one wonder "why not just *not* pass it into this function, then?"

exercises/book-store/example.rs
coriolinus: No, but it does get captured by the test runner, so I didn't bother to take it out. That said, I don't object to removing it.

exercises/book-store/tests/book-store.rs
coriolinus: I'll take out the comments aimed at the exercise implementor.

exercises/book-store/tests/book-store.rs
coriolinus: This is a note to the exercise implementor, but in this case I think we should leave it for clarity for those students who examine the tests.

On test generation, the generator collects the cases. If the case doesn't already have an `input` property, it generates one as a tuple containing all the unknown properties in the case. However, this discards information: as properties, these items have names; as tuple fields, they do not. The comment is inserted to help the implementer decipher the input tuple.

exercises/book-store/tests/book-store.rs
coriolinus: As commented above, the `input` tuple contains all the case properties which we don't know. In this case, I believe that the `targetgrouping` field was meant to be an _output_ hint.

I considered having the student return their groups and comparing them to the target groupings, but I decided against it. Checking the computed lowest price only gives the same assurance as checking it with the groups, but with a considerably simpler interface.

exercises/book-store/tests/book-store.rs
petertseng: I was about to suggest

> Ah, then this should be `process_total_case(basket: Vec<usize>, targetgrouping: Vec<Vec<usize>>, expected: f64)`

but now I hesitate a little bit about that suggestion because then it might not be clear which of these are inputs, whereas it is clar if they're named `input:`

It's also unfortuntae because `targetgrouping` is not an input. Maybe it shouldn't be in the canonical-data.json at all.

petertseng: but we can continue this discussion pretending it's an input, based on what we would like to see for exercises that actually do have multiple inputs.

coriolinus: The principle I used when writing the generator is that in cases where we require manual intervention from the test implementor, the effort required should be minimal. The actual generated code looks like this:

```rust
fn process_{property}_case<I, O>(input: I, expected: O) {
    // typical implementation:
    // assert_eq!(
    //     student_{property}_func(input),
    //     expected
    // )
    unimplemented!()
}
```

While it would be possible to break that out into a more complex function signature, as you say, it becomes less obvious for the test implementer which parameters form the actual inputs. I prefer to leave them as a simple tuple, for now.

PR #391, opened 2017-11-11
Add exercise reverse-string

coriolinus:
Used the `reverse-string` exercise to test the modifications to the exercise generation script; figured there was no particular reason not to finish it and submit the results.

If nothing else, this is an easy exercise for people to start learning about the difference between `&str` and `String`, and the natural implementation is a good way for people to start learning about iterators.


coriolinus:
Something to consider: should we add a Rust-specific property `reverse_in_place` with signature `fn reverse_in_place(input: &mut String)`? It would add a bit more work for students, but would help really drive home the difference between an always-immutable string slice `str` and a mutable `String`. It would also serve as an introduction to the notion of mutable input variables.

coriolinus:
Just spent some time attempting to implement `fn reverse_in_place(_: &mut String)`, and decided it's not appropriate for this exercise:

- Implementing this without `unsafe` requires allocating a temporary `String`, which defeats the point of modification in place. Worse, students might not already know that it defeats the point of modification in place, and implement it with allocation. We can't really check for allocation in a test, so we'd have taught them bad habits.
- Implementing this without allocation, with `unsafe`, requires fancy tricks: `mem::swap`, `mem::replace`, `mem::uninitialized`, and the like. It also introduces the chance to panic if the input includes characters more than 8 bits wide. Given that this is meant to be a difficulty 1 exercise for new Rust students, I think we're better off just sticking to the exercise as it already is.

PR #394, opened 2017-11-13
perfect-numbers: Update tests from table-based to macro-based

### LINKED ISSUE #392: consider table-driven tests
In many tests, we have some number of input/output pairs, and the validation we perform on the input/output pairs is the same. For the purpose of not repeating ourselves, we often extract the validation into one function.

For further deduplication, and for not adding so many test functions, we may consider the approach taken in https://github.com/exercism/rust/blob/master/exercises/perfect-numbers/tests/perfect-numbers.rs, a table-driven test.

Let us decide some set of circumstances under which table-based tests are ideally to be used (all the time, none of the time, or something in between).

coriolinus:
This PR is in response to #392; closes #392.

Perfect-numbers currently uses table-based testing. For reasons outlined
in https://github.com/exercism/rust/issues/392#issuecomment-343865993
I believe that table-based testing is not the appropriate way to do things.

This is what macro-based tests might look like. Are they abstruse and
opaque? Yes. However, they at least become individual tests, and I
believe that this is _better than the status quo_.


coriolinus:
```shell
The command "sh ./_test/ensure-stubs-compile.sh" exited with 0.
$ sh ./_test/count-ignores.sh
/home/travis/build/exercism/rust/exercises/perfect-numbers/tests/perfect-numbers.rs: Has 2 tests and 11 ignores (should be 1)
The command "sh ./_test/count-ignores.sh" exited with 1.
```

Clearly, there will need to be some tooling updates if we decide to go this route.

coriolinus:
Given that there's been no discussion or objection in the last 6 days, I'll merge this once Travis approves.

PR #398, opened 2017-11-21
Implemented exercise simple-linked-list

kedeggel:
It's my first submit, so I'm sure something is missing. Just tell me and I'll fix it.


exercises/simple-linked-list/src/lib.rs
```diff
@@ -0,0 +1,36 @@
+pub struct SimpleLinkedList<T> {
+    // Delete this field
+    // _dummy is needed to avoid unused parameter error during compilation
+    _dummy: T
+    // head
+    // size
+}
+
+// Implementation of basic functions
+// Use SimpleLinkedList as Stack (LIFO - Last In, First Out)
```
coriolinus: `impl<T> SimpleLinkedList<T>` will always be an implementation of basic functions for `SimpleLinkedList<T>`; we don't need that comment. The tests should ensure that it has stack semantics, so we can also eliminate that comment. 

The goal here is to maximize the degree to which students get to figure things out for themselves without prompting.

exercises/simple-linked-list/tests/simple-linked-list.rs
```diff
@@ -0,0 +1,63 @@
...
+    list.push(1);
+    list.push(2);
+    assert_eq!(list.pop(), Some(2), "Element must be 2");
+    assert_eq!(list.pop(), Some(1), "Element must be 1");
+    assert_eq!(list.pop(), None, "No element should be contained in list");
+}
+
+#[test]
+#[ignore]
+fn peek_returns_current_element() {
```
coriolinus: Nitpick: `peek_returns_head_element` would be more descriptive than `peek_returns_current_element`. If you feel strongly otherwise, please feel free to ignore this comment.

exercises/simple-linked-list/tests/simple-linked-list.rs
```diff
@@ -0,0 +1,63 @@
...
+
+#[test]
+fn test_new_list_is_empty() {
+    let list: SimpleLinkedList<u32> = SimpleLinkedList::new();
+    assert_eq!(list.size(), 0, "list's size must be 0");
+}
+
+#[test]
+#[ignore]
+fn push_increments_size() {
```
coriolinus: All test functions should have names starting with `test_`.

exercises/simple-linked-list/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+    pub fn pop(&mut self) -> Option<T> {
+        unimplemented!()
+    }
+
+    pub fn peek(&self) -> Option<&T> {
+        unimplemented!()
+    }
+
+    // This is where you might consider to bind the generic type parameter with a trait
+    pub fn from_slice(slice: &[T]) -> Self {
```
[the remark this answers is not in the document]
kedeggel:
```
75 | impl<T> From<[T]> for SimpleLinkedList<T>
   |         ^^^^^^^^^ `[T]` does not have a constant size known at compile-time
   |
   = help: the trait `std::marker::Sized` is not implemented for `[T]`
```
This is way I didn't implemented it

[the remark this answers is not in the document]
coriolinus: Right, sorry; the impl has to be [a little more complex](https://play.rust-lang.org/?gist=03825bfd7841ffc4eb054e1913b6f5d0&version=stable):

```rust
impl<'a, T> From<&'a [T]> for SimpleLinkedList<T> where T: Clone  {
    fn from(_: &[T]) -> SimpleLinkedList<T> {
        unimplemented!()
    }
}
```

The general point, which is that for pedagogical rigor we should use appropriate traits instead of dedicated methods, stands.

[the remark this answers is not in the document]
kedeggel: I had this idea, too. I've already implemented it, but when testing it the following error occurs:
```
58 |     let mut list = SimpleLinkedList::from(&slice);
   |                    ^^^^^^^^^^^^^^^^^^^^^^ the trait `std::convert::From<&[&str; 4]>` is not implemented for `simple_linked_list::SimpleLinkedList<_>`
   |
   = help: the following implementations were found:
             <simple_linked_list::SimpleLinkedList<T> as std::convert::From<&'a [T]>>
   = note: required by `std::convert::From::from`
```
So I think `fn from_slice` is the better solution

exercises/simple-linked-list/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+The simplest kind of linked list is a singly linked list. Each element in the
+list contains data and a "next" field pointing to the next element in the list
+of elements.
+
+This variant of linked lists is often used to represent sequences or
+push-down stacks (also called a LIFO stack; Last In, First Out).
+
+As a first take, lets create a singly linked list to contain the range (1..10),
+and provide functions to reverse a linked list and convert to and from arrays.
```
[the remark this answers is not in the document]
kedeggel: - `impl<T> From<SimpleLinkedList<T>> for Vec<T>`: The `error [E0210]` occurs, so I've decided to test for `impl<T> Into<Vec<T>> for SimpleLinkedList<T>`

exercises/simple-linked-list/src/lib.rs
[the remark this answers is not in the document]
coriolinus: Arrays and slices are not quite the same thing. 

There are a few ways around this. We could improve the trait's bounds to cover what we're after, [which works](https://play.rust-lang.org/?gist=0b5560c45499c0339093a60e17f58a75&version=stable):

```rust
impl<'a, S, T> From<&'a S> for SimpleLinkedList<T>
where S: AsRef<[T]>,
      T: Clone,      
{
    fn from(_: &S) -> SimpleLinkedList<T> {
        unimplemented!()
    }
}
```

If we were writing a production library, that's probably the path we'd take. However, that might be a bit intimidating for the student, even if we gave them the stub. It's friendlier to them to use another approach. Luckily, that's a pretty simple matter of [changing the calling pattern](https://play.rust-lang.org/?gist=42757091d364b1d930ce1f91f5a18ab2&version=stable): `let _: SimpleLinkedList<_> = [1, 2, 3].as_ref().into();` works even with the previous, simpler impl.

exercises/simple-linked-list/tests/simple-linked-list.rs
```diff
@@ -0,0 +1,89 @@
...
+#[test]
+#[ignore]
+fn test_from_slice() {
+    let array = ["1", "2", "3", "4"];
+    let mut list = SimpleLinkedList::from(array.as_ref());
+    assert_eq!(list.pop(), Some("1"));
+    assert_eq!(list.pop(), Some("2"));
+    assert_eq!(list.pop(), Some("3"));
+    assert_eq!(list.pop(), Some("4"));
+}
```
coriolinus: The tests as written reverse the order of an input slice, and again when emitting a vector. While that's _consistent_, in that `vec![1, 2, 3] == SimpleLinkedList::from([1, 2, 3].as_ref()).into()`, I'd expect that to cause some confusion for students. If you build a vec and a list in the same way, they should be the same:

```rust
let mut v = Vec::new();
let mut s = SimpleLinkedList::new();
for i in vec![1, 2, 3] {
    v.push(i);
    s.push(i);
}
assert_eq!(v, s.into());
```

Please change the semantics of the traits converting from slices and to vectors so that they preserve the order of elements, and update the tests accordingly.

config.json
```diff
@@ -243,6 +243,18 @@
       "topics": [
         "Result"
       ]
+    },
+	{
+      "uuid": "10923b0b-c726-44a7-9e02-b775e7b8b237",
+      "slug": "simple-linked-list",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 4,
```
kedeggel: What difficulty would be appropiate?

kedeggel: What difficulty would be appropiate?

kedeggel: What difficulty would be appropiate?

coriolinus:
I had no problems with the difficulty you assigned, 4, or the exercise's
position within the list.

exercises/simple-linked-list/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+    fn from(item: &[T]) -> Self {
+        unimplemented!()
+    }
+}
+
+impl<T> Into<Vec<T>> for SimpleLinkedList<T>
+where
+    T: Clone,
+{
+    fn into(self) -> Vec<T> {
```
coriolinus: If `into` consumes `self`, then we don't need the trait bound `T: Clone`; we can just return the original objects.

exercises/simple-linked-list/.meta/hints.md
```diff
@@ -0,0 +1,35 @@
...
+Also implements this struct `Node` with the following fields:
+```
+struct Node<T> {
+    data: T,
+    next: Option<Box<Node<T>>>,
+}
+```
+`data` contains the stored data, and `next` points to the following node (if available) or None.  
+
+#### Why Option<__Box__<Node<T>>> and not just Option<Node<T>>?
```
coriolinus: If this is contained in a level 2 heading (`##`), it should be a level 3 heading (`###`).

exercises/simple-linked-list/.meta/hints.md
```diff
@@ -0,0 +1,35 @@
+## Implementation Hints
+
+Do not use the struct `SimpleLinkedList` just as a wrapper around a `Vec`, but use the fields `head` and `len`.
```
coriolinus: - Consider phrasing as "Do not implement..." instead of "Do not use..."
- Possibly replace this sentence as follows:

> Do not implement the strict `SimpleLinkedList` as a wrapper around a `Vec`. Instead, allocate nodes on the heap.

Some students may want to omit the `len` field in favor of recursive computation on the fly, and we have no reason to forbid that design choice.

exercises/simple-linked-list/.meta/hints.md
```diff
@@ -0,0 +1,35 @@
...
+Do not use the struct `SimpleLinkedList` just as a wrapper around a `Vec`, but use the fields `head` and `len`.  
+This might look similar to this:
+```
+pub struct SimpleLinkedList<T> {
+    head: Option<Box<Node<T>>>,
+    len: usize,
+}
+```
+The `head` field points to the first element (Node) of this linked list.  
+Also implements this struct `Node` with the following fields:
```
coriolinus: Instead of "Also implements...", consider phrasing this as "This implementation also requires a struct `Node` with the following fields:"

exercises/simple-linked-list/.meta/hints.md
```diff
@@ -0,0 +1,35 @@
+## Implementation Hints
+
+Do not use the struct `SimpleLinkedList` just as a wrapper around a `Vec`, but use the fields `head` and `len`.  
+This might look similar to this:
```
coriolinus: Consider phrasing as "This might be implemented as:"

exercises/simple-linked-list/src/lib.rs
kedeggel: I agree that T doesn't need to be clonable in this function, but I cannot omit the trait bound here because then I get the error `the trait bound 'T: std::clone::Clone' is not satisfied`.
But you're right: in my example.rs it's not necessary to clone the data.

coriolinus: I can't replicate that error using only `src/lib.rs`.

However, I see where it's hit you in `example.rs`. The thing is, you've got `T: Clone` bounds in several places where they're not actually necessary. [Playground example](https://play.rust-lang.org/?gist=2f0b2301b793e9c59185109aae25cd06&version=stable) where they're removed and everything runs successfully.

In order to provide a good example for the students, let's omit the `T: Clone` bound here.

kedeggel: Okay, thank you very much for this playgrund example. I didn't know there can be several impl blocks for the same struct with different trait bounds. Now it's clear. I'm changing it and hope everything will be alright then.

config.json
```diff
@@ -243,6 +243,18 @@
...
+      "topics": [
+		"lists",
```
petertseng: I would suggest that the indentation here use spaces instead of tabs, otherwise formatters such as https://github.com/exercism/rust/blob/master/bin/init_exercise.py will cause this line to change for a potentially unrelated change.

config.json
```diff
@@ -243,6 +243,18 @@
...
+        "type_conversion",
+		"struct"
```
petertseng: > I would suggest that the indentation here use spaces instead of tabs, otherwise formatters such as https://github.com/exercism/rust/blob/master/bin/init_exercise.py will cause this line to change for a potentially unrelated change.

same.

config.json
```diff
@@ -243,6 +243,18 @@
       "topics": [
         "Result"
       ]
+    },
+	{
```
petertseng: > I would suggest that the indentation here use spaces instead of tabs, otherwise formatters such as https://github.com/exercism/rust/blob/master/bin/init_exercise.py will cause this line to change for a potentially unrelated change.

same

exercises/simple-linked-list/Cargo.toml
```diff
@@ -0,0 +1,5 @@
+[package]
+name = "simple_linked_list"
+version = "1.0.0"
```
petertseng: since this is not based on any problem-specifications 1.0.0 (https://github.com/exercism/problem-specifications/tree/master/exercises/simple-linked-list not existing), this must be 0.0.0 or 0.1.0 (your choice), so that when 1.0.0 is created we know that we need to check whether we are in compliance with it.

If we do not do this, we could get out of date with problem-specifications and not know about it.

config.json
```diff
@@ -244,6 +244,18 @@
...
+      "uuid": "10923b0b-c726-44a7-9e02-b775e7b8b237",
+      "slug": "simple-linked-list",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 4,
+      "topics": [
+        "lists",
+        "type_conversion",
+        "struct"
+        ]
```
petertseng: this should be indented two spaces less than it currently is, so that it's on the same column as "topics". the other exercises follow this pattern.

kedeggel:
I squashed the commits

PR #399, opened 2017-11-23
Implement exercise diffie-hellman

hekrause:


exercises/diffie-hellman/Cargo.lock
```diff
@@ -0,0 +1,48 @@
+[[package]]
```
coriolinus: Please include an exercise-specific `.gitignore` file which excludes `Cargo.lock`.

exercises/diffie-hellman/example.rs
```diff
@@ -0,0 +1,18 @@
+extern crate rand;
+
+use rand::distributions::{IndependentSample, Range};
+
+pub fn private_key(p: &u64) -> u64 {
```
coriolinus: The `u64` type is `Copy`, so it's more idiomatic and convenient to pass by value instead of by reference. 

Please update the tests and the example such that all `u64`s are passed by value instead of by reference.

exercises/diffie-hellman/tests/diffie-hellman.rs
```diff
@@ -0,0 +1,58 @@
...
+use diffie_hellman::*;
+
+#[test]
+fn test_private_key_in_range_key() {
+    let primes: Vec<u64> = vec![5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47, 773, 9091,
+                                38321, 6131329];
+    let private_keys: Vec<u64> = primes.iter().map(|x| private_key(*x)).collect();
+
+    for i in 0..primes.len() {
+        assert!(1 < private_keys[i] && private_keys[i] < primes[i]);
```
petertseng: On https://en.wikipedia.org/wiki/Diffie%E2%80%93Hellman_key_exchange I see a quote that says:

> Alice picks a random natural number a, where 1 ≤ a < n, and sends g<sup>a</sup> to Bob.

So, it seems like 1 would be valid, but this line disallows it. Is 1 in fact allowed, or did I misunderstand?

exercises/diffie-hellman/example.rs
```diff
@@ -0,0 +1,18 @@
...
+use rand::distributions::{IndependentSample, Range};
+
+pub fn private_key(p: u64) -> u64 {
+    assert!(p < u64::max_value());
+    Range::new(2, p).ind_sample(&mut rand::thread_rng())
+}
+
+pub fn public_key(p: u64, g: u64, a: u64) -> u64 {
+    assert_ne!(g.checked_mul(a), None);
+    (g * a) % p
```
petertseng: On https://en.wikipedia.org/wiki/Diffie%E2%80%93Hellman_key_exchange I see a quote that says:

> Alice chooses a secret integer a = 4, then sends Bob A = g<sup>a</sup> mod p

So it seems to me that the `*` should be a `**` instead. Is it in fact `**` or did I misunderstand the page?

exercises/diffie-hellman/tests/diffie-hellman.rs
```diff
@@ -0,0 +1,58 @@
...
+}
+
+#[test]
+#[ignore]
+fn test_public_key_correct() {
+    let p: u64 = 23;
+    let g: u64 = 5;
+
+    let private_key: u64 = 6;
+    let expected: u64 = 7;
```
petertseng: On https://en.wikipedia.org/wiki/Diffie%E2%80%93Hellman_key_exchange I see a quote that says:

> Alice chooses a secret integer a = 4, then sends Bob A = g<sup>a</sup> mod p

It seems like thus the expected result should be 8. Is it in fact 8 or did I misunderstand the page?

exercises/diffie-hellman/example.rs
```diff
@@ -0,0 +1,18 @@
...
+}
+
+pub fn secret(p: u64, b_pub: u64, a: u64) -> u64 {
+    assert_ne!(b_pub.checked_mul(a), None);
+    (b_pub * a) % p
```
petertseng: On https://en.wikipedia.org/wiki/Diffie%E2%80%93Hellman_key_exchange I see a quote that says:

> Alice computes s = B<sup>a</sup> mod p

So it seems to me that the `*` should be a `**` instead. Is it in fact `**` or did I misunderstand the page?

exercises/diffie-hellman/tests/diffie-hellman.rs
```diff
@@ -0,0 +1,58 @@
...
+
+#[test]
+#[ignore]
+fn test_secret_key_correct() {
+    let p: u64 = 92233987;
+    let g: u64 = 5371081;
+
+    let private_key = 75233987;
+    let public_key = public_key(p, g, private_key);
+    let secret = secret(p, public_key, private_key);
```
petertseng: so, to confirm, this is showing what happens when someone effectively performs a DH key exchange with themselves, right? Since they're using their own public key and their own private key.

It does not seem to me like this is reflective of a likely real-world usage, but as you know since I am ignorant of DH, there could simply be a use that I am not aware of.

exercises/diffie-hellman/tests/diffie-hellman.rs
hekrause: Hmm I think then the [description](https://github.com/hekrause/problem-specifications/blob/master/exercises/diffie-hellman/description.md) is wrong.

exercises/diffie-hellman/example.rs
hekrause: Oh yes, my bad I never used this form of power(```**```) I always use ```^```.

exercises/diffie-hellman/example.rs
hekrause: Oh yes, my bad I never used this form of power(```**```) I always use ```^```.

exercises/diffie-hellman/tests/diffie-hellman.rs
hekrause: No, you are right.

exercises/diffie-hellman/tests/diffie-hellman.rs
hekrause: In the last test ```test_changed_secret_key()``` they exchange their public keys.

grafik

You see that A decrypts Bs public key with As private key and p. 
B decrypts As public key with Bs private key and p. 

If successful, they share the same secret key/session key.
From the [wiki](https://en.wikipedia.org/wiki/Diffie%E2%80%93Hellman_key_exchange): 
```Now s is the shared secret key and it is known to both Alice and Bob, but not to Eve.```

File: exercises/diffie-hellman/tests/diffie-hellman.rs (hunk 3 above)
Comment (hekrause) [reply exercises/diffie-hellman/tests/diffie-hellman.rs]: @petertseng 
Ah I think in this section of the [wiki](https://en.wikipedia.org/wiki/Diffie%E2%80%93Hellman_key_exchange) they are taking ```a``` from a pool of known numbers.
We are working here with primes where 1 is no prime.

File: exercises/diffie-hellman/tests/diffie-hellman.rs [hunk 8]
```diff
@@ -0,0 +1,60 @@
...
+use diffie_hellman::*;
+
+#[test]
+fn test_private_key_in_range_key() {
+    let primes: Vec<u64> = vec![5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47, 773, 967,
+                                3461, 6131];
+    let private_keys: Vec<u64> = primes.iter().map(|x| private_key(*x)).collect();
+
+    for i in 0..primes.len() {
+        assert!(1 <= private_keys[i] && private_keys[i] < primes[i]);
```
petertseng: > Hmm I think then the [description](https://github.com/hekrause/problem-specifications/blob/master/exercises/diffie-hellman/description.md) is wrong.
> I think in this section of the [wiki](https://en.wikipedia.org/wiki/Diffie%E2%80%93Hellman_key_exchange) they are taking ```a``` from a pool of known numbers.
We are working here with primes where 1 is no prime and therefore should not be available in the range.

I understand. In that case I have changed my mind, let us not allow 1.

the reason I do not allow 1 as a private key is not because it is composite (otherwise, we would also have to disallow all other composite numbers, and this line does no such thing), but rather because the description says so, and perhaps it is for because 1 seems like a weak key (identity in exponentiation)

exercises/diffie-hellman/tests/diffie-hellman.rs
```diff
@@ -0,0 +1,60 @@
...
+
+#[test]
+#[ignore]
+fn test_secret_key_correct() {
+    let p: u64 = 11;
+    let g: u64 = 7;
+
+    let private_key = 7;
+    let public_key = public_key(p, g, private_key);
+    let secret = secret(p, public_key, private_key);
```
petertseng: > > so, to confirm, this is showing what happens when someone effectively performs a DH key exchange with themselves, right? Since they're using their own public key and their own private key.
> 
> In the last test `test_changed_secret_key()` they exchange their public keys.

I agree and now I am wanting to talk about `test_secret_key_correct`.

It appears the purpose of it is to make sure the `secret` function is correct. This is a good purpose; there absolutely should be a test that confirms this.

There is an interesting thing about the inputs to this function. We're asking for (the secret that results from (my own
 private key) and (the public key corresponding to my own private key)), which means I have performed a DH key exchange with myself. Again, now I show my ignorance, *maybe* this is a reasonable thing to do, because supposedly now I have a secret that I and only I know (it's `g ** (a ** 2) mod p`), I just don't know about it. If it is unusual to DH key exchange with myself, then I would not want the tests to show it as an example, since it would give people the wrong idea.

So I was thinking of having one of four things happen:

* Show me something that tells that it makes sense to DH key exchange with myself
* That I was actually mistaken, this is not a DH key exchange with myself
* Use a different (but still fixed!) private key in the call to `public_key` instead of the same one used in `secret`
* Use a fixed value for `public_key` instead of calling the `public_key` function here (in this case `g` can be deleted)

hekrause: ```That I was actually mistaken, this is not a DH key exchange with myself```

There is a public key exchange between Alice(A) and Bob(B).
After the public key exchange, they both can generate a **secret session key** from their **own private key** and the **opposite public key**.
Alice can then verify that the message is from Bob because he has encrypted the message with the secret session key which only they both share.

petertseng: > **their own private key** and **the opposite public key**

I do not believe this is an opposite public key. Because the same `private_key` is used in `let public_key = public_key(p, g, private_key)` and `let secret = secret(p, public_key, private_key)`, The public key was derived from the same private key. For it to be an opposite public key, there must be `let public_key = public_key(p, g, some_other_private_key)`

petertseng: In anticipation of a possible response "it just so happens that by random chance, Alice and Bob have the same private key", I'll accept that by adding a fifth option:

* Add a line that says something like `let other_private_key = private_key; // by some coincidence, it just so happens that the two parties have the same private key` to make it clear that there is in fact a second party to this exchange, but they just happen to have the same key and it is not by design that they always have the same key.

petertseng: EBWODP: with the fifth option, then there would also be `let public_key = public_key(p, g, other_private_key)` of course.

hekrause: Ah now I get where the problem is. I thought you would still talk about the ```test_changed_secret_key()``` but you were talking about the ```test_secret_key_correct()```.
Yes, you are right there should be a public key from b and a private key from a for example.

exercises/diffie-hellman/src/lib.rs
```diff
@@ -0,0 +1,12 @@
+
```
petertseng: an extra newline that I believe could be removed here

petertseng: (ignore the fact that this comment shows in the wrong place. The newline I intended has been removed, and the one this comment is currently on should not be removed)

hekrause: Yes, then it may be better if you could do the branch/commit cleaning.

PR #400, opened 2017-11-25
Implement exercise saddle-points

hekrause:


hekrause:
I don't know why the ```ensure-readmes-are-updated.sh``` script has a problem with the README.md. 
I generated it with the command ```./configlet generate rust --only saddle-points``` and then pushed it.

exercises/saddle-points/README.md
```diff
@@ -0,0 +1,66 @@
...
+  |---------
+0 | 9  8  7
+1 | 5  3  2     <--- saddle point at (1,0)
+2 | 6  6  7
+```
+
+It has a saddle point at (1, 0).
+
+It's called a "saddle point" because it is greater than or equal to
+every element in its row and the less than or equal to every element in
```
petertseng: and the less than

petertseng: that means you did not take https://github.com/exercism/problem-specifications/commit/2f6a515ed1b1a05477988ce18c3954e1669feda2 into account when generating the README. Please regenerate taking it into account.

exercises/saddle-points/Cargo.toml
```diff
@@ -0,0 +1,5 @@
+[package]
+name = "saddle-points"
+version = "1.1.0"
```
coriolinus: This version should match the version in [`canonical-data.json`](https://github.com/exercism/problem-specifications/blob/master/exercises/saddle-points/canonical-data.json#L3). That means that as of the current `master` branch of `problem-specifications`, it should be `1.0.0`.

exercises/saddle-points/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+
+pub fn find_saddle_points(input: Vec<Vec<u64>>) -> Vec<(u64, u64)>{
```
coriolinus: There's no need to transfer ownership of the input for `find_saddle_points` to work. We want to teach people how to write idiomatic Rust, so in this case, this function should take `&[&[u64]]` (reference to slice of reference to slice of u64). `Vec<Vec<u64>>` should coerce transparently to the slice form.

coriolinus: Naming the variable `input` causes a compiler warning due to the unused variable. While this doesn't break CI, I'd prefer to skip that by renaming the variable to `_input` or simply `_`.

exercises/saddle-points/tests/saddle-points.rs
```diff
@@ -0,0 +1,68 @@
+extern crate saddle_points;
+
+use saddle_points::*;
+
+#[test]
+fn identify_single_saddle_point() {
```
coriolinus: All test names should begin with `test_`.

exercises/saddle-points/tests/saddle-points.rs
```diff
@@ -0,0 +1,68 @@
...
+fn identify_empty_matrix() {
+    let input: Vec<Vec<u64>> = vec![vec![]];
+    let expected: Vec<(u64, u64)> = Vec::new();
+
+    assert_eq!(expected, find_saddle_points(input));
+}
+
+#[test]
+#[ignore]
+fn identify_missing_saddle_point() {
```
coriolinus: Use of "missing" in this test name implies that one should be present. Please update the name of this function to make it clear that there are expected to be none.

I believe this is based on the canonical test "Can identify lack of saddle points when there are none". If so, consider naming this test based on that name. The test names don't need to precisely map to the canonical test descriptions, but it's best when there is at least some correspondence.

exercises/saddle-points/tests/saddle-points.rs
```diff
@@ -0,0 +1,68 @@
...
+    let input: Vec<Vec<u64>> = vec![vec![1, 2, 3],
+                                    vec![3, 2, 1],
```
coriolinus: If this test is based on the canonical test "Can identify lack of saddle points when there are none", this row of input data diverges from the [equivalent row in the canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/saddle-points/canonical-data.json#L39). Is there a reason for this?

exercises/saddle-points/tests/saddle-points.rs
```diff
@@ -0,0 +1,68 @@
...
+fn identify_bottom_right_saddle_point() {
+    let input: Vec<Vec<u64>> = vec![vec![8, 7, 9],
+                                    vec![6, 7, 6],
+                                    vec![3, 2, 5]];
+    assert_eq!(vec![(2,2)], find_saddle_points(input));
+}
+
+#[test]
+#[ignore]
+fn non_quadratic_matrix_high() {
```
coriolinus: Non-imperative quibble: change `*quadratic*` test names to use the term `square` instead. While "quadratic" isn't wrong, it has associations with the operation of squaring more than the shape of a matrix. Also, it's usually better to name things for clarity, and "square" is a much more common word.

If you feel strongly that the term should remain, it's fine; I don't feel that this is an essential change, just one I'd prefer.

exercises/saddle-points/example.rs
```diff
@@ -0,0 +1,23 @@
+
+pub fn find_saddle_points(input: Vec<Vec<u64>>) -> Vec<(u64, u64)>{
+    let mut saddle_points: Vec<(u64, u64)> = Vec::new();
+
+    let row_index = input.len();
```
petertseng: Although this is indeed an index, it is also not the index of any existing element in the input (it is the index of the first element beyond it), so to avoid confusion I would call this `height` instead.

exercises/saddle-points/example.rs
```diff
@@ -0,0 +1,23 @@
...
+    let column_index = input[0].len();
```
petertseng: > Although this is indeed an index, it is also not the index of any existing element in the input (it is the index of the first element beyond it), so to avoid confusion I would call this `height` instead.

Same, except now `width`.

exercises/saddle-points/example.rs
```diff
@@ -0,0 +1,23 @@
...
+
+    for i in 0..row_index {
+        for j in 0..column_index {
+            let row = &input[i];
```
petertseng: since this is only based on `i` and not `j`, it would be possible to assign it in the outer loop and not the inner one. Unless that makes the code less understandable.

exercises/saddle-points/example.rs
```diff
@@ -0,0 +1,23 @@
...
+            let column = input.iter().map(|x| x[j]).collect::<Vec<u64>>();
+
+            let max = row.iter().max().unwrap();
```
petertseng: since this is only based on `row`, which is only based on `i` and not `j`, it would be possible to assign it in the outer loop and not the inner one. Unless that makes the code less understandable.

exercises/saddle-points/example.rs
```diff
@@ -0,0 +1,23 @@
...
+            let min = column.iter().min().unwrap();
+            let value = input[i][j];
+
+            if value >= *max && value <= *min {
+                saddle_points.push((i as u64, j as u64));
+            }
+        }
+    }
+    return saddle_points;
```
petertseng: I suggest removing the `return` and the semicolon.

However, I cannot find a justification for this suggestion in https://aturon.github.io nor https://github.com/rust-lang-nursery/fmt-rfcs/blob/master/guide/guide.md which means I have no grounds backing up this suggestion. Nevertheless, it is suggested.

exercises/saddle-points/tests/saddle-points.rs
hekrause: No, it was a typo. :-)

exercises/saddle-points/example.rs
hekrause: If the matrix is empty the two for loops prevent an unwrap on  ```row``` in line 13.

coriolinus:
Thanks @hekrause for updating these. As mentioned before (https://github.com/exercism/rust/pull/400#pullrequestreview-79030613), there still isn't a test which validates that solutions can return multiple saddle points if they exist; there's a case for that already in the canonical data. Please add at least that case.

exercises/saddle-points/tests/saddle-points.rs
```diff
@@ -3,66 +3,65 @@ extern crate saddle_points;
 use saddle_points::*;
 
 #[test]
-fn identify_single_saddle_point() {
-    let input: Vec<Vec<u64>> = vec![vec![9, 8, 7],
-                                    vec![5, 3, 2],
-                                    vec![6, 6, 7]];
+fn test_identify_single_saddle_point() {
+    let input: &[&[u64]] = &[&[9, 8, 7],
```
coriolinus: Interesting; I hadn't intended that you change the tests when I asked you to change the signature; a reference to `Vec<Vec<u64>>` should just work for a function which requires `&[&[u64]]`. 

I'm not sure whether to ask you to change this back. As of right now, I'm undecided: on the one hand, it's interesting to see that slice reference literals work at all. On the other hand, idiomatic code would be to specify the inputs as `Vec<Vec<u64>>` and then pass the reference and let the `Deref` trait do the work for you.

I'd like to solicit opinions from the other maintainers about which of these forms are better style.

exercises/saddle-points/example.rs
```diff
@@ -0,0 +1,24 @@
...
+
+    let width = input.len();
+    let height = input[0].len();
+
+    for i in 0..width {
+        for j in 0..height {
+
+            let column = input.iter().map(|x| x[j]).collect::<Vec<u64>>();
+            let row = &input[i];
+            let max = row.iter().max().unwrap();
```
petertseng: > > since this is only based on row, which is only based on i and not j, it would be possible to assign it in the outer loop and not the inner one. Unless that makes the code less understandable.
>
> If the matrix is empty the two for loops prevent an unwrap on row in line 13.

Indeed, if a row is empty. In that case, I will now only optionally ask for `row` to be moved to the outer loop (it still can be), and agree that I was wrong to ask for this `max` line to be moved; it cannot be.

exercises/saddle-points/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+
```
petertseng: > this empty line *could* be removed, and I think it should be. The fact that there aren't any `use` lines shouldn't affect things I don't think, because if they were present, we wouldn't start the file with an empty line either.

same

exercises/saddle-points/example.rs
```diff
@@ -0,0 +1,24 @@
+
```
petertseng: this empty line *could* be removed, and I think it should be. The fact that there aren't any `use` lines shouldn't affect things I don't think, because if they were present, we wouldn't start the file with an empty line either.

exercises/saddle-points/tests/saddle-points.rs
petertseng: Since it is idiomatic, I would keep the `vec` then. Will be even more convinced if finding a source that tells us this, but I cannot.

We do have quite a few instances of using slices in tests in other exercises, but potentially changing them is a story for another day

exercises/saddle-points/example.rs
hekrause: Yes you are right. I think in the way it is now it is also more readable.

exercises/saddle-points/tests/saddle-points.rs
hekrause: So you mean in the tests should be vectors as in the first version but I should call the function with the ```&[&[u64]]``` parameters?And the slices should be from the vectors?

coriolinus: Yes: leave the function signature as it is, which expects slices of slices, and change the tests so that they allocate `Vec`s and pass a reference to the function.

exercises/saddle-points/tests/saddle-points.rs
```diff
@@ -0,0 +1,103 @@
+extern crate saddle_points;
+
+use saddle_points::*;
+
+#[test]
+fn test_identify_single_saddle_point() {
+    let vector: Vec<Vec<u64>> = vec![vec![9, 8, 7],
+                                     vec![5, 3, 2],
+                                     vec![6, 6, 7]];
+    let input: &[&[u64]] = &[&vector[0][..],
```
coriolinus: This way of forming the input slice of slices is ugly, and is my fault. I inadvertently gave you bad advice. The `find_saddle_points` signature should actually be `pub fn find_saddle_points(_: &[Vec<u64>]) -> Vec<(u64, u64)>`.

I was wrong to tell you that it should accept an `&[&[u64]]`, because the compiler doesn't automatically dereference the inner elements into a slice; the inner elements need to remain `Vec<u64>`. At the same time, the outer layer should be `&[_]` because it is idiomatic to pass a slice reference, not to transfer ownership of a `Vec`. 

In other words, the function signature and tests should work [as in this playground example](https://play.rust-lang.org/?gist=dcc7e8c78a38e261baa4ed8a921e074f&version=stable). 

I apologize for the bad advice. Now please change the signature one more time to hopefully clean this up once and for all.

coriolinus: Probably the best function signature for real projects would look like [this](https://play.rust-lang.org/?gist=90f6eb37932f7d5d01afb9515d94eb5b&version=stable):

```rust
pub fn find_saddle_points<X, Y, T>(_: &Y) -> Vec<(T, T)>
where
    Y: AsRef<[X]>,
    X: AsRef<[T]>,
{
    unimplemented!()
}
```

However, that would be an intimidating signature to require of new students. I still prefer `&[Vec<u64>]`.

exercises/saddle-points/example.rs
```diff
@@ -0,0 +1,24 @@
...
+            let column = input.iter().map(|x| x[j]).collect::<Vec<u64>>();
+            let row = &input[i];
+
+            let max = row.iter().max().unwrap();
+            let min = column.iter().min().unwrap();
+
+            let value = input[i][j];
+
+            if value >= *max && value <= *min {
+                saddle_points.push((i as u64, j as u64));
```
coriolinus: In fact, we could eliminate the casts here by having the function return `Vec<(usize, usize)>`. I prefer that over the current version.

PR #401, opened 2017-11-27
Add macro implementation exercise

### LINKED ISSUE #397: Macros by Example exercise
Macros are a powerful part of a Rust programmer's toolkit, and [macros by example](https://doc.rust-lang.org/reference/macros-by-example.html) are a relatively simple way to access this power. However, to beginners in the language, they're just scary black magic; while `println!` and `vec!` are frequently used, they're not something you'd write on your own. At least, that's how I felt when I was starting in the language.

The Rust track needs an exercise which is best solved by using this tool, as a way of introducing people to the concept. To some extent, we have exercises which are implicitly similar if macros are available. For example, the [decimal example](https://github.com/exercism/rust/blob/master/exercises/decimal/example.rs) is implemented in terms of macros. However, there's nothing currently which explicitly suggests the use of macros in a case where they'd make sense.

The elephant in the room WRT a macro exercise is that there is no way to guarantee that the student has actually used macros in their solution; anything which can be generated by macro can be hand-written by a student with a sufficiently high threshold for tedium. 

My intent is to use this issue as a space to bat back and forth some proposals for macro exercises, and collaboratively come to a design which makes sense.

coriolinus:
Closes #397.

This exercise requires that students implement a `hashmap!` macro
to simplify creation of hashmap literals. The intent is to gently
introduce macros-by-example by having them implement a macro which
is both simple but useful.

The [maplit crate](https://crates.io/crates/maplit) which inspired
this exercise uses the [MIT license](https://opensource.org/licenses/MIT),
which as far as I can tell permits this derivative use of its code.

Used first edition book link in the readme because second edition's
macros chapter is not yet written.

### DIFF
--- a/exercises/macros/.meta/description.md
+++ b/exercises/macros/.meta/description.md
@@ -1,4 +1,3 @@
-
 Macros are a powerful part of a Rust programmer's toolkit, and [macros by example](https://doc.rust-lang.org/reference/macros-by-example.html) are a relatively simple way to access this power. Let's write one!
 
 ## Context
--- a/exercises/macros/.meta/description.md
+++ b/exercises/macros/.meta/description.md
@@ -1,0 +1,30 @@
+
+Macros are a powerful part of a Rust programmer's toolkit, and [macros by example](https://doc.rust-lang.org/reference/macros-by-example.html) are a relatively simple way to access this power. Let's write one!
+
+## Context
+
+What is a macro? [Wikipedia](https://en.wikipedia.org/wiki/Macro_(computer_science)) describes it thus:
+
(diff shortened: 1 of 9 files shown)


exercises/macros/.meta/description.md
```diff
@@ -0,0 +1,30 @@
+
```
petertseng: I feel like this extra line could go, unless it is very important to have two newlines after the title

exercises/macros/README.md
```diff
@@ -0,0 +1,71 @@
+# Macros
+
+
```
petertseng: (which will cause this newline to go too)

PR #402, opened 2017-11-27
Implemented exercise series

kedeggel:


exercises/series/src/lib.rs
```diff
@@ -0,0 +1,15 @@
+pub struct Series {}
```
coriolinus: Why this interface? I'm not categorically rejecting it at this time, but I'd want a solid reason why the tested interface shouldn't be simply `pub fn series(digits: &str, len: usize) -> Option<Vec<String>>`.

config.json
```diff
@@ -317,6 +317,17 @@
         "char"
       ]
     },
+    {
+      "uuid": "9de405e1-3a05-43cb-8eb3-00b81a2968e9",
+      "slug": "series",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 4,
```
coriolinus: If we use a simpler, function-based interface, there's likely a one-line solution using [`Vec::windows`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.windows), which makes me question the 4 difficulty here a bit.

If we use the struct-based interface, most of the complexity of the exercise comes from implementing the struct itself, which isn't that interesting.

kedeggel:
I think you‘re right. First I wanted to try another approach where I needed this struct, but then I rejected this and forgot to remove it

config.json
kedeggel: 👍

exercises/series/src/lib.rs
kedeggel: I think you‘re right. First I wanted to try another approach where I needed this struct, but then I rejected this and forgot to remove it.

coriolinus: In fact, we don't even need the `Option`; we could have the function just return a 0-length `Vec` if the consumer requests a series of length greater than the input length, and lose no information.

kedeggel: I like my approach with Result because this is one of the few rusty things in this exercise.

coriolinus: Using `Result` to use more Rust features when it's not the natural interface is an antipattern.

There are plenty of exercises for which `Option` and `Result` make sense, and are implied by the requirements of the problem. This is not one of them: you know before ever entering the function exactly how many items you expect to get back from it.

kedeggel: What would be expected if `len` = 0? A 0-length `Vec`?

coriolinus: My instinct in that case would actually be to expect `vec!["".to_string(); digits.len() + 1]`.

Reasoning: in all other valid cases, you expect a `Vec` composed of `len`-length sequences, which contains `digits.len() + 1 - len`-length strings. Therefore, this sticks to the pattern.

config.json
kedeggel: I changed difficulty to 1 because there are only the difficulties 1, 4, 7, and 10 in use (according to the last point in https://github.com/exercism/rust/blob/master/README.md#contributing-a-new-exercise).

kedeggel:
Squashed

PR #406, opened 2017-12-04
parallel-letter-frequency: fix broken links in README.md

sjwarner-bp:
fixes two broken links in README.md

### DIFF
--- a/exercises/parallel-letter-frequency/README.md
+++ b/exercises/parallel-letter-frequency/README.md
@@ -33,11 +33,11 @@ If you are using rustup.rs:
 rustup run nightly cargo bench
 ```
 
-- [Benchmark tests](https://doc.rust-lang.org/1.1.0/book/benchmark-tests.html)
+- [Benchmark tests](https://doc.rust-lang.org/stable/unstable-book/library-features/test.html)
 
 Learn more about nightly Rust:
 
-- [Nightly Rust](https://doc.rust-lang.org/nightly/)
+- [Nightly Rust](https://doc.rust-lang.org/book/first-edition/release-channels.html)
 - [Rustup: Working with nightly](https://github.com/rust-lang-nursery/rustup.rs#working-with-nightly-rust)
 
 
--- a/exercises/parallel-letter-frequency/README.md
+++ b/exercises/parallel-letter-frequency/README.md
@@ -33,11 +33,11 @@ If you are using rustup.rs:
 rustup run nightly cargo bench
 ```
 
-- [Benchmark tests](https://doc.rust-lang.org/book/benchmark-tests.html)
+- [Benchmark tests](https://doc.rust-lang.org/1.1.0/book/benchmark-tests.html)
 
 Learn more about nightly Rust:
 
-- [Nightly Rust](https://doc.rust-lang.org/book/nightly-rust.html)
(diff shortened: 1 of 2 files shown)


exercises/parallel-letter-frequency/README.md
```diff
@@ -33,11 +33,11 @@ If you are using rustup.rs:
 rustup run nightly cargo bench
 ```
 
-- [Benchmark tests](https://doc.rust-lang.org/book/benchmark-tests.html)
+- [Benchmark tests](https://doc.rust-lang.org/1.1.0/book/benchmark-tests.html)
```
coriolinus: Is there a particular reason for using the "1.1.0" edition of the book, instead of the generic first-edition link? (<https://doc.rust-lang.org/book/first-edition/>)?

exercises/parallel-letter-frequency/README.md
```diff
@@ -33,11 +33,11 @@ If you are using rustup.rs:
...
 
 Learn more about nightly Rust:
 
-- [Nightly Rust](https://doc.rust-lang.org/book/nightly-rust.html)
+- [Nightly Rust](https://doc.rust-lang.org/nightly/)
```
coriolinus: This link doesn't point to documentation _about_ nightly Rust, but to the documentation overview page for the day's nightly build.

exercises/parallel-letter-frequency/README.md
sjwarner-bp: No particular reason - would you agree with [https://doc.rust-lang.org/book/first-edition/testing.html](https://doc.rust-lang.org/book/first-edition/testing.html) as the right place to link to?

exercises/parallel-letter-frequency/README.md
sjwarner-bp: Ah, my mistake. Would you agree with [https://doc.rust-lang.org/book/first-edition/release-channels.html](https://doc.rust-lang.org/book/first-edition/release-channels.html) as a good place to direct users? I cannot see anything more specific regarding nightly releases.

coriolinus: Yes, with the caveat that once the second-edition book's "nightly" appendix is written, that would be better yet. As of right now, your first-edition link is the best option available.

exercises/parallel-letter-frequency/README.md
coriolinus: That link doesn't appear to contain the word "benchmark" at all, making it IMO a poor choice for a link purporting to describe benchmark tests. 

Oddly enough, I can't currently find any current first- or second- edition chapter which talks about benchmark tests. The most current version I could find is <https://doc.rust-lang.org/1.16.0/book/benchmark-tests.html>; apparently the chapter was removed without replacement in 1.17.0. 

My advice for now would be to provisionally use the 1.16.0 link for now without giving up the search for something current.

coriolinus: Aha, found the correct link for this: <https://doc.rust-lang.org/stable/unstable-book/library-features/test.html>. Please use this one.

coriolinus:
The changes look good to me, but Travis won't pass it yet, because the README no longer matches what's auto-generated. You should be able to fix Travis by making these changes also to `HINTS.md`. Once that's done and Travis passes, we should be good to go.

PR #417, opened 2017-12-17
isbn-verifier: Include more succinct description of ISBN-10

petertseng:
Fixes https://github.com/exercism/problem-specifications/issues/1011.

A simple definition of what the system should do before getting in to
exactly what the digits of an ISBN mean would be good, in my opinion.

https://github.com/exercism/problem-specifications/pull/1019


petertseng:
Huh, guess the regenerate README only picked up this one for this repo. Fine with me.

Note that Bob is not included but that requires a change to the example solution so it deserves separate treatment.

PR #418, opened 2017-12-17
bob 1.1.0: Add yelling question rule

petertseng:
Closes https://github.com/exercism/problem-specifications/issues/943.

It was possible for a message directed to Bob to be both a question and
yelling, which meant it was unclear which rule should take effect.
Resolve ambiguity by simply adding a rule for that case.

https://github.com/exercism/problem-specifications/pull/1025


petertseng:
Oh just great. I typo'd the branch name. Oh well, too late to do anything about it now.

PR #348, opened 2017-08-29
adding links for Rust Recommended Learning Resources

gaurikholkar:
Adding a few links that helped me in the process of learning Rust

PR #419, opened 2017-12-31
Add rust-learning resource

ctjhoa:


ctjhoa:
rust-learning is reference in the official rust-lang documentation (https://www.rust-lang.org/en-US/documentation.html) and I think that it could help people with your exercises.

coriolinus:
Hi @ctjhoa, thanks for the contribution! I'm sorry it's taken so long to respond to this, but with the holidays there have been other claims on my time, and I suspect the same is true of the other moderation team members.

I am slightly opposed to merging this commit because I don't find big lists of links helpful. However, I am willing to be overridden if other maintainers feel that this could improve the track.

jgilray:
I think that since this link is fundamentally different than the others already there, it is probably a good addition.

petertseng:
Question as food for thought: Would a PR that simply adds something similar to "there is a list of learning materials at https://www.rust-lang.org/en-US/documentation.html" be accepted?

As a reminder, this link being added is one of the three links that would have been added by https://github.com/exercism/rust/pull/348 . One may state whether any reasons that applied then apply now.

If I think about it in a way similar to the web of trust model: I see that rust-learning is linked to from https://www.rust-lang.org/en-US/documentation.html. One may reasonably assume that I trust the maintainers of the rust-lang.org and that documentation page on it to list good resources. Since I haven't personally looked at the rust-learning list (my next opportunity to do so is probably on the order of "a few months from now"), I personally would not sign their keys. So I can't provide this in a "I personally endorse this link, it's goooood stuff" kind of way, and the best way I would be able to offer is "someone I think is reasonable has reasonably chosen to provide this link and maybe you'll find it helpful and maybe you won't, I can't do anything about it".

So I do not override the decision not to merge, but I wouldn't be opposed if someone else overrides it.

petertseng:
EBWODP

> "I personally endorse this link, it's goooood stuff" kind of way, and the best way I would be able to offer is "someone I think is reasonable has reasonably chosen to provide this link and maybe you'll find it helpful and maybe you won't, I can't do anything about it".

So the question is which of these two models we assume the http://exercism.io/languages/rust/learning has and I assumed it was the former, but maybe people would rather assume it's the latter.

coriolinus:
> Would a PR that simply adds something similar to "there is a list of learning materials at https://www.rust-lang.org/en-US/documentation.html" be accepted?

I'd support that, for two reasons:

1. It is in fact the official rust documentation page
2. It's short enough to take in at a glance; there's no need to search for anything. 

Neither reason applies to the link in this PR.

I like the web of trust model @petertseng proposes, and agree with his conclusion: I'm willing to endorse the official source, but am wary of endorsing this third-party repo as anything other than "the official documentation page links here".

coriolinus:
From current contributors, we have one vote in favor right now, one against, and a food for thought question. This issue has been open for two months. 

I am not comfortable being the tiebreaker in this case, because I've already voted. I do want to resolve this issue one way or another, though. If there is no further discussion on this by Friday 16 Feb, I'll close without merge. Otherwise, if there is discussion and some other course is chosen, I'll happily abide by it.

IanWhitney:
I support linking to just the Rust learning materials page.

ijanos:
I also support linking just the official Rust page, that way we do not have the pressure to keep our list up to date.

coriolinus:
Closing without merge: consensus is that we might add a link to https://www.rust-lang.org/en-US/documentation.html, but we will not link to an unofficial external repository which is a list of links.

PR #420, opened 2018-01-07
isbn-verifier: Add test for long numeric ISBN

nederdirk:
My initial [solution](http://exercism.io/submissions/fbfae36628fc4deb907fced6808362c2) would've failed this test case (a valid extended 11-test, adding an extra number, which would've been multiplied by a factor of 0, instead of X), so I added this case and a test for the length of the number.

### DIFF
--- a/exercises/isbn-verifier/tests/isbn-verifier.rs
+++ b/exercises/isbn-verifier/tests/isbn-verifier.rs
@@ -81,3 +81,8 @@ fn test_invalid_isbn_too_long() {
 fn test_invalid_isbn_with_check_digit_X_instead_of_0() {
     assert!(!is_valid_isbn("3-598-21515-X"));
 }
+
+#[test]
+fn test_invalid_isbn_too_long_with_number() {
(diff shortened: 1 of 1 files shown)


ijanos:
Hi! Thank you for your contribution. Please note that we have a separate repository for discussing changes in the exercises and their tests, please raise and issue or pull request there first, then we can go on accepting this. 

https://github.com/exercism/problem-specifications/tree/master/exercises/isbn-verifier

Otherwise, this pull request looks good to me, the CI failed because it expected this test to be `#[ignore]`-ed, we can fix that later :)

coriolinus:
@nederdirk to add to @ijanos' comment, I also believe that this looks like a good PR. We'd like you to make another PR to the [problem-specifications repo](https://github.com/exercism/problem-specifications/pulls) and have that merged upstream before merging this. That way, all tracks can benefit from the improvement, which contains nothing Rust-specific. 

Would you like a hand with the mechanics of that process?

coriolinus:
There's been no motion on this in over 5 months; it's timed out.

PR #421, opened 2018-01-14
scrabble-score: improve non-english letter tests

### LINKED ISSUE #264: Incomplete test coverage for scrabble-score
Sorry if I make any mistakes here. This is my first issue on exercism, and one of my first on GitHub. Furthermore, I'm actually rather new to TDD.

# Proposals:
## Proposal 1:
Add the following test to [scrabble-score](http://exercism.io/tracks/rust/exercises/scrabble-score):

    #[test]
    fn german_letters_do_not_score() {
        assert_eq!(score("STRASSE"), 7, "\"SS\" should score 2.");
        assert_eq!(score("STRAßE"), 5, "'ß' should score 0");
    }

## Proposal 2:
Additionally, modify the `non_english_scrabble_letters_do_not_score` test to something like:

    #[test]
    fn non_english_scrabble_letters_do_not_score() {
        assert_eq!(score("pinata"), 8, "'n' should score 1");
        assert_eq!(score("piñata"), 7, "'ñ' should score 0");
    }

# Motivation:
## Motivation for Proposal 1:
The exercise `scrabble-score` contains a test for ensuring only "English" letters is used, namely `non_english_scrabble_letters_do_not_score`. I assume that means all non-English letters shouldn't score. For example, the "word" `"ß"` should score 0!

Here's the problem: I originally thought the obvious solution is:

    fn codepoint_score(cp: char) -> u32 {
      if !cp.is_uppercase() {
        panic!("This method assumes uppercase!");
      } else { match cp {
        'A'|'E'|'I'|'O'|'U'|'L'|'N'|'R'|'S'|'T' => 1,
        'D'|'G'                                 => 2,
        'B'|'C'|'M'|'P'                         => 3,
        'F'|'H'|'V'|'W'|'Y'                     => 4,
        'K'                                     => 5,
        'J'|'X'                                 => 8,
        'Q'|'Z'                                 => 10,
        _                                       => 0,
      }}
    }

    pub fn score(word: &str) -> u32 {
      word.chars().map(|c| c.to_uppercase()).map(codepoint_score).sum()
    }

It doesn't compile due to `to_uppercase()` having an output-type of `ToUppercase` rather than `char`. So I resolved the compilation error by changing my code to:

    fn codepoint_score(cp: char) -> u32 {
        ...
    }

    fn letter_score(letter: char) -> u32 {
        letter.to_uppercase().map(codepoint_score).sum()
    }

    pub fn score(word: &str) -> u32 {
        word.chars().map(letter_score).sum()
    }

The above code correctly passes `non_english_scrabble_letters_do_not_score`, but also correctly fails the following test:

    #[test]
    fn german_letters_do_not_score() {
        assert_eq!(score("STRASSE"), 7, "\"SS\" should score 2.");
        assert_eq!(score("STRAßE"), 5, "'ß' should score 0");
    }

I don't know if anybody else would go this route. However since in Rust's `char.to_uppercase` returns a `ToUppercase`, which can be obviously checked one codepoint at a time. And since the `scrabble-score` exercise, contains the test `scoring_is_case_insensitive`, I at least think that such an approach might occur more often. Even if it doesn't. Including a test that tests for something like this might encourage other students to think about some more "out of the box" problems during coding and/or testing.

My proposal is therefore to add a test that would fail if any character that `char.to_uppercase` converts to two or more English letters gets a score greater than 0.

## Motivation for Proposal 2:
1. The test has two assertions, therefore if the test ever fails, the user needs to look up the line reference to determine why it failed. Simply adding the proposed information to the assertion should inform the user what went wrong without him needing to lookup the line reference.
2. The proposed version of the test is more self-documenting (i.e. it is easier for a reader to understand how each assertion in the test works).
3. This introduces a careful reader to the optional third parameter of `assert!`.

# Example:
For some reason I can't seem to attach files. See iteration >=3 of <http://exercism.io/submissions/b0f318070ab34df2aa3a03b0ecfcc0bb> for a complete example.

# Updates:
Several updates due to problems related to my being incapable attaching a file.

coriolinus:
Resolves #264.

Implements both of @dexterlemmer's proposals from that issue.

### DIFF
--- a/exercises/scrabble-score/tests/scrabble-score.rs
+++ b/exercises/scrabble-score/tests/scrabble-score.rs
@@ -71,6 +71,6 @@ fn all_letters_work() {
 #[test]
 #[ignore]
 fn german_letters_do_not_score() {
-    assert_eq!(score("STRASSE"), 7, "\"SS\" should score 2.");
+    assert_eq!(score("STRASSE"), 7, "\"SS\" should score 2");
     assert_eq!(score("STRAßE"), 5, "'ß' should score 0");
 }
--- a/exercises/scrabble-score/tests/scrabble-score.rs
+++ b/exercises/scrabble-score/tests/scrabble-score.rs
@@ -52,8 +52,8 @@ fn long_mixed_case_word() {
 #[test]
 #[ignore]
 fn non_english_scrabble_letters_do_not_score() {
-    assert_eq!(score("pinata"), 8);
-    assert_eq!(score("piñata"), 7);
+    assert_eq!(score("pinata"), 8, "'n' should score 1");
+    assert_eq!(score("piñata"), 7, "'ñ' should score 0");
 }
 
 #[test]
@@ -66,4 +66,11 @@ fn empty_words_are_worth_zero() {
 #[ignore]
 fn all_letters_work() {
     assert_eq!(score("abcdefghijklmnopqrstuvwxyz"), 87);
+}
+
+#[test]
+#[ignore]
+fn german_letters_do_not_score() {
+    assert_eq!(score("STRASSE"), 7, "\"SS\" should score 2.");
+    assert_eq!(score("STRAßE"), 5, "'ß' should score 0");
 }


exercises/scrabble-score/tests/scrabble-score.rs
```diff
@@ -67,3 +67,10 @@ fn empty_words_are_worth_zero() {
 fn all_letters_work() {
     assert_eq!(score("abcdefghijklmnopqrstuvwxyz"), 87);
 }
+
+#[test]
+#[ignore]
+fn german_letters_do_not_score() {
+    assert_eq!(score("STRASSE"), 7, "\"SS\" should score 2.");
```
petertseng: inconsistency: In all other lines added/modified in this PR, there was no trailing period, unlike this line which does.

PR #424, opened 2018-02-02
rna-transcription: Make DNA and RNA validate input on creation.

derekdreery:
PR to address issue #423.

The tests are re-ordered because it is natural to test validation before `to_rna` using the changed data structures.

PR #426, opened 2018-02-04
Inline the exercise README insert

kytrinyx:
Now that we're operating with a track-wide template for generating exercise
READMEs we no longer need a separate markdown document to use for hints
and instructions that apply to all the exercises in a track. These instructions
can be defined directly in the template.

The first commit regenerates the exercise READMEs to pick up tweaks to problem descriptions, which have been updated with various formatting changes, whitespace changes, and clarifications to the description.

That gives us a clean diff when inlining the template and regenerating.

Ref: https://github.com/exercism/meta/issues/94


petertseng:
I hereby propose these commit messages for the regenerated READMEs. If you reject this proposal, all you have to do is push cde2a0c to this branch again.

I was thinking of asking the question "so if we want to override a problem-specifications description without having to duplicate the track hint what do we do?" but the mechanism for that is to put `.meta/description.md`, great. One of these days I may even propose using a .patch file since I don't want to think of an entire file if I'm only changing a few lines from problem-specifications, but this a thing to think about later.

kytrinyx:
These commit messages look great to me. 🚢

petertseng:
Hello, in case anyone reading this in the future is concerned that #424 and #426 both contain "regenerate README" commits, let me allay your fears now. They were in fact identical commits (same SHAs) so `git merge` will only include one copy of each in the `master` tree, so all is well!

PR #429, opened 2018-02-06
reverse-string: Add test case for wide characters

### LINKED ISSUE #428: Additional test for reverse-string
There is a method of passing all the tests in the reverse-string excercise that fails completely in other situations. This is the solution I came up with, and realized afterwards that it will fail to reverse wide characters.
```rust
pub fn reverse(source: &str) -> String {
    if source == "" { return source.to_owned() }
    let mut rev = source.to_owned();
    let len = rev.len() - 1;
    let mut i = 0;
    while i < len / 2 {
        unsafe { rev.as_bytes_mut().swap(i, len - i); }
        i += 1;
    }
    rev
}
```
If all characters in the string provided are exactly 1 byte, this works as expected. However, if `source` was, for example, "運動" it fails because there are characters of multiple bytes. Obviously I'm aware this is a bad solution, however, I think it would be beneficial to have a test that prevents students from making this mistake and not realizing that it's a mistake at all. So, I propose that the following test is added:
```rust
#[test]
// wide characters
fn test_wide() {
    process_reverse_case("運動", "動運")
}
```

rrredface:
Closes #428 

"運動" changed to "子猫" as per request.

### DIFF
--- a/exercises/reverse-string/.meta/hints.md
+++ b/exercises/reverse-string/.meta/hints.md
@@ -1,0 +1,3 @@
+## Bonus
+Test your function on this string: `uüu` and see what happens. Try to write a function that properly
+reverses this string. Hint: grapheme clusters
--- a/exercises/reverse-string/README.md
+++ b/exercises/reverse-string/README.md
@@ -6,6 +6,11 @@ For example:
 input: "cool"
 output: "looc"
 
+## Bonus
+Test your function on this string: `uüu` and see what happens. Try to write a function that properly
+reverses this string. Hint: grapheme clusters
+
+
 ## Rust Installation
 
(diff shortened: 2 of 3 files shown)


coriolinus:
Travis failed because you didn't actually regenerate the readme after adding the hint. It's pretty easy:

- If you don't yet have configlet and are on a unixy system, go to the root of the rust repository and run `bin/fetch-configlet`. 
- If you don't yet have configlet and want to manually install, get it from [here](https://github.com/exercism/configlet/releases)
- From the root of the rust repository, run `configlet generate . -o reverse-string`. This will generate the readme for the reverse-string exercise.

Commit the changes it makes and push them; this will trigger a Travis rerun.

PR #432, opened 2018-02-15
robot-name: Move rand dependency to Cargo-example.toml

### LINKED ISSUE #431: Robot-name exercise needs a proper rand dependency
Right now, the version specification in the `Cargo.toml` is `0.3.12`, which due to how semver works, downloads `0.3.22`, which is a reexport of rand `0.4.0`. This might cause confusion for people not familiar with `rand`.

There are two ways to fix this:

- Update `Cargo.toml` to use `rand = "0.4.0" which is the newer version of `rand`.
- Fix the version number like so: `rand = "= 0.3.12", which will have better compatibility with older versions of Rust.

coriolinus:
Closes #431

Exercise has no library stubs which require the rand crate, so
dependency is removed. Students should learn to use crates.io
to search for dependencies such as this on their own.

### DIFF
--- a/exercises/robot-name/Cargo-example.toml
+++ b/exercises/robot-name/Cargo-example.toml
@@ -3,3 +3,4 @@ name = "robot-name"
 version = "0.0.0"
 
 [dependencies]
+rand = "0.3.12"


coriolinus:
travis

Travis seems to be having trouble at the moment: according to its metrics, the build ran for ~20 min after a delay of about the same time. Per the commit clock on github, I checked this in about an hour ago, which leaves another ~20 min unaccounted for.

Anyone know what might be causing this, or should we just chalk it up to a random traffic spike on the day that a new version of stable rust was released?

petertseng:
The only thing I found was on https://www.traviscistatus.com/

> As a result, we are currently seeing a backlog of public/open source builds on all our infrastructures.

I couldn't find any other information

PR #434, opened 2018-02-23
react: Express callback expectations precisely after each `set_value`

petertseng:
react: Express callback expectations precisely after each `set_value`

**Problem statement**:

Consider a test with two `set_value` calls and which expects that a
callback has, ultimately, been called with the two values, one for each
`set_value`.

The tests currently do not check that one value was added during each
`set_value` call. For all we know, maybe an implementation:

* magically manages to predict the future and calls the callback twice
  on the first `set_value` call, with the correct value.
* calls the callback zero times on the first `set_value` call, but twice
  on the second `set_value` call.

To more precisely define the `set_value` expectations, this commit uses
a `Cell`-based implementation to test callbacks.

---

**Discussion**:

Is this clearer to students than the old vector-based approach? It is
certainly true that the expectations are closer to the line of code that
could cause the expectations to fail; each `set_value` now shows exactly
what callback calls are expected as a result of it. I would argue this
is does increase test clarity.

Some questions I will have:

* Is there any potential for student confusion if they were to read the
  implementation of the `Expector` class? Any way to alleviate that
  confusion?
* A better name for `Expector`. It was chosen arbitrarily simply because
  it is something that expects.
* Whether this can in fact be done without Cell. I tried my best but
  could not find a way.

It behooves me to explain why I could not. If the reason is already obvious to you because of your knowledge of Rust, you can probably skip this part, then.

If the callback mutably borrows something (say, to record what value it was called with), the rest of the test code cannot borrow that same thing to check that the values are correct. Same if we try to go vice-versa: If the test code mutably borrows something to signal to the callback "I am going to call set_value now so it is safe for you to let one more value through", the callback cannot have immutably borrowed it to check that signal. So it seems the only sharing the two can do is through both of them immutably borrowing something.

If they must both immutably borrow something, it seems that at least one of them will nevertheless need to mutate something. ~~Thus, it seems to be the case that we must achieve dynamic (runtime-checked, instead of compile-time-checked) borrowing, by using a RefCell.~~
After experimentation, we can use a Cell, preferable because now there is no dynamic borrowing.

### DIFF
--- a/exercises/react/tests/react.rs
+++ b/exercises/react/tests/react.rs
@@ -2,6 +2,39 @@ extern crate react;
 
 use react::*;
 
+struct CallbackRecorder {
+    // Note that this `Cell` is https://doc.rust-lang.org/std/cell/
+    // a mechanism to allow internal mutability,
+    // distinct from the cells (input cells, compute cells) in the reactor
+    value: std::cell::Cell<Option<isize>>,
+}
+
+/// A CallbackRecorder helps tests whether callbacks get called correctly.
+/// You'll see it used in tests that deal with callbacks.
+/// The names should be descriptive enough so that the tests make sense,
+/// so it's not necessary to fully understand the implementation,
+/// though you are welcome to.
+impl CallbackRecorder {
+    fn new() -> Self {
+        CallbackRecorder {
+            value: std::cell::Cell::new(None),
+        }
+    }
+
+    fn expect_to_have_been_called_with(&self, v: isize) {
+        assert_ne!(self.value.get(), None, "Callback was not called, but should have been");
+        assert_eq!(self.value.replace(None), Some(v), "Callback was called with incorrect value");
+    }
+
+    fn expect_not_to_have_been_called(&self) {
+        assert_eq!(self.value.get(), None, "Callback was called, but should not have been");
+    }
+
+    fn callback_called(&self, v: isize) {
+        assert_eq!(self.value.replace(Some(v)), None, "Callback was called too many times; can't be called with {}", v);
+    }
+}
+
 #[test]
 fn input_cells_have_a_value() {
     let mut reactor = Reactor::new();
@@ -102,17 +135,13 @@ fn error_setting_a_compute_cell() {
 #[test]
 #[ignore]
 fn compute_cells_fire_callbacks() {
-    // This is a bit awkward, but the closure mutably borrows `values`.
-    // So we have to end its borrow by taking reactor out of scope.
-    let mut values = Vec::new();
-    {
-        let mut reactor = Reactor::new();
-        let input = reactor.create_input(1);
-        let output = reactor.create_compute(&[input], |v| v[0] + 1).unwrap();
-        assert!(reactor.add_callback(output, |v| values.push(v)).is_ok());
-        assert!(reactor.set_value(input, 3).is_ok());
-    }
-    assert_eq!(values, vec![4]);
+    let cb = CallbackRecorder::new();
+    let mut reactor = Reactor::new();
+    let input = reactor.create_input(1);
+    let output = reactor.create_compute(&[input], |v| v[0] + 1).unwrap();
+    assert!(reactor.add_callback(output, |v| cb.callback_called(v)).is_ok());
+    assert!(reactor.set_value(input, 3).is_ok());
+    cb.expect_to_have_been_called_with(4);
 }
 
 #[test]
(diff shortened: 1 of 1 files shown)


petertseng:
EBWODP:

> I tried my best but could not find a way.

It behooves me to explain why I could not. If the reason is already obvious to you because of your knowledge of Rust, you can probably skip this part, then.

If the callback mutably borrows something (say, to record what value it was called with), the rest of the test code cannot borrow that same thing to check that the values are correct. Same if we try to go vice-versa: If the test code mutably borrows something to signal to the callback "I am going to call `set_value` now so it is safe for you to let one more value through", the callback cannot have immutably borrowed it to check that signal. So it seems the only sharing the two can do is through both of them immutably borrowing something.

If they must both immutably borrow something, it seems that at least one of them will nevertheless need to mutate something. Thus, it seems to be the case that we must achieve dynamic (runtime-checked, instead of compile-time-checked) borrowing, by using a RefCell.

petertseng:
The possible outcomes that I expect from this discussion are the two that might immediately come to mind.

* Decide that this is an improvement, and use this implementation for all (or maybe only some?) of the tests in this file. I volunteer to make the change if this is so.
* Decide that this is not an improvement, and close this PR after having reached this conclusion.

petertseng:
EBWODP:

> enforce that set_value is indeed being called at precisely the right times

This is a massive error  that completely changes the meaning of this PR. The correct text is "enforce that the callbacks are indeed being called at precisely the right times"

petertseng:
EBWODP: 

> one test was converted

This doesn't seem like enough to get a feel for how this change will affect. I have now done two more.

petertseng:
This tells me that this is possible in Rust, which allays my fears about making such tests canonical as I suggest in https://github.com/exercism/problem-specifications/pull/1194#issuecomment-367816386 .I will probably decide it's better to be precise in the canonical data, even if the Rust track chooses not to do this.

coriolinus:
I need some more time to think this over; I do not yet have an opinion about this. 

I do, however, have a bit of dogshedding: it wasn't obvious to me at first what the distinction between `have_value` and `value` were. It feels to me like that would idiomatically be expressed in terms of an Option; was there a reason not to simply write the following?

```rust
struct Expector {
    value: RefCell<Option<isize>>,
}
```

exercises/react/tests/react.rs
```diff
@@ -101,32 +101,29 @@ fn error_setting_a_compute_cell() {
...
+            value: RefCell::new(None),
         }
     }
 
     fn expect_to_have_been_called_with(&self, v: isize) {
-        assert!(*self.have_value.borrow(), "Callback was not called, but should have been");
-        assert_eq!(*self.value.borrow(), v, "Callback was called with incorrect value");
-        *self.have_value.borrow_mut() = false;
+        assert_ne!(*self.value.borrow(), None, "Callback was not called, but should have been");
+        assert_eq!(*self.value.borrow(), Some(v), "Callback was called with incorrect value");
```
petertseng: I'd consider using https://doc.rust-lang.org/std/cell/struct.RefCell.html#method.replace in this line to remove the need for the line below it.

exercises/react/tests/react.rs
```diff
@@ -101,32 +101,29 @@ fn error_setting_a_compute_cell() {
...
-            value: RefCell::new(0),
+            value: RefCell::new(None),
         }
     }
 
     fn expect_to_have_been_called_with(&self, v: isize) {
-        assert!(*self.have_value.borrow(), "Callback was not called, but should have been");
-        assert_eq!(*self.value.borrow(), v, "Callback was called with incorrect value");
-        *self.have_value.borrow_mut() = false;
+        assert_ne!(*self.value.borrow(), None, "Callback was not called, but should have been");
```
petertseng: not necessary, given the below line, but maybe helpful for a better error message???

exercises/react/tests/react.rs
```diff
@@ -101,32 +101,29 @@ fn error_setting_a_compute_cell() {
...
     fn expect_not_to_have_been_called(&self) {
-        assert!(!*self.have_value.borrow(), "Callback was called, but should not have been");
+        assert_eq!(*self.value.borrow(), None, "Callback was called, but should not have been");
     }
 
     fn callback_called(&self, v: isize) {
-        assert!(!*self.have_value.borrow(), "Callback was called too many times");
-        *self.value.borrow_mut() = v;
-        *self.have_value.borrow_mut() = true;
+        assert_eq!(*self.value.borrow(), None, "Callback was called too many times; can't be called with {}", v);
```
petertseng: also possible use of https://doc.rust-lang.org/std/cell/struct.RefCell.html#method.replace

exercises/react/tests/react.rs
coriolinus: Yes, I think that would be natural.

exercises/react/tests/react.rs
coriolinus: Agreed.

exercises/react/tests/react.rs
coriolinus: Sure; to present a human-friendly error message in test code is definitely worth an extra line and a single repeated branch.

exercises/react/tests/react.rs
```diff
@@ -99,20 +99,42 @@ fn error_setting_a_compute_cell() {
     assert!(reactor.set_value(output, 3).is_err());
 }
 
+use std::cell::RefCell;
+struct Expector {
```
petertseng: `CallbackRecorder`, maybe.

petertseng:
Ah, as much as I think this is a cool thing to have, I see that https://blog.rust-lang.org/2018/02/15/Rust-1.24.html tells me `replace` was only stabilised in 1.24. Caution may be needed.

petertseng:
Now that I understand `Cell`and`RefCell` instead, I understand now that I can use `Cell`. I think `Cell`'s operations have been stable for longer (https://doc.rust-lang.org/std/cell/struct.Cell.html#method.replace has 1.17.0 by the name so I assume that's the release where it was made stable)

petertseng:
and https://blog.rust-lang.org/2017/04/27/Rust-1.17.html confirms the thought about Cell `replace` 1.17.

exercises/react/tests/react.rs
```diff
@@ -99,20 +99,41 @@ fn error_setting_a_compute_cell() {
     assert!(reactor.set_value(output, 3).is_err());
 }
 
+struct Expector {
+    value: std::cell::Cell<Option<isize>>,
```
petertseng: I'm only doing this because I thought If I just use `cell` here (with a `use std::cell::Cell` before, of course), what if a submitted implementation (which we `use react::*`) has a `Cell` ? will the conflicting names named in a `use` cause problems? I admit I never tried.

petertseng: It does **not** cause problems if `Cell` is specifically named:

```rust
mod m1 {
    #[derive(Debug)]
    pub struct A {
        pub a: usize,
    }
}

mod m2 {
    #[derive(Debug)]
    pub struct A {
        pub a: usize,
    }
}

use m1::*;
use m2::A;

fn main() {
    println!("Hello, world! {:?}", A{a: 1});
}
```

So, could go back to `use std::cell::Cell`, but perhaps I'll leave it this way for clarity in case someone doesn't know the rules. I didn't know off the top of my head.

coriolinus:
I've finally had the chance to think about this for long enough to understand it. I think these changes improve the exercise: 

- expected behavior is tested more precisely
- the mechanism of tests is a bit more elegant
- it gives students a working example of how (and when) to use `Cell`.

I'm not approving in the Github sense largely because this PR is a demo, not yet complete. It doesn't apply to all tests.

The most compelling objection is that this increases the cognitive load for students who want to understand how the tests work. My response to that is that this is a level 10 exercise, one of only three in the Rust track IIRC. Students who attempt this exercise are expected to be comfortable with advanced idioms, and they have the opportunity to learn something useful by studying this construction. 

Thank you, @petertseng, for coming up with this; it's very cool. I like it.

petertseng:
> The most compelling objection is that this increases the cognitive load for students who want to understand how the tests work. 

My response here is that I hope the names of the functions should make this clear, but I also propose to add some comments to help.

Commits coming that will use this for all the callback tests; I'll rely on GitHub's standard notifications for that. I do propose that I squash all these commits when merging.

petertseng:
A note I've discovered: All `let cb =` MUST come before `let mut reactor =`, otherwise there will be an error like the following:

```
error[E0597]: `cb` does not live long enough
   --> tests/react.rs:133:46
    |
133 |     assert!(reactor.add_callback(output, |v| cb.callback_called(v)).is_ok());
    |                                          --- ^^ borrowed value does not live long enough
    |                                          |
    |                                          capture occurs here
...
137 | }
    | - borrowed value dropped before borrower
    |
    = note: values in a scope are dropped in the opposite order they are created
```

So we'll have to live with the fact that the `let cb` will be at the top, rather than, say, right before the callback gets added.

petertseng:
Feel free to ask for extra newlines to be added to make the tests easier to read.

petertseng:
It is good manners for me to say that the improvement of more precisely saying which callbacks should be called for each `set_value` is not Rust-specific, and that https://github.com/exercism/problem-specifications/pull/1196 will apply it to all tracks.

petertseng:
EBWODP

> will apply it to all tracks.

That's a little deceptive. I mean it will apply it to the canonical data.

petertseng:
After rebasing for #454 I believe that `git diff 4fc6e7c e925ff2 -- exercises/react` told me that the only changes are to remove the useless `vec!`.

petertseng:
To avoid confusion on what I intend the final product's commit message to be (and to support another upcoming change to react), I've squashed down to one commit. If anyone wishes to see the old history, you may see at https://github.com/petertseng/exercism-rust/commits/e925ff28cba1abfc41956bb0382f265a886130ac. I do not believe any of that is worth keeping as separate commits since they were all exploratory.

petertseng:
As you can see, I am proposing moving `CallbackRecorder` to right before the first test involving callbacks, instead of at the very top of the file. Let me know if that's better.

coriolinus:
I will not merge now, so as to allow potential movement of the documentation block.

petertseng:
Confirmed via `cargo new a` and the following in `a/src/lib.rs/`

```rust
pub struct hi {
    a: usize,
}

/// hi is an impl
impl hi {
    fn a() -> usize {
        5
    }
}
```

`cargo doc` has no doc for a

(but it does if you put `/// hi is a struct` above the struct)

petertseng:
correction: the **index** has no doc for `hi`.

The specific documentation page for `hi`, however, does say, under the `impl hi` section, "hi is an impl".

For our case, it is better to put the doc on the struct.

petertseng:
Although I did not rebase this, the stub has not changed and by pushing a new commit this had the effect of making Travis test the merge result (https://travis-ci.org/exercism/rust/jobs/362275367 tested commit https://github.com/exercism/rust/commit/330f7d368b8c0fa765db473144c6e20929570052 which merged https://github.com/exercism/rust/commit/664236c77fc880073f809ccd1903ea408795381e, my commit, into https://github.com/exercism/rust/commit/ccc6f0fe471f68034ff6dd9f68401db027b2094dm, current master)

exercises/react/tests/react.rs
```diff
@@ -127,95 +156,96 @@ fn error_adding_callback_to_nonexistent_cell() {
...
+
+    let mut reactor = Reactor::new();
+    let input = reactor.create_input(11);
+    let output = reactor.create_compute(&[input], |v| v[0] + 1).unwrap();
+
+    let callback = reactor.add_callback(output, |v| cb1.callback_called(v)).unwrap();
+    assert!(reactor.add_callback(output, |v| cb2.callback_called(v)).is_ok());
+
+    assert!(reactor.set_value(input, 31).is_ok());
+    cb1.expect_to_have_been_called_with(32);
```
petertseng: Note as to how this applies to other languages:

Other languages would have been perfectly fine to use their language's equivalent of the old vector-based implementation, because they are free to check that the vector `== [32]` here. It was solely because I was unable to to do in Rust that I chose to use a Cell.

Other languages *need not* use Cell, but may still choose to if they feel it makes their tests more descriptive.

What all languages can indeed benefit from is some way of testing that the callbacks were called with value `32` (instead of waiting until the end to test this), no matter how that is implemented.

PR #436, opened 2018-02-25
gigasecond: Update dependency to conform to usage guidelines

cbzehner:
This updates the dependencies for the gigasecond exercise to conform to the [usage guidelines](https://docs.rs/chrono/0.4.0/chrono/#usage) for the [chrono](https://docs.rs/chrono/0.4.0/chrono) cargo package:
> Avoid using use chrono::*; as Chrono exports several modules other than types. If you prefer the glob imports, use the following instead:

I realize using `chrono::*` might have been an intentional decision to avoid having to expose beginners to importing `time::Duration` as well, but I think it probably makes more sense to mirror a real environment closely when exposing people to these exercises.

I tried to make the minimum number of changes here and ensured that `_test/check-exercises.sh` was able to validate the gigasecond exercise before submitting.

### DIFF
--- a/exercises/gigasecond/.meta/hints.md
+++ b/exercises/gigasecond/.meta/hints.md
@@ -1,0 +1,1 @@
+If you're unsure what operations you can perform on `DateTime<Utc>` take a look at the [chrono crate](https://docs.rs/chrono/0.4.0/chrono/) which is listed as a dependency in the `Cargo.toml` file for this exercise.
--- a/exercises/gigasecond/README.md
+++ b/exercises/gigasecond/README.md
@@ -4,6 +4,9 @@ Calculate the moment when someone has lived for 10^9 seconds.
 
 A gigasecond is 10^9 (1,000,000,000) seconds.
 
+If you're unsure what operations you can perform on `DateTime<Utc>` take a look at the [chrono crate](https://docs.rs/chrono/0.4.0/chrono/) which is listed as a dependency in the `Cargo.toml` file for this exercise.
+
+
 ## Rust Installation
 
 Refer to the [exercism help page][help-page] for Rust installation and learning
--- a/exercises/gigasecond/src/lib.rs
+++ b/exercises/gigasecond/src/lib.rs
@@ -1,7 +1,7 @@
 extern crate chrono;
-use chrono::*;
+use chrono::{DateTime, Utc};
 
 // Returns a Utc DateTime one billion seconds after start.
 pub fn after(start: DateTime<Utc>) -> DateTime<Utc> {
-    unimplemented!()
+    unimplemented!("What time is a gigasecond later than {}", start);
 }
--- a/exercises/gigasecond/example.rs
+++ b/exercises/gigasecond/example.rs
@@ -1,5 +1,5 @@
 extern crate chrono;
-use chrono::*;
+use chrono::{DateTime, Duration, Utc};
 
 pub fn after(start: DateTime<Utc>) -> DateTime<Utc> {
     start + Duration::seconds(1_000_000_000)
--- a/exercises/gigasecond/tests/gigasecond.rs
+++ b/exercises/gigasecond/tests/gigasecond.rs
@@ -13,7 +13,7 @@ extern crate gigasecond;
  * In order to use the crate, your solution will need to start with the two following lines
 */
 extern crate chrono;
-use chrono::*;
+use chrono::{TimeZone, Utc};
 
 #[test]
 fn test_date() {


exercises/gigasecond/Cargo.toml
```diff
@@ -4,4 +4,5 @@ version = "1.1.0"
 
 [dependencies]
 chrono = "0.4"
+time = "0.1"
```
coriolinus: The `time` crate has been depreciated. We shouldn't be adding new dependencies to it. Instead, let's use the `chrono` reexport of `Duration`; that way, when that library updates to use `std::time::Duration`, we upgrade with it for free.

exercises/gigasecond/src/lib.rs
```diff
@@ -1,7 +1,10 @@
-extern crate chrono;
-use chrono::*;
```
coriolinus: Agree that we should remove the glob import as bad style, but I think we can do better. What about:

```rust
use chrono::{DateTime, Duration};
```

This way we don't have to mess with a second dependency or globs.

coriolinus: This applies both to the lib stub and the example.

cbzehner:
Ah! That's much cleaner. I didn't realize `time` was deprecated. I'll update this PR with your suggestions.

exercises/gigasecond/src/lib.rs
```diff
@@ -1,7 +1,7 @@
-extern crate chrono;
-use chrono::*;
-
-// Returns a Utc DateTime one billion seconds after start.
-pub fn after(start: DateTime<Utc>) -> DateTime<Utc> {
-    unimplemented!()
-}
+extern crate chrono;
+use chrono::{DateTime, Duration, Utc};
```
cbzehner: `Duration` is an unused import here, but when I ran `cargo check` from `exercises/gigasecond` it didn't output any warnings so I left in in since it's a useful hint. Lmk what you think of this.

coriolinus: I have a mild preference to exclude `Duration` based only on an instinct that it's good for students to have to look at the docs at least a little bit. By providing the appropriate dependency crate and the stub which we have, we've already reduced the required work from the student to near nothing. 

I'm willing to be overridden on this if people feel strongly otherwise, though.

petertseng: The original angle that I wanted this exercise to take was to be some sort of introduction to crates (specifically for Rust) in addition to time/duration (applies to this exercise no matter what the language). That is what I thought in https://github.com/exercism/rust/pull/79#issuecomment-198557453. Note that at the time there was no stub at all. I may have forgotten about that when reviewing https://github.com/exercism/rust/pull/306 which added the stub. Encouraging a look at the docs seems good. As demonstrated by my having implicitly changed my mind about this a few times over the years, I don't feel strongly about it.

ijanos: I say remove it but add a note to the hints file with a link to the documentation, so students who get lost in the documentation (or don't know how to use it yet) won't get stuck.

cbzehner:
I'll update this to remove Duration and provide a hint linking to the documentation.

exercises/gigasecond/.meta/hints.md
```diff
@@ -0,0 +1 @@
+If you're unsure what operations you can perform on `DateTime<Utc>` take a look at the [chrono crate](https://docs.rs/chrono/0.4.0/chrono/) which is listed as a dependency in the `Cargo.toml` file for this exercise.
```
cbzehner: I wasn't sure if there was a specific format  for `.meta/hints.md` files. I looked at a couple exercism repos on Github but didn't see any. Let me know if this is alright.

coriolinus: There's no specific format necessary there; it's all just human-readable text. 

You will, however, need to regenerate the README for this to pass CI:

- Get the configlet tool. On mac/linux you can just use the bin/fetch-configlet script; on Windows, you have to find it and install it manually.
- Run `bin/configlet generate . --only gigasecond` to regenerate the readme
- Commit the new README.

cbzehner: Thanks for the guidance. I just updated the README.

exercises/gigasecond/README.md
```diff
@@ -4,6 +4,9 @@ Calculate the moment when someone has lived for 10^9 seconds.
 
 A gigasecond is 10^9 (1,000,000,000) seconds.
 
+If you're unsure what operations you can perform on `DateTime<Utc>` take a look at the [chrono crate](https://docs.rs/chrono/0.4.0/chrono/) which is listed as a dependency in the `Cargo.toml` file for this exercise.
```
cbzehner: I'm still a bit concerned this isn't a clear enough hint to walk users to adding `Duration` as an import from the `chrono` crate. When I look at the [crate documentation on Duration]:(https://docs.rs/chrono/0.4.0/chrono/#duration)
> Chrono currently uses the time::Duration type from the time crate to represent the magnitude of a time span...
It implies to me as a beginner in Rust that I should be doing the same thing `chrono` does and adding `time` to my `Cargo.toml` so I can also use the `time::Duration` type.

Duration doesn't appear in [reexports](https://docs.rs/chrono/0.4.0/chrono/#reexports) and the function most people end up using for this exercise `Duration::from` isn't apparent anywhere on the page.

Do y'all have any suggestions on how to help this hint provide a bit more structured guidance for people new to the Cargo ecosystem?

coriolinus:
The crate documentation for Duration is actually here
<https://docs.rs/chrono/0.4.0/chrono/struct.Duration.html>. The section you
linked to is indeed present in the chrono docs, but AFAICT it's stale.

In the end, people will just need to experiment. I haven't tried myself,
but I believe that attempting to use a time::Duration where a
chrono::Duration is expected will just fail with recent versions of chrono.
If people try that, and it fails, they'll need to continue looking at the
documentation until they find the solution which does work.

cbzehner:
Failed the test due to removing `Duration` from the `example.rs` file. Updated to resolve test failures.

exercises/gigasecond/README.md
petertseng: > I'm still a bit concerned this isn't a clear enough hint to walk users to adding `Duration` as an import from the `chrono` crate.

I see what you mean because there's a big dedicated section about durations being from `time` and being unclear on what relationship this has with https://docs.rs/chrono/0.4.0/chrono/struct.Duration.html.

I don't have any definitive answers, but how about some ideas?

* Shall we link to the relevant `+` operator? https://docs.rs/chrono/0.4.0/chrono/struct.DateTime.html#method.add-1
* Shall we link to the current Duration? https://docs.rs/chrono/0.4.0/chrono/struct.Duration.html

PR #437, opened 2018-02-25
nth-prime: Provide a function template

cbzehner:
I found it a bit confusing when I pulled down this exercise and didn't have a template for the function. Ended up pulling the function name and type out of the `tests/nth-prime.rs` file but I think having this in place would help myself and others.

One alternate approach would be to use a return type of `Result<usize, &'a str>` to allow `Err("Some message")` values.

The full function definition would then be:
```
pub fn nth<'a>(n: usize) -> Result<usize, &'a str> {
    unimplemented!()
}
```

I wasn't sure if this was better, worse or of no consequence since it introduces the `'a` lifetime annotations as well.

I tried to make the minimum number of changes here and ensured that `_test/check-exercises.sh` was able to validate the nth-prime exercise before submitting.

### DIFF
--- a/exercises/nth-prime/src/lib.rs
+++ b/exercises/nth-prime/src/lib.rs
@@ -1,0 +1,3 @@
+pub fn nth(n: u32) -> Option<u32> {
+    unimplemented!("What is the {}th prime number?", n)
+}
--- a/exercises/nth-prime/tests/nth-prime.rs
+++ b/exercises/nth-prime/tests/nth-prime.rs
@@ -2,29 +2,29 @@ extern crate nth_prime as np;
 
 #[test]
 fn test_first_prime() {
-    assert_eq!(np::nth(1), Ok(2));
+    assert_eq!(np::nth(1), Some(2));
 }
 
 #[test]
 #[ignore]
 fn test_second_prime() {
-    assert_eq!(np::nth(2), Ok(3));
+    assert_eq!(np::nth(2), Some(3));
 }
 
 #[test]
 #[ignore]
 fn test_sixth_prime() {
-    assert_eq!(np::nth(6), Ok(13));
+    assert_eq!(np::nth(6), Some(13));
 }
 
 #[test]
 #[ignore]
 fn test_big_prime() {
-    assert_eq!(np::nth(10001), Ok(104743));
+    assert_eq!(np::nth(10001), Some(104743));
 }
 
 #[test]
 #[ignore]
 fn test_zeroth_prime() {
-    assert!(np::nth(0).is_err());
+    assert_eq!(np::nth(0), None);
 }
--- a/exercises/nth-prime/example.rs
+++ b/exercises/nth-prime/example.rs
@@ -9,10 +9,10 @@ fn is_prime(n: u32) -> bool {
    return true;
 }
 
-pub fn nth(n: u32) -> Result<u32, ()> {
+pub fn nth(n: u32) -> Option<u32> {
     match n {
-        0 => Err(()),
-        1 => Ok(2),
+        0 => None,
+        1 => Some(2),
         _ => {
             let mut count: u32 = 1;
             let mut candidate: u32 = 1;
@@ -22,7 +22,7 @@ pub fn nth(n: u32) -> Result<u32, ()> {
                     count += 1;
                 }
             }
-            Ok(candidate)
+            Some(candidate)
         }
     }
 }


exercises/nth-prime/src/lib.rs
```diff
@@ -0,0 +1,3 @@
+pub fn nth(n: u32) -> Result<u32, ()> {
```
coriolinus: `rustc` should complain when asked to compile this stub, because `n` isn't actually used in the function. (`rustc` doesn't know that this is a stub function designed to be replaced by students!) 

While we don't have tooling in place which requires that the stubs and examples compile without warnings, the goal is to get there eventually. As such, please change `n` to either `_` or `_n`.

coriolinus: WRT the error type of the result, I'm of two minds.

On the one hand, `Result<T, ()>` is bad style: if you can get away with an empty tuple as your error type, you should be using `Option<T>` instead. 

On the other hand, it's legitimate to want to familiarize students with `Result`, even if we actually don't care what error type they use.

On the gripping hand, this exercise is number 6 of currently 80 exercises in Rust, per

```bash
$ jq '{slug: .exercises[].slug}' config.json | grep slug | nl | grep nth-prime
     6    "slug": "nth-prime"
```

Students attempting this exercise are still expected to be in the early stages of Rust learning. This implies that we _do_ want to provide a stub for them, and _do not_ want them to need to deal with traits, generics, or other advanced features. 

I think the signature I'd like to see looks like this:

```rust
pub fn nth(_: usize) -> Result<usize, &'static str> {
    unimplemented!()
}
```

(`'static` is a special lifetime for things which are known to last the entire lifetime of the program, such as string literals.) This still allows the `Err("Some message")`  usage without requiring people to write a generic lifetime just yet.

However, I'm not 100% convinced of my own correctness yet in this case, and would like to invite comments from other maintainers about this proposal.

petertseng: I agree with not showing `Result<T, ()>` early.

Now, it does lead me to suggest a possible option: Let's just change this execise to use `Option` instead of `Result`. Don't need `Result` yet.

`Result<usize, &'static str>` is OK to start. I'll attempt to predict some possible objections to them and respond:

* students may not know what `&'static` is.
  * Okay, then maybe we can explain it like we did in https://github.com/exercism/rust/blob/master/exercises/hello-world/src/lib.rs#L1
* it's bad style to use a string for an error; instead it should be a type that the caller can inspect to determine the cause of the error
  * Okay, maybe we can set the example to use a `NoSuchPrime` type, that's fine. Should the test suite explicitly check for that type, OR should it just look for `is_err`?

coriolinus: I'd be fine with just using `Option<T>`. That broadens the scope of this PR, but in a way that I'd be willing to accept. 

@cbzehner would you be willing to rework the exercise such that tests expect an `Option`?

cbzehner: Yes, I'd be happy to.

cbzehner:
Thanks, I'll try and split the code into two commits today or tomorrow!

exercises/nth-prime/tests/nth-prime.rs
```diff
@@ -2,29 +2,29 @@ extern crate nth_prime as np;
...
 fn test_big_prime() {
-    assert_eq!(np::nth(10001), Ok(104743));
+    assert_eq!(np::nth(10001), Some(104743));
 }
 
 #[test]
 #[ignore]
 fn test_zeroth_prime() {
-    assert!(np::nth(0).is_err());
+    assert_eq!(np::nth(0), None);
```
cbzehner: Changing this from `assert!` to `assert_eq!` was a good call, the error message goes from:
```
---- test_zeroth_prime stdout ----
	thread 'test_zeroth_prime' panicked at 'assertion failed: np::nth(0).is_none()', tests/nth-prime.rs:28:5
```
to the much clearer
```
---- test_zeroth_prime stdout ----
	thread 'test_zeroth_prime' panicked at 'assertion failed: `(left == right)`
  left: `Some(1)`,
 right: `None`', tests/nth-prime.rs:28:5
```

You can see the [full output](https://gist.github.com/cbzehner/5a7dede3a62048457aabebefc4bc007a#file-is_none_vs_none-sh) of both options here.

cbzehner:
Thanks for the help and suggestions y'all! I really appreciate all the work you put in getting this to a merge-ready state!

petertseng:
And thank you as well!

PR #440, opened 2018-02-27
prime-factors: Add solution template

neganp:


exercises/prime-factors/src/lib.rs
```diff
@@ -0,0 +1,3 @@
+pub fn factors(n: u32) -> Vec<u32> {
+    unimplemented!()
```
coriolinus: Please update to reference `n` so that there is no unused variable warning, [like this](https://github.com/exercism/rust/pull/438#issuecomment-368357841).

neganp: Oh nice! I didn't know you could do that. :+1:

neganp:
Cool, thanks! :slightly_smiling_face:

pravic:
@neganp Have you reviewed your code? The test uses a big number which does not fit into `u32`:

https://github.com/exercism/rust/blob/80d29881944fc06172ca21dbb699149c3f88a797/exercises/prime-factors/tests/prime-factors.rs#L42-L44

neganp:
Oh, good catch. Weird that it passed all my tests :confused: 

It looks like a `u64` is big enough on my platform. Would that work?

neganp:
Wait, this doesn't even pass my tests; how embarrassing :anguished: .

coriolinus:
This raises the question: why didn't Travis catch this?

neganp:
I looks like some of the tests don't run on at least some of the builds: https://travis-ci.org/exercism/rust/jobs/347483074#L2213

coriolinus:
I think that's a bug which we should fix.

PR #441, opened 2018-03-01
collatz-conjecture: Use Option<T> instead of Result<T, str>

petertseng:
**Preface**: I'd like to open this PR as a discussion of when we should use Option vs Result. The result of this discussion might also be applicable to perfect-numbers.

---

Since there is only one kind of error that can occur (a non-positive
number), no extra information is given by the `str`, and I do not like
to encourage using `str` as an error type.

I examined a user poll of when to use Result vs Option:
https://www.reddit.com/r/rust/comments/2j0k21/rust_option_vs_result/

One user says to use Option when there is only one failure mode, which
supports the approach taken in this commit.

However, in contrast, there are others who say Result is for any case
when you need to explain the failure. It is possible that we may want to
explain the failure even if there is only one way for a function to
fail.

So the question is whether this error warrants an explanation. Perhaps
any necessary explanation can be produced in the documentation... and
the documentation here does say take any **positive** integer.

### DIFF
--- a/exercises/collatz-conjecture/tests/collatz-conjecture.rs
+++ b/exercises/collatz-conjecture/tests/collatz-conjecture.rs
@@ -4,29 +4,29 @@ use collatz_conjecture::*;
 
 #[test]
 fn test_1() {
-    assert_eq!(Ok(0), collatz(1));
+    assert_eq!(Some(0), collatz(1));
 }
 
 #[test]
 #[ignore]
 fn test_16() {
-    assert_eq!(Ok(4), collatz(16));
+    assert_eq!(Some(4), collatz(16));
 }
 
 #[test]
 #[ignore]
 fn test_12() {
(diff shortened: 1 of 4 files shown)


exercises/collatz-conjecture/tests/collatz-conjecture.rs
```diff
@@ -4,29 +4,29 @@ use collatz_conjecture::*;
...
 fn test_1000000() {
-    assert_eq!(Ok(152), collatz(1000000));
+    assert_eq!(Some(152), collatz(1000000));
 }
 
 #[test]
 #[ignore]
 fn test_0() {
-    assert!(collatz(0).is_err());
+    assert_eq!(collatz(0), None);
```
petertseng: switched left/right here. should switch it to be consistent.

PR #435, opened 2018-02-24
no-op declare compliance with problem-specifications for any input object change

petertseng:
In https://github.com/exercism/problem-specifications/issues/996 it was proposed that inputs will be moved to the input object.

It was enacted across all the exercises.

For those exercises for which the input type is the only reason the Rust track is no longer up to date, those exercises are now declared to be up to date.

**No attempt** was made to update any exercise that was out of date for **any other reason**. The author has no interest in writing the test generators at this time, nor making any changes by hand.

The up-to-date determination was done in the usual manner (https://github.com/petertseng/exercism-problem-specifications/blob/up-to-date/up-to-date/rust.rb).

All commit contents were generated by the following script.
With the exception of queen-attack and robot-simulator (written by hand), all commits messages were also generated by the following script

```ruby
require 'json'

pr = ARGV.last

repo_root = `git rev-parse --show-toplevel`.chomp
exercise = ARGV.size > 1 ? ARGV.first.delete(?/) : `pwd`.chomp.split(?/).last
ver = JSON.parse(File.read("#{repo_root}/../problem-specifications/exercises/#{exercise}/canonical-data.json"))['version']

version_file = "#{repo_root}/exercises/#{exercise}/Cargo.toml"

lines = File.readlines(version_file)
File.open(version_file, ?w) { |f|
  lines.each { |l|
    f.write(l.include?('version') ? "version = \"#{ver}\"\n" : l)
  }
}

m = "#{exercise} #{ver}: no-op declare compliance with problem-specifications

input object

https://github.com/exercism/problem-specifications/pull/#{pr}"

`git commit --all -m '#{m}'`

puts m
```

--- 

Question:

https://github.com/exercism/discussions/issues/133#issuecomment-367964538

> If test version lived in a separate file in `.meta`, I'd find it pretty easy to accidentally forget to update it when updating an exercise.

So, does this also apply to the version being in Cargo.toml?

PR #446, opened 2018-03-04
check-exercises: Output progress dot when DENYWARNINGS

petertseng:
Otherwise, Travis may assume we're hung if we don't produce output in
10+ minutes. Normally we don't think checking all exercises takes this
long, but this prepares for the future and prevents spurious failures in
instances when outside circumstances cause slowness.

This approach was suggested in response to such a failure in
https://github.com/exercism/rust/pull/435#issuecomment-368254549

### DIFF
--- a/_test/check-exercises.sh
+++ b/_test/check-exercises.sh
@@ -45,6 +45,9 @@ for exercise in $files; do
       # Compiler errors will still be shown though.
       # Both flags are necessary to keep things quiet.
       ./bin/test-exercise $directory --quiet --no-run
+      # Output a progress dot; otherwise Travis may assume we're hung,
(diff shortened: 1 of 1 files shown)


_test/check-exercises.sh
```diff
@@ -45,6 +45,9 @@ for exercise in $files; do
       # Compiler errors will still be shown though.
       # Both flags are necessary to keep things quiet.
       ./bin/test-exercise $directory --quiet --no-run
+      # Output a progress dot; otherwise Travis may assume we're hung,
+      # if we don't produce output in > 10 mins.
+      echo -n '.'
```
petertseng: nope, bad, now `$?` will always be 0. Move this before `bin/test-exercise` please.

PR #448, opened 2018-03-08
Added armstrong-numbers exercise

shingtaklam1324:
Currently, it is after `series` and before `collatz-conjecture`. I'm not 100% sure on the positioning but the exercise is complete

Problem Specification
---
https://github.com/exercism/problem-specifications/tree/master/exercises/armstrong-numbers

### DIFF
--- a/exercises/armstrong-numbers/Cargo.toml
+++ b/exercises/armstrong-numbers/Cargo.toml
@@ -1,0 +1,6 @@
+[package]
+name = "armstrong_numbers"
+version = "1.0.0"
+authors = ["Shing Tak Lam <shingtaklam1324@gmail.com>"]
+
+[dependencies]
--- a/exercises/armstrong-numbers/Cargo.toml
+++ b/exercises/armstrong-numbers/Cargo.toml
@@ -1,6 +1,3 @@
 [package]
 name = "armstrong_numbers"
-version = "1.0.0"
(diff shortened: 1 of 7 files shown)


exercises/armstrong-numbers/src/lib.rs
```diff
@@ -0,0 +1,3 @@
+pub fn is_armstrong_number(_num: u32) -> bool {
```
coriolinus: Was there a particular reason to use `u32` instead of `usize`?

exercises/armstrong-numbers/Cargo.toml
```diff
@@ -0,0 +1,3 @@
+[package]
+name = "armstrong_numbers"
+version = "0.0.0"
```
coriolinus: This should match the version from `canonical-data.json`: `1.0.0`.

coriolinus:
Thank you for your contribution, @shingtaklam1324! I've made a few comments, which appear above. In general, this looks good; I should have time for a proper review in a few hours.

exercises/armstrong-numbers/src/lib.rs
shingtaklam1324: It's strictly because of the tests. They overflow `u16` but could fit comfortably in `u32`

Issue #442, opened 2018-03-01
Remove underscores from exercise stub 


cbzehner:
Per discussion that began on https://github.com/exercism/rust/pull/438. 

Some of the exercise stubs currently use `_` as a parameter for their input. This is confusing for beginners since `_` binds any pattern (it serves as the default case in matches) and is prefixed to a variable to suppress an underscore.

This is currently used in function stubs as a way of reducing the initial number of compiler warnings encountered on the Rust track and most of these stubs have a body of `unimplemented!()` and nothing more.

However, in most cases we should be able to easily turn the function input into a regular parameter _and_ avoid compiler warnings by referencing it in the body of our function. So the normal stub:
```
pub fn exercise(_: &str) -> bool {
  unimplemented!()
}
```
becomes
```
pub fn exercise(input: &str) -> bool {
  unimplemented!("There's an easy way to reference your input variable: {}", input)
}
```

A quick `rg -e 'pub fn \w+\(_' --count | sort | cut -f 1 -d '/'` shows which exercises are affected (caveat: this might miss some).
* [ ] alphametics
* [ ] book-store
* [ ] crypto-square
* [ ] decimal
* [ ] diffie-hellman
* [ ] isbn-verifier
* [ ] poker
* [ ] reverse-string
* [ ] series

petertseng:
I looked for all multi-line stub definitions:

```
$ git grep fn exercises/*/src/lib.rs | grep -v '{'
exercises/two-bucket/src/lib.rs:pub fn solve(capacity_1: u8,
```

I verified that two-bucket has no underscores.

There are two categories of things we'd like to remove: plain underscore `_` and underscore-prefixed names such as `_input`.

The former:

```
$ git grep 'fn.*(_:'    (03-17 22:30)
exercises/alphametics/src/lib.rs:pub fn solve(_: &str) -> Option<HashMap<char, u8>> {
exercises/book-store/src/lib.rs:pub fn lowest_price(_: &[usize]) -> usize {
exercises/crypto-square/src/lib.rs:pub fn encrypt(_: &str) -> String {
exercises/decimal/src/lib.rs:    pub fn try_from(_: &str) -> Option<Decimal> {
exercises/poker/src/lib.rs:pub fn winning_hands<'a>(_: &[&'a str]) -> Option<Vec<&'a str>> {
$ git grep 'fn.*,_:'    (03-17 22:31)
zsh: exit 1     git grep 'fn.*,_:'
```

[...]

PR #449, opened 2018-03-09
Added Complex Number exercise

shingtaklam1324:
Currently it is after `isbn-verifier` and before `perfect-numbers`.

Problem Specification
---
https://github.com/exercism/problem-specifications/blob/master/exercises/complex-numbers


exercises/complex-number/.meta/description.md
```diff
@@ -0,0 +1,3 @@
+Describe your exercise here
```
coriolinus: You should remove this file; it's confusing configlet.

coriolinus: You'll need to regenerate the README once you do this.

exercises/complex-number/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+pub struct Complex {
+    // Add in the fields
+}
+
+impl Complex {
+    pub fn new(_re: f64, _im: f64) -> Complex {
+        unimplemented!()
+    }
+
+    pub fn exp(&self) -> Complex {
```
coriolinus: I wouldn't know what needs to be implemented in this operation. I see `exp` and think exponentiation, but there's no argument passed in, and that's not what the example does.

Please add a short comment to this function describing what operation should be performed.

exercises/complex-number/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+        unimplemented!()
+    }
+
+    // Return the real part
```
coriolinus: All the doc comments look good, but they should begin with triple forward slashes (`///`) so that `rustdoc` can use them.

exercises/complex-number/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+        unimplemented!()
+    }
+    
+    // Return the conjugate
+    pub fn conjugate(&self) -> Complex {
+        unimplemented!()
+    }
+}
+
+// Implement `std::ops` for `Complex`
```
coriolinus: I want to invite discussion from the other maintainers: should we provide stubs for the ops functions, or require students to look up and import them?

In the [decimal exercise](https://github.com/exercism/rust/tree/master/exercises/decimal) we do not provide the stubs, but that is significantly later in the track. I could be persuaded in either direction for this exercise. What do you guys think?

Note that because that exercise doesn't include the stubs, it requires the use of `.meta/ALLOWED_TO_NOT_COMPILE`; this would also require that in case we choose to take out these stubs.

exercises/complex-number/src/lib.rs
shingtaklam1324: I decided to use `Complex::exp` for `e^(a + bi)` to be consistent with what the [standard library](file:///home/shingtaklam1324/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/share/doc/rust/html/std/primitive.f64.html#method.exp) does there. It seems to be more common to use `exp` for `e^x` and `pow(i / f / c)` for `a^b` within Rust's ecosystem

shingtaklam1324:
I see what the problem with this was...

When I was using `init_exercises.py`, I passed in `complex-number` instead of `complex-numbers`, and it couldn't find the files. I'll fix that manually now.

shingtaklam1324:
huh? Travis is complaining about an outdated README but as far as I can tell, the README + Hints match up...

shingtaklam1324:
If we do decide to add more test cases, that would become an issue, but with the current set of tests, all of the square roots are *integers*. The ones using `within_range` is because of the `sin` and `cos` functions of `std::f64`, where `sin(PI) != 1` and `cos(PI) != 0`.

exercises/complex-numbers/.meta/hints.md
```diff
@@ -0,0 +1,2 @@
+- To implement `ops`, take a look at overloading operators in [Standard Library Documentation for `std::ops`](https://doc.rust-lang.org/std/ops/index.html).
```
coriolinus: It may be worth inserting a header line here, to better differentiate the sections in the output document, i.e.

```markdown
## Hint
```

This is not essential and will not affect the remainder of this review.

exercises/complex-numbers/src/lib.rs
```diff
@@ -0,0 +1,71 @@
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct Complex {
+    // Add in the fields
+}
+
+impl Complex {
+    pub fn new(_re: f64, _im: f64) -> Complex {
```
coriolinus: Please rewrite this to remove the underscores per #442/#476.

coriolinus: Please also apply this to all initial-underscore variables in the file.

shingtaklam1324:
I'm not really that free at the moment, so I'll do this as soon as I'm free, probably when exams end (2 mo.>)

coriolinus:
@shingtaklam1324 Are you still interested in working on this?

shingtaklam1324:
Yeah, but I'm busy with exams atm and won't be free until 20th June.

shingtaklam1324:
Actually I don't think I will have enough time to work on this so I'm gonna close it for now.

PR #451, opened 2018-03-10
book-store: use integer number of cents, not floating-point numbers

### LINKED ISSUE #450: book-store should use integer number of cents, not floating-point numbers.
As suggested in https://github.com/exercism/problem-specifications/issues/964 and ratified in https://github.com/exercism/problem-specifications/pull/1089, book-store should use integer number of cents. The Rust track should do the same.

TASK:

* Change the tests in https://github.com/exercism/rust/blob/master/exercises/book-store/tests/book-store.rs to use integer number of cents instead of floating-point numbers by multiplying by 100. The reason we do this is to avoid testing exact floating-point equality, a practice we do not want to encourage by example. For example, `8.0` should change to `800`.
* Change the stub in https://github.com/exercism/rust/blob/master/exercises/book-store/src/lib.rs to use an integer type instead of `f64`, so that students downloading this exercise will receive a stub listing the correct type.
* Change the example in https://github.com/exercism/rust/blob/master/exercises/book-store/example.rs to perform calculations in terms of integers, so that we can verify that the changed tests were in fact changed correctly.
* Increase the version in https://github.com/exercism/rust/blob/master/exercises/book-store/Cargo.toml#L3 to match https://github.com/exercism/problem-specifications/blob/master/exercises/book-store/canonical-data.json#L3 so that we can keep track of the fact that we are up to date.

coriolinus:
Closes #450.

### DIFF
--- a/exercises/book-store/example.rs
+++ b/exercises/book-store/example.rs
@@ -5,13 +5,10 @@ use std::hash::{Hash, Hasher};
 use std::mem;
 use std::cell::RefCell;
 
-extern crate noisy_float;
-use noisy_float::prelude::*;
-
 type Book = usize;
 type GroupedBasket = Vec<Group>;
-type Price = f64;
-const BOOK_PRICE: Price = 8.0;
+type Price = usize;
+const BOOK_PRICE: Price = 800;
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 struct Group(RefCell<BTreeSet<Book>>);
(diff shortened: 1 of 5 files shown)


exercises/book-store/example.rs
```diff
@@ -96,10 +93,9 @@ fn hash_of(basket: &GroupedBasket) -> u64 {
 
 pub fn lowest_price(books: &[Book]) -> Price {
     DecomposeGroups::new(books)
-        .map(|gb| r64(basket_price(&gb)))
```
petertseng: immediate reaction: Huh? Where did `r64` come from?

After reading docs: Oh, it's https://docs.rs/noisy_float/0.1.4/noisy_float/types/fn.r64.html . Carry on then.

coriolinus: TBH it caught me the same way just now when I was looking over the code. It's useful that this PR eliminates that dependency anyway, because if it hadn't, it may have been worth filing a follow-up issue to clarify what was happening there.

Issue #458, opened 2018-03-11
`crypto-square` exercise tests contradict `README.md`


Phippsaurus:
In the [`README.md`](https://github.com/exercism/rust/blob/master/exercises/crypto-square/README.md) for the exercise `crypto-square` it says:

> Output the encoded text in chunks.  Phrases that fill perfect rectangles
`(r X c)` should be output `c` chunks of `r` length, separated by spaces.
Phrases that do not fill perfect rectangles will have `n` empty spaces.
Those spaces should be distributed evenly, added to the end of the last
`n` chunks.

The [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/crypto-square/canonical-data.json) gives an example for this:

```json
    {
      "description": "54 character plaintext results in 7 chunks, the last two with trailing spaces",
      "property": "ciphertext",
      "input": {
        "plaintext": "If man was meant to stay on the ground, god would have given us roots."
      },
      "expected": "imtgdvs fearwer mayoogo anouuio ntnnlvt wttddes aohghn  sseoau "
}
```

Contrarily the [rust implementation](https://github.com/exercism/rust/blob/master/exercises/crypto-square/tests/crypto-square.rs) of that test looks like this:

```rust
fn test_example() {
    test(
        "If man was meant to stay on the ground, god would have given us roots.",
        "imtgdvs fearwer mayoogo anouuio ntnnlvt wttddes aohghn sseoau",
    )
}
```

I was about to create a pull request for that issue, but there are further discrepancies which are not quite clear. The rust tests strip numbers from the input, where the canonical data leaves them in. Furthermore the rust version strips non-ASCII characters, where the canonical data has no example for that case.

If somebody could clarify what the exercise is actually supposed to look like, I will adjust it accordingly.

DanT-Git:
lines 50-52 in the README use the word "spaces" but really mean missing characters or short lines. The requirement is to distribute the jaggedness so the encoded rows at the bottom are at most 1 char short of the first row(s).

I had trouble with this, so I made my own `test_jagged()` test.  Your code fails these tests, btw:

```rust
test("aaaa bbbb cccc d", "abcd abc abc abc");
test("aaaa bbbb cccc dd", "abcd abcd abc abc");
test("aaaa bbbb cccc ddd", "abcd abcd abcd abc");
```

If I can figure out how, I will do a pull request with this clarification and added tests.

Phippsaurus:
[...]

PR #459, opened 2018-03-12
Clarified README.md and added tests for jaggedness requirement. #458

dantho:
Fixes issue #458 (mostly)

### DIFF
--- a/exercises/crypto-square/README.md
+++ b/exercises/crypto-square/README.md
@@ -47,9 +47,10 @@ imtgdvsfearwermayoogoanouuiontnnlvtwttddesaohghnsseoau
 
 Output the encoded text in chunks.  Phrases that fill perfect rectangles
 `(r X c)` should be output `c` chunks of `r` length, separated by spaces.
-Phrases that do not fill perfect rectangles will have `n` empty spaces.
-Those spaces should be distributed evenly, added to the end of the last
-`n` chunks.
+Phrases that do not fill perfect rectangles will be `n` characters short.
+To minimize the jaggedness, the missing chars should be distributed evenly,
+added to the end of the last `n` chunks. The bottom row(s) should be, at most,
+one char short of the top row(s).
 
 ```text
 imtgdvs fearwer mayoogo anouuio ntnnlvt wttddes aohghn  sseoau 
--- a/exercises/crypto-square/tests/crypto-square.rs
+++ b/exercises/crypto-square/tests/crypto-square.rs
@@ -65,6 +65,14 @@ fn test_ignore_non_ascii_chars() {
 
 #[test]
 #[ignore]
+fn test_minimize_jaggedness() {
+    test("aaaa bbbb cccc d", "abcd abc abc abc");
+    test("aaaa bbbb cccc dd", "abcd abcd abc abc");
+    test("aaaa bbbb cccc ddd", "abcd abcd abcd abc");
+}
+
+#[test]
+#[ignore]
 fn test_long() {
     test(
         r#"


exercises/crypto-square/tests/crypto-square.rs
```diff
@@ -63,6 +63,14 @@ fn test_ignore_non_ascii_chars() {
     );
 }
 
+#[test]
+#[ignore]
+fn test_minimize_jaggedness() {
+    test("aaaa bbbb cccc d", "abcd abc abc abc");
```
shingtaklam1324: This test seems to have failed travis ci

exercises/crypto-square/README.md
```diff
@@ -47,9 +47,10 @@ imtgdvsfearwermayoogoanouuiontnnlvtwttddesaohghnsseoau
 
 Output the encoded text in chunks.  Phrases that fill perfect rectangles
 `(r X c)` should be output `c` chunks of `r` length, separated by spaces.
-Phrases that do not fill perfect rectangles will have `n` empty spaces.
-Those spaces should be distributed evenly, added to the end of the last
-`n` chunks.
+Phrases that do not fill perfect rectangles will be `n` characters short.
```
shingtaklam1324: The Description.md and README.md needs to match for configlet to not throw an error. Shouldn't this mean that the [exercise specification](https://github.com/exercism/problem-specifications/blob/master/exercises/crypto-square/description.md) will need to be updated first?

petertseng: indeed. And this would be a good thing to do because (as http://x.exercism.io/problems/crypto-square tells me) the Rust track is not the only track that would benefit from a changed explanation. To share our work so that others can benefit, the change should be made to https://github.com/exercism/problem-specifications/blob/master/exercises/crypto-square/description.md first, before being made here.

exercises/crypto-square/README.md
```diff
@@ -47,9 +47,10 @@ imtgdvsfearwermayoogoanouuiontnnlvtwttddesaohghnsseoau
...
+To minimize the jaggedness, the missing chars should be distributed evenly,
+added to the end of the last `n` chunks. The bottom row(s) should be, at most,
```
petertseng: I understand the intent, though the meaning of adding a missing character might imply that the bottom row gets longer. I'd phrase it either in terms of taking a character away or purely in terms of the bottom rows being shorter (which this PR is doing)

exercises/crypto-square/README.md
dantho: OK, I set off to fix the [exercise specification](https://github.com/exercism/problem-specifications/blob/master/exercises/crypto-square/description.md) and update [canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/crypto-square/canonical-data.json) with my jaggedness tests.  I learned that the Rust implementation DOES NOT MATCH the specification!

Here's an item from the formal spec:
```json
{
  "description": "54 character plaintext results in 7 chunks, the last two with trailing spaces",
  "property": "ciphertext",
  "input": {
    "plaintext": "If man was meant to stay on the ground, god would have given us roots."
  },
  "expected": "imtgdvs fearwer mayoogo anouuio ntnnlvt wttddes aohghn  sseoau "
}
```

And here's the rust implementation of this test:
```rust
#[test]
fn test_example() {
    test(
        "If man was meant to stay on the ground, god would have given us roots.",
        "imtgdvs fearwer mayoogo anouuio ntnnlvt wttddes aohghn sseoau",
    )
}
```
Note that the "expected" outputs DIFFER -- specifically the extra spaces on the last two chunks are missing for the rust implementation.  This inconsistency is a big part of the problem this PR was supposed to fix.  I will clarify the problem description in the exercise spec, then, raise the inconsistency as a separate issue in the rust implementation.

exercises/crypto-square/README.md
dantho: Turns out the Rust implementation of the exercism spec got this requirement wrong.  See my comment, above.

exercises/crypto-square/README.md
dantho: Here's my [exercism specification PR](https://github.com/exercism/problem-specifications/pull/1210).

dantho:
The problem turns out to be more complicated that unclear documentation.  See ongoing discussion in issue #458.

PR #460, opened 2018-03-12
Updated Difficulties in use in README.md

shingtaklam1324:
Difficulty three is in use by `saddle-points` and `isogram`

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ Note that:
 
 - `README.md` may be [regenerated](https://github.com/exercism/docs/blob/master/maintaining-a-track/regenerating-exercise-readmes.md) from Exercism data. The generator will use the `description.md` from the exercise directory in the [problem-specifications repository](https://github.com/exercism/problem-specifications/tree/master/exercises), then any hints in `.meta/hints.md`, then the [Rust-specific instructions](https://github.com/exercism/rust/blob/master/config/exercise-readme-insert.md). The `## Source` section comes from the `metadata.yml` in the same directory.  Convention is that the description of the source remains text and the link is both name and hyperlink of the markdown link.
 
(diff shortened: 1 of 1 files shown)


coriolinus:
I believe there also exist come difficulty 6 exercises.

Is it worth continuing to enumerate all difficulties in use? If we have a ten point scale, in my mind, we may as well use all ten points.

petertseng:
Yeah too bad, the cat's been out of the bag for a long time. That's really too bad because 1 4 7 10 was the easiest way to transition out of https://github.com/exercism/rust/blob/aea470a76a9c6f0005adebb2a00d969512da3b6e/problems.md and I hate having to AP around which difficulty a problem should be. Shrug.

coriolinus:
```shell
$ jq '.exercises[] | .difficulty' config.json | sort -n | uniq -c
      2 null
     18 1
      2 3
     45 4
      2 6
      4 7
      8 10
```

We currently use 6 of the 10 possible difficulties, the vast majority of which are either 1 or 4. This suggests two things:

1. We should update the readme in one of two ways:
    - Enumerate all current difficulties: 1, 3, 4, 6, 7, 10
    - Rephrase to eliminate the enumeration, and have people just assign a plausible difficulty to their exercises. I prefer this solution.
2. We should give some thought to increasing the number of challenging exercises in this track. They're the fun ones, once someone slogs through all 60+ exercises at and below difficulty 4.

petertseng:
I can't really justify listing `1, 3, 4, 6, 7, 10` in the README since there doesn't seem to be much rhyme or reason in why those numbers are included and 2, 5, 8, and 9 are excluded. 1, 4, 7, 10 made sense because it advanced from 1 to 10 in equally-spaced increments, and it was easier to make a decision when there were fewer choices. But I should have made Travis CI enforce that if I wanted that to remain the case.

Using more difficulties may help us understand how to order exercises better. I can't guarantee I'll spend a lot of time thinking where the boundaries between any two adjacent numbers should be, but I'll probably notice if an exercise is rated much too low or high.

Issue #462, opened 2018-03-15
react: Use specific error types in stub


petertseng:
Discussion in https://github.com/exercism/rust/issues/444#issuecomment-371985900 . Instead of using `()`, use a separate error type for each of the four public functions.

petertseng:
I have started work on this.

PR #463, opened 2018-03-16
react: return domain-specific errors (DO NOT MERGE, reference for #434 #464 merge ONLY)

petertseng:
Closes https://github.com/exercism/rust/issues/462.
Closes #464 by mutual exclusion.

Note to reviewers: Since this will inevitably conflict with https://github.com/exercism/rust/pull/434, I chose to make all commits on top of https://github.com/exercism/rust/pull/434. This assumes that https://github.com/exercism/rust/pull/434 is getting merged before this one.

You should ignore the commit from #434 in your review of this PR.

If you would instead prefer the changes described in this PR be merged before #434 (perhaps you can review this one very quickly but require more time to review #434), you must instead review #464. Then #434 will rebased on #464.

I have no particular preference on whether the five commits comprising this PR should be squashed (but please do not squash them together with #434).

### DIFF
--- a/exercises/react/example.rs
+++ b/exercises/react/example.rs
@@ -91,14 +91,11 @@ impl <'a, T: Copy + PartialEq> Reactor<'a, T> {
     }
 
     pub fn add_callback<F: FnMut(T) -> () + 'a>(&mut self, id: CellID, callback: F) -> Option<CallbackID> {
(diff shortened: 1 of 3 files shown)


petertseng:
No need to review this one, this one'll be used as reference for how I rebase whichever one gets merged second.

PR #464, opened 2018-03-16
react: return domain-specific errors

### LINKED ISSUE #462: react: Use specific error types in stub

petertseng:
Closes #462.

Details are in individual commit messages.

I have no particular preference on whether the five commits comprising this PR should be squashed.

### DIFF
--- a/exercises/react/example.rs
+++ b/exercises/react/example.rs
@@ -88,14 +88,14 @@ impl <'a, T: Copy + PartialEq> Reactor<'a, T> {
         })
     }
 
-    pub fn add_callback<F: FnMut(T) -> () + 'a>(&mut self, id: CellID, callback: F) -> Result<CallbackID, &'static str> {
+    pub fn add_callback<F: FnMut(T) -> () + 'a>(&mut self, id: CellID, callback: F) -> Option<CallbackID> {
         match self.cells.get_mut(id) {
             Some(c) => {
                 c.callbacks_issued += 1;
                 c.callbacks.insert(c.callbacks_issued, Box::new(callback));
-                Ok(c.callbacks_issued)
+                Some(c.callbacks_issued)
             },
-            None => Err("Can't add callback to nonexistent cell"),
+            None => None,
         }
     }
 
--- a/exercises/react/example.rs
+++ b/exercises/react/example.rs
@@ -9,6 +9,12 @@ pub enum SetValueError {
     ComputeCell,
 }
 
+#[derive(Debug, PartialEq)]
+pub enum RemoveCallbackError {
+    NonexistentCell,
+    NonexistentCallback,
+}
+
 struct Cell<'a, T: Copy> {
     value: T,
     last_value: T,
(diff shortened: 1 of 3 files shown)


exercises/react/src/lib.rs
```diff
@@ -11,6 +11,12 @@ pub enum SetValueError {
     ComputeCell,
 }
 
+#[derive(Debug, PartialEq)]
+pub enum RemoveCallbackError {
+    NonexistentCell,
```
petertseng: note: It apparently is the case that we never test the `NonexistentCell` case. That can be dealt with later.

petertseng:
I decided to dispense with any reviewer-facing considerations on which of #464 or #463 to review. I decide to offer only this PR #464 for review, and I will rebase either this PR or #434 accordingly, now that I have #463 as a reference for what the final product should look like.

exercises/react/example.rs
```diff
@@ -46,11 +58,11 @@ impl <'a, T: Copy + PartialEq> Reactor<'a, T> {
         self.cells.len() - 1
     }
 
-    pub fn create_compute<F: Fn(&[T]) -> T + 'a>(&mut self, dependencies: &[CellID], compute_func: F) -> Result<CellID, &'static str> {
+    pub fn create_compute<F: Fn(&[T]) -> T + 'a>(&mut self, dependencies: &[CellID], compute_func: F) -> Result<CellID, CellID> {
         // Check all dependencies' validity before modifying any of them,
         // so that we don't perform an incorrect partial write.
-        if !dependencies.iter().all(|&id| id < self.cells.len()) {
-            return Err("Nonexistent input");
+        if let Some(&invalid) = dependencies.iter().find(|&dep| *dep >= self.cells.len()) {
```
petertseng: There is in fact not a specific reason I wrote `find(|&dep| *dep >= self.cells.len())` as opposed to one of these two alternatives:

```
find(|&&dep| dep >= self.cells.len())
find(|dep| **dep >= self.cells.len())
```

I really don't know what I'm doing and which is better, so I need help on that.

exercises/react/tests/react.rs
```diff
@@ -206,10 +206,10 @@ fn removing_a_callback_multiple_times_doesnt_interfere_with_other_callbacks() {
     let output = reactor.create_compute(&[input], |v| v[0] + 1).unwrap();
     let callback = reactor.add_callback(output, |v| cb1.callback_called(v)).unwrap();
     assert!(reactor.add_callback(output, |v| cb2.callback_called(v)).is_some());
-    // We want the first remove to be Ok, but we don't care about the others.
+    // We want the first remove to be Ok, but the others should be errors.
```
petertseng: strictly speaking, this line shouldn't be in this PR; it's been wrong ever since I did https://github.com/exercism/rust/pull/191/commits/2602332e4d787dbe0c60eddf48a25c90ab82fce2 as the last commit right before merging https://github.com/exercism/rust/pull/191

petertseng:
Thanks for that!

PR #466, opened 2018-03-16
Switch travis config away from container-based infrastructure

kytrinyx:
The sudo: false option runs on older infrastructure; experimenting with sudo: true to see how it affects
the build time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ script:
 - "./bin/fetch-configlet"
 - "sh ./_test/ensure-readmes-are-updated.sh"
 - "./bin/configlet lint ."
-sudo: false
+sudo: true
 rust:
   - stable
   - beta


petertseng:
Since https://docs.travis-ci.com/user/reference/overview/ tells us the bootup time for `sudo: false` has approximate boot time of 1-6s as opposed to `sudo: true` having 20-50s, this implies the `sudo: true` option would have to have a significant speedup in order to result in benefits.

I would prefer not to deceptively write the `.travis.yml` to ask for `sudo` when it is not required, since it makes it harder for a reader of the file to understand what this project needs to build. If `sudo: true` is to be merged because of X benefit, a comment before that line should explicitly explain what X is.

kytrinyx:
@petertseng Yeah, I hear you! I just wanted to run an experiment—I didn't expect this to be merged (I should have said that up front).

Issue #444, opened 2018-03-03
What E to use in stubs for Result<T, E>?


petertseng:
In https://github.com/exercism/rust/issues/269 we decided that exercises should have stubs. In https://github.com/exercism/rust/issues/371 we further decided that the provided stubs should compile.

For all exercises that have function(s) expecting a `Result<T, E>`, this poses a question: What type should go in the stub for `E`?

* Use `()`, but have a comment explaining that the type *can* be changed. Currently used in https://github.com/exercism/rust/blob/master/exercises/react/src/lib.rs (rather too late, I would think).
* Use `()`, with no additional explanations.
* Use a nonsense placeholder such as `PUT_YOUR_ERROR_TYPE_HERE`. Not yet used in this track, because it would cause the stub not to compile. We are able to mark stubs as allowed to not compile (https://github.com/exercism/rust/blob/master/exercises/decimal/.meta/ALLOWED_TO_NOT_COMPILE) but it should be used sparingly.
* Use a domain-specific error type that is already defined for the student. Currently used in https://github.com/exercism/rust/blob/master/exercises/forth/src/lib.rs.
* Use a domain-specific error type, but stub out the definition. For example, perhaps we would have a `BowlingError` but leave `BowlingError` as simply `pub struct BowlingError {}`.
* Use `&'static str` with the idea that it's to be used as an error message. Currently used in https://github.com/exercism/rust/blob/master/exercises/variable-length-quantity/src/lib.rs
* Use an arbitrary choice, decided by the contributor who contributes the stub.
* Your suggestion here.

I'm starting to shy away from using `str` as an error type since it is difficult to programmatically inspect, but now need to think about which of the other options I prefer. Perhaps my reason is bogus and using `str` is okay (let me know!). A contributor once also said this as a tip, but did not provide reasoning, so I cannot evaluate the tip. https://github.com/exercism/rust/pull/259#issuecomment-28088196

Need to think about the goals: Are we trying to teach something via these stubs? Or is it precisely the minimum code needed to get it to compile?

petertseng:
I think the goals are:

* If stubs compile, students can run `cargo test` without having to reconstruct expected signatures from each test file.
* Since it's better to let CI enforce the fact that stubs compile, ALLOWED_TO_NOT_COMPILE should be used sparingly as possible.
[...]

PR #468, opened 2018-03-16
ocr-numbers: Return domain-specific error

petertseng:
We prefer domain-specific errors to empty tuple because the former are
programmatically inspectable. We would prefer to encourage students to
use domain-specific errors rather than empty tuple.

As discussed in #444.

---

One slight disadvantage is now it appears we have to be opinionated about what information is contained by the errors. The `InvalidRowCount` seems relatively uncontroversial, but one might imagine that the `InvalidColumnCount` could also include which line had an invalid column count. But, careful, that's hard to test because there are multiple lines with invalid column count, so should we accept all of them?

If we wanted to let students change what data the error variants will carry (perhaps they like to provide more information than what the test asks for), I guess we can have the tests just do a  `match { Error::InvalidRowCount(_) => ... }` and that would probably work (I haven't tested it). I'm not sure how much of a demand there is for this.

### DIFF
--- a/exercises/ocr-numbers/example.rs
+++ b/exercises/ocr-numbers/example.rs
@@ -2,16 +2,20 @@ use std::collections::BTreeMap;
 
 #[derive(Debug, PartialEq)]
 pub enum Error {
-    InvalidRowCount,
-    InvalidColumnCount,
+    InvalidRowCount(usize),
+    InvalidColumnCount(usize),
 }
 
 pub fn convert(input: &str) -> Result<String, Error> {
-    if input.lines().count() % 4 != 0 {
-        return Err(Error::InvalidRowCount);
+    let line_count = input.lines().count();
+    if line_count % 4 != 0 {
+        return Err(Error::InvalidRowCount(line_count));
     }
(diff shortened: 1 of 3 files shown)


coriolinus:
For the reason you mentioned, that line numbers get complicated quickly, I'd prefer to exclude them. We also don't include invalid column numbers in `InvalidRowCount`, and it seems like these errors should be symmeetrical.

petertseng:
last reviewed commit was: 81b003fe41a68ec284cd9182c86d76ecb01739d0 (GitHub shows this in the "View Changes" link beside the approval)

`git diff 81b003f 4e4978a exercises/ocr-numbers` says only diff in the ocr-numbers exercise since that review is the README regeneration

petertseng:
In cases where I can document that the diff does not meaningfully differ since last review, I will take the last review as authoritative and be merging.

petertseng:
diff-of-diff: `diff -u <(git diff -u 81b003f~2 81b003f) <(git diff -u 4e4978a~2 4e4978a)` shows no changes at all.

PR #469, opened 2018-03-16
all-your-base: Return domain-specific error

petertseng:
We prefer domain-specific errors to empty tuple because the former are
programmatically inspectable. We would prefer to encourage students to
use domain-specific errors rather than empty tuple.

As discussed in #444:
https://github.com/exercism/rust/issues/444

### DIFF
--- a/exercises/all-your-base/example.rs
+++ b/exercises/all-your-base/example.rs
@@ -1,5 +1,12 @@
 pub type Digit = u32;
 
+#[derive(Debug, PartialEq)]
+pub enum Error {
+    InvalidInputBase,
+    InvalidOutputBase,
+    InvalidDigit(Digit),
+}
+
 ///
 /// Convert a number between two bases.
 ///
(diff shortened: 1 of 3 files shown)


exercises/all-your-base/example.rs
```diff
@@ -31,15 +38,18 @@ pub type Digit = u32;
...
+        return Err(Error::InvalidInputBase);
+    }
+    if to_base < 2 {
+        return Err(Error::InvalidOutputBase);
     }
 
     // check that all digits are in the correct range specified by the base
-    if digits.as_ref().iter().any(|&num| num >= from_base) {
-        return Err("Digits invalid for input base");
+    if let Some(&invalid) = digits.as_ref().iter().find(|&num| *num >= from_base) {
```
petertseng: There is in fact not a specific reason I wrote `find(|&num| *num >= from_base)` as opposed to one of these two alternatives:

```
find(|&&num| num >= from_base)
find(|num| **num >= from_base)
```

I really don't know what I'm doing and which is better, so I need help on that.

coriolinus: I am also unaware of any particular style guide suggestion as to which of those alternatives is ideal, but I like the symmetry of having equal numbers of `&` and `*`.

petertseng:
last reviewed: e62e97ee3371fa8dd8fd32b6677756240b4e0ecf

regular diff: `git diff e62e97ee3371fa8dd8fd32b6677756240b4e0ecf 99e951d exercises/all-your-base` : README regeneration only.

diff-of-diff: `diff -u <(git diff e62e97ee3371fa8dd8fd32b6677756240b4e0ecf~ e62e97ee3371fa8dd8fd32b6677756240b4e0ecf) <(git diff 99e951d~ 99e951d)` - no change

petertseng:
thus, I'm taking the most recent approval as authorittative

petertseng:
I see that the stub compilation check has made itself useful already.

PR #470, opened 2018-03-16
 largest-series-product: Expect specific errors; add stub

petertseng:
To encourage students to be specific about error information.

Since specific errors were just added, stub is provided to help get
student running.

We decided that stubs are a good idea in
#269

We decided that domain-specific errors are a good idea in
#444

### DIFF
--- a/exercises/largest-series-product/example.rs
+++ b/exercises/largest-series-product/example.rs
@@ -1,10 +1,16 @@
-pub fn lsp(string_digits: &str, span: usize) -> Result<u64, String> {
+#[derive(Debug, PartialEq)]
+pub enum Error {
+    SpanTooLong,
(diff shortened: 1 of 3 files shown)


petertseng:
Previous approval: eafd08111190608135613d1cd4ea9b4c5a5cd0e5

Diff: `git diff eafd08111190608135613d1cd4ea9b4c5a5cd0e5 781d0a1 exercises/largest-series-product` shows README regen only

diff-of-diff: `diff -u <(git diff eafd08111190608135613d1cd4ea9b4c5a5cd0e5~2 eafd08111190608135613d1cd4ea9b4c5a5cd0e5) <(git diff 781d0a1~2 781d0a1)` no diff

PR #474, opened 2018-03-17
Swaps all `_` only parameters for `input`

### LINKED ISSUE #442: Remove underscores from exercise stub

Baelyk:
Resolves #442 

Changed the parameter to `input`, and uses it in `unimplemented!()` to silence the unused variable warning.

Some of the exercises reported in the issue used `_` as a prefix and not as a lone parameter (e.g. in isbn-verifier, `_isbn`).  I wasn't if those were also a problem, so I didn't change them.

Changed exercises:
- [x] alphametics
- [x] book-store
- [x] crypto-square
- [x] decimal
- [ ] diffie-hellman
- [ ] isbn-verifier
- [x] poker
- [ ] ~reverse-string~
- [ ] series

(reverse-string was already changed)

### DIFF
--- a/exercises/alphametics/src/lib.rs
+++ b/exercises/alphametics/src/lib.rs
@@ -1,5 +1,5 @@
 use std::collections::HashMap;
 
-pub fn solve(_: &str) -> Option<HashMap<char, u8>> {
-    unimplemented!()
+pub fn solve(input: &str) -> Option<HashMap<char, u8>> {
+    unimplemented!("Solve the alphametric {}", input)
 }
--- a/exercises/alphametics/src/lib.rs
+++ b/exercises/alphametics/src/lib.rs
@@ -1,5 +1,5 @@
 use std::collections::HashMap;
 
 pub fn solve(input: &str) -> Option<HashMap<char, u8>> {
-    unimplemented!("Solve the alphametric {}", input)
+    unimplemented!("Solve the alphametric {:?}", input)
 }
--- a/exercises/alphametics/src/lib.rs
+++ b/exercises/alphametics/src/lib.rs
@@ -1,5 +1,5 @@
 use std::collections::HashMap;
 
 pub fn solve(input: &str) -> Option<HashMap<char, u8>> {
-    unimplemented!("Solve the alphametric {:?}", input)
+    unimplemented!("Solve the alphametic {:?}", input)
 }
--- a/exercises/book-store/src/lib.rs
+++ b/exercises/book-store/src/lib.rs
@@ -1,3 +1,3 @@
 pub fn lowest_price(input: &[usize]) -> usize {
-    unimplemented!("Find the lowest price of the bookbasket: {}", input)
+    unimplemented!("Find the lowest price of the bookbasket: {:?}", input)
 }
--- a/exercises/book-store/src/lib.rs
+++ b/exercises/book-store/src/lib.rs
@@ -1,3 +1,3 @@
-pub fn lowest_price(_: &[usize]) -> usize {
-    unimplemented!()
+pub fn lowest_price(input: &[usize]) -> usize {
+    unimplemented!("Find the lowest price of the bookbasket: {}", input)
 }
--- a/exercises/book-store/src/lib.rs
+++ b/exercises/book-store/src/lib.rs
@@ -1,3 +1,3 @@
-pub fn lowest_price(input: &[usize]) -> usize {
-    unimplemented!("Find the lowest price of the bookbasket: {:?}", input)
+pub fn lowest_price(books: &[usize]) -> usize {
+    unimplemented!("Find the lowest price of the bookbasket with books {:?}", books)
 }
--- a/exercises/crypto-square/src/lib.rs
+++ b/exercises/crypto-square/src/lib.rs
@@ -1,3 +1,3 @@
 pub fn encrypt(input: &str) -> String {
-    unimplemented!("Encrypt {} using a square code", input)
+    unimplemented!("Encrypt {:?} using a square code", input)
 }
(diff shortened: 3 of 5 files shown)


IanWhitney:
Thanks! You're close. I think if update  the `{  }` to `{:?}` the tests should pass.

Baelyk:
Okay! The tests now pass!

petertseng:
> Resolves #438

Please use #442 instead

petertseng:
> Some of the exercises reported in the issue used _ as a prefix and not as a lone parameter (e.g. in isbn-verifier, _isbn). I wasn't if those were also a problem, so I didn't change them.

I'm happy to reduce the scope of #442 to say only change the exact variable name `_`. As we can see in #438, the problem was it's not obvious to the student that it's necessary to change `_` since it can never be referenced.

In that case, we'd also want to open a separate issue to change the underscore-prefixed variable names, because it is although it is easy to see how to reference those variables, it is undesirable. By convention underscore-prefixed names indicate unused variables, and the only reason the underscores are present is to silence unused variable warnings.

In a similar vein, we'd like to remove any `#[allow(unused_variables)]` that we see.

Baelyk:
Oops I meant 442...

Is it okay if I open that issue?

petertseng:
> Oops I meant 442...

Don't forget to amend your commit message as well (not just your pull request description, which you have now edited). If not, I request the person who merges this please remember to change the issue number from 438 to 442.

> Is it okay if I open that issue?

I must ask to make sure it's ambiguous. You mean you want to open an issue that asks for underscore-prefixed variable names to be removed? If so, cool.

Baelyk:
Okay I updated the unimplemented text and squashed it down to one commit (I couldn't find another way to amend the original commit).

exercises/alphametics/src/lib.rs
```diff
@@ -1,5 +1,5 @@
 use std::collections::HashMap;
 
-pub fn solve(_: &str) -> Option<HashMap<char, u8>> {
-    unimplemented!()
+pub fn solve(input: &str) -> Option<HashMap<char, u8>> {
+    unimplemented!("Solve the alphametric {:?}", input)
```
petertseng: `alphametric` contradicts the spelling of the exercise, so the `r` should be removed

(I'm OK with this being `input`).

exercises/book-store/src/lib.rs
```diff
@@ -1,3 +1,3 @@
-pub fn lowest_price(_: &[usize]) -> usize {
-    unimplemented!()
+pub fn lowest_price(input: &[usize]) -> usize {
```
petertseng: `input` doesn't tell me what the meaning of the input is. I'd like to see `books`.

exercises/crypto-square/src/lib.rs
```diff
@@ -1,3 +1,3 @@
-pub fn encrypt(_: &str) -> String {
-    unimplemented!()
+pub fn encrypt(input: &str) -> String {
```
petertseng: I'm OK with this being `input`.

Another possibility I would accept is `plaintext`.

exercises/decimal/src/lib.rs
```diff
@@ -4,7 +4,7 @@ pub struct Decimal {
 }
 
 impl Decimal {
-    pub fn try_from(_: &str) -> Option<Decimal> {
-        unimplemented!()
+    pub fn try_from(input: &str) -> Option<Decimal> {
```
petertseng: I think I'm OK with `input` here too.

exercises/poker/src/lib.rs
```diff
@@ -2,6 +2,6 @@
 ///
 /// Note the type signature: this function should return _the same_ reference to
 /// the winning hand(s) as were passed in, not reconstructed strings which happen to be equal.
-pub fn winning_hands<'a>(_: &[&'a str]) -> Option<Vec<&'a str>> {
-    unimplemented!()
+pub fn winning_hands<'a>(input: &[&'a str]) -> Option<Vec<&'a str>> {
```
petertseng: I'd like this to tell me what each element of the slice is, so I'll say `hands` or `input_hands`

exercises/decimal/src/lib.rs
```diff
@@ -4,7 +4,7 @@ pub struct Decimal {
...
+        unimplemented!("Create a new decimal with a value of {:?}", input)
```
petertseng: I think I prefer you use `{}` here, since it will say `Create a new decimal with a value of 1.1` rather than `Create a new decimal with a value of "1.1"` and I'd rather see it without the quotes.

petertseng:
I request the one who merges this squash the commits.

IanWhitney:
Squashed & merged. Thanks for the help @Baelyk!

PR #477, opened 2018-03-18
Remove variables that start with an underscore

### LINKED ISSUE #476: Variable names that start with an underscore should be removed
Variables that start with underscores (e.g. `_isbn` in isbn-verify) should be removed. The reason they are there is to silence unused variable warnings, and there is a better way of doing that (see #442).

Exercises I know of from working on #474:

- [ ] diffie-hellman
- [ ] isbn-verifier
- [ ] series

Baelyk:
Resolves #476

Removed the underscore and included the variable in the `unimplemented!()` macro to get around unused variable warnings.


exercises/diffie-hellman/src/lib.rs
```diff
@@ -1,11 +1,11 @@
-pub fn private_key(_p: u64) -> u64 {
-    unimplemented!()
+pub fn private_key(p: u64) -> u64 {
+    unimplemented!("Pick a private key greater than and less than {}", p)
```
petertseng: This is missing something after `greater than`

(Since it would be difficult for a number to be simultaneously greater than `p` and less than `p`, which is what this sentence means if there's nothing after `greater than`)

exercises/diffie-hellman/src/lib.rs
```diff
@@ -1,11 +1,11 @@
...
+pub fn public_key(p: u64, g: u64, a: u64) -> u64 {
+    unimplemented!("Calculate public key using prime numbers {} and {}, and private key {}", p, g, a)
 }
 
-pub fn secret(_p: u64, _b_pub: u64, _a: u64) -> u64 {
-    unimplemented!()
-}
+pub fn secret(p: u64, b_pub: u64, a: u64) -> u64 {
+    unimplemented!("Calculate secret key using prime number {}, public key {}, and private key {}", p, b_pub, a)
+}
```
petertseng: there was a trailing newline in this file. Since this PR should be focused on dealing with underscore-prefixed names and not with whitespace, the trailing newline should not be removed in this PR

(also the trailing newline should stay anyway, by convention)

(Same comment for all other files)

PR #478, opened 2018-03-19
WIP: ensure-stubs-compile: Deny warnings in stubs

petertseng:


petertseng:
(note that the `touch $dir/src/lib.rs` is now no longer necessary, since the `sed` causes the stub to have been modified)

coriolinus:
Good point. You're the author of this PR, so you should feel free to adjust it as required; my intent was simply to eliminate the merge conflict so we could see if the checks would now pass.

coriolinus:
Oops: of course merging #497 hasn't changed this PR's Travis results, because this will need to be rebased onto the current head of master for that to work.

petertseng:
eh, any sort of rebase to make travis make a new merge commit; it just doesn't do that if the commit has already been tested. I wonder if there's a way. But rebasing on master's the most natural way anyway, since it's the most likely thing anyone would do for any other reason

PR #485, opened 2018-03-27
Improve README in simple-linked-list exercise

pravic:

PR #486, opened 2018-04-04
Fix formatting in simple-linked-list

pravic:
As pointed out in #485.

### DIFF
--- a/exercises/simple-linked-list/.meta/hints.md
+++ b/exercises/simple-linked-list/.meta/hints.md
@@ -15,7 +15,7 @@ struct Node<T> {
     next: Option<Box<Node<T>>>,
 }
 ```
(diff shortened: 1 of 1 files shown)


coriolinus:
Please incorporate these changes into #485 instead of using a second PR. Try this sequence of commands:

```sh
git checkout patch-1
git merge patch-2
git push
```

pravic:
Oh, come on.

coriolinus:
Rationale: merging a single PR is an atomic operation. Merging two PRs is not. In the former case, the codebase transitions from one consistent state to a different one. In the latter, there exists some amount of time during which the codebase is in an inconsistent state. 

More concretely, Travis won't approve either of these commits in isolation. I won't merge a PR which Travis doesn't approve. However, by combining them into a single PR, I'd expect them to pass Travis.

PR #472, opened 2018-03-17
Remove link to rust team page

Baelyk:
Resolves #414

Removed the link to the rust team page because it didn't work for non-team members.

Left the text "rust team members" but removed the link.

Issue #414, opened 2017-12-08
Link to rust team page is broken


phansch:
Almost all of the exercises link to the [Rust Team Page](https://github.com/orgs/exercism/teams/rust), but the link returns a 404.

When I go to [orgs/exercism/teams](https://github.com/orgs/exercism/teams) it tells me `You’re not a member of any teams in this organization.`

[There's a bunch of exercises with the link](https://github.com/exercism/rust/search?utf8=%E2%9C%93&q=%22https%3A%2F%2Fgithub.com%2Forgs%2Fexercism%2Fteams%2Frust%22&type=)

I guess this could also be an issue with other language repositories?

petertseng:
The intention behind putting that link there was good, but the execution could not match it, since teams are only visible to organisation members. GitHub probably has their reasons for doing this; maybe it simplifies logic of team views since some team members might choose not to publicise their membership.

I have two reasonable options to suggest:

1. Remove link with no replacement.
2. Replace link with https://github.com/exercism/rust/blob/master/config/maintainers.json.
3. Your suggestion goes here

coriolinus:
I thought that the point of maintaining config/maintainers.json was that
exercism was going to put up a more human-friendly page built from that
data, at some point. (Possibly only in nexercism?) If anything, I'd point
the link there.

Baelyk:
I hope this isn't considered grave digging, but

I think we should link to the [Getting Help](http://exercism.io/languages/rust/help) page on exercism.io.

Would that be a satisfactory solution?

In mentions the support Gitter which seems active enough and this issues page.

petertseng:
That's an interesting choice that http://exercism.io/languages/rust/help chooses to link to https://github.com/exercism/exercism.io/issues instead of https://github.com/exercism/rust/issues but depending on the situation, either of the two might be appropriate.

I can't think of any reasons to strongly support or strongly oppose adding a link to http://exercism.io/languages/rust/help.

The link to https://github.com/orgs/exercism/teams was added in https://github.com/exercism/rust/pull/138 for the purpose of showing students the track. The link to the list of maintainers is not essential for that purpose (but may make the repo seem more desirable?). Since it doesn't work, we lose little by removing it. It should be removed. The text can stay, it just doesn't need to be a link.

PR #487, opened 2018-04-04
regenerate all READMEs to remove link to GitHub team

petertseng:
https://github.com/exercism/rust/pull/472 removed the link from the
template because only org members can access the link. This commit
regenerates the README to do the same.

Discussion in https://github.com/exercism/rust/issues/414

It's my belief that this is the only change to the READMEs (no other
README changes from problem-specifications are pending) because upon
reverting #472 and attempting to regenerate, no other changes appear.


petertseng:
I predict that the up-to-date README check would block all other PRs to any other exercise, so this PR is necessitated first.

PR #488, opened 2018-04-04
crypto-square: Bring exercise implementation in line with README

### LINKED ISSUE #458: `crypto-square` exercise tests contradict `README.md`

coriolinus:
The initial implementation of this exercise diverged from the
README: it stripped out the extra padding spaces specified, and
it eliminated all non-alphabetical characters instead of just
getting rid of punctuation.

This commit rectifies both of those diversions. This closes #458.
However, in doing so it uses a method only introduced in Rust 1.24;
only those people who are relatively recent in their Rust compilers
will be able to compile the example.

As the example is not primarily for students, this should be acceptable.

### DIFF
--- a/exercises/crypto-square/example.rs
+++ b/exercises/crypto-square/example.rs
@@ -40,10 +40,18 @@ fn prepare(input: &str) -> Vec<char> {
     output.extend(
         input
             .chars()
-            .filter(|&c| c.is_ascii() && !c.is_whitespace())
-            .map(|c| c.to_ascii_lowercase())
-            .filter(|&c| c >= 'a' && c <= 'z'),
+            .filter(|&c| c.is_ascii() && !c.is_whitespace() && !c.is_ascii_punctuation())
+            .map(|c| c.to_ascii_lowercase()),
     );
+
+    // add space padding to the end such that the actual string returned
+    // forms a perfect rectangle
+    let (r, c) = dimensions(output.len());
+    let padding_qty = (r * c) - output.len();
+    for _ in 0..padding_qty {
+        output.push(' ');
+    }
+
     output.shrink_to_fit();
 
     output
--- a/exercises/crypto-square/tests/crypto-square.rs
+++ b/exercises/crypto-square/tests/crypto-square.rs
@@ -10,7 +10,6 @@ fn test_empty_input() {
     test("", "")
 }
 
-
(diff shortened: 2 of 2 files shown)


coriolinus:
Note that the example implementation does still strip non-ascii characters from the input. This is not specified by the README. However, the test which required that behavior has been removed, and no tests requiring strict compliance with the README have been added. As general policy here is to avoid non-ascii character tests unless there's a good reason otherwise, this is not expected to become an issue.

coriolinus:
At the time I initially implemented the exercise, there existed no canonical test data; I invented my own. I therefore rate it as highly unlikely that the tests are sufficiently in line with the canonical test data to declare compliance.

PR #490, opened 2018-04-07
circular buffer: remove allow(unused_must_use)

### LINKED ISSUE #475: circular_buffer: Shall we remove `allow(unused_must_use)`?
You will observe that https://github.com/exercism/rust/blob/master/exercises/circular-buffer/tests/circular-buffer.rs contains an `allow(unused_must_use)` to avoid compiler warnings (and thus a CI rejection, since our CI rejects warnings) from the fact that many Results from `write` are ignored.

Shall we instead use `let _ = buffer.write(...)` ?

The advantage is that it is more precise about exactly which lines are ignoring a `Result` rather than casting a wide net over the entire test module.

It would also allow us to remove an `mod tests` that is present in only circular-buffer probably for the sole reason of scoping the `#[allow(...)]`

The disadvantage is that since there may be other reasons to use `let _ = ...` it makes it hard to grep the entire repo to determine what's ignoring a Result and what is using `let _ = ...` for other reasons.

The `circular-buffer` was added in https://github.com/exercism/rust/pull/52 at which time there was no discussion of this point.

The other exercise that prominently uses `let _ = ...` is Bowling; the discussion may be seen around https://github.com/exercism/rust/pull/213#issuecomment-253384376 and a few surrounding comments (such as https://github.com/exercism/rust/pull/213#discussion_r83556987)

coriolinus:
This PR removes the compiler directive. Instead, all otherwise unused
`Result`s are now `unwrap()`'d.

Closes #475.

### DIFF
--- a/exercises/circular-buffer/tests/circular-buffer.rs
+++ b/exercises/circular-buffer/tests/circular-buffer.rs
@@ -1,113 +1,121 @@
 extern crate circular_buffer;
+use circular_buffer::{CircularBuffer, Error};
 
-#[allow(unused_must_use)]
-mod tests {
+#[test]
+fn error_on_read_empty_buffer() {
+    let mut buffer = CircularBuffer::<char>::new(1);
+    assert_eq!(Err(Error::EmptyBuffer), buffer.read());
+}
 
-    use circular_buffer::{CircularBuffer, Error};
+#[test]
+#[ignore]
+fn write_and_read_back_item() {
+    let mut buffer = CircularBuffer::new(1);
+    // Normally, using `unwrap` like this would be pretty bad practice:
+    // the call panics on error, which is bad behavior on the part of an
+    // executable or library.
+    //
+    // It's less of a problem within a test, for two reasons:
+    //  1. We control the entire test environment, and so should be able to
+    //     predict exactly when errors will and will not occur.
+    //  2. A panic here simply fails the test instead of aborting in the middle
+    //     of unrelated code.
+    //
+    // Finally, it underscores the fact that this isn't actually the behavior
+    // under test at the moment: when we want to test some behavior, we use assert.
+    // `unwrap()`, in this context, means that we're just trying to get to the point.
+    buffer.write('1').unwrap();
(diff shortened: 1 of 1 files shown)


exercises/circular-buffer/tests/circular-buffer.rs
```diff
@@ -1,113 +1,121 @@
...
+    // It's less of a problem within a test, for two reasons:
+    //  1. We control the entire test environment, and so should be able to
+    //     predict exactly when errors will and will not occur.
+    //  2. A panic here simply fails the test instead of aborting in the middle
+    //     of unrelated code.
+    //
+    // Finally, it underscores the fact that this isn't actually the behavior
+    // under test at the moment: when we want to test some behavior, we use assert.
+    // `unwrap()`, in this context, means that we're just trying to get to the point.
+    buffer.write('1').unwrap();
```
petertseng: This is the very first time `write` is encountered in this test file, and the function even has `write` in its name. That makes me wonder about the "this isn't actually the behavior under test" part of the comment.

The previous situation was that the `write` was being tested, but only indirectly: If the `write` had failed unexpectedly, we would see the subsequent `read` return an unexpected value.

Now, we are testing that the `write` succeeds, as we expect it to. I think this is better: It is good to notify as early as possible when we know there are failures.

If this is not a situation where we consider `write` part of the behaviour under test... under what kind of situation would we actually consider it to be?

I suppose my own answer to this question would be: a test entitled "full buffer can't be written to" as in https://github.com/exercism/problem-specifications/blob/master/exercises/circular-buffer/canonical-data.json#L116-L133, or something similar.

coriolinus: I see. I think I'm discovering a preference for `assert!(buffer.write(...).is_ok())`, because it's very explicit about what we expect, and it means that we can eliminate the entire section of comments justifying the use of unwrap.

petertseng: I'm good with that as well.

PR #493, opened 2018-04-08
Remove #[allow(unused_variables)]

### LINKED ISSUE #481: Use `unimplemented!` to reference variables, instead of `allow(unused_variables)`
We desire for our stubs to compile *without warnings*, as that helps us set a better example and prevent problems like #461 (a stub that provides a u32 where a u64 is needed).

In order for us to get there, these files that have `allow(unused_variables)` in them should stop using `allow(unused_variables)` and instead use the model established in https://github.com/exercism/rust/pull/438 https://github.com/exercism/rust/pull/474 and https://github.com/exercism/rust/pull/477, where in the `unimplemented!` macro we write a message saying what the function should do and reference the arguments.

Here are the files that should be modified as a result of this:

```
$ git grep unused_variables
exercises/all-your-base/src/lib.rs:#[allow(unused_variables)]
exercises/grade-school/src/lib.rs:#[allow(unused_variables)]
exercises/ocr-numbers/src/lib.rs:#[allow(unused_variables)]
exercises/react/src/lib.rs:#[allow(unused_variables)]
exercises/robot-simulator/src/lib.rs:    #[allow(unused_variables)]
exercises/robot-simulator/src/lib.rs:    #[allow(unused_variables)]
exercises/space-age/src/lib.rs:#![allow(unused_variables)]
```

You may choose to either work on these one at a time, or send one PR that does this for all the exercises.

coriolinus:
Closes #481.

Note that in the case of the react exercise, it wasn't possible to follow the standard procedure of adding a use of the variables in the unimplemented block. This is because react uses the generic types `F` and `T`, which are not bounded by the `Debug` trait. In that case, I fell back to the underscore-prefix pattern; students who reach a level 10 exercise should not be confused by it.

### DIFF
--- a/exercises/react/src/lib.rs
+++ b/exercises/react/src/lib.rs
@@ -1,5 +1,3 @@
-#[allow(unused_variables)]
-
 // Because these are passed without & to some functions,
 // it will probably be necessary for these two types to be Copy.
 pub type CellID = ();
@@ -20,7 +18,7 @@ pub enum RemoveCallbackError {
 pub struct Reactor<T> {
     // Just so that the compiler doesn't complain about an unused type parameter.
     // You probably want to delete this field.
-    dummy: T,
+    dummy: ::std::marker::PhantomData<T>,
 }
 
 // You are guaranteed that Reactor will only be tested against types that are Copy + PartialEq.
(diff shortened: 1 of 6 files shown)


exercises/react/src/lib.rs
```diff
@@ -99,6 +97,10 @@ impl <T: Copy + PartialEq> Reactor<T> {
     //
     // A removed callback should no longer be called.
     pub fn remove_callback(&mut self, cell: CellID, callback: CallbackID) -> Result<(), RemoveCallbackError> {
-        unimplemented!()
+        unimplemented!(
+            "Remove the callback identified by the CallbackId {:?} from the cell {:?}",
```
petertseng: `CallbackId`'s capitalisation is inconsistent with the fact that `CallbackID` is used a few lines above

PR #494, opened 2018-04-08
 ensure-stubs-compile: Deny warnings in tests

### LINKED ISSUE #461: Travis allows merges which include overflowing literals
This bit us in #440.

petertseng:
We'd eventually like to deny all warnings from the stubs themselves as
well, but this is a good intermediate state.
    
Immediate motivation is avoiding an overflowing literal, but I imagine
there are other unforeseen benefits as well.
    
Note that this required backing up and restoring the tests directory,
since we are editing it.

Closes #461 

Note for reviewers: This PR contains a commit that re-breaks the stub, to validate that this check in fact prevents such stubs. I'd like to see the CI fail with that commit, then that commit reverted, then the CI pass. The entire thing should be squashed when merging so that the net change is only to ensure-stubs-compile.

### DIFF
--- a/_test/ensure-stubs-compile.sh
+++ b/_test/ensure-stubs-compile.sh
@@ -16,12 +16,24 @@ for dir in $repo/exercises/*/; do
     # Touch the src/lib.rs file so that we surely recompile using the stub.
     touch $dir/src/lib.rs
 
+    # Backup tests; this script will modify them.
+    cp -r $dir/tests $dir/tests.orig
+
+    # Deny warnings in the tests that may result from compiling the stubs.
+    # This helps avoid, for example, an overflowing literal warning
+    # that could be caused by a stub with a type that is too small.
+    sed -i -e '1i #![deny(warnings)]' $dir/tests/*.rs
+
     if [ -f $allowed_file ]; then
       echo "$exercise's stub is allowed to not compile"
     elif ! (cd $dir && cargo test --quiet --no-run); then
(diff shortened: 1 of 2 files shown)


petertseng:
Oh, that is a subtlety.

The warning is in the *test* file. Okay, guess I'll add the `deny` to both!

petertseng:
Oh, awkward, now that I intend for this PR to have two commits in its merged state, I can't simply put a commit on, revert it, and ask for it to be squashed. Uh, okay, I'll show Travis CI links instead.

https://travis-ci.org/exercism/rust/builds/363763273 failed.
The commit was https://github.com/exercism/rust/commit/ef50ee5a19e2a754a58cdf6af01a452f9511137d . As you can see, it changes `u64` back to `u32`. 

In the Travis log, I see:

```
error: literal out of range for u32
  --> tests/prime-factors.rs:44:24
   |
44 |     assert_eq!(factors(93819012551), vec![11, 9539, 894119]);
   |                        ^^^^^^^^^^^
   |
note: lint level defined here
  --> tests/prime-factors.rs:1:9
   |
1  | #![deny(warnings)]
   |         ^^^^^^^^
   = note: #[deny(overflowing_literals)] implied by #[deny(warnings)]
```

petertseng:
Better idea is just to put `ensure-stubs-compile: continue early if allowed not to compile` in its own PR, really. I'll *start* on that, but since that's going to require three additional Travis CI builds, I'm just going to take this PR as-is if it gets approved in the meantime.

PR #497, opened 2018-04-08
Eliminate stub warnings

### LINKED ISSUE #484: use `unimplemented!` to reference variables, instead of having an unused variable warning
We desire for our stubs to compile *without warnings*, as that helps us set a better example and prevent problems like #461 (a stub that provides a u32 where a u64 is needed).

In order for us to get there, these files that have unused variables should instead use the model established in https://github.com/exercism/rust/pull/438 https://github.com/exercism/rust/pull/474 and https://github.com/exercism/rust/pull/477, where in the `unimplemented!` macro we write a message saying what the function should do and reference the arguments.

Here are the exercises that should be modified as a result of this (Determined via manual inspection of https://github.com/exercism/rust/pull/478:

- [ ] beer-song
- [ ] bob
- [ ] collatz-conjecture
- [ ] difference-of-squares
- [ ] forth
- [ ] grade-school
- [ ] grains
- [ ] leap
- [ ] pascals-triangle
- [ ] perfect-numbers
- [ ] proverb
- [ ] raindrops
- [ ] react
- [ ] saddle-points
- [ ] simple-linked-list
- [ ] sum-of-multiples
- [ ] two-bucket
- [ ] variable-length-quantity

For each exercise, the relevant file is the src/lib.rs file. For example, https://github.com/exercism/rust/blob/master/exercises/bob/src/lib.rs for bob.

You may choose to either work on these one at a time, or send one PR that does this for all the exercises.

coriolinus:
Closes #484 

Includes `_test/check-stubs.sh`, which discovers any compilation errors in the stubs.

Eliminates all stub warnings, not just unused variables.

simple-linked-list uses generics which are not bounded by `Debug`, so we can't use the standard "print it in `unimplemented!()` trick there. We fall back on underscore-prefixing affected variables in this case.

### DIFF
--- a/exercises/beer-song/src/lib.rs
+++ b/exercises/beer-song/src/lib.rs
@@ -1,7 +1,7 @@
 pub fn verse(n: i32) -> String {
-    unimplemented!()
+    unimplemented!("emit verse {}", n)
 }
 
 pub fn sing(start: i32, end: i32) -> String {
-    unimplemented!()
+    unimplemented!("sing verses {} to {}, inclusive", start, end)
 }
--- a/exercises/bob/src/lib.rs
+++ b/exercises/bob/src/lib.rs
@@ -1,3 +1,3 @@
 pub fn reply(message: &str) -> &str {
-    unimplemented!()
+    unimplemented!("have Bob reply to the incoming message: {}", message)
 }
--- a/_test/check-stubs.sh
+++ b/_test/check-stubs.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+# This checks every stub which exists, and emits all of their warnings.
+# To get a list of exercises for which there are warnings on compilation,
+# run it like this:
+#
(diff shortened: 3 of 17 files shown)


exercises/beer-song/src/lib.rs
```diff
@@ -1,7 +1,7 @@
 pub fn verse(n: i32) -> String {
-    unimplemented!()
+    unimplemented!("emit verse {}", n)
 }
 
 pub fn sing(start: i32, end: i32) -> String {
-    unimplemented!()
+    unimplemented!("sing verses {} to {}", start, end)
```
petertseng: could specify whether `end` is inclusive

exercises/bob/src/lib.rs
```diff
@@ -1,3 +1,3 @@
 pub fn reply(message: &str) -> &str {
-    unimplemented!()
+    unimplemented!("have bob reply to the incoming message: {}", message)
```
petertseng: could write Bob not bob

Issue #473, opened 2018-03-17
react: Investigate more type safety for input vs compute cells


petertseng:
A `Reactor` may issue `CellID`s in response to either a `create_input` or `create_compute`.

A `Reactor` accepts `CellID`s in the following circumstances:

* `create_compute` so that the caller specifies the dependencies of the compute cell to be created. These dependencies may be either input or compute cells.
* `value`. May be either input or compute cell.
* `set_value`. Only valid for an input cell.
* `add_callback`/`remove_callback`. Only valid for a compute cell. There is no technical barrier against an input cell having callbacks, but the specification does not call for it and it is never tested.

I believe it is technically feasible to have slightly more safety, as follows:

* `create_input` issues only `InputCellID`s.
* `create_compute` issues only `ComputeCellID`s
* `add_callback`/`remove_callback` accept `ComputeCellID` only.
* `set_value` accepts `InputCellID` only.
* `create_compute` and `value` need to accept `enum CellID { Input(InputCellID), Compute(ComputeCellID) }`

Should this be done? Is this a reasonable API and/or any other reasons it's a good/bad idea?

If we do this, is it further possible to test at compile-time that `InputCellID` and `ComputeCellID` cannot be assigned to each other? Even if it is possible, should we do that? I encourage making them unassignable to each other for more safety, but I don't want students to feel too restricted.

If I do not hear objections or think of any myself, I will investigate what the code looks like. I will not be offended if there were initially no objections but objections come to light after the seeing the code.

petertseng:
Initial investigation of what the code looks like present in https://github.com/petertseng/exercism-rust/commits/react-cell-ids . Note that I based it off a post-#434 and post-#464 state, since there would otherwise be too many merge conflicts that I'm not willing to write two versions of.

Also note that as of this writing, now `set_value` should consider returning something other than `SetValueError` since there is only one failure mode (cell does not exist); the other failure mode (setting a compute cell) is eliminated by the type system. That has not yet been decided.

coriolinus:
Having now actually attempted to implement an example of my API suggestion, I see that:

- `HasValue` makes more sense applied to `Cell` than to `CellID`
- `CellID` probably needs to remain an enum distinguishing the two cell types
- There is already a comment on `Reactor.value` to the effect that putting the value on the cell type was investigated and discarded for reasons of implementation complexity.

[...]

PR #499, opened 2018-04-09
react: Distinguish cell IDs

petertseng:
Closes https://github.com/exercism/rust/issues/473

~~I strongly advise that this should not be merged yet, for at least the following reason:~~

`set_cell_value` now can only fail in one way (cell does not exist) since it is no longer possible to designate an input cell to it. Therefore, it no longer needs to return `SetValueError`. We can just return a boolean, even.

Except for that point, I guess everything else is how I expect it to be.

I'm also happy to hear ideas about being able to move `set_value` onto input cells directory, in which case I don't see the need for this PR.

### DIFF
--- a/exercises/react/example.rs
+++ b/exercises/react/example.rs
@@ -1,7 +1,29 @@
 use std::collections::HashMap;
 
-pub type CellID = usize;
+/// `InputCellID` is a unique identifier for an input cell.
+pub type InputCellID = usize;
+/// `ComputeCellID` is a unique identifier for a compute cell.
+/// Values of type `InputCellID` and `ComputeCellID` should not be mutually assignable,
+/// demonstrated by the following tests:
+///
+/// ```compile_fail
+/// let mut r = react::Reactor::new();
+/// let input: react::ComputeCellID = r.create_input(111);
+/// ```
+///
+/// ```compile_fail
+/// let mut r = react::Reactor::new();
+/// let input = r.create_input(111);
+/// let compute: react::InputCellID = r.create_compute(&[react::CellID::Input(input)], |_| 222).unwrap();
+/// ```
+pub type ComputeCellID = usize;
 pub type CallbackID = usize;
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum CellID {
+    Input(InputCellID),
+    Compute(ComputeCellID),
+}
 
 #[derive(Debug, PartialEq)]
 pub enum SetValueError {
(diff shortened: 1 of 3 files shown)


petertseng:
EBWODP

> I strongly advise that this should not be merged

I mean to add the word "yet'. I'll fix the issue eventually.

petertseng:
Well, I guess this is it.

exercises/react/example.rs
```diff
@@ -1,12 +1,13 @@
 use std::collections::HashMap;
 
-pub type CellID = usize;
+pub type InputCellID = usize;
+pub type ComputeCellID = usize;
```
coriolinus: I note that these are typedefs, not newtypes. As such, the compiler treats `InputCellID`, `ComputeCellID`, and `usize` interchangeably. If we want it to enforce real type safety for us, we'll need to use newtypes:

```rust
pub struct InputCellID(usize);
pub struct ComputeCellID(usize);
// the above two types _cannot_ be interchanged
```

exercises/react/src/lib.rs
```diff
@@ -1,12 +1,13 @@
 // Because these are passed without & to some functions,
-// it will probably be necessary for these two types to be Copy.
-pub type CellID = ();
+// it will probably be necessary for these three types to be Copy.
+pub type InputCellID = ();
+pub type ComputeCellID = ();
```
coriolinus: It might be worth declaring these as 

```rust
pub struct InputCellID();
pub struct ComputeCellID();
```

to suggest the newtype pattern instead of the typedef pattern. 

Interestingly, I believe that the tests don't care; they should work whether these are type aliases or newtypes. Still, if we're in this for the type safety, we may as well do it right.

petertseng:
I would like to experiment with adding a compile_fail test (docmented in https://doc.rust-lang.org/stable/rustdoc/documentation-tests.html) to ensure IDs aren't mutually assignable.

exercises/react/src/lib.rs
```diff
@@ -16,8 +15,10 @@ pub type InputCellID = ();
 /// let input = r.create_input(111);
 /// let compute: react::InputCellID = r.create_compute(&[react::CellID::Input(input)], |_| 222).unwrap();
 /// ```
-pub type ComputeCellID = ();
-pub type CallbackID = ();
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct ComputeCellID();
+#[derive(Clone, Copy, Debug, PartialEq)]
```
petertseng: Note that the tests are going to need all of these. The reason why `Debug` and `PartialEq` should be obvious.

The reason for Copy (and therefore why Clone) is:

```
error[E0382]: use of moved value: `callback`
   --> tests/react.rs:203:52
    |
201 |     assert!(reactor.remove_callback(output, callback).is_ok());
    |                                             -------- value moved here
202 |     for _ in 1..5 {
203 |         assert_eq!(reactor.remove_callback(output, callback), Err(RemoveCallbackError::NonexistentCallback));
    |                                                    ^^^^^^^^ value used here after move
    |
    = note: move occurs because `callback` has type `react::CallbackID`, which does not implement the `Copy` trait
```

petertseng:
Thanks for reviewing!

PR #501, opened 2018-04-15
Palindrome Product Exercise

Menkir:
Hey Guys,

i just added the exercise palindrome product [https://github.com/exercism/problem-specifications/blob/master/exercises/palindrome-products/description.md](url)

Please give me feedback :)


coriolinus:
Thanks @Dev7353 for this work! 

Just going to summarize the Travis results here, as you may not be familiar with interpreting its output:

- All but the first test should have `#[ignore]` attributes
- The filename of the README should be `README.md`, not `README.MD` (note capitalization)

I'll perform a proper review later once Travis passes, but here are some items I noted in a quick glance:

- You haven't assigned a difficulty in `config.json`. It actually surprises me that `configlet lint` hasn't flagged that as an issue. Please assign an appropriate difficulty and insert the exercise in the appropriate location within `config.json`.
- I see that you've placed this exercise in `config.json` among the difficulty 10 exercises. That ranking surprises me; the example is fewer than 100 lines of straightforward code. I'm willing to be convinced that that difficulty is appropriate, but I'd like to hear your justification for it.
- Track [policy](https://github.com/exercism/rust/issues/269#issuecomment-343203839) prefers an actual stub in the stub file, not simply a blank file.

Dev7353:
Thanks, @coriolinus for the detailed feedback. I'm wondered as well why lint doesn't flag anything.
The reason why I set the difficulty for 10 was retrospectively nonsense. I struggled a bit with the duration of the testing because I needed to check millions of numbers so the test was running very long. Because I don't find a way to decrease the testing time I set the difficulty higher.

coriolinus:
We'll see what we can do about the testing run time. There are options to run the automated tests in release mode, which may be applicable here, and it may prove possible to improve the algorithm used in the example (though I haven't actually looked at that in any detail yet, so this is only speculation). In the meantime, you should pick a difficulty more in line with how challenging this exercise is to solve.

Yes, you'll need to specify all appropriate types and return values in the file. The [Forth exercise](https://github.com/exercism/rust/blob/master/exercises/forth/src/lib.rs), another difficulty-10 problem, does it well: all necessary types are defined, including the error enum, in a way which does not in any way reduce the difficulty of completing the exercise.

Dev7353:
Hi @coriolinus ,
I just checked out the travis log and it says:

>  Exercises with out-of-date README.md:$
>   palindrome-products

What does that mean? What is out of date?

coriolinus:
@Dev7353 That error means that README.md is out of sync with the data from upstream. The best solution here is to generate the readme instead of attempting to edit it by hand, because it's not always obvious what the differences are. Assuming you're on a unixy system, you can do so like this:

```shell
# start in the project root
# get configlet and put it into the bin directory
bin/fetch-configlet
bin/configlet generate . --only "palindrome-product"
```

exercises/palindrome-products/example.rs
```diff
@@ -0,0 +1,95 @@
...
+        let div = n/number;
+        if n % number == 0 &&  div <= max && div >= min && !factors.contains(&(div, number)){
+            factors.push((number, n/number));
+        }
+    }
+    factors
+}
+
+fn get_smallest_palindrome(min: i32, max:i32)-> Result<i32, Error>{
+    let l:Vec<i32> = (min*min .. max*max).collect();
```
coriolinus: This approach isn't very efficient, particularly as the difference between min and max grows. You'd do better with a nested for loop:

```rust
for i in min..(max+1) {
    for j in i..(max + 1) {
        if is_palindrome(i*j) {
            ...
        }
    }
}
```

exercises/palindrome-products/src/lib.rs
```diff
@@ -0,0 +1,17 @@
+#[derive(Debug, PartialEq)]
+pub enum Error{
+    Empty,
+    RangeFailure,
+}
+#[derive(Debug, PartialEq)]
+pub struct ReturnVals{
+    pub result: i32,
+    pub factors: Vec<(i32, i32)>,
```
coriolinus: It is not obvious to me what value there is in returning the factors here. Best case, there exist certain implementations for which one can get the factorization more or less for free when generating the palindrome. Worst case, the student has to go back and factorize their solutions after generation, which is at best an orthogonal problem.

I see that the canonical data has them. There's no real explanation of why. I therefore propose that for the Rust track, we diverge from the canonical data by dropping the factors from the implementation and the tests.

coriolinus: If we do this, we'd end up with a much nicer type for Palindrome:

```rust
pub type Palindrome = u64;
```

exercises/palindrome-products/src/lib.rs
```diff
@@ -0,0 +1,17 @@
+#[derive(Debug, PartialEq)]
+pub enum Error{
+    Empty,
+    RangeFailure,
+}
+#[derive(Debug, PartialEq)]
+pub struct ReturnVals{
+    pub result: i32,
```
coriolinus: No negative number can be a palindrome, so let's use an unsigned type for the palindromes.

exercises/palindrome-products/src/lib.rs
```diff
@@ -0,0 +1,17 @@
+#[derive(Debug, PartialEq)]
+pub enum Error{
+    Empty,
+    RangeFailure,
+}
+#[derive(Debug, PartialEq)]
+pub struct ReturnVals{
```
coriolinus: This should be named `Palindrome`, which is much more descriptive than `ReturnVals`.

exercises/palindrome-products/src/lib.rs
coriolinus: Let's use `u64`, in fact, because it's easier than you might expect to overflow a `u32`. It's fine if `min` and `max` are `u32`, however; this helps reduce the chance of an overflow.

petertseng:
To keep this PR focused on adding the Palindrome Product exercise instead of unrelated changes, let's not `configlet fmt` in this PR please.

exercises/palindrome-products/src/lib.rs
```diff
@@ -0,0 +1,12 @@
+pub type Palindrome = u64;
+pub fn get_palindrome_products(min: u64, max: u64) -> Vec<Palindrome>{
+    unimplemented!("Find all palindromic numbers which are products of numbers in the inclusive range ({}..{})", min, max)
+}
+
+pub fn min(palindromes: Vec<Palindrome>)->Option<Palindrome>{
```
coriolinus: Accepting `Vec<Palindrome>` here transfers ownership of that vector, negating the potential performance improvement of computing once and then getting min/max.

It would be possible to accept a `&Vec<Palindrome>` instead and get that back, but that's not considered idiomatic: it locks the caller into a `Vec` implementation, when it may be more efficient for them to generate a slice in other ways.

It's therefore best to just accept a slice reference directly: `&[Palindrome]`.

This applies to `max` as well.

exercises/palindrome-products/example.rs
```diff
@@ -0,0 +1,33 @@
...
+                palindromes.push(i*j);
+            }
+        }
+    }
+    palindromes
+}
+
+pub fn min(palindromes: Vec<Palindrome>)->Option<Palindrome>{
+    if palindromes.is_empty(){
+        return None;
```
petertseng: is it necessary? https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.min already returns `None` if empty.

exercises/palindrome-products/example.rs
```diff
@@ -0,0 +1,33 @@
...
+    }
+    palindromes.iter().cloned().min()
+}
+
+pub fn max(palindromes: Vec<Palindrome>)->Option<Palindrome>{
+    if palindromes.is_empty(){
+        return None;
```
petertseng: is it necessary? https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.max already returns `None` if empty.

exercises/palindrome-products/example.rs
Menkir: Oh, yeah, you're right. Thank you for the feedback!

exercises/palindrome-products/src/lib.rs
Menkir: I'm struggling a bit with the lifetimes. So when I return a slice over a vector i.e. palindrome.as_slice()  in get_palindrome_product() then I need to specify that the slice lives long enough. But I have no reference lifetime in my arguments. Should I give min and max as references or is there a more elegant approach?

coriolinus: You don't really need to use lifetimes here for this to work. I've thrown up a quick branch to demonstrate: [lib.rs](https://github.com/coriolinus/exercism_rust/blob/palindrome-products/exercises/palindrome-products/src/lib.rs), [palindrome-products.rs](https://github.com/coriolinus/exercism_rust/blob/palindrome-products/exercises/palindrome-products/tests/palindrome-products.rs)

coriolinus:
@Menkir On my machine, testing in release mode takes just over 8 seconds, while debug mode takes almost 2 minutes. That's a big enough discrepancy that it's worth testing this exercise in release mode. To accomplish this, please add a file `.meta/test-in-release-mode` which contains a short comment describing the timing.

This is important because the CI runs every exercise's tests multiple times, and for those which take non-trivial time, this can really speed up the CI results.

exercises/palindrome-products/src/lib.rs
coriolinus: There are a few performance improvements in there:

- [5](https://github.com/coriolinus/exercism_rust/blob/palindrome-products/exercises/palindrome-products/src/lib.rs#L5): `j` begins its lower bound at `i` instead of `min`. This eliminates roughly half the search space.
- [15, 19](https://github.com/coriolinus/exercism_rust/blob/palindrome-products/exercises/palindrome-products/src/lib.rs#L15): clone only the single item we care about, not the entire list. Doesn't matter so much for `Copy` types like `u64`, but for largeish things like strings, this really does speed things up.
- [26](https://github.com/coriolinus/exercism_rust/blob/palindrome-products/exercises/palindrome-products/src/lib.rs#L26): Perform `to_string()` only once. Do not copy the string. Producing a string requires heap allocation, which isn't quick. 
- [26](https://github.com/coriolinus/exercism_rust/blob/palindrome-products/exercises/palindrome-products/src/lib.rs#L26): Iterate over the bytes of the string instead of the characters. This is faster because when you use the `chars()` string iterator, it must produce a four-byte `char` on each iteration. For strings for which all chars are bytes, converting it into a byte vec and then iterating is faster. Naturally, if you can't guarantee that all chars in your string are in fact bytes, then you don't have any choice but to pay that price. Here, however, we know that all digits produced by the base-10 representation of a number are in fact a single byte wide.

Another performance improvement which I haven't implemented in this branch, but which should roughly halve the time required:

- Restructure tests to persist the list of palindrome products for min and max. That is, instead of having a pair of tests:

    ```rust
    #[test]
    fn smallest_palindrome_double_digits() {
        assert_eq!(min(&get_palindrome_products(10, 99)), Some(121));
    }
    #[test]
    fn largest_palindrome_double_digits() {
        assert_eq!(max(&get_palindrome_products(10, 99)), Some(9009));
    }
    ```

    We could persist the list of palindromes in a single test:

    ```rust
    #[test]
    fn palindromes_double_digits() {
        let pp = get_palindrome_products(10, 99);
        assert_eq!(min(&pp), Some(121));
        assert_eq!(max(&pp), Some(9009));
    }
    ```

    The part of these tests which takes the longest is searching a large number-space for palindrome products. By persisting the results of that search between the min and max checks, we can significantly reduce the time requires.

    There's a tradeoff to that approach, of course: it reduces the granularity of the tests. For long-running tests such as these, though, I believe that it's probably worth it.

Note that only the performance improvement which I didn't actually implement requires the use of a reference in the function API.

coriolinus: Another unimplemented performance improvement: zip up only the first half of the byes checked in `is_palindrome` with the second half of the bytes checked there. Right now, each byte pair is checked twice; that's not necessary.

coriolinus: It's worth mentioning that non-reference function arguments have fundamentally different semantics than in, say, go. In go, a non-reference function argument produces a copy, which would be expensive for a vector. In Rust, it _moves_ the variable. This is fast, but it means that the calling code no longer has access to that variable.

Menkir:
@coriolinus 

> Please exclude the configlet fmt changes, though: they introduce quite a lot of unrelated noise to this PR.

Would it be enough to revert this commit by a new one or should i rebase it?

coriolinus:
Simple reversion will be fine; we're going to squash-merge this in the end.

exercises/palindrome-products/tests/palindrome-products.rs
```diff
@@ -0,0 +1,58 @@
+extern crate palindrome_products;
+use palindrome_products::*;
+
+#[test]
+#[ignore]
```
coriolinus: The first test in the tests file should be the one not ignored. There are at least two ways to accomplish this: 

- move the `single_digits` test to the top of the file.
- ignore `single_digits` and unignore `empty_result_for_smallest_palindrome`.

I do not have a preference between these options.

Menkir:
@coriolinus  thank you for the big support. 
I should note that i used rustfmt 0.60 and it say that a line of size 101 chars is not allowed (only <= 100). So i split the string by hand. Maybe this is the issue?

coriolinus:
Ok, in that case, just undo the string split; we can deal with a long line
from a string literal now and then.

PR #502, opened 2018-04-16
Adjust config.json to return exercise difficulties to one of 1, 4, 7, 10

coriolinus:
Closes #460 by exclusion.

Performed with the following command line:

```shell
jq '.exercises |= map(
    if .difficulty == 3 then
       .difficulty=4
    elif .difficulty == 6 then
       .difficulty=7
    else
        .
    end
)' config.json > config.json.new && \
diff config.json config.json.new ; \
mv config.json.new config.json
```

### DIFF
--- a/config.json
+++ b/config.json
@@ -200,7 +200,7 @@
       "slug": "saddle-points",
       "core": false,
       "unlocked_by": null,
-      "difficulty": 3,
+      "difficulty": 4,
       "topics": [
         "vectors",
         "iterators"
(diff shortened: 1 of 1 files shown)


coriolinus:
It wasn't originally my preferred solution, but it grew on me. I'd rather take action and move on than leave it hanging.

petertseng:
I'm down. No harm in picking something and then changing it later on if it doesn't work.

If maybe 1, 4, 7, 10 aren't enough levels, perhaps it can be 1 (for hello world only) and then the five even levels. Not as much brainpower/AP as having to divide between 10 levels.

PR #503, opened 2018-04-16
Verify exercise difficulties

coriolinus:
Encompasses #502 and should probably be merged afterward.


coriolinus:
This PR should be squashed when merging.

PR #519, opened 2018-04-23
Diamond Exercise

Menkir:


Menkir:
@coriolinus i've got some weird cargo output:
______________________________________________________________________________
Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running /Users/kmg/projects/rust/exercises/diamond/target/debug/deps/diamond-9106e305cfc2658d

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

   Doc-tests diamond

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
______________________________________________________________________________

It seems that cargo cannot map the resources from diamond exercise. What do you think?

petertseng:
> running 0 tests

I'm sorry if this is a false alarm, but this wouldn't be because the directory should be called `tests`, wouldn't it?

coriolinus:
@Menkir try @petertseng 's suggestion; it seems plausible to me.

Menkir:
@coriolinus @petertseng 
Do you think the testsuite is reasonable?

coriolinus:
@Menkir yes, given that the test suite matches the tests of the canonical data, I'd say they look reasonable. 

I don't get notifications when you change the title or add commits, so once this is complete and passes Travis, feel free to ping me for a review.

Menkir:
@coriolinus & @petertseng 
Feedback please :)

exercises/diamond/example.rs
```diff
@@ -0,0 +1,63 @@
...
+        if e == c {
+            break;
+        }
+    }
+
+    //build second half
+    let mut rev = result.clone();
+    rev.pop(); //remove middle pice to avoid duplicates
+    for e in rev.iter().rev().into_iter() {
+        result.push((*e).clone());
```
coriolinus: You could probably avoid this second clone by using [`drain`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.drain) on `rev`.

exercises/diamond/example.rs
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    result
+}
+
+fn get_line(e: char, c: char) -> String {
```
coriolinus: It's not obvious to me reading this code what `e` and `c` are supposed to _mean_.

Here, it's fine; given that the code works, we can consider this to be more or less write-only code. If I were reviewing this at work, I'd demand better variable names, so that future maintainers will have a better chance of understanding it.

exercises/diamond/example.rs
```diff
@@ -0,0 +1,63 @@
...
+    r
+}
+
+fn get_letter_line(e: char) -> String {
+    let mut r = String::new();
+    let odd = (0..)
+        .filter(|x| x % 2 != 0)
+        .take(26)
+        .collect::<Vec<usize>>()
+        [ABC.find(e).unwrap()];
```
coriolinus: This will panic if someone passes in a lower-case letter. That would be fine if you had sanitized the input by either rejecting lower-case letters or converting everything to uppercase, but if that happens in this file, I've missed it.

exercises/diamond/example.rs
```diff
@@ -0,0 +1,63 @@
...
+    }
+    r
+}
+
+fn get_letter_line(e: char) -> String {
+    let mut r = String::new();
+    let odd = (0..)
+        .filter(|x| x % 2 != 0)
+        .take(26)
+        .collect::<Vec<usize>>()
```
coriolinus: Instead of this `take`->`collect`->`index` sequence, why not just use [`nth`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.nth)?

exercises/diamond/example.rs
```diff
@@ -0,0 +1,68 @@
+static ABC: &'static str = "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
+use std::ascii::AsciiExt;
+
+pub fn get_diamond(diamond_char: char) -> Vec<String> {
+    let diamond_char = AsciiExt::to_ascii_uppercase(&diamond_char);
+    if ABC.find(diamond_char).is_none() {
+        panic!("{:?} is not a valid char.", diamond_char);
```
coriolinus: I get the motivation here, but it's best to leave panics as a worst case after absolutely everything else has failed. It makes more sense to me to just return an empty vector in this case.

PR #520, opened 2018-04-25
Twofer exercise

AvasDream:
Hello, i created my first exercise and would really appreciate your feedback.

### DIFF
--- a/config.json
+++ b/config.json
@@ -13,6 +13,17 @@
         "println!"
       ]
     },
+	{
+      "uuid": "c6631a2c-4632-11e8-842f-0ed5f89f718b",
+      "slug": "twofer",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 1,
+      "topics": [
+        "Strings",
+		"match"
+      ]
+    },
     {
       "uuid": "f880b1ef-8f66-41f3-bb89-f39a4ed592a2",
       "slug": "gigasecond",
--- a/config.json
+++ b/config.json
@@ -18,15 +18,15 @@
         "println"
       ]
     },
-	{
-      "uuid": "c6631a2c-4632-11e8-842f-0ed5f89f718b",
+    {
       "slug": "twofer",
+      "uuid": "c6631a2c-4632-11e8-842f-0ed5f89f718b",
       "core": false,
       "unlocked_by": null,
       "difficulty": 1,
       "topics": [
-        "Strings",
-		"match"
+        "match",
+        "strings"
       ]
     },
     {
--- a/exercises/twofer/Cargo.toml
+++ b/exercises/twofer/Cargo.toml
@@ -1,5 +1,5 @@
 [package]
 name = "twofer"
-version = "0.1.0"
+version = "1.2.0"
 
 [dependencies]
--- a/exercises/twofer/Cargo.toml
+++ b/exercises/twofer/Cargo.toml
@@ -1,0 +1,5 @@
+[package]
+name = "twofer"
+version = "0.1.0"
+
+[dependencies]
--- a/exercises/twofer/example.rs
+++ b/exercises/twofer/example.rs
@@ -1,5 +1,5 @@
 pub fn twofer(name: &str)-> String {
-    return match name {
+    match name {
         "" => "One for you, one for me.".to_string(),
         _ => format!("One for {}, one for me.",name),
     }
--- a/exercises/twofer/example.rs
+++ b/exercises/twofer/example.rs
@@ -1,0 +1,6 @@
+pub fn twofer(name: &str)-> String {
+    return match name {
+        "" => "One for you, one for me.".to_string(),
+        _ => format!("One for {}, one for me.",name),
+    }
+}
--- a/exercises/twofer/src/lib.rs
+++ b/exercises/twofer/src/lib.rs
@@ -1,0 +1,3 @@
+pub fn twofer(_name: &str)-> String {
+    unimplemented!("Solve the exercise here.");
+}
--- a/exercises/twofer/src/lib.rs
+++ b/exercises/twofer/src/lib.rs
@@ -1,3 +1,3 @@
-pub fn twofer(_name: &str)-> String {
-    unimplemented!("Solve the exercise here.");
+pub fn twofer(name: &str)-> String {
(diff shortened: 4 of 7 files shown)


exercises/twofer/Cargo.toml
```diff
@@ -0,0 +1,5 @@
+[package]
+name = "twofer"
+version = "0.1.0"
```
coriolinus: This should match the [version in `canonical-data.json`](https://github.com/exercism/problem-specifications/blob/master/exercises/two-fer/canonical-data.json#L3).

exercises/twofer/example.rs
```diff
@@ -0,0 +1,6 @@
+pub fn twofer(name: &str)-> String {
+    return match name {
```
coriolinus: `return` is unnecessary here.

exercises/twofer/src/lib.rs
```diff
@@ -0,0 +1,3 @@
+pub fn twofer(_name: &str)-> String {
```
coriolinus: Instead of using an underscore prefix for the variable, try to fit it into the `unimplemented` statement. It's fine to get cute with it: `unimplemented!("One for {}, one for me.", name)`.

config.json
```diff
@@ -12,6 +12,17 @@
...
+    },
+	{
+      "uuid": "c6631a2c-4632-11e8-842f-0ed5f89f718b",
+      "slug": "twofer",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 1,
+      "topics": [
+        "Strings",
+		"match"
```
coriolinus: I'd suggest using the topic "string interpolation" here instead of "match": solutions will almost certainly use string interpolation, but I'd expect a good portion of implementers to use an if statement to solve this instead of a match.

coriolinus: This is a suggestion, not a requirement; feel free to ignore if desired.

coriolinus:
I thought about that. To my mind, using an Option here actually increases
the complexity: given Some(""), what should be returned?

petertseng:
> given Some(""), what should be returned?

I yield on that point, because while *I personally* have no qualms with saying that the answer should be `"One for , one for me"`, the point of your asking that question is to point out that the API incites such questions, and is not to ask what the exact answer to the question is.

I mentioned that I didn't like Some/None in our hello-world in https://github.com/exercism/problem-specifications/issues/290#issuecomment-237998505. I would say the same thing here.

Okay, in that case my only request for us keeping this exercise (I'd also like to explicitly call out the option of simply not having this exercise in this track) is one of the two following:

* have a comment along the lines of saying that there are other ways in Rust to indicate the absence of a value rather than using a sentinel value, but those ways will not be covered in this exercise for simplicity
* explain to me that such a comment is actually not necessary for certain reasons (don't confuse the student with unnecessary information, etc., I'll let the person who explains it to me come up with the reason why it's not necessary)

coriolinus:
Right, that makes sense to me. I think that, were I to write the comment,
I'd structure it as an analogy: for some functions operating on numbers, 0
is outside the domain of the function, and can safely be used as a null
value. For others, 0 is within the domain, so Option _must_ be used if null
values are possible.

Even in the former case, it's still better practice to use an Option to
indicate the null value, but it's common to see this in code translated
from languages which do not possess monadic types. Likewise, there are two
categories of functions operating on strings: those for which the empty
string is within the domain of the function, and those for which it is not.
Twofer is an example of the latter. Even though it's better practice to use
an Option, we are intentionally structuring the API to avoid that
requirement to ease the burden on new learners.

AvasDream:
Hi, i made the requested changes. 
I personally thought about this exercise as some kind of second "hello world" and introduction to pattern matching. In my opinion the Result/Some stuff is a more advanced concept and covered in later exercises.

coriolinus:
@AvasDream We're performing some track maintenance in https://github.com/exercism/rust/pull/528, and it is likely that this is going to require some action from you through no fault of your own. Sorry about that.

There are two possible ways to ensure that this PR still passes Travis once #528 is merged:

- Manually edit your entry in `config.json` to ensure that it conforms to the ordering specified in https://github.com/exercism/meta/issues/95#issuecomment-341991512 and enforced by #528 ([example](https://github.com/coriolinus/exercism_rust/blob/7311a7d726b14e1bbf49a5386ecd4addca82da77/config.json#L86-L98)).

- Via configlet:
    1. Update configlet; the necessary feature is in version 3.9.0 and higher
    2. Run `configlet fmt .` from the track root
    3. Commit its changes to your branch

    Note that if you choose this method, then before #528 is merged, it will produce large unrelated diffs in `config.json` and `config/maintainers.json`. Those are expected, and should vanish once #528 is merged.

exercises/twofer/tests/two-fer.rs
```diff
@@ -0,0 +1,19 @@
+extern crate twofer;
+use twofer::twofer;
+
+#[test]
+fn empty_string() {
+    assert_eq!(twofer(""), "One for you, one for me.");
```
petertseng: > explain to me that such a comment is actually not necessary for certain reasons (don't confuse the student with unnecessary information, etc., I'll let the person who explains it to me come up with the reason why it's not necessary)

I will consider the following part of https://github.com/exercism/rust/pull/520#issuecomment-386879901 to comprise this explanation:

> thought about this exercise as some kind of second "hello world" and introduction to pattern matching. In my opinion the Result/Some stuff is a more advanced concept and covered in later exercises.

I counter that it is still possible to say as a comment in this test "using the empty string as a sentinel value is not preferred in Rust since we have better ways to express the absence of a value, but we will learn those ways later in the track" without having to specify what those alternative ways are.

I cannot comment on what is the best way to teach students these concepts; I would have to spent time on improving my skills and knowledge before I can effectively spend time to think about and comment on how to teach.

Thus, it does not make sense for me to require any changes before this PR should be merged.

coriolinus:
@AvasDream I've jumped in and made the necessary changes to `config.json` by merging master and then applying `configlet fmt .`. I was holding off from merging until those changes happened: if I'd merged without those changes, Travis would fail for everyone else until someone went in and made them. I got sick of waiting, so I did it myself. Hope you don't mind. 

I intend to merge as soon as Travis reports green status.

PR #521, opened 2018-04-27
[WIP] Spiral-Matrix Exercise

Menkir:
- add Cargo.toml
	- add entry in config.json

### DIFF
--- a/config.json
+++ b/config.json
@@ -768,6 +768,17 @@
       "topics": []
     },
     {
+      "uuid": "8dea3473-36f4-4228-b24b-bee8d9389167",
+      "slug": "spiral-matrix",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 4,
+      "topics": [
+        "matrix",
+        "nested structures"
+      ]
+    },
+    {
       "uuid": "0a33f3ac-cedd-4a40-a132-9d044b0e9977",
       "slug": "poker",
       "core": false,
--- a/exercises/spiral-matrix/Cargo.toml
+++ b/exercises/spiral-matrix/Cargo.toml
@@ -1,0 +1,4 @@
+[package]
+name = "spiral-matrix"
+version = "1.1.0"
+
--- a/exercises/spiral-matrix/README.md
+++ b/exercises/spiral-matrix/README.md
@@ -1,0 +1,63 @@
+# Spiral Matrix
+
+Given the size, return a square matrix of numbers in spiral order.
+
+The matrix should be filled with natural numbers, starting from 1
+in the top-left corner, increasing in an inward, clockwise spiral order,
+like these examples:
+
+###### Spiral matrix of size 3
+
+```text
+1 2 3
+8 9 4
+7 6 5
+```
+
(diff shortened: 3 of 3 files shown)


coriolinus:
Hi @Menkir. Please leave the five tests separate. While it can be interesting to parametrically generate test cases, to do so for this problem would necessarily give the student an example solution to copy. 

In general, a useful design principle for unit tests is that they should be as cheap and small as possible. Usually this means that each test tests only one thing, and each thing is covered by one or more tests. While this does lead to some repetition, that's fine; unit tests are one area in which we don't generally attempt to follow the DRY principle.

Menkir:
@coriolinus 
Any suggestions for the [tests]?(https://github.com/exercism/rust/pull/521/commits/53c30297fc730a7fba8f4cbaa38185c01023d550)

Menkir:
😄 No, i meant something like code improvements to satisfy the exercism best pratices.

Menkir:
@coriolinus  feedback pls 😃

exercises/spiral-matrix/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+pub fn get_spiral(size: usize) -> Vec<Vec<usize>> {
```
coriolinus: I'm not entirely certain that `get_spiral` is the best name. People will need to either use `spiral_matrix::get_spiral`, which is repetitive, or `get_spiral` alone, which is vague, and also falsely implies that the spiral is being fetched from somewhere instead of being generated.

Possibly just naming the function `spiral_matrix` would be best.

exercises/spiral-matrix/example.rs
```diff
@@ -0,0 +1,42 @@
+pub fn get_spiral(size: usize) -> Vec<Vec<usize>> {
+    let mut matrix: Vec<Vec<usize>> = Vec::with_capacity(size);
+    for i in 0..size {
+        matrix.push(Vec::with_capacity(size));
+        for _ in 0..size {
+            matrix[i].push(0);
+        }
+    }
+
+    let num_concentric_squares: usize = (size as f64 / 2.0).round() as usize;
```
coriolinus: Trailing decimals here will always be either `.0` or `.5`. I had to think for a bit to remember what `round` did with `.5`. Using [`ceil`](https://doc.rust-lang.org/std/primitive.f64.html#method.ceil) would be a bit more explicit.

exercises/spiral-matrix/example.rs
```diff
@@ -0,0 +1,42 @@
+pub fn get_spiral(size: usize) -> Vec<Vec<usize>> {
+    let mut matrix: Vec<Vec<usize>> = Vec::with_capacity(size);
+    for i in 0..size {
```
coriolinus: There's a shorthand for initializing a vector with `N` copies of type `T`: `vec![T; N]`. You could use it here to avoid the loop: `vec![vec![0; size]; size]`.

I'm not certain, but I suspect that the compiler can optimize array initialization of this form to something quicker than a nested loop.

PR #525, opened 2018-05-06
Adding simple-cipher

miken500:


coriolinus:
Travis failed because README.md is out of sync with the description from the problem specifications repo. The simplest way to fix this is to download configlet and have it regenerate the readme: `configlet generate . --only simple-cipher`.

exercises/simple-cipher/tests/simple-cipher.rs
```diff
@@ -0,0 +1,136 @@
...
+
+#[test]
+#[ignore]
+fn cipher_can_decode_with_given_key() {
+    assert_eq!(decode(KEY, "abcdefghij"), Some("aaaaaaaaaa".to_string()));
+}
+
+#[test]
+#[ignore]
+fn cipher_is_reversible_given_key() {
```
petertseng: It does not make sense for me to request this to change here, since this was just taken from other tracks.

Please excuse me as I use this PR as a staging ground for comments I am to make to the problem-specifications PR adding canonical tests.

This test is only a demonstration of possibilities, and has no power to reject any implementations that the previous tests have not already rejected.

Is the advantage of demonstrating possibilities worth having the test?

I would say it is not.

exercises/simple-cipher/tests/simple-cipher.rs
```diff
@@ -0,0 +1,136 @@
...
+#[test]
+#[ignore]
+fn encode_returns_none_with_any_numeric_key() {
+    let key = "abcd345ef";
+    assert_eq!(encode(key, PLAIN_TEXT), None);
+}
+
+#[test]
+#[ignore]
+fn encode_returns_none_with_empty_key() {
```
petertseng: clarify: should it also return `None` with a key that is non-empty but shorter than the message? If so, is there a test for that? (I may have missed it in my haste, so please excuse my inattentiveness)

What about the same tests for decode?

miken500: A key that is shorter than message is fine. A one character key would be equivalent to a Caesar cipher. The default key in the tests is shorter than the message so we do test that case. 

All these return none tests are only testing the encode function, but yes, it would be better to test both encode and decode. Before I add this, is there a preference for adding a test for each encode/decode case or combining the assert statements in one test?

petertseng: > Before I add this, is there a preference for adding a test for each encode/decode case or combining the assert statements in one test?

I am thinking it would be better to have them be separate, and my reason is this: If both asserts were to fail, this makes it possible to determine which one failed from just the test name, instead of having to look at the line number.

I wrote than and then realised that even if they two asserts were combined in one test, it may be possible to determine which one failed from reading the test failure message. In that case that makes my preference weaker, though I guess it still exists.

miken500:
Could you help with this error? I didn't touch the readme file and I tried to update it again but there were no changes from what is in the commit.

PR #530, opened 2018-05-10
hamming: Update return type

akiekintveld:
This is effectively a difference of two sizes, so we should return `usize`

### DIFF
--- a/exercises/hamming/src/lib.rs
+++ b/exercises/hamming/src/lib.rs
@@ -1,6 +1,6 @@
 /// Return the Hamming distance between the strings,
-/// or an error if the lengths are mismatched.
-pub fn hamming_distance(s1: &str, s2: &str) -> Result<usize, ()> {
+/// or None if the lengths are mismatched.
+pub fn hamming_distance(s1: &str, s2: &str) -> Option<usize> {
     unimplemented!("What is the Hamming Distance between {:?} and {:?}",
                    s1, s2);
 }
--- a/exercises/hamming/src/lib.rs
+++ b/exercises/hamming/src/lib.rs
@@ -1,6 +1,6 @@
 /// Return the Hamming distance between the strings,
 /// or an error if the lengths are mismatched.
-pub fn hamming_distance(s1: &str, s2: &str) -> Result<u64, ()> {
+pub fn hamming_distance(s1: &str, s2: &str) -> Result<usize, ()> {
     unimplemented!("What is the Hamming Distance between {:?} and {:?}",
                    s1, s2);
 }
--- a/exercises/hamming/tests/hamming.rs
+++ b/exercises/hamming/tests/hamming.rs
@@ -2,41 +2,41 @@ extern crate hamming;
 
 #[test]
 fn test_no_difference_between_empty_strands() {
-    assert_eq!(hamming::hamming_distance("", ""), Ok(0));
+    assert_eq!(hamming::hamming_distance("", ""), Some(0));
 }
 
 #[test]
 #[ignore]
 fn test_no_difference_between_identical_strands() {
-    assert_eq!(hamming::hamming_distance("GGACTGA", "GGACTGA"), Ok(0));
+    assert_eq!(hamming::hamming_distance("GGACTGA", "GGACTGA"), Some(0));
 }
 
 #[test]
 #[ignore]
 fn test_complete_hamming_distance_in_small_strand() {
-    assert_eq!(hamming::hamming_distance("ACT", "GGA"), Ok(3));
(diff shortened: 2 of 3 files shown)


akiekintveld:
I noticed a lot of solutions (mine included) ended up casting their final result from a `usize` to a `u64`. It's not a huge issue it just makes the function a bit nicer and clearer if the return type is `usize`. Since `usize` is used for string character counts, it makes sense to also use it for the distance to eliminate the unnecessary casting.

akiekintveld:
Fantastic! While we're looking at the example, the `Result::Err` type appears to be `&'static str`, while the problem template's is `()`. I'm not sure which is "better", but the `&'static str` in the example seems nicer to me. Otherwise if it should be `()` it might be better to return an `Option` than a `Result`.

coriolinus:
You're right. We're slowly weeding out bogus uses of `Result` where an `Option` would do, and we haven't caught this one yet. 

Because fixing that in a separate PR would cause merge conflicts, I'm ok with you updating this PR to also replace the `Result` return type with `Option`, so long as that change is in a separate commit. If you choose not to, it would be nice if you would open an issue as a reminder for someone else to take care of it.

coriolinus:
FWIW, `Result<T, ()>` is almost always an antipattern; `Result<T, &'static str>` isn't much better. In the first case, it's better to replace with an `Option`; in the second, it depends on how many error types are returned. If only one, it also should become an `Option`; otherwise, it should become a domain-specific Error enum.

exercises/hamming/src/lib.rs
```diff
@@ -1,6 +1,6 @@
 /// Return the Hamming distance between the strings,
 /// or an error if the lengths are mismatched.
```
petertseng: is "an error" still OK here? It's not clear that `None` is "an error", but since the word "error" is not capitalised (i.e. it is not explicitly referring to `Result::Err`), it might make sense if one squints.

exercises/hamming/tests/hamming.rs
```diff
@@ -2,41 +2,41 @@ extern crate hamming;
...
 fn test_larger_distance() {
-    assert_eq!(hamming::hamming_distance("ACCAGGG", "ACTATGG"), Ok(2));
+    assert_eq!(hamming::hamming_distance("ACCAGGG", "ACTATGG"), Some(2));
 }
 
 #[test]
 #[ignore]
 fn test_first_string_is_longer() {
-    assert!(hamming::hamming_distance("AAA", "AA").is_err());
+    assert!(hamming::hamming_distance("AAA", "AA").is_none());
```
petertseng: For an implementation that incorrectly gives `Some(_)` here, which of these two gives a clearer test failure message?

```rust
assert!(hamming::hamming_distance("AAA", "AA").is_none());
```

```rust
assert_eq!(hamming::hamming_distance("AAA", "AA"), None);
```

All I request is a confirmation that someone has experimented with both possibilities and accepted that whichever possibility is chosen by this PR is indeed better, with the reason why better

akiekintveld: After experimenting with both, the second option produces much nicer failure messages. I've updated the PR with this change.

exercises/hamming/src/lib.rs
akiekintveld: Good catch. I've updated it from "an error" to "None".

petertseng:
I say :+1:  to the final state of this PR.

Administrative question about how it should be merged: I think I understand the reasoning for why this should remain two commits, the two commits as follows:

* First commit should just change `u64` to `usize`, which only changes the stub to bring it in line with the other components of the exercise
* Second commit changes `Result<T, _>` to `Option<T>`, and all related changes required to get Travis CI to pass

I think the reasoning is that these two changes are unrelated and so shouldn't get mixed up with one another. Okay.

I *think* this result is achievable from the current state by squashing together every commit except the first. Since GitHub is not able to automatically perform this, manual action will need to be taken.

Note for myself: Current commit is fa1969b994afa8c9436217d85d2637e314a3e531

So this comment serves as a reminder that manual action may be necessary. I'm willing to make it happen if so, if nobody else does that before, say, the 17th.

petertseng:
Ah, right. Sorry to have forgotten about that and therefore caused an unnecessary delay. It's important enough that I'll spend this effort now to do it... but not important enough that I would have been displeased if it had been one commit. I should have specified that (again, my failure to do so caused an unnecssary delay). In either event, it is now two commits. I'll give some time to see whether this is to your liking, then merge it

coriolinus:
No worries! This separation of commits looks good to me; merge when ready!

PR #532, opened 2018-05-10
Fix rustc 1.26 warnings

coriolinus:
This rustc release has deprecated the `AsciiExt` trait.
It turns out that the diamond exercise used that trait, and
therefore no longer compiles under DENYWARNINGS=1.

This removes its use of the deprecated trait.


coriolinus:
Because without this patch no Travis job can currently pass, and with it, they can, I count this as urgent enough that I'm going to self-merge it without waiting for review.

petertseng:
Retroactive review

👍

ijanos:
:+1:

PR #533, opened 2018-05-12
Add a bonus test in reverse string

ccouzens:
The Readme talks about a bonus test case with grapheme clusters, but it
wasn't in the test suite.
Looking through the submitted solutions, very few of them attempted
reversing on grapheme clusters.
This might encourage more people to attempt it.

### DIFF
--- a/exercises/reverse-string/tests/reverse-string.rs
+++ b/exercises/reverse-string/tests/reverse-string.rs
@@ -62,3 +62,11 @@ fn test_a_palindrome() {
 fn test_wide_characters() {
     process_reverse_case("子猫", "猫子");
 }
+
+#[test]
+#[ignore]
+#[cfg(feature="graphemes")]
+/// grapheme clusters
+fn test_grapheme_clusters() {
+    process_reverse_case("uüu", "uüu");
+}
--- a/exercises/reverse-string/tests/reverse-string.rs
+++ b/exercises/reverse-string/tests/reverse-string.rs
@@ -65,7 +65,7 @@ fn test_wide_characters() {
 
 #[test]
 #[ignore]
-#[cfg(feature="graphemes")]
+#[cfg(feature="grapheme")]
 /// grapheme clusters
 fn test_grapheme_clusters() {
     process_reverse_case("uüu", "uüu");
--- a/exercises/reverse-string/.meta/hints.md
+++ b/exercises/reverse-string/.meta/hints.md
@@ -1,3 +1,10 @@
 ## Bonus
 Test your function on this string: `uüu` and see what happens. Try to write a function that properly
 reverses this string. Hint: grapheme clusters
+
+To get the bonus test to run, remove the ignore flag (`#[ignore]`) from the
+last test, and execute the tests with:
+
+```bash
+$ cargo test --features grapheme
+```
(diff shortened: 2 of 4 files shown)


exercises/reverse-string/tests/reverse-string.rs
```diff
@@ -62,3 +62,11 @@ fn test_a_palindrome() {
 fn test_wide_characters() {
     process_reverse_case("子猫", "猫子");
 }
+
+#[test]
+#[ignore]
+/// bonus only
+/// grapheme clusters
+fn test_grapheme_clusters() {
+    process_reverse_case("uüu", "uüu");
```
ccouzens: This isn't displaying correctly in Firefox 60.
Chrome correctly shows the strings as being symmetrical

exercises/reverse-string/tests/reverse-string.rs
```diff
@@ -62,3 +62,11 @@ fn test_a_palindrome() {
 fn test_wide_characters() {
     process_reverse_case("子猫", "猫子");
 }
+
+#[test]
+#[ignore]
+/// bonus only
```
coriolinus: Instead of changing the example, introducing a dependency in `Cargo.toml`, etc, why not use conditional compilation to enforce that this is a bonus feature? I.e.

```rust
#[test]
#[ignore]
#[cfg(feature="graphemes")]
fn test_grapheme_clusters() { ... }
```

Examples of how to use custom features [here](https://stackoverflow.com/a/27634313/504550).

coriolinus: You'd need to add a comment explaining to students how to run this test.

ccouzens:
Thanks for the features suggestion @coriolinus.

I've gone ahead and made the change suggested.

bahelms:
Yes! I read that in the README but forgot about doing it because it wasn't in the test suite. Once my tests pass I submit. That should be one of the tests, I think.

exercises/reverse-string/tests/reverse-string.rs
```diff
@@ -62,3 +62,11 @@ fn test_a_palindrome() {
 fn test_wide_characters() {
     process_reverse_case("子猫", "猫子");
 }
+
+#[test]
+#[ignore]
+#[cfg(feature="graphemes")]
```
petertseng: `graphemes` here, but `grapheme` in the README - which is intended?

I cannot run this test with either option currently:

```$ cargo test --features grapheme -- --ignored
   Compiling reverse_string v1.0.0 (file:///home/pt/src/exercism/rust/exercises/reverse-string)
    Finished dev [unoptimized + debuginfo] target(s) in 1.9 secs
     Running target/debug/deps/reverse_string-5106bfa554d6b4ff

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

     Running target/debug/deps/reverse_string-a3483aefc3b6f53e

running 5 tests
test test_a_palindrome ... ok
test test_a_word ... ok
test test_a_capitalized_word ... ok
test test_a_sentence_with_punctuation ... ok
test test_wide_characters ... ok

test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 1 filtered out

   Doc-tests reverse_string

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

(Note that the `test_grapheme_clusters` was not run)


```
$ cargo test --features graphemes -- --ignored 
error: Package `reverse_string v1.0.0 (file:///home/pt/src/exercism/rust/exercises/reverse-string)` does not have these features: `graphemes`
```

ccouzens: My bad, thanks for spotting.

I've corrected it to be singular `grapheme` throughout

PR #542, opened 2018-06-08
Add a template to the `accumulate` exercise

neganp:
With an empty `src/lib.rs` the test suite won't compile until a
learner has written a more or less complete solution. This means they
can't pass the first test until they've understood the whole suite.

This commit adds stub implementations of the `map_function` and
`map_closure` functions so that the test suite compiles as each test
is passed.

The template I wrote uses generics types instead of having all arguments be
`i32`s. This seemed more natural to me, but I'm happy to change the template
to be more concrete if that's preferable.

### DIFF
--- a/exercises/accumulate/src/lib.rs
+++ b/exercises/accumulate/src/lib.rs
@@ -1,8 +1,8 @@
-pub fn map_function<T, R>(xs: Vec<T>, f: &Fn(T) -> R) -> Vec<R> {
+pub fn map_function<T, R>(_xs: Vec<T>, _f: &Fn(T) -> R) -> Vec<R> {
     panic!("Implement your solution here")
 }
 
-pub fn map_closure<F, T, R>(xs: Vec<T>, f: F) -> Vec<R>
+pub fn map_closure<F, T, R>(_xs: Vec<T>, _f: F) -> Vec<R>
 where
     F: Fn(T) -> R,
 {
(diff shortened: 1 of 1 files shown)


neganp:
Thanks for the feedback! I've updated it to use `unimplemented`; that silenced the unused variable errors in my setup, but I can add them to the unimplemented messages if it's still a problem.

I would agree that learning to write the type signatures is just as important as learning to write the function bodies. As far as I can see, "Simple Linked List" is the only previous exercise where generic types come up, and it has a fairly detailed template. The learner does write their own types later in "Sublist", which doesn't have a template. It's also part of the supplemental tests in "Binary Search" where the learner must write code that's generic for refs and borrows.

With that broader context, and since the type signatures in this exercise are pretty simple (and are therefore good candidates for learners to practice on) I think it's fine to leave this template out. I'll retract the PR.

petertseng:
I wonder whether it would be useful to leave a comment in the stub saying "We're asking you to write the type signatures yourself", for this exercise (and maybe other exercises where we do ask this). This most clearly signals intent, whereas both maintainers and students may might forget this intent months down the road (as more exercises get stubs).

(Stubs containing *only* comments are also exempted from the must-compile rule, according to https://github.com/exercism/rust/blob/master/_test/ensure-stubs-compile.sh#L10)

I haven't yet decided whether it'd be a good idea, so feel free to continue discussion in this PR, despite it being closed, on the worth of this idea

neganp:
I think that makes sense. We could include a link to the relevant portion of the documentation, either the [generics chapter](https://doc.rust-lang.org/book/second-edition/ch10-00-generics.html) of The Rust Programming Language, or the section on [generic functions](https://doc.rust-lang.org/1.8.0/book/generics.html) in the original book.

PR #548, opened 2018-06-13
twofer: Added escape characters test

ZapAnton:
While browsing the solutions for `twofer` exercise,
I have found that a lot of them use such checks as 
`match name {"" => "you", ...` 
or 
`if name == ""`

This checks would not work for cases when name is a string with a single space `" "`,
or a string with escape characters like `"\n"`.

So I suggest a new test that addresses these cases. 

While it adds complexity for the final solution, I think people 
would benefit from it since this sort of problem is really common in programming.

### DIFF
--- a/exercises/twofer/tests/two-fer.rs
+++ b/exercises/twofer/tests/two-fer.rs
@@ -8,7 +8,7 @@ fn empty_string() {
 
 #[test]
 #[ignore]
-fn alice() { 
+fn alice() {
     assert_eq!(twofer("Alice"), "One for Alice, one for me.");
 }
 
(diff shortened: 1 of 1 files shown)


coriolinus:
Twofer is a weird exercise in large part because we assign special meaning to the empty string `""`. This is not idiomatic Rust. We discussed it at some length in https://github.com/exercism/rust/pull/520, and decided at the time that the exercise was early enough that it was simpler to just assign the special meaning than to introduce the `Option` type.

This PR pushes us farther down that path, and I'm not sure I like it. @ZapAnton there's nothing wrong with your code from a practical standpoint, but from a pedagogical one, I'm not sure we want to go there just to ensure that students know about the `trim` method. 

My inclination is to close without merging, but my conviction is moderately low. If there are good arguments in favor of this, I am willing to hear them. 

If there is no other discussion, I will close this without merging not sooner than Friday the 22nd.

PR #549, opened 2018-06-17
acronym: Add test for empty string

thvdburgt:
This will add a test with an empty string as input to the acronym exercise.


petertseng:
I'd suggest this case get added to https://github.com/exercism/problem-specifications/blob/master/exercises/acronym/canonical-data.json as well.

No need to wait for it to get added to problem-specifications before adding it here though, since we can add whatever cases we want

PR #550, opened 2018-06-18
Adds stub definition to the ETL exercise

tushartyagi:
The `lib.rs` file of ETL exercise is missing the function stub. Adding it in this commit.


exercises/etl/src/lib.rs
```diff
@@ -0,0 +1,5 @@
+use std::collections::BTreeMap;
+
+pub fn transform(h: &BTreeMap<i32, Vec<char>>) -> BTreeMap<char, i32> {
+    unimplemented!("How will you transform the tree {:?}?", h)
```
coriolinus: Please express this in the form of an imperative statement: "Transform this tree: {:?}" or similar.

PR #553, opened 2018-06-19
gigaseconds: RustFmt-ed tests source file

ZapAnton:
Since README for language track mentions that `rustfmt` should be applied for Rust sources.


ZapAnton:
I noticed that some other exercises could also be formatted. 
Should I open another PR where I format every exercise?

coriolinus:
Sure, if you're willing, that would be helpful. If you're going to do batch improvements, please build them all into a single PR instead of opening a new one for every exercise.

PR #556, opened 2018-06-20
Rustfmt-ed existing exercises

ZapAnton:
Didn't touch `diamond`, `minesweeper` and `rectangles`, because in these exercises non-standard formatting was used for illustration purpose.

Didn't touch `gigasecond` because of #553


ZapAnton:
Also wondered if there any tool for automatic exercise formatting? Because in this PR I did it all by hand and some sort of shell-script would made this job easier.

coriolinus:
I don't know of any existing tools, but it should be pretty simple to whip up a quick bash script. Something like this:

```sh
for ex in exercises/*; do
  (
    echo $ex
    cd "$ex"
    cargo fmt
  )
done
```

Note that I haven't actually tested that script at all, but it (or something very like it) should work fine.

coriolinus:
Merging despite Travis failure: travis is complaining about certain READMEs being out of date, and this PR doesn't touch those READMEs.

PR #557, opened 2018-06-20
Added script for rustfmt-ing existing exercises

ZapAnton:
Should be useful when new exercises are added or when rustfmt is updated

### DIFF
--- a/bin/format_exercises
+++ b/bin/format_exercises
@@ -1,7 +1,9 @@
 #!/bin/bash
 # Format existing exercises using rustfmt
 
-EXERCISES_PATH="../exercises/"
+RUST_TRACK_REPO_PATH=$(cd "$(dirname "$0")/.." && pwd)
+
+EXERCISES_PATH="${RUST_TRACK_REPO_PATH}/exercises/"
 
 # Some exercises use custom formatting for illustration purpose and should not be formatted. 
 IGNORED_EXERCISES=(
--- a/bin/format_exercises
+++ b/bin/format_exercises
@@ -3,7 +3,7 @@
 
 EXERCISES_PATH="../exercises/"
 
-# Some exercises use custsom formatting for illustration purpose and should not be formatted. 
+# Some exercises use custom formatting for illustration purpose and should not be formatted. 
 IGNORED_EXERCISES=(
 	"diamond" 
 	"minesweeper" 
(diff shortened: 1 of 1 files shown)


bin/format_exercises
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+# Format existing exercises using rustfmt
+
+EXERCISES_PATH="../exercises/"
```
coriolinus: This will only work if you are in the `bin` directory when you run this script. Please change it to base the path off the repo root. See [here](https://github.com/exercism/rust/blob/master/_test/check-configlet-fmt.sh#L6) for an example of how to do that.

bin/format_exercises
```diff
@@ -0,0 +1,31 @@
...
+
+# Some exercises use custsom formatting for illustration purpose and should not be formatted.
```
coriolinus: Typo: `custsom` -> `custom`

bin/format_exercises
```diff
@@ -0,0 +1,31 @@
...
+	"diamond" 
+	"minesweeper" 
+	"rectangles"
+)
+
+cd "$EXERCISES_PATH"
+
+# Iterate over every exercise directory and if it is not ignored - format it
+for exercise_dir in ./*; do
+	exersice_name=${exercise_dir##*/}
```
coriolinus: Typo: `exersice_name` -> `exercise_name`

bin/format_exercises
```diff
@@ -0,0 +1,31 @@
...
+	if [[ " ${IGNORED_EXERCISES[*]} " == *"$exersice_name"* ]]; then
+		echo "$exersice_name - Ignored"
+	else
+		echo "$exersice_name - Formatting"
+
+		cd "$exercise_dir"
+
+		cargo fmt
+
+		cd ".."
```
coriolinus: This will work, but I usually find it more elegant to put commands which need to be in a particular directory into a subshell instead of simply moving up one level in the directory tree. That way, things still work if not all items in the list have the same parent directory.

bin/format_exercises
```diff
@@ -0,0 +1,31 @@
...
+IGNORED_EXERCISES=(
+	"diamond" 
+	"minesweeper" 
+	"rectangles"
+)
+
+cd "$EXERCISES_PATH"
+
+# Iterate over every exercise directory and if it is not ignored - format it
+for exercise_dir in ./*; do
```
coriolinus: Why not `for exercise_dir in ${EXERCISES_PATH}/*; do`? That would make it less dependent on the particular location in which this script is being run.

ZapAnton:
Done, @coriolinus.

PR #558, opened 2018-06-20
Add topics to those exercises missing them

coriolinus:
Contributes to https://github.com/exercism/rust/issues/543

### DIFF
--- a/config.json
+++ b/config.json
@@ -82,7 +82,10 @@
       "core": false,
       "unlocked_by": null,
       "difficulty": 1,
-      "topics": null
+      "topics": [
+        "loops",
+        "primes"
+      ]
     },
     {
       "slug": "bob",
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -796,7 +799,11 @@
       "core": false,
       "unlocked_by": null,
       "difficulty": 4,
-      "topics": null
+      "topics": [
+        "ascii",
+        "regular_expressions",
```
petertseng: not necessarily, but I don't have grounds for asking this to be removed because I don't think there are clear standards on what topics should be put. Should a topic be put if it is *possible* that a solution might use it? Or should it only be put if a solution *must* use it (perhaps because the tests require it), etc.? Or are we supposed to exercise judgment and include a topic if we think *any reasonable* solution would use it? But what constitutes *reasonable*? etc.

We shouldn't discuss this for too long. Long enough to figure out what questions we have, then we should present these questions to someone who would know. Who would know?

coriolinus: I agree with all of your comments; topics are too underspecified for my taste. Perhaps actual usage on exercism v2 will help us determine some more concrete guidelines for topic selection.

I added it as a topic as a recommendation to students, on the assumption that students will see the topics before beginning the exercise. My own implementation doesn't use regular expressions, and it was much more complicated than the example, which does.

PR #559, opened 2018-06-20
Rewrite about section to conform to formatting guidelines

coriolinus:
This contributes to https://github.com/exercism/rust/issues/543.


docs/ABOUT.md
```diff
@@ -1,7 +1,15 @@
-Rust aims to bring modern language design and an advanced type system to systems programming. Rust does not use a garbage collector, but instead ensures safety and optimizes code at compile time. The concept of [ownership](https://doc.rust-lang.org/book/second-edition/ch04-00-understanding-ownership.html) is how Rust achieves its largest goal, memory safety.
+Rust is a systems programming language that runs blazingly fast, prevents segfaults, and guarantees thread safety.
+It aims to bring modern language design and an advanced type system to systems programming.
+Rust does not use a garbage collector, using advanced static analysis to provide deterministic drops instead.
```
petertseng: is it fair for me to wonder whether a general reader might not know what "drop" is?

if this is concern is unfounded, I hav enothing more to say, everything else looks OK as is.

coriolinus: It's a reasonable concern, but I don't know how better to express the mechanism by which Rust avoids the need for a garbage collector. People who have studied this a little will know what drops are; people who have not will probably just glaze past that sentence.

PR #560, opened 2018-06-20
Exercise tree

### LINKED ISSUE #547: twofer: Exercise name conflict
Discovered this problem when I tried to generate README 
with `configlet` - it searches `problem-specification` for exercise metadata, 
but cannot find it since in that project the exercise is named `tho-fer` not `twofer`.

It appears this problem also breaks Travis builds

### LINKED ISSUE #552: Should we remove the twofer exercise entirely?
The twofer exercise assigns special meaning to the empty string, which is not good practice in rust. We came to a close decision to use that instead of the `Option` type, but it still doesn't feel right.

We have a ton of exercises toward the beginning of the track; the track would not suffer for removing one. 

Thoughts?

coriolinus:
Organize exercises into a tree structure. Contributes to https://github.com/exercism/rust/issues/543. Closes #552 by completion and closes #548 and closes #547 by exclusion.

Individual commits contain the reasoning about individual exercises. The general approach was to group them into topics, and have the easiest member introduce a topic. 

Due to the large number of individual topics chosen, we ended up with a moderately high number of core exercises. I believe that this number is not excessive, but am very willing to discuss the arrangement. This is a draft, but I do not harbor the delusion that it is perfect.

All non-deprecated exercises are either core or depend on another exercise:

```sh
$ jq '.exercises[] | select((.deprecated|not) and (.core|not) and (.unlocked_by|not))' config.json
```

Core exercises:

```sh
$ jq '.exercises[] | select((.deprecated|not) and .core) | .slug' config.json
"hello-world"
"pythagorean-triplet"
"saddle-points"
"clock"
"luhn"
"atbash-cipher"
"bracket-push"
"sublist"
"space-age"
"macros"
"poker"
"anagram"
"minesweeper"
"parallel-letter-frequency"
"forth"
$ jq '.exercises[] | select((.deprecated|not) and .core) | .slug' config.json | wc -l
15
```


coriolinus:
This also eliminates the twofer exercise entirely, and should therefore close https://github.com/exercism/rust/issues/552 and https://github.com/exercism/rust/pull/548.

coriolinus:
Turns out that the required structure is a vine, not a tree. Who knew?

I'll rework these later.

coriolinus:
@exercism/rust This now passes all its checks, and is the last thing blocking #543, so I'd appreciate it if you guys could take a look at it.

petertseng:
As a first step to reviewing this. I know that there is a graphical visualiser for the new exercise structure. Given that every exercise is core or unlocked by core, I think I don't need the full graphical visualiser (and/or install whatever is necessary to get it running) and I just want to write a textual one. So here is one real quick:

```ruby
require 'json'

r = JSON.parse(ARGV.empty? ? File.read("#{__dir__}/config.json") : ARGF.read)

cores = []
unlocks = Hash.new { |h, k| h[k] = [] }

r['exercises'].each { |x|
  if x['deprecated']
    puts "[deprecated] #{x['slug']}"
  elsif x['core']
    cores << x['slug']
  else
    unlocks[x['unlocked_by']] << x['slug']
  end
}

cores.each { |x|
  puts x
  unlocks[x].each { |u| puts "    #{u}" }
}
```

And the output:

```
[deprecated] two-fer
[deprecated] nucleotide-codons
[deprecated] hexadecimal
hello-world
    gigasecond
    leap
    raindrops
    reverse-string
    nth-prime
    bob
    beer-song
    proverb
    difference-of-squares
    sum-of-multiples
    grains
    prime-factors
    armstrong-numbers
pythagorean-triplet
    series
    collatz-conjecture
    diffie-hellman
saddle-points
    isogram
    say
    run-length-encoding
    isbn-verifier
    perfect-numbers
    hamming
    scrabble-score
    pangram
    all-your-base
    allergies
    variable-length-quantity
    pig-latin
clock
    simple-linked-list
    pascals-triangle
    nucleotide-count
    etl
    acronym
    sieve
    rna-transcription
    triangle
    grade-school
    binary-search
    robot-simulator
    queen-attack
    bowling
    tournament
    alphametics
    two-bucket
    spiral-matrix
    palindrome-products
luhn
    largest-series-product
    word-count
    accumulate
    roman-numerals
    phone-number
    diamond
atbash-cipher
    crypto-square
    rotational-cipher
    simple-cipher
bracket-push
    luhn-from
sublist
    custom-set
space-age
    luhn-trait
    wordy
macros
poker
    decimal
    book-store
    dominoes
anagram
    protein-translation
    robot-name
    ocr-numbers
    react
minesweeper
    rectangles
    circular-buffer
parallel-letter-frequency
forth
```

petertseng:
There is the following conflict between my responsibilities and my time:

It should be my responsibility as a track maintainer to make sure the track progression is good.

However, the time I currently have on my hands means it is difficult for me to think deeply about any proposed scheme, or engage in discourse about whether this progression or that progression be better.

At present, I am able to give an Approval that simply means "this is a reasonable track progression with no obvious errors", but warn that I truly have given little thought as to whether it is better or worse than any other. Another factor in this is that I feel I will make some mistake no matter what I may propose, and it may make more sense to simply try it out and refine it based on feedback. The obvious risk is that if there are egregious errors, the first batch of students will have a bumpier path than if I had taken the proper time to come up with a good progression.

I am currently forced to take the action of a low-effort Approval, acknowledging the risk that the contrast in time spent can be seen as disrespectful. One person puts in the work, expects others to put in at least some effort, and the other person doesn't come through. I apologise for the disrespect this action may cause.

It is interesting that Luhn is not the exercise to unlock its two variants (luhn-from and luhn-trait), but I surmise there must be a good reason for that.

petertseng:
Since edit history is now shown, I suppose this is less necessary, but for the sake of email I'd like it to remain on the record that I edited:

> come up with a good progression

engage in discourse on improving this admitted draft

coriolinus:
> There is the following conflict between my responsibilities and my time

I agree wholeheartedly. Topics are complex, and ideally, we could all sit down and hash out a good progression over the course of a week. Instead, what we have is the progression that I could put together over the course of a few hours. 

We all have external responsibilities and other claims on our time. Do not feel bad if you cannot devote more time to this than you already have.

> it may make more sense to simply try it out and refine it based on feedback

That was my hope, when putting it together. The chance that I got it perfect on the first try seems low. We can always change it later.

> It is interesting that Luhn is not the exercise to unlock its two variants

They are in different categories. I forget which category I assigned to Luhn itself, but luhn-from (and bracket-push) are about implementing external traits, and luhn-trait (and space-age) are about potentially implementing your own.

petertseng:
Previous script failed to show the `core: false` `unlocked_by: null` exercises. New script that does:

```ruby
require 'json'

r = JSON.parse(ARGV.empty? ? File.read("#{__dir__}/config.json") : ARGF.read)

cores = []
unlocks = Hash.new { |h, k| h[k] = [] }

r['exercises'].each { |x|
  if x['deprecated']
    puts "[deprecated] #{x['slug']}"
  elsif x['core']
    cores << x['slug']
  else
    unlocks[x['unlocked_by']] << x['slug']
  end
}

unlocks[nil].each { |x|  puts "[free] #{x}" }

cores.each { |x|
  puts x
  unlocks[x].each { |u| puts "    #{u}" }
}
```

Since there are no such non-deprecated exercises for the Rust track, the output is unchanged.

PR #561, opened 2018-06-22
diffie-hellman: Added big numbers tests

ZapAnton:
Mainly to prevent using `a as u32` solution, since it yields false results when a is larger then u32::MAX

### DIFF
--- a/exercises/diffie-hellman/.meta/hints.md
+++ b/exercises/diffie-hellman/.meta/hints.md
@@ -1,0 +1,2 @@
+One possible solution for this exercise is to implement your own modular exponentiation function.
+To learn more about it refer to the [following page](https://en.wikipedia.org/wiki/Modular_exponentiation).
--- a/exercises/diffie-hellman/README.md
+++ b/exercises/diffie-hellman/README.md
@@ -38,7 +38,8 @@ The calculations produce the same result!  Alice and Bob now share
 secret s.
 
 One possible solution for this exercise is to implement your own modular exponentiation function.
-To learn more about it refer to the [following page][modular-exponentiation].
+To learn more about it refer to the [following page](https://en.wikipedia.org/wiki/Modular_exponentiation).
+
 
 ## Rust Installation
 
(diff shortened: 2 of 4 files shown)


coriolinus:
Travis is complaining because the example solution does not succeed on these new tests. The example does not do anything which obviously restricts its range to `u32`. 

Prerequisite to a proper review will be getting this PR to pass Travis, which in this case will mean adjusting either the tests or the example solution such that the example passes the tests.

ZapAnton:
Is the example solution the one located in `rust/exercises/diffie-hellman/example.rs`? 
Then it actually restricts the range of `a` to `u32`, for example in this function:
```rust
pub fn public_key(p: u64, g: u64, a: u64) -> u64 {
    g.pow(a as u32) % p
}
```
Should i rewrite this example to use modular exponentiation?

coriolinus:
Yes, that's the example solution.

I suppose it's up to you how to handle this. There are at least a few
options here:
- keep the new tests, and rewrite the example to use modular exponentiation.
- scrap the new tests and rewrite the function signature to use `u32`

If the example solution needs to become much more complex in order to
handle the new tests appropriately, it's also worth reevaluating the
exercise difficulty.

ZapAnton:
I'm inclined to keep the new tests and rewrite example, since I cannot see a point in exercise, which only briefly introduces some concept from cryptography and then asks you to use `*` and  `%` to pass it. 

At least with new tests you can learn something about faster exponentiation methods and how casting integers affects your variable, if nothing else.  Or is my line of reasoning flawed?

ZapAnton:
Done, @coriolinus.
I also added link for modular exponentiation, in case somebody wanted to use it.

coriolinus:
Thanks! I'll take a look later tonight.

coriolinus:
Travis is complaining about an out-of-date README. Would you regenerate that, please?

ZapAnton:
I tried using `bin/configlet generate . -o diffie-hellman`, but it only deleted my changes to the exercise's README. Should I use another command?

coriolinus:
No, you have the right command. It's just that you were adding your changes in the wrong place. Sorry about that; I should have pointed you in the right direction sooner.

Add your changes to a new file `.meta/hints.md`, and then rerun the configlet generate command. That will bring them into the readme in a way that Travis won't complain about.

ZapAnton:
Fixed - Travis does not complain now

ZapAnton:
Thanks. 

One more question if you can: how are existing solutions affected when test cases are changed? Do they get some sort of notification?

coriolinus:
No, they don't; all these changes affect only new students attempting the
exercise in the future.

One thing on my personal to-do list, once Exercism 2.0 is released, is to
delete my current solutions and go back through and re-do every exercise.
Many of the exercises have gotten better, and I have certainly gotten
better at Rust than my first time through the track!

ZapAnton:
Merge reminder @coriolinus

Issue #551, opened 2018-06-18
Some exercises are missing templates in lib.rs


tushartyagi:
I guess the issue is self explanatory. There are some exercises which are missing the template/function stub in the `lib.rs` file. For example, the `lib.rs` of [etl](https://github.com/exercism/rust/blob/master/exercises/bracket-push/src/lib.rs) is empty. #512 is about the same issue and it is still open.

I can raise PRs as and when I come across such exercises. But wanted to understand what would go inside the `unimplemented!()` description? Will the best description do, or are we following any pattern.

I've created a PR for this one: https://github.com/exercism/rust/pull/550

coriolinus:
The full list of exercises with empty stub files can be generated with 

```sh
find exercises/ -name "lib.rs" -empty | cut -d'/' -f2
```

As of right now, that list is

- [ ] accumulate
- [ ] acronym
- [ ] allergies
- [ ] anagram
- [ ] binary-search
- [ ] bracket-push
- [ ] circular-buffer
- [ ] clock
- [ ] custom-set
- [ ] dominoes
- [ ] etl
- [ ] hexadecimal
- [ ] luhn-from
- [ ] luhn-trait
- [ ] minesweeper
- [ ] nucleotide-codons
- [ ] nucleotide-count
- [ ] parallel-letter-frequency
- [ ] phone-number
- [ ] pig-latin
- [ ] protein-translation
- [ ] queen-attack
- [ ] rectangles
- [ ] rna-transcription
- [ ] robot-name
- [ ] roman-numerals
- [ ] rotational-cipher
- [ ] sieve
- [ ] sublist
- [ ] tournament
- [ ] triangle
- [ ] wordy

In general, track maintainers will not add stubs for these exercises, but we're happy to review PRs which add appropriate stubs.

neganp:
I also noticed this. As discussed in #542, there are some exercises where adding even a boilerplate template with the appropriate types would take something away from the exercise. Even if the tests dictate what types you need to use, I think reading the tests and figuring out what you need is a useful exercise. In some of these cases, a little note explaining what you're being asked to do might be more appropriate. 

That said, there are some exercises (`custom-set` is one I worked through recently) where you have to write out a whole bunch of boilerplate before you can get your first test to run. In that case, I might push a little harder for a template, especially if there are other exercises where you're doing the type-sleuthing yourself.

petertseng:
[...]

PR #565, opened 2018-07-02
nucleotide-count: Added template to stub file

### LINKED ISSUE #512: nucleotide-count has no template
`nucleotide/src/lib.rs` should be

```Rust

/// Count the number of occurrences of each of the four nucleotides
/// G, A, C, T.
pub fn nucleotide_counts(dna: &str) -> Result<HashMap<char, usize>, ()> {
    unimplemented!("Count the occurrences of G, A, C, T in {}.", dna);
}

/// Count the number of occurrences of a given nucleotide,
/// which must be either G, A, C or T.
pub fn count(nuc: char, dna: &str) -> Result<usize, ()> {
    unimplemented!("Count the occurrences of {} in {}.", nuc, dna);
}
```

Also, these should arguably return `Option` instead of `Result`, or just `panic!()` with a failed `assert!()` on bad input.

ZapAnton:
Contributes to #551.

I guess it also closes #512, since no further comments were made about rewriting  returned `Result` or using `Option` / `panic!`

### DIFF
--- a/exercises/nucleotide-count/example.rs
+++ b/exercises/nucleotide-count/example.rs
@@ -3,11 +3,15 @@ use std::collections::HashMap;
 static VALID_NUCLEOTIDES: &'static str = "ACGT";
 
 pub fn count(nucleotide: char, input: &str) -> Result<usize, char> {
-    let valid = |x: char| { VALID_NUCLEOTIDES.contains(x) };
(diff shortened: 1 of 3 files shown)


ZapAnton:
Done, @coriolinus

PR #566, opened 2018-07-03
Formatted existing examples with rustfmt

ZapAnton:
Also updated `bin/format_exercises` script to format examples.

Side-note: `format_exercises` fails on the old directory left from `twofer` exercise, that contains only `Cargo.lock` and `target`. Should this directory be removed?

### DIFF
--- a/bin/format_exercises
+++ b/bin/format_exercises
@@ -25,6 +25,8 @@ for exercise_dir in ${EXERCISES_PATH}/*; do
 			cd "$exercise_dir"
 
 			cargo fmt
+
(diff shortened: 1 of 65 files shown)


coriolinus:
Yes, the leftover directory should be removed. Git can't remove directories containing only `.gitignore`'d files; as far as it's concerned, they don't exist anyway.

coriolinus:
Thanks for doing this! I'll schedule this one for Friday.

ZapAnton:
Regarding `twofer` directory: I will remove it in the separate PR then, to decrease noise in this one

PR #567, opened 2018-07-04
rotational-cipher: Added template to stub file

ZapAnton:
Contributes to #551

### DIFF
--- a/exercises/rotational-cipher/src/lib.rs
+++ b/exercises/rotational-cipher/src/lib.rs
@@ -1,1 +1,7 @@
-
+pub fn rotate(input: &str, key: u8) -> String {
+    unimplemented!(
+        "How would input text '{}' transform when every letter is shifted using key '{}'?",
+        input,
+        key
+    );
+}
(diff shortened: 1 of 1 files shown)


exercises/rotational-cipher/src/lib.rs
```diff
@@ -1 +1,7 @@
-
+pub fn rotate(input: &str, key: u8) -> String {
```
coriolinus: None of the tests shift by negative keys, but there are plenty of people who would try shifting by `-1` before thinking of using `25`. I'd make `key` an `i8`, just as a hint to the students.

ZapAnton:
Changed the `key` type. Should generalize the function and increase learning for the students

ZapAnton:
Off-topic, if I may: Would it be practical to add some sort of web-hook or bot for this repo to schedule automatic PR merges and other tasks (like on the Rust language repo)?

coriolinus:
There's been some discussion about integrating a github app to do that, but the available options haven't convinced us as yet. I'm wary of giving any external app write permission to the repo.

PR #568, opened 2018-07-05
isbn-verifier: Added tests for long string and special characters

ZapAnton:
Added two tests: one with special characters and one with long digit string. 

The motivation behind these tests is that it is possible to write a solution, that simply filters input string and counts the sum without checking the input string length, and still pass all the tests(for example `test_invalid_isbn_too_long` test, that contains valid isbn and only one extra `X` that could be filtered)

### DIFF
--- a/exercises/isbn-verifier/tests/isbn-verifier.rs
+++ b/exercises/isbn-verifier/tests/isbn-verifier.rs
@@ -77,6 +77,18 @@ fn test_invalid_isbn_too_long() {
 
 #[test]
 #[ignore]
+fn test_valid_digits_invalid_length() {
+    assert!(!is_valid_isbn("35982150881"));
+}
+
+#[test]
+#[ignore]
+fn test_special_characters() {
+    assert!(!is_valid_isbn("!@#%!@"));
+}
+
+#[test]
+#[ignore]
 #[allow(non_snake_case)]
 fn test_invalid_isbn_with_check_digit_X_instead_of_0() {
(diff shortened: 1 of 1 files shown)


exercises/isbn-verifier/tests/isbn-verifier.rs
```diff
@@ -75,6 +75,18 @@ fn test_invalid_isbn_too_long() {
...
+#[test]
+#[ignore]
+fn test_valid_digits_invalid_length() {
+    assert!(!is_valid_isbn("35982150881"));
+}
+
+#[test]
+#[ignore]
+fn test_special_characters() {
+    assert!(!is_valid_isbn("!@#%!@"));
```
coriolinus: What implementation would have accepted this in the first place?

ZapAnton:
```rust
isbn.chars()
        .filter(|ch| (ch.is_digit(10)) || (*ch == 'X' && isbn.find(*ch).unwrap() == isbn.len() - 1))
        .map(|ch| {
            if ch == 'X' {
                10
            } else {
                ch.to_digit(10).unwrap()
            }
        })
        .enumerate()
        .fold(0, |acc, (i, num)| acc + (10 - (i as u32)) * num) % 11 == 0
```

coriolinus: Huh. You're right. I revoke my objection.

PR #571, opened 2018-07-06
accumulate: Added template to stub file

ZapAnton:
Contributes to #551 

I really doubt that the template is needed here, since the point of the exercise is to teach students about closures and how to pass them as function arguments and fully implementing the functions reduces learning.

If this is the case, then perhaps something like this would suffice:

```rust
/*
/// What should _function type be?
pub fn map_function(input: Vec<i32>, _function: ???) -> Vec<i32> {
    unimplemented!("Transform input vector {:?} using passed function", input);
}

/// What should _closure type be? 
pub fn map_closure(input: Vec<i32>, _closure: ???) -> Vec<i32> {
    unimplemented!("Transform input vector {:?} using passed closure", input);
}
*/
```
Or without multi-line comments and `ALLOWED_TO_NOT_COMPILE`?

### DIFF
--- a/exercises/accumulate/.meta/ALLOWED_TO_NOT_COMPILE
+++ b/exercises/accumulate/.meta/ALLOWED_TO_NOT_COMPILE
@@ -1,0 +1,2 @@
+Stub doesn't compile because there is no type specified for _function and _closure arguments. 
+This exercise teaches students about the use of function pointers and closures, therefore specifying the type of arguments for them would reduce learning.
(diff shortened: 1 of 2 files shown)


coriolinus:
Agree that the point of the exercise is for students to figure out the signatures for themselves. I don't like putting the stubs into comments; this is what `ALLOWED_TO_NOT_COMPILE` is for. I like the idea of using that.

ZapAnton:
Done, @coriolinus

PR #576, opened 2018-07-09
triangle: Added template to stub file

ZapAnton:
Contributes to #551

### DIFF
--- a/exercises/triangle/src/lib.rs
+++ b/exercises/triangle/src/lib.rs
@@ -1,1 +1,19 @@
+pub struct Triangle;
 
+impl Triangle {
+    pub fn build(sides: [u64; 3]) -> Result<Triangle, ()> {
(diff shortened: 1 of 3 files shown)


exercises/triangle/src/lib.rs
```diff
@@ -1 +1,19 @@
+pub struct Triangle;
 
+impl Triangle {
+    pub fn build(sides: [u64; 3]) -> Result<Triangle, ()> {
```
coriolinus: I'm aware that the signature isn't your fault, but I don't like the practice of returning `Result<T, ()>`; that should idiomatically be an `Option<T>` instead. 

If you're willing, I would appreciate it if you would update the exercise to use an Option here. If not, feel free to ignore.

ZapAnton: Will do

ZapAnton:
The exercise is now updated, @coriolinus

PR #579, opened 2018-07-17
roman-numerals: Added template to the stub file

ZapAnton:
Contributes to #551

### DIFF
--- a/exercises/roman-numerals/src/lib.rs
+++ b/exercises/roman-numerals/src/lib.rs
@@ -8,8 +8,8 @@ impl Display for Roman {
     }
 }
 
-impl From<i32> for Roman {
-    fn from(num: i32) -> Self {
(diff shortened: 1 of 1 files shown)


exercises/roman-numerals/src/lib.rs
```diff
@@ -1 +1,15 @@
...
 
+pub struct Roman;
+
+impl Display for Roman {
+    fn fmt(&self, _f: &mut Formatter) -> Result {
+        unimplemented!("Return a roman-numeral string representation of the Roman object");
+    }
+}
+
+impl From<i32> for Roman {
```
coriolinus: Given that I've never heard of a negative roman numeral, and none of the tests use negative numbers, it's probably better to implement `From<u32>`.

ZapAnton:
Done, @coriolinus

PR #581, opened 2018-07-19
binary-search: Added template to the stub file

ZapAnton:
Contributes to #551


exercises/binary-search/src/lib.rs
```diff
@@ -1 +1,7 @@
-
+pub fn find(array: &[i32], key: i32) -> Option<usize> {
+    unimplemented!(
+        "Using the binary search algorithm, find the element '{}' in the array '{:?}' and return it's index.",
```
coriolinus: No apostrophe in possessive "its".

PR #582, opened 2018-07-19
Fix ignore counts

coriolinus:
Currently, Travis always fails. This is my fault: I edited the [`_test/count-ignores.sh`](https://github.com/exercism/rust/blob/master/_test/count-ignores.sh) script to identify `#[ignore]` and `#[test]` attributes which appear anywhere on a line, not just at its beginning. This identified two exercises which had commented-out tests, which were not ignored: binary-search and triangle.

It would have been possible to revert to the old behavior, but that was also demonstrably wrong, incorrectly failing #577. Instead, I've added `#[ignore]` flags to the appropriate tests, uncommented them, and put them behind a feature gate, as we've done before for the reverse-string exercise. 

Travis does not currently check that feature-gated tests pass, but in case we add that capability in the future, I've ensured that the example solutions do correctly solve the feature-gated tests.

Requesting relatively high-priority review, as I am reluctant to merge any other PR until Travis passes.

### DIFF
--- a/exercises/binary-search/.meta/hints.md
+++ b/exercises/binary-search/.meta/hints.md
@@ -25,7 +25,12 @@ are some additional things you could try.
 - Additionally this find function can work not only on slices, but at the
   same time also on a Vec or an Array.
 
-You can find tests (commented out) for these bonus tasks in the test file.
+To run the bonus tests, remove the `#[ignore]` flag and execute the tests with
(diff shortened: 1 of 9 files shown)


coriolinus:
Thanks for the quick review! 

I hadn't even realized that `float_example.rs` existed. I'll add an issue to remove it.

PR #586, opened 2018-07-23
Updated commands to download luhn exercise

ZapAnton:
Updates commands in README for the new version of CLI

### DIFF
--- a/exercises/luhn-trait/.meta/description.md
+++ b/exercises/luhn-trait/.meta/description.md
@@ -4,11 +4,15 @@ Before doing this exercise you should probably do the original Luhn exercise and
 
 To get the original Luhn exercise, run
 
-> `exercism download --exercise=luhn --track=rust`
+```shell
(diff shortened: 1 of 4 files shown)


exercises/luhn-trait/.meta/description.md
```diff
@@ -6,7 +6,9 @@ To get the original Luhn exercise, run
 
 > `exercism download --exercise=luhn --track=rust`
 
-To get the "Luhn: Using the From Trait" exercise, run `exercism fetch rust luhn-from`
+To get the "Luhn: Using the From Trait" exercise, run
+
+> `exercism download --exercise=luhn-from --track=rust`
```
coriolinus: Instead of using blockquotes around a code span, it's better to use a code block. I.e. Instead of 

        > `exercism download --exercise=luhn-from --track=rust`


It's better to do this:

        ```shell
        exercism download --exercise=luhn-from --track=rust
        ```

This applies to all shell examples.

PR #588, opened 2018-07-25
Added dot-dsl exercise to the track

ZapAnton:
This PR adds `dot-dsl` exercise.

The main motivation behind this is to have an easy exercise that introduces Rust module system.
The implementation also allows to introduce the `builder` design pattern.

There is no canonical data for this exercise, so I simply ported tests from the [Python track](https://github.com/exercism/python/tree/master/exercises/dot-dsl), which could introduce some non-idiomatic code.

The `example.rs` file is somewhat cluttered, because I had to move all modules to one file.

Was not sure if I should add some hints to the README, so I left it as it is.


exercises/dot-dsl/tests/dot-dsl.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    assert_eq!(graph.edges, vec![Edge::new("a", "b")]);
+}
+
+#[test]
+#[ignore]
+fn test_graph_with_one_attribute() {
+    let graph = Graph::new().with_attrs(&[("foo", "1")]);
+
+    let mut expected_attrs = HashMap::new();
```
coriolinus: The [`maplit` crate](https://crates.io/crates/maplit) is good for test cases like this.

exercises/dot-dsl/.meta/description.md
```diff
@@ -0,0 +1,22 @@
+Write a Domain Specific Language similar to the Graphviz dot language.
```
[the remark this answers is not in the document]
ZapAnton: For some reason `configlet generate`  did not generate README unless i manually copied the description from `problem-specification`

exercises/dot-dsl/tests/dot-dsl.rs
ZapAnton: I thought external dependencies were not permitted in the test suite?

coriolinus: They're discouraged but not strictly forbidden, and I don't recall a written policy. What it really comes down to is a judgment call on the part of the track moderators. In my judgment, using `maplit` macros increases test clarity, so I allow it.

exercises/dot-dsl/.meta/description.md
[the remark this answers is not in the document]
coriolinus: That's interesting. Could it be that `configlet generate` is referencing a local copy of the `problem-specifications` repo which has fallen out of date?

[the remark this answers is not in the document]
ZapAnton: I actually updated the description of `dot-dsl` on the `problem-specification` repo just recently, so no, it is up to date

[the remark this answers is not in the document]
petertseng: I could not replicate this behaviour - `generate` does cause `dot-dsl/README.md` to be written, even without this file. I would prefer not to have this file because we don't want to have to maintain a duplicate version of it.

ZapAnton:
The problems from the review have been addressed, though it seems like Travis is having some unforeseen errors.

In addition to the future suggestions/requests I would like to ask about 2 moments:

- The public access to the `Graph` struct fields.
It feels like it belongs in the Python exercise, rather then in the Rust exercise. Should some `getter` methods be added?

- The point of the exercise.
Right now the main selling point of the added exercise is the introduction of `module` and `builder-pattern` concepts, but the task itself is minimal - it just asks to make some struct constructors. Is this acceptable, or should the task be more sophisticated?

exercises/dot-dsl/.meta/hints.md
```diff
@@ -0,0 +1,7 @@
+## Builder pattern
+
+This exercise expects you to build several structs using `builder pattern`.
+In short, this pattern allows you to split the construction function of you struct, that contains a lot of arguments, into
```
coriolinus: Typo: "you struct" -> "your struct".

exercises/dot-dsl/.meta/hints.md
```diff
@@ -0,0 +1,7 @@
+## Builder pattern
+
+This exercise expects you to build several structs using `builder pattern`.
+In short, this pattern allows you to split the construction function of your struct, that contains a lot of arguments, into 
+several separate functions. This approuch gives you the means to make compact but highly-flexible struct construction and
```
petertseng: approach

PR #589, opened 2018-07-25
bracket-push: Added template to the stub file

ZapAnton:
Contributes to #551

### DIFF
--- a/exercises/bracket-push/src/lib.rs
+++ b/exercises/bracket-push/src/lib.rs
@@ -1,6 +1,6 @@
 pub struct Brackets;
 
-impl From<&'static str> for Brackets {
+impl<'a> From<&'a str> for Brackets {
     fn from(input: &str) -> Self {
         unimplemented!("From the '{}' input construct a new Brackets struct", input);
     }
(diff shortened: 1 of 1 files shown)


exercises/bracket-push/src/lib.rs
```diff
@@ -1 +1,13 @@
+pub struct Brackets;
 
+impl From<&'static str> for Brackets {
```
coriolinus: I bet this would work, and more generally, without the `'static` lifetime bound.

ZapAnton: It does not actually. When I try to compile the stub i get the following output:
```rust
error[E0106]: missing lifetime specifier
 --> src/lib.rs:3:11
  |
3 | impl From<&str> for Brackets {
  |           ^ expected lifetime parameter

```

coriolinus: Right! Sorry about that. Well, let's at least make it a generic lifetime bound, so that someone could generate an arbitrary non-static `&str` and the `From` trait would still work:

```rust
impl<'a> From<&'a str> for Brackets { ... }
```

PR #590, opened 2018-07-26
queen-attack: Added template to the stub file

ZapAnton:
Contributes to #551 

Should I update the test suite to ensure that the invalid file/rank is returned from the `ChessPosition::new()` method?

### DIFF
--- a/exercises/queen-attack/src/lib.rs
+++ b/exercises/queen-attack/src/lib.rs
@@ -5,8 +5,12 @@ pub struct ChessPosition;
 pub struct Queen;
 
 impl ChessPosition {
-    pub fn new(rank: i32, file: i32) -> Result<Self, i32> {
-        unimplemented!("Construct a ChessPosition struct, given the following rank, file: ({}, {}). If the position is invalid return error that contains wrong rank/file.", rank, file);
+    pub fn new(rank: i32, file: i32) -> Option<Self> {
+        unimplemented!(
+            "Construct a ChessPosition struct, given the following rank, file: ({}, {}). If the position is invalid return None.",
+            rank,
+            file
+        );
     }
 }
 
--- a/exercises/queen-attack/src/lib.rs
+++ b/exercises/queen-attack/src/lib.rs
@@ -1,5 +1,7 @@
+#[derive(Debug)]
 pub struct ChessPosition;
 
+#[derive(Debug)]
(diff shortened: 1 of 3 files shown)


exercises/queen-attack/src/lib.rs
```diff
@@ -1 +1,19 @@
...
+
+impl ChessPosition {
+    pub fn new(rank: i32, file: i32) -> Result<Self, i32> {
+        unimplemented!("Construct a ChessPosition struct, given the following rank, file: ({}, {}). If the position is invalid return error that contains wrong rank/file.", rank, file);
+    }
+}
+
+impl Queen {
+    pub fn new(_position: ChessPosition) -> Self {
+        unimplemented!("Given the chess position, construct a Queen struct.");
```
coriolinus: Please include the `position` variable so the leading underscore is unnecessary: `"Given the chess position {:?}, construct a Queen struct."`. This will require a `#[derive(Debug)]` annotation on `ChessPosition`.

exercises/queen-attack/src/lib.rs
```diff
@@ -1 +1,19 @@
...
+    }
+
+    pub fn can_attack(&self, _other: &Queen) -> bool {
+        unimplemented!("Determine if this Queen can attack the other Queen");
```
coriolinus: Please include the `other` variable so the leading underscore is unnecessary: `"Determine if this queen can attack the other Queen ({:?})"`. This will require a `#[derive(Debug)]` annotation on `Queen`.

exercises/queen-attack/src/lib.rs
```diff
@@ -1 +1,19 @@
+pub struct ChessPosition;
 
+pub struct Queen;
+
+impl ChessPosition {
+    pub fn new(rank: i32, file: i32) -> Result<Self, i32> {
```
coriolinus: `Result<Self, i32>` is ambiguous: what should the return value be when someone calls `ChessPosition::new(-2, -1)`?

There are several ways in which we might handle this in the real world: use a builder pattern, so it's unambiguous  which step raised an error; return an enum with `Rank`, `File`, and `Both` variants, and more. Both of those possibilities require more work from the student than I believe this exercise should.

Instead, would you be willing to update the stub, example, and test suite such that `new` returns `Option<Self>`?

PR #591, opened 2018-07-29
check-configlet-fmt.sh is now run only when config.json is modified

### LINKED ISSUE #544: configlet fmt check: Should only run when a JSON file is changed
If it runs unconditionally, a new `configlet` release can cause unrelated failures on PRs that touch nothing related to the JSON files, but we run the risk of being out of date.

Similar tradeoff that we made when checking READMEs: It only checks the exercises that were changed in PR.

ZapAnton:
Closes #544

### DIFF
--- a/_test/check-configlet-fmt.sh
+++ b/_test/check-configlet-fmt.sh
@@ -3,6 +3,26 @@
 # This ensures that config.json and config/maintainers.json are compatible
 # with the output of configlet fmt.
 
+# Check if config.json or maintainers.json were modified
+current_branch_name="$(git rev-parse --abbrev-ref HEAD)"
+
+check_pattern="config.json\|config/maintainers.json"
+
+if [ "$current_branch_name" != "master" ]; then
+    # Check the changes on the current branch against master branch
+    git diff --name-only master | grep "$check_pattern"
+else
+    # Check the commits on the master branch made during the week
+    # This is because Travis cron is set to test the master branch weekly.
+    git diff --name-only "@{7 days ago}" | grep "$check_pattern"
+fi
+
+if [ $? != 0 ]; then
+    echo "config.json or maintainers.json were not changed - configlet fmt is aborted."
+
+    exit 0
+fi
+
 repo=$(cd "$(dirname "$0")/.." && pwd)
 configlet="${repo}/bin/configlet"
 


_test/check-configlet-fmt.sh
```diff
@@ -9,6 +9,15 @@ configlet="${repo}/bin/configlet"
 c="${repo}/config.json"
 m="${repo}/config/maintainers.json"
 
+# Check if config.json was modified
+git diff-tree --no-commit-id --name-only -r HEAD^ | grep "$c"
```
petertseng: since this only checks `config.json` but `configlet fmt` modifies `maintainers.json` and `config.json`, I think this now makes it possible for a PR to give `maintainers.json` nonstandard formatting if it only modifies `maintainers.json` and not `config.json`. is that true?

ZapAnton:
I also wanted to ask about the check method in this PR.

Currently this script checks if `config.json` was changed in the last commit:

```shell
git diff-tree --no-commit-id --name-only -r HEAD | grep "$c"
```

This presents a problem, when a batch of commits is pushed, e.g. `commitA` -> `commitB` -> `push`
and `config.json` was changed in the `commitA` - the current script would not detect that.

I could change the script to check the changes in the current branch against `master` branch:

```shell
git diff --name-status master "$(git rev-parse --abbrev-ref HEAD)" | grep "$c"
```
This would detect changed in the whole branch.

The new problem is that it would not detect changes, if commits are pushed directly to `master`

The solution I see is to check for branch name, and if it is `master`, check the last commit. It would work, but only if Travis is launched for every commit, made directly to `master`.

Is the above algorithm appropriate for the linked issues, or would you like to see additional logic in it?

_test/check-configlet-fmt.sh
petertseng: > Should I now add a check for maintainers.json?

Yes, I think it has to be. Since the goal is that the two files should have standard formatting

petertseng:
> checks if config.json was changed in the last commit:
> when a batch of commits is pushed, e.g. commitA -> commitB -> push and config.json was changed in the commitA - the current script would not detect that.
> I could change the script to check the changes in the current branch against master branch:

Yes, I think it has to be this way (check against master branch)

> The new problem is that it would not detect changes, if commits are pushed directly to master

I think that should be a rare occurrence. I do not see a need to even deal with this case at all. I do not intend to ever push directly to master.

Another reason as to why: One reason CI wants to exist is because in Exercism v2, changes to `master` to live almost immediately. Thus, it makes sense to use CI to make sure that a potential change won't cause a problem.

If we ever need to push directly to master, we're already too late for CI to catch any problems, since it has already gone live. But I acknowledge that in this case, maybe CI can retroactively tell us of the problem so we know to fix it faster. So, if this can be achieved without making the logic too complex, I don't object.

> but only if Travis is launched for every commit, made directly to master.

Ever since https://github.com/exercism/discussions/issues/174#issuecomment-318855073 , Travis is not made for every commit made to master. The only compromise I was able to make was to run Travis weekly against master.

ZapAnton:
The final algorithm is this:
- if not on `master` branch check the changes on the current branch against the `master` branch
- if on `master` branch check the changes made during the week

off-topic: Is there a preferred naming format for commits on non-exercise parts of track? 
E.g. "slug: Changes made to "slug" exercise"

_test/check-configlet-fmt.sh
```diff
@@ -9,6 +9,25 @@ configlet="${repo}/bin/configlet"
 c="${repo}/config.json"
 m="${repo}/config/maintainers.json"
 
+# Check if config.json or maintainers.json were modified
+current_branch_name="$(git rev-parse --abbrev-ref HEAD)"
+
+check_pattern="config.json\|config/maintainers.json"
```
petertseng: something I don't care to have fixed: the . are going to match any character, which means this will fire if we change a file named configxjson

I don't want to sacrifice clarity to have this fixed because I find it unlikely we will ever have a file with that name.

_test/check-configlet-fmt.sh
```diff
@@ -9,6 +9,25 @@ configlet="${repo}/bin/configlet"
 c="${repo}/config.json"
```
petertseng: Because the newly added code does not use `$c`, `$m`, `$repo`, and `$configlet`, and I think it is better to define variables close to where they are used: I think all the newly added code should go above all four of those definitions.

The potential room for disagreement is if someone thinks it's better to define all variables at the top, but I think this only holds true for variables that we may want to change often. I can see `configlet` wanting to change, but nothing else.

petertseng:
Thanks! Nice idea about the `"@{7 days ago}"`, taught me something new.

PR #592, opened 2018-07-29
Fix gitignores

lutostag:
Found that some of the exercises had different gitignores. Caused me to almost commit the /target/ files to my exercises when running through them (reverse string was the first I ran into with the problem).

Made sure they all have the same now:
```
# Generated by Cargo
# will have compiled files and executables
/target/
**/*.rs.bk

# Remove Cargo.lock from gitignore if creating an executable, leave it for libraries
# More information here http://doc.crates.io/guide.html#cargotoml-vs-cargolock
Cargo.lock
```

Also updated the init_exercise.py to have the same gitignore for future created projects.

You can check they are all the same (except for grep) which I also included the *.txt because it seemed like some of the tests may have large text files there... by running
`md5sum exercises/*/.gitignore | sort`

Hope this is helpful.

Thanks!


exercises/bob/README.md
```diff
@@ -13,6 +13,8 @@ anything.
 
 He answers 'Whatever.' to anything else.
 
+Bob's conversational partner is a purist when it comes to written communication and always follows normal rules regarding sentence punctuation in English.
```
coriolinus: This has already been updated by https://github.com/exercism/rust/pull/597. Please remove this change from this PR.

exercises/grep/.gitignore
```diff
@@ -1,4 +1,9 @@
...
+*.txt
+# Generated by Cargo
+# will have compiled files and executables
+/target/
+**/*.rs.bk
+
+# Remove Cargo.lock from gitignore if creating an executable, leave it for libraries
 # More information here http://doc.crates.io/guide.html#cargotoml-vs-cargolock
 Cargo.lock
-*.txt
```
coriolinus: This line should not be removed; the grep exercise generates some text files, and while it is meant to clean up after itself, it might not in the event of panics.

lutostag: It's moved to the top line. Still there. I'll move it back to the bottom

lutostag: Using your explanation as a comment there too. Thanks!

PR #594, opened 2018-07-31
phone-number: Added template to the stub file

ZapAnton:
Contributes to #551

### DIFF
--- a/exercises/phone-number/src/lib.rs
+++ b/exercises/phone-number/src/lib.rs
@@ -1,1 +1,6 @@
-
+pub fn number(user_number: &str) -> Option<String> {
+    unimplemented!(
+        "Given the number entered by user '{}', convert it into SMS-friendly format. If the entered number is not a valid NANP number, return None.",
+        user_number
+    );
+}


exercises/phone-number/src/lib.rs
```diff
@@ -1 +1,6 @@
-
+pub fn number(user_number: &str) -> Option<String> {
+    unimplemented!(
+        "Given the number entered by user '{}', convert it into SMS-friendly format. If the entered number is not a valid NANP number, return None.",
```
petertseng: I do see that NANP is described in the README, though I don't think SMS is mentioned. I do happen to know what SMS is. Is it OK to leave SMS in?

coriolinus: I think it's reasonable to assume that most peopele in the modern world will know what SMS means.

ZapAnton: The SMS part is taken from the README (first sentence):

```text
Phone Number

Clean up user-entered phone numbers so that they can be sent SMS messages.

```

PR #595, opened 2018-07-31
wordy: Added template to the stub file

ZapAnton:
Contributes to the #551 

The returned type of the `WordProblem::answer` function is `Result<i32, String>`, but perhaps `Option<i32>` is more suited here?

### DIFF
--- a/exercises/wordy/src/lib.rs
+++ b/exercises/wordy/src/lib.rs
@@ -1,1 +1,16 @@
+pub struct WordProblem;
 
+impl WordProblem {
+    pub fn new(command: &str) -> Self {
+        unimplemented!(
+            "From the '{}' command construct a new WordProblem struct.",
+            command
+        );
+    }
+
+    pub fn answer(&self) -> Result<i32, String> {
+        unimplemented!(
+            "Return the result of the transmitted command, or an error, if the command is invalid."
+        );
+    }
+}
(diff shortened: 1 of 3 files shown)


petertseng:
> perhaps `Option<i32>` is more suited here?

yes, I think so. At the time I wrote https://github.com/exercism/rust/issues/444#issue-302020058 I was thinking strings for the E in `Result<T, E>` are not programmatically inspectable so would be better replaced by a type that is, or `None`.

ZapAnton:
Done

exercises/wordy/src/lib.rs
```diff
@@ -1 +1,16 @@
...
+    pub fn new(command: &str) -> Self {
+        unimplemented!(
+            "From the '{}' command construct a new WordProblem struct.",
+            command
+        );
+    }
+
+    pub fn answer(&self) -> Option<i32> {
+        unimplemented!(
+            "Return the result of the transmitted command, or an error, if the command is invalid."
```
petertseng: I think this should talk about `None` instead of `error`, now, right?

I'll prefer that change be amended into the Result -> Option commit so all the necessary changes for Result->Option are in one commit. It's small enough that amending won't cause a review burden.

coriolinus: Agreed.

PR #603, opened 2018-08-06
protein-translation: Added template to the stub file

ZapAnton:
Contributes to #551 

The returned type seems like it could be changed from `Result<&str, &str>` to `Option<&str>`, but
I would wait for an approval, before changing anything.

### DIFF
--- a/exercises/protein-translation/example.rs
+++ b/exercises/protein-translation/example.rs
@@ -12,11 +12,7 @@ pub fn parse<'a>(pairs: Vec<(&'a str, &'a str)>) -> CodonInfo<'a> {
 
 impl<'a> CodonInfo<'a> {
     pub fn name_for(&self, codon: &str) -> Option<&'a str> {
-        if let Some(acid) = self.actual_codons.get(&codon) {
-            Some(acid)
-        } else {
-            None
-        }
+        self.actual_codons.get(&codon).map(|&acid| acid)
     }
 
     pub fn of_rna(&self, strand: &str) -> Option<Vec<&'a str>> {
--- a/exercises/protein-translation/example.rs
+++ b/exercises/protein-translation/example.rs
@@ -11,20 +11,21 @@ pub fn parse<'a>(pairs: Vec<(&'a str, &'a str)>) -> CodonInfo<'a> {
 }
 
 impl<'a> CodonInfo<'a> {
-    pub fn name_for(&self, codon: &str) -> Result<&'a str, &'static str> {
-        match self.actual_codons.get(&codon) {
-            Some(name) => Ok(name),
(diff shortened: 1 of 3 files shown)


exercises/protein-translation/src/lib.rs
```diff
@@ -1 +1,21 @@
+pub struct CodonsInfo;
 
+impl CodonsInfo {
+    pub fn name_for(&self, codon: &str) -> Result<&'static str, &'static str> {
```
coriolinus: I'm not thrilled by returning `&'static str` in either position of the result; that's not normally great practice. While it works here, it wouldn't work if, for example, one were reading data from a file. 

I'm really tempted to try to get away with returning `&str` here and accepting it in `parse`, but that feels a bit mean. Probably the right thing to do here is make `CodonsInfo` generic over `'a`, and add a `std::marker::PhantomData<'a>` field to keep the compiler from complaining about unused generic types. Of course, that would also need a short comment explaining that the student should get rid of it once they have their own field using `'a`.

coriolinus: Agree that returning `Option` makes more sense here than returning `Result`. Please put the edits which change the return type into a separate commit. This also applies to the return type of `of_rna`.

exercises/protein-translation/example.rs
```diff
@@ -11,20 +11,21 @@ pub fn parse<'a>(pairs: Vec<(&'a str, &'a str)>) -> CodonInfo<'a> {
 }
 
 impl<'a> CodonInfo<'a> {
-    pub fn name_for(&self, codon: &str) -> Result<&'a str, &'static str> {
-        match self.actual_codons.get(&codon) {
-            Some(name) => Ok(name),
-            None => Err("Invalid"),
+    pub fn name_for(&self, codon: &str) -> Option<&'a str> {
+        if let Some(acid) = self.actual_codons.get(&codon) {
```
ZapAnton: This part feels really wrong (get `Some` to return `Some`), but if I used only `self.actual_codons.get(&codon)`,
the compiler would complain that the `Option<&&str>` is returned instead of `Option<&str>`. 
Any other way to bypass the problem?

coriolinus: My own instinct would be to use `self.actual_codons.get(&codon).map(|&c| c)`, but there are a few other ways to handle it as well.

ZapAnton: The example should simply compile and pass tests,
so perhaps there is no need for more advanced concepts 
and it would be enough if I changed `if let` with `map`?

Apart from that are there any more changes you would request?

coriolinus: No, that should be the last change, I think.

PR #605, opened 2018-08-06
custom-set: Added template to the stub file

ZapAnton:
Contributes to #551

### DIFF
--- a/exercises/custom-set/src/lib.rs
+++ b/exercises/custom-set/src/lib.rs
@@ -1,1 +1,61 @@
+use std::fmt::Debug;
+use std::marker::PhantomData;
 
+#[derive(Debug, PartialEq)]
+pub struct CustomSet<T> {
+    // This field is here to make the template compile and not to
+    // complain about unused type parameter 'T'. Once you start
+    // solving the exercise, delete this field and the 'std::marker::PhantomData'
+    // import.
+    phantom: PhantomData<T>,
+}
+
+impl<T: Debug> CustomSet<T> {
+    pub fn new(input: &[T]) -> Self {
+        unimplemented!(
+            "From the given input '{:?}' construct a new CustomSet struct.",
+            input
+        );
+    }
+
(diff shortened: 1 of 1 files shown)


exercises/custom-set/src/lib.rs
```diff
@@ -1 +1,61 @@
...
+        );
+    }
+
+    pub fn is_empty(&self) -> bool {
+        unimplemented!("Determine if the CustomSet struct is empty.");
+    }
+
+    pub fn is_disjoint(&self, other: &Self) -> bool {
+        unimplemented!(
+            "Determine if the CustomSet struct and othe struct '{:?}' are disjoint.",
```
ZapAnton: Spotted a typo `othe struct` should probably be `the other struct`. To reduce the noise, I would amend the fix to
the initial commit.

coriolinus: Good spot! I missed that. Yes, please amend that into the single commit in this PR.

ZapAnton:
Done.
I should really find some grammar check plugin for Vim, the amount of typos in the PRs is embarrassing...

exercises/custom-set/src/lib.rs
```diff
@@ -1 +1,61 @@
...
+#[derive(Debug, PartialEq)]
+pub struct CustomSet<T> {
+    // This field is here to make the template compile and not to
+    // complain about unused type parameter 'T'. Once you start
+    // solving the exercise, delete this field and the 'std::marker::PhantomData'
+    // import.
+    phantom: PhantomData<T>,
+}
+
+impl<T: Debug> CustomSet<T> {
```
petertseng: OK to add `<T: Debug>` here, since a `T` is used in so many `unimplemented` messages? Or should it be left out?

ZapAnton: The problem with the `Debug` trait bound seems to appear in a lot of exercises. Instead of having separate discussion in every PR, would you mind creating an issue about it? So that there would be one fixed solution.

PR #606, opened 2018-08-06
tournament: Added template to the stub file

ZapAnton:
Contributes to #551

### DIFF
--- a/exercises/tournament/src/lib.rs
+++ b/exercises/tournament/src/lib.rs
@@ -1,1 +1,6 @@
-
+pub fn tally(match_results: &str) -> String {
+    unimplemented!(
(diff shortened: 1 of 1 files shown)


kunicmarko20:
Currently doing this and I was confused that I got an empty file, thank you!

exercises/tournament/src/lib.rs
```diff
@@ -1 +1,6 @@
-
+pub fn tally(matches_results: &str) -> String {
```
coriolinus: Quibble: the argument should be `match_results`; you only need to pluralize the second word, for reasons of English grammar which are out of scope here.

PR #607, opened 2018-08-07
parallel-letter-frequency: Added template to the stub file

ZapAnton:
Contributes to #551 

Was not sure about `worker_count` type:  on one hand it is a variable that hold the count of something, so
`i32` is natural here, on the other hand this variable would be used to chunk the input data in separate parts, therefore a lot of casts like `as usize`.

### DIFF
--- a/exercises/parallel-letter-frequency/src/lib.rs
+++ b/exercises/parallel-letter-frequency/src/lib.rs
@@ -1,1 +1,12 @@
+use std::collections::HashMap;
 
+pub fn frequency(input: &[&str], worker_count: i32) -> HashMap<char, u32> {
+    unimplemented!(
+        "Count the frequency of letters in the given input '{:?}'. Ensure that you are using {} to process the input.",
+        input,
+        match worker_count {
+            1 => format!("1 worker"),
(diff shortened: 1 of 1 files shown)


exercises/parallel-letter-frequency/src/lib.rs
```diff
@@ -1 +1,12 @@
+use std::collections::HashMap;
 
+pub fn frequency(input: &[&str], worker_count: i32) -> HashMap<char, u32> {
```
coriolinus: WRT the `worker_count` type: in this case I would probably use `usize`, though I wouldn't chunk the inputs naively. Chunking every `worker_count` letters and sending one char at a time to the workers would be slower than single-threaded, because of the time spent synchronizing a send.

The rationale for using `usize` instead of `u32` is simply that at some level, there needs to be a vector of workers, and the size limit for a vector is the pointer size cap.

coriolinus: One more note: the output should probably be `HashMap<char, usize>` on the basis that a `str` can contain at most `usize` bytes. This won't affect any of the tests here, but it's good practice.

PR #608, opened 2018-08-07
circular-buffer: Added template to the stub file

ZapAnton:
Contributes to the #551 

It is unclear what `write` method should return on success, so the return type is `Result<(), Error>`.

Perhaps `()` could be replaced with `usize` as to indicate the index of the successful write?

### DIFF
--- a/exercises/circular-buffer/src/lib.rs
+++ b/exercises/circular-buffer/src/lib.rs
@@ -1,1 +1,44 @@
+use std::fmt::Debug;
+use std::marker::PhantomData;
 
+pub struct CircularBuffer<T: Debug> {
+    // This field is here to make the template compile and not to
+    // complain about unused type parameter 'T'. Once you start
+    // solving the exercise, delete this field and the 'std::marker::PhantomData'
+    // import.
+    field: PhantomData<T>,
+}
+
+#[derive(Debug, PartialEq)]
+pub enum Error {
+    EmptyBuffer,
+    FullBuffer,
+}
+
+impl<T: Debug> CircularBuffer<T> {
+    pub fn new(capacity: usize) -> Self {
+        unimplemented!(
+            "Construct a new CircularBuffer with the capacity to hold {}.",
+            match capacity {
+                1 => format!("1 element"),
+                _ => format!("{} elements", capacity),
+            }
+        );
+    }
+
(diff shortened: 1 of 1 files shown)


exercises/circular-buffer/src/lib.rs
```diff
@@ -1 +1,44 @@
...
+    field: PhantomData<T>,
+}
+
+#[derive(Debug, PartialEq)]
+pub enum Error {
+    EmptyBuffer,
+    FullBuffer,
+}
+
+impl<T: Debug> CircularBuffer<T> {
```
petertseng: same comment as with `sublist` - I know that this `Debug` constraint is added purely so that we can reference variables in `unimplemented` message. We made a different choice in https://github.com/exercism/rust/commit/c5f7a433c5e954aa78162bfcab14eab96510bc4f and I'll consider whether it's better to use underscore prefix here. Input appreciated.

exercises/circular-buffer/src/lib.rs
```diff
@@ -1 +1,44 @@
...
+        unimplemented!(
+            "Construct a new CircularBuffer with the capacity to hold {}.",
+            match capacity {
+                1 => format!("1 element"),
+                _ => format!("{} elements", capacity),
+            }
+        );
+    }
+
+    pub fn write(&mut self, element: T) -> Result<(), Error> {
```
petertseng: I think `Result<(), Error>` makes sense and is the accepted way, as https://doc.rust-lang.org/std/io/trait.Write.html#method.write_all does this as well. While it would be possible to return the index of the inserted value, I think the API presented in this exercise is not conducive to a client of this code being able to do anything with that information, so I wouldn't add it.

exercises/circular-buffer/src/lib.rs
coriolinus: I think that we should not bound `T: Debug` here, and instead use an underscore variable. Perhaps it's worth identifying which of the exercises first uses an underscore variable in the track and adding a comment explaining why that is used, instead of a `Debug` trait bound.

exercises/circular-buffer/src/lib.rs
coriolinus: Agree: `Result<(), Error>` is fine and idiomatic.

PR #609, opened 2018-08-08
minesweeper: Added template to the stub file

ZapAnton:
Contributes to #551

### DIFF
--- a/exercises/minesweeper/src/lib.rs
+++ b/exercises/minesweeper/src/lib.rs
@@ -1,1 +1,3 @@
-
+pub fn annotate(minefield: &[&str]) -> Vec<String> {
+    unimplemented!("\nAnnotate the given minefield with the number of surrounding mines in each square (blank if there are no surrounding mines):\n{:#?}\n", minefield);
+}
--- a/exercises/minesweeper/src/lib.rs
+++ b/exercises/minesweeper/src/lib.rs
@@ -1,3 +1,3 @@
 pub fn annotate(minefield: &[&str]) -> Vec<String> {
-    unimplemented!("\nAnnotate the given minefield with the number of surrounding mines in each square (blank if there are no surrounding mines):\n{:#?}\n", minefield);
+    unimplemented!("\nAnnotate each square of the given minefield with the number of mines that surround said square (blank if there are no surrounding mines):\n{:#?}\n", minefield);
 }


exercises/minesweeper/src/lib.rs
```diff
@@ -1 +1,3 @@
-
+pub fn annotate(minefield: &[&str]) -> Vec<String> {
+    unimplemented!("\nAnnotate the given minefield with the number of surrounding mines in each square (blank if there are no surrounding mines):\n{:#?}\n", minefield);
```
petertseng: can you confirm: does this look better with the leading newline?

petertseng: I was pondering this sentence. If someone read the phrase "surrounding mines in each square" as one unit (rather than the way the sentence should actually be read), it may not be the best.

I'll present another possibility: "Annotate each square of the given minefield with the number of surrounding mines"

Is it better, worse, makes no difference? I may need some input here because English is not my first language

ZapAnton: `cargo test` with newline:
```text
---- mine_surrounded_by_spaces stdout ----
thread 'mine_surrounded_by_spaces' panicked at 'not yet implemented: 
Annotate the given minefield with the number of surrounding mines in each square (blank if there are no surrounding mines):
[
    "   ",
    " * ",
    "   "
]
', src/lib.rs:2:5
```
`cargo test` without newline:
```text
---- mine_surrounded_by_spaces stdout ----
thread 'mine_surrounded_by_spaces' panicked at 'not yet implemented: Annotate the given minefield with the number of surrounding mines in each square (blank if there are no surrounding mines):
[
    "   ",
    " * ",
    "   "
]
', src/lib.rs:2:5
```

I would say that the newline is needed here, so that the helper text is positioned near the minefield 'image'

ZapAnton: I think your wording is better.
Perhaps only a small change: "Annotate each square of the given minefield with the number of  mines that surround said square"

petertseng: 👍 in that case, I'll ask whether `rectangles` would look better with a newline. if so, feel free to add it there as well, sorry I did not wait to give you a chance before merging `rectangles`

petertseng: That's a great change!

PR #611, opened 2018-08-08
dominoes: Added template to the stub file

ZapAnton:
Contributes to #551


ZapAnton:
There is already `type Domino = (usize, usize)` in the test suite, so if the student wants to use it, they will include the type in their solution. I would not change the `chain` signature, so not to restrict the potential solutions.

petertseng:
OK, in that case I think that's enough time passed since the PR was submitted for any comments to be made. We'll merge as is, and thank you for another.

PR #613, opened 2018-08-08
luhn-from: Added template to the stub file

ZapAnton:
Contributes to #551

### DIFF
--- a/exercises/luhn-from/src/lib.rs
+++ b/exercises/luhn-from/src/lib.rs
@@ -11,8 +11,8 @@ impl Luhn {
 /// by hand for the every other type presented in the test suite,
 /// but your solution will fail if a new type is presented.
 /// Perhaps there exists a better solution for this problem?
-impl From<&'static str> for Luhn {
-    fn from(input: &'static str) -> Self {
+impl<'a> From<&'a str> for Luhn {
+    fn from(input: &'a str) -> Self {
         unimplemented!("From the given input '{}' create a new Luhn struct.", input);
     }
 }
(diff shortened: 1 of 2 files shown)


exercises/luhn-from/src/lib.rs
```diff
@@ -1 +1,18 @@
...
+impl Luhn {
+    pub fn is_valid(&self) -> bool {
+        unimplemented!("Determine if the current Luhn struct contains a valid credit card number.");
+    }
+}
+
+/// Here is the example of how the From trait could be implemented
+/// for the &str type. Naturally, you can implement this trait
+/// by hand for the every other type presented in the test suite,
+/// but you solution will fail if a new type is presented.
```
petertseng: "you solution" -> "your solution"

exercises/luhn-from/src/lib.rs
```diff
@@ -1 +1,18 @@
...
+/// Perhaps there exists a better solution for a problem?
```
petertseng: "a problem" -> "this problem" or "the problem"

exercises/luhn-from/src/lib.rs
```diff
@@ -1 +1,18 @@
...
+        unimplemented!("Determine if the current Luhn struct contains a valid credit card number.");
+    }
+}
+
+/// Here is the example of how the From trait could be implemented
+/// for the &str type. Naturally, you can implement this trait
+/// by hand for the every other type presented in the test suite,
+/// but your solution will fail if a new type is presented.
+/// Perhaps there exists a better solution for this problem?
+impl From<&'static str> for Luhn {
```
coriolinus: If we claim we are demoing this implementation for the `&str` type, then we should do it properly: `impl<'a> From<&'a str> for Luhn { ... }`. 

In general, it is bad practice to require the `'static` lifetime. There are exceptions, but they're rare, and not something that students will encounter.

PR #614, opened 2018-08-08
luhn-trait: Added template to the stub file

ZapAnton:
Contributes to #551

### DIFF
--- a/exercises/luhn-trait/src/lib.rs
+++ b/exercises/luhn-trait/src/lib.rs
@@ -1,1 +1,14 @@
+pub trait Luhn {
+    fn valid_luhn(&self) -> bool;
+}
 
+/// Here is the example of how to implement custom Luhn trait
+/// for the &str type. Naturally, you can implement this trait
+/// by hand for the every other type presented in the test suite,
+/// but you solution will fail if a new type is presented.
+/// Perhaps there exists a better solution for a problem?
+impl Luhn for &'static str {
+    fn valid_luhn(&self) -> bool {
+        unimplemented!("Determine if '{}' is a valid credit card number.", self);
+    }
+}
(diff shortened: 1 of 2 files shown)


exercises/luhn-trait/src/lib.rs
```diff
@@ -1 +1,14 @@
+pub trait Luhn {
+    fn valid_luhn(&self) -> bool;
+}
 
+/// Here is the example of how to implement custom Luhn trait
+/// for the &str type. Naturally, you can implement this trait
+/// by hand for the every other type presented in the test suite,
+/// but you solution will fail if a new type is presented.
```
petertseng: "you solution" should probably be "your solution", right?

exercises/luhn-trait/src/lib.rs
```diff
@@ -1 +1,14 @@
...
+/// Perhaps there exists a better solution for a problem?
```
petertseng: I think "a problem" should be "this problem" or "the problem", otherwise it might not refer to the problem presented in the previous sentence.

ZapAnton:
The `&'static str` type is here for demonstration purpose of how you implement a trait for specific type and was chosen simply because the `&str` is present in the first test case. 

For the purpose of the template it could be easily replaced with `String` (since the template can not compile in either case).

Should I make the change?

exercises/luhn-trait/src/lib.rs
```diff
@@ -1 +1,14 @@
+pub trait Luhn {
+    fn valid_luhn(&self) -> bool;
+}
 
+/// Here is the example of how to implement custom Luhn trait
+/// for the &str type. Naturally, you can implement this trait
+/// by hand for the every other type presented in the test suite,
+/// but your solution will fail if a new type is presented.
+/// Perhaps there exists a better solution for this problem?
+impl Luhn for &'static str {
```
coriolinus: Same comment as in `luhn-from`: if we claim we're going to implement this for `&str`, then we should do it properly, with `impl<'a> Luhn for &'a str { ... }`.

PR #616, opened 2018-08-09
ensure-lib-src-rs-exist: Updated the script to check if the stub file is empty

### LINKED ISSUE #551: Some exercises are missing templates in lib.rs

ZapAnton:
As discussed in #551 the script in this PR now has a check to ensure that the stub files of the exercises
are not empty.

The reasons why `ensure-lib-src-rs-exist.sh` was chosen for this PR and not `ensure-stubs-compile.sh` are:
- The content of the stub file is not a concern of `ensure-stubs-compile.sh` - it should check if the stub compiles, even if it is empty
- When `ensure-lib-src-rs-exist.sh` checks for the stub and thinks it exists only because the file exists (even though it could be empty) is sort of false-positive

The `[ "$(cat "$dir/src/lib.rs")" == "" ]` line was added for the check, because when `rustfmt` works on an empty file it inserts the `0a` symbol in it, and the `-s` flag check is useless in this case.

In case this PR is merged, closes #551

### DIFF
--- a/_test/ensure-lib-src-rs-exist.sh
+++ b/_test/ensure-lib-src-rs-exist.sh
@@ -15,18 +15,19 @@ IGNORED_EXERCISES=(
 )
 
 for dir in $repo/exercises/*/; do
-  exercise=$(basename "$dir")
-  if [ ! -f "$dir/src/lib.rs" ]; then
-	# https://github.com/exercism/rust/pull/270
-	echo "$exercise is missing a src/lib.rs; please create one (an empty file is acceptable)"
-	missing="$missing\n$exercise"
+	exercise=$(basename "$dir")
+
+	if [ ! -f "$dir/src/lib.rs" ]; then
+		# https://github.com/exercism/rust/pull/270
+		echo "$exercise is missing a src/lib.rs; please create one (an empty file is acceptable)"
+		missing="$missing\n$exercise"
 	else
 		#Check if the stub file is empty
 		if [ ! -s "$dir/src/lib.rs" ] || [ "$(cat "$dir/src/lib.rs")" == "" ] && [[ " ${IGNORED_EXERCISES[*]} " != *"$exercise"* ]]; then
 			echo "$exercise has src/lib.rs stub file, but it is empty."
 			empty_stub="$empty_stub\n$exercise"
 		fi
-  fi
+	fi
 done
 
 if [ -n "$missing" ]; then
--- a/_test/ensure-lib-src-rs-exist.sh
+++ b/_test/ensure-lib-src-rs-exist.sh
@@ -4,16 +4,45 @@ repo=$(cd "$(dirname "$0")/.." && pwd)
 
 missing=""
 
+empty_stub=""
+
+check_status=0
+
+IGNORED_EXERCISES=(
+	"two-fer" #deprecated
+	"nucleotide-codons" #deprecated
+	"hexadecimal" #deprecated
+)
+
 for dir in $repo/exercises/*/; do
   exercise=$(basename "$dir")
-  if [ ! -f $dir/src/lib.rs ]; then
-    # https://github.com/exercism/rust/pull/270
-    echo "$exercise is missing a src/lib.rs; please create one (an empty file is acceptable)"
-    missing="$missing\n$exercise"
+  if [ ! -f "$dir/src/lib.rs" ]; then
+	# https://github.com/exercism/rust/pull/270
+	echo "$exercise is missing a src/lib.rs; please create one (an empty file is acceptable)"
+	missing="$missing\n$exercise"
+	else
+		#Check if the stub file is empty
+		if [ ! -s "$dir/src/lib.rs" ] || [ "$(cat "$dir/src/lib.rs")" == "" ] && [[ " ${IGNORED_EXERCISES[*]} " != *"$exercise"* ]]; then
+			echo "$exercise has src/lib.rs stub file, but it is empty."
+			empty_stub="$empty_stub\n$exercise"
+		fi
   fi
 done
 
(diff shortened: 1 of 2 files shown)


ZapAnton:
Seems like the script could not be run, because of the syntax error in the list creation.
Is it a shell syntax conflict? Because the script worked in my shell.

coriolinus:
Arrays are bash syntax not available in `sh`. This script is [currently being run via `sh`](https://github.com/exercism/rust/blob/c53330c3d84ace6c15107e5e2b4795de2e998416/.travis.yml#L5). Recommended fix: run it with bash.

ZapAnton:
For consistency perhaps every other script should be run with the `bash` too?

coriolinus:
I don't have strong feelings on that either way. Just be careful not to
break something which uses `sh`-specific features not available in bash.

ZapAnton:
The script now works as intended, and Travis fails because there are still exercises which lack the template in the stub file.

For now I will wait for a comment from maintainers whether this should be rebased or simply merged when other templates are added (or any more changes should be made).

Making the syntax and calls of the Travis shell scripts would be moved to the other issue, since they are not a concern in this specific PR.

_test/ensure-lib-src-rs-exist.sh
```diff
@@ -4,16 +4,46 @@ repo=$(cd "$(dirname "$0")/.." && pwd)
...
+check_status=0
+
+IGNORED_EXERCISES=(
+	"two-fer" #deprecated
+	"nucleotide-codons" #deprecated
+	"hexadecimal" #deprecated
+)
+
 for dir in $repo/exercises/*/; do
-  exercise=$(basename "$dir")
```
petertseng: are these changes purely for re-indenting the lines?

I would ask that re-indent be its own commit (or not included at all)

_test/ensure-lib-src-rs-exist.sh
```diff
@@ -4,16 +4,46 @@ repo=$(cd "$(dirname "$0")/.." && pwd)
...
-  fi
+	exercise=$(basename "$dir")
+
+	if [ ! -f "$dir/src/lib.rs" ]; then
+		# https://github.com/exercism/rust/pull/270
+		echo "$exercise is missing a src/lib.rs; please create one (an empty file is acceptable)"
+		missing="$missing\n$exercise"
+	else
+		#Check if the stub file is empty
+		if [ ! -s "$dir/src/lib.rs" ] || [ "$(cat "$dir/src/lib.rs")" == "" ] && [[ " ${IGNORED_EXERCISES[*]} " != *"$exercise"* ]]; then
```
petertseng: (despite the above comment, I understand that these are new lines)

_test/ensure-lib-src-rs-exist.sh
```diff
@@ -4,16 +4,46 @@ repo=$(cd "$(dirname "$0")/.." && pwd)
...
+	check_status=1
+fi
+
+if [ -n "$empty_stub" ]; then
+	printf "\nExercises with empty src/lib.rs stub file:$empty_stub\n"
+
+	check_status=1
+fi
+
+if [ "$check_status" -ne 0 ]; then
```
petertseng: this could potentially be written as `exit $check_status` - was this intentionally written the proposed way for clarity, or some other reason?

ZapAnton: Two reasons actually:
- For clarity
- Separation of concern: `check_status` is used for internal script logic `exit 0/1` is used for external system(Travis in this case)

If this feels like unnecessary logic to you I will change it to `exit $check_status`

_test/ensure-lib-src-rs-exist.sh
ZapAnton: May I ask why?
The lines I have added for this PR have their own indent and if i leave the above line as they were, it would hurt readability.

petertseng: > May I ask why?

Well, I thought that if a line is changed in a way that isn't consistent with the commit message, it would cause confusion, and make it harder to understand the git history.

> if i leave the above line as they were, it would hurt readability.

indeed, so I think I should not suggest they are left as they are. Putting the indentation change in its own commit should accomplish both the goals though (understand history, and keep the indentation consistent)

petertseng: Note that I understand I am causing non-trivial hardship here, by asking for the re-indenting be its own commit. It is not necessarily the most pleasant experience to have to redo the commits to split them in two.

I offer this remedy:

Since I'm the one who cares most about this, I offer to make it happen. However I can only execute this offer at a time more convenient to me. Luckily we still have time before all exercises have their stubs, so we are not in a huge hurry.

There is no obligation to accept this offer, of course.

petertseng: > make it harder to understand the git history

and make it harder to review. Now a reviewer needs to think to see whether a line was changed *only* for reindenting, or if it has a change other than reindent. A separate commit makes it clear to the reviewer.

ZapAnton:
@petertseng I have rewritten my commits to extract the indenting changes into separate commit. Does it look good to you?

ZapAnton:
I have rebased against the `master` brach and now Travis is all good

Issue #356, opened 2017-10-04
Link to first edition of the rust book


janczer:
This is the link to first edition of the rust book:

https://github.com/exercism/rust/blob/45ac147de3277552dbaba3f00e52ed6fd36c14ab/exercises/gigasecond/tests/gigasecond.rs#L11

I'm not sure, maybe link to second edition would be more useful:

```
https://doc.rust-lang.org/book/second-edition/ch02-00-guessing-game-tutorial.html#generating-a-random-number
```

ijanos:
You are right. You can create a pull request if you wish and we will merge it.

shaleh:
This should also happen in the Getting Started doc in hello-world

petertseng:
OK, I guess since the https://doc.rust-lang.org/book/ says "We suggest reading the second edition and then checking out the first edition later to pick up some of the more esoteric parts of the language." I guess it's prudent for all our links to go to the second edition.

It looks like this issue should be kept open until all links to specific parts in the first edition are replaced with second edition.

I think those are:

```
config/exercise-readme-insert.md:Make sure to read the [Crates and Modules](https://doc.rust-lang.org/stable/book/crates-and-modules.html) chapter if you
config/exercise-readme-insert.md:[crates-and-modules]: http://doc.rust-lang.org/stable/book/crates-and-modules.html
docs/ABOUT.md:Rust aims to bring modern language design and an advanced type system to systems programming. Rust does not use a garbage collector, but instead ensures safety and optimizes code at compile time. The concept of [ownership](https://doc.rust-lang.org/book/ownership.html) is how Rust achieves its largest goal, memory safety.
docs/TESTS.md:The directory must be named `src` and the file must be named `lib.rs` otherwise your code will not compile. For more details, check out the rustlang book [chapter on crates and modules](http://doc.rust-lang.org/stable/book/crates-and-modules.html)
docs/TESTS.md:Because Rust checks all code at compile time you may find that your tests won't compile until you write the required code. Even `ignore`d tests are checked at compile time. You can [comment out](https://doc.rust-lang.org/stable/book/comments.html) tests that won't compile by starting each line with a `//`. Then, when you're ready to work on that test, you can un-comment it.
exercises/bracket-push/HINTS.md:- Lifetimes and Structs: https://doc.rust-lang.org/book/lifetimes.html#impl-blocks
[...]

PR #617, opened 2018-08-13
Update "The Rust Book" links to 2018-edition.

joshuayoerger:
Update `The Rust Programming Language` link to point to 2018-edition.

> The 2018 edition of the book is a "living version" of the book; based on the second edition, it will be updated as Rust updates. If you're not sure what version of the book to read, you should prefer this version. [(Source)](https://doc.rust-lang.org/book/)

### DIFF
--- a/docs/LEARNING.md
+++ b/docs/LEARNING.md
@@ -1,6 +1,6 @@
 ## Learning Rust
 
-* [The Rust Programming Language](https://doc.rust-lang.org/book/second-edition/) is a great resource for getting started with Rust as well as diving deeper into specific features of Rust.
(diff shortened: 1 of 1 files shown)


petertseng:
I think we should merge this (but I think the commit message should be updated so that we're saying we're linking to 2018 in the first line, not just "Update LEARNING.md" in the first line), but also simultaneously file an issue to replace other second-edition links (like what we did for first->second https://github.com/exercism/rust/issues/356). There are a lot of them, `git grep -i second-edition | wc -l` says 382, but 3*(num_exercises+1) of them are generated so easy enough.

PR #618, opened 2018-08-13
accumulate: make exercise more idiomatic

Emerentius:
The distinction between fn pointers and closures is unusal and unnecessary. The common solution with generics works for both anyway.

Furthermore, it was evident from the old examples that this was never about
function pointers (`fn(i32) -> i32`) like the hint claimed, but about trait objects (`&Fn(i32) ->
i32`). Those also implement the `Fn` trait and therefore they, too, work with the generic version.

Edit: I just thought of, maybe I should also change the input and output type. With equal types everything can be done in place which I think takes away from the exercise, it's just iteration, function calling and mutation. A general accumulate has to convert between types which forces creation of a new `Vec`. If the input type is not `Copy`, this also ties into move semantics and ownership.

What do you think?

### DIFF
--- a/exercises/accumulate/example.rs
+++ b/exercises/accumulate/example.rs
@@ -1,18 +1,9 @@
-pub fn map_function(values: Vec<i32>, f: &Fn(i32) -> i32) -> Vec<i32> {
-    let mut out: Vec<i32> = vec![];
-    for val in values {
-        out.push(f(val))
-    }
-    out
-}
-
-pub fn map_closure<F>(values: Vec<i32>, f: F) -> Vec<i32>
+pub fn map<F>(mut values: Vec<i32>, f: F) -> Vec<i32>
 where
     F: Fn(i32) -> i32,
 {
-    let mut out: Vec<i32> = vec![];
-    for val in values {
-        out.push(f(val))
+    for val in &mut values {
+        *val = f(*val);
     }
-    out
+    values
 }
--- a/exercises/accumulate/tests/accumulate.rs
+++ b/exercises/accumulate/tests/accumulate.rs
@@ -1,24 +1,16 @@
 extern crate accumulate;
 
-use accumulate::{map_closure, map_function};
+use accumulate::map;
 
 fn square(x: i32) -> i32 {
     x * x
-}
-
-fn abs_val(x: i32) -> i32 {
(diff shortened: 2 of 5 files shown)


exercises/accumulate/example.rs
```diff
@@ -1,18 +1,9 @@
-pub fn map_function(values: Vec<i32>, f: &Fn(i32) -> i32) -> Vec<i32> {
-    let mut out: Vec<i32> = vec![];
-    for val in values {
-        out.push(f(val))
-    }
-    out
-}
-
-pub fn map_closure<F>(values: Vec<i32>, f: F) -> Vec<i32>
+pub fn map<F>(mut values: Vec<i32>, f: F) -> Vec<i32>
```
petertseng: Interesting that it mutates `values`. On the one hand, this behaviour is unlike traditional `map`. On the other hand, the tests can't tell the difference, and they also do not preclude a non-mutating implementation. This is worth its own issue and not worth holding up this PR.

exercises/accumulate/tests/accumulate.rs
```diff
@@ -1,94 +1,86 @@
 extern crate accumulate;
 
-use accumulate::{map_closure, map_function};
+use accumulate::map;
 
 fn square(x: i32) -> i32 {
     x * x
 }
 
-fn abs_val(x: i32) -> i32 {
```
petertseng: I considered asking for this change to be its own commit, but I suppose it's a reasonable thing to roll in since all the lines that reference `abs_val` are already changed in this PR anyway.

petertseng:
> With equal types everything can be done in place which I think takes away from the exercise, it's just iteration, function calling and mutation. A general accumulate has to convert between types which forces creation of a new Vec. If the input type is not Copy, this also ties into move semantics and ownership.

Ah, right, that'd all be valuable to add, for the reasons you've discussed. I was going to open an issue for it anyway, but you can feel free to make a PR for it.

exercises/accumulate/example.rs
petertseng: > This is worth its own issue

https://github.com/exercism/rust/issues/628

petertseng:
> maybe I should also change the input and output type

indeed, https://github.com/exercism/rust/issues/629

PR #619, opened 2018-08-14
Added binary-search-tree to the track

ZapAnton:
This PR adds `binary-search-tree` exercise to the track.

The main motivation behind the exercise is to introduce the `unsafe` subset of Rust, namely working with raw pointers.

There is a couple of questions in addition to future requests/suggestions:
- Should the binary tree in the exercise contain generic data or `i32` will suffice?
- Right now the only thing that makes students to use `unsafe` is a return type of the `get_root` method and a hint in the README. Should the usage of `unsafe` be more enforced on the students (for instance adding a struct in the template that contains raw pointers, although it is far too limiting)?
- In case this exercise is accepted and merged should the similar  exercise be added, but without the `unsafe` part, that introduces the usage of `Rc`, `RefCell`, `Weak` etc. (just like `luhn-from` and `luhn-trait` case)?


coriolinus:
@ZapAnton I'm on vacation right now and don't expect to have time to properly review this for another few weeks; please be patient. In the meantime, my answers to your discussion questions:

- student-implemented collections should be generic. 
- we shouldn't force `unsafe` on students. The hint in the README should suffice to let them know that `unsafe` is a way do to this soundly, but it is also valuable practice to implement this using `Rc<Cell<_>>`.
- `get_root` should return `&mut TreeNode`, not a raw pointer. That's unidiomatic for the purpose of forcing the student to implement the exercise in a particular way.
- Instead of adding a second exercise, I'd prefer expanding on the hint: say that it's possible to solve this in a relatively straightforward way with `unsafe`, but it's also possible to handle it with `Rc<Cell<_>>`. It might even be worth building a feature-gated benchmark that students with a nightly compiler can run, so that if they want to build both implementations, they can compare the performance. Let the student pick the implementation path; they'll get something from it either way.

Emerentius:
The ownership model is hierarchical and only non-hierarchical data structures or data accesses have problems with the borrow checker. A tree can be implemented entirely in safe Rust.

`insert_node` does not require `&self`. It's better as an associated method.
`get_root(&self) -> *mut` this looks suspiciously like shared, mutable state.

I've refactored the code into safe Rust. You can see it here:
https://github.com/Emerentius/rust-1/tree/add_binary_search_tree/exercises/binary-search-tree

Feel free to merge it into your own branch.

ZapAnton:
I have added the idiomatic code changes to the exercise as per @coriolinus request.

I feel that I can not progress further before the following moments will be addressed:

- Should the additional structure be added to the tree, namely parent node for `TreeNode` and a delete method for the tree. That will make students to create more complex solutions, since the current version of the exercise could be solved using `safe` Rust without a lot of magic as @Emerentius pointed out.

- The point of `unsafe` in the exercise. If we agree that it is a valuable topic to learn, then I fear we might miss the learning opportunity in this exercise by not fully focusing on the `unsafe` solution, because implementing recursive data structures is the easiest way to use `unsafe`, other being some chosen methods from std like `String::from_raw_parts` (which are not that good for learning) or using the FFI (which may require the implementation of Rust-specific exercise on the track).

exercises/binary-search-tree/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+
+    pub fn insert(&mut self, element: i32) {
+        unimplemented!(
+            "Insert the '{}' element into the BinarySearchTree struct.",
+            element
+        );
+    }
+}
+
+impl<T> Into<Vec<T>> for BinarySearchTree<T> {
```
coriolinus: It's more general to implement `IntoIterator` for `BinarySearchTree<T>`; you can still get a `Vec<T>` pretty easily with a `collect()` call, but you could also get a number of other container types.

PR #620, opened 2018-08-14
brackets: simplify required interface

Emerentius:
I'm just going to quote Peter Tillemans' post from the slack channel. He put it very nicely.

>Peter Tillemans [11 août à 23 h 35]
Regarding the Brackets exercise, the way the challenge is made, the student is forced to use an object-like interface where you have to initialise the 'object' before calling a method on it. This 'object' ends usually up with either the result of the brackets balance or the input text or some intermediate value of the data transformation pipeline. This challenge lends itself well to a purely functional solution. I also feel it pushes students to use stateful solutions where none are warranted. (Actually @mawis pointed this out originally).

The hint in the Readme points students towards using lifetimes, but most just accept the `impl From<&'a str> for Brackets` as is. They don't work with lifetimes themselves, they only conform to the interface.

### DIFF
--- a/exercises/bracket-push/src/lib.rs
+++ b/exercises/bracket-push/src/lib.rs
@@ -1,13 +1,3 @@
-pub struct Brackets;
-
-impl<'a> From<&'a str> for Brackets {
-    fn from(input: &str) -> Self {
-        unimplemented!("From the '{}' input construct a new Brackets struct", input);
-    }
-}
-
-impl Brackets {
-    pub fn are_balanced(&self) -> bool {
-        unimplemented!("Check if your Brackets struct contains balanced brackets");
-    }
+pub fn brackets_are_balanced(string: &str) -> bool {
+    unimplemented!("Check if the string \"{}\" contains balanced brackets", string);
 }
--- a/exercises/bracket-push/tests/bracket-push.rs
+++ b/exercises/bracket-push/tests/bracket-push.rs
@@ -1,82 +1,82 @@
 extern crate bracket_push;
 
-use bracket_push::*;
+use bracket_push::brackets_are_balanced;
 
 #[test]
 fn paired_square_brackets() {
-    assert!(Brackets::from("[]").are_balanced());
+    assert!(brackets_are_balanced("[]"));
 }
 
 #[test]
 #[ignore]
 fn empty_string() {
-    assert!(Brackets::from("").are_balanced());
+    assert!(brackets_are_balanced(""));
 }
 
 #[test]
 #[ignore]
 fn unpaired_brackets() {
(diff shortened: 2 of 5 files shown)


exercises/bracket-push/tests/bracket-push.rs
```diff
@@ -1,88 +1,88 @@
 extern crate bracket_push;
 
-use bracket_push::*;
+use bracket_push::brackets_are_balanced;
```
petertseng: note: If the crate name were `brackets`, I would have wanted the name to be `brackets::are_balanced` to avoid repetition. However, looks like we are using `bracket_push`. So I think `brackets_are_balanced` has to be the name.

 It can be worth a discussion whether to change `bracket_push` (why is it even called bracket-push??? https://github.com/exercism/problem-specifications/issues/693 ) but this PR does not need to depend on it.

exercises/bracket-push/Cargo.toml
```diff
@@ -1,3 +1,3 @@
 [package]
 name = "bracket-push"
-version = "1.2.0"
+version = "1.3.0"
```
petertseng: this should not change until the test cases match https://github.com/exercism/problem-specifications/blob/master/exercises/bracket-push/canonical-data.json 1.3.0. It should go in a separate PR.

Emerentius: Ah ok, so that's what it's used for.

Emerentius:
I've reverted the change to the version.

exercises/bracket-push/src/lib.rs
```diff
@@ -1,13 +1,3 @@
...
-    fn from(input: &str) -> Self {
-        unimplemented!("From the '{}' input construct a new Brackets struct", input);
-    }
-}
-
-impl Brackets {
-    pub fn are_balanced(&self) -> bool {
-        unimplemented!("Check if your Brackets struct contains balanced brackets");
-    }
+pub fn brackets_are_balanced(_string: &str) -> bool {
```
petertseng: I understand that the leading underscore is to say it is OK that it is unused.

I would like to suggest instead the model established in https://github.com/exercism/rust/pull/438 - call it `string` and reference it in the `unimplemented` message. The benefit of the approach is that it avoids a potential student misunderstanding, thinking that the variable should remain named `_string`.

Otherwise, this will fall under https://github.com/exercism/rust/issues/476 which asked us to remove all instances of variables with leading underscores

Emerentius: Yeah, that sounds reasonable. It felt wrong to put it there in the first place to get past the (denied) warning.

petertseng:
Thanks!

PR #621, opened 2018-08-14
saddle-points: add test for edge case

Emerentius:
A commonly unhandled case is when two equally maximum elements
occur in a row that are both minima in their respective columns.

There is another test case already that has two equally maximum elements, but when a student checks only one element in a row and uses a max finder that always returns the right-most element of multiple equally large elements (like `Iterator::max`), they happen to get the right result by luck.


Emerentius:
I've copied this test from the python track actually. If I understand the system correctly,  I think this exercise's test suite is still from version 1.1 of the problem specifications which have since changed to 1.3.
So I'm thinking, I should just update the exercise's version in `Cargo.toml` and the rest of the test suite to conform to the new specifications. Is that right?

coriolinus:
That would be very helpful. Thanks!

Emerentius:
I've updated the test suite. It was hard to see the matrices when all the rows where kept in one line so I split them apart. Sorry for the diff noise. It turns out that the track had several extensions of the standard tests. I replaced 2 with the overlapping ones from `problem-specifications` and kept one more around.
While I was at it, I removed all the redundant `test_` prefixes of the tests.

PR #622, opened 2018-08-14
leap: add more test cases

Emerentius:
The exercise has just 5 test cases. This can easily let some bugs slip through.
This adds a check for a range of years (1600 to 1699) and a few additional centuries. The range is kept short so the user isn't spammed with a huge list of incorrect years with wrong code. It also avoids having the full solution in the tests where students could stumble upon it.

I got this faulty code from a student which passes all the old tests:  
```rust
year % 4 == 0 && !year % 100 == 0 || year % 400 == 0
```

### DIFF
--- a/exercises/leap/tests/leap.rs
+++ b/exercises/leap/tests/leap.rs
@@ -14,12 +14,27 @@ fn test_any_old_year() {
 #[test]
 #[ignore]
 fn test_century() {
+    assert_eq!(leap::is_leap_year(1700), false);
+    assert_eq!(leap::is_leap_year(1800), false);
     assert_eq!(leap::is_leap_year(1900), false);
 }
 
 #[test]
 #[ignore]
 fn test_exceptional_centuries() {
+    assert_eq!(leap::is_leap_year(1600), true);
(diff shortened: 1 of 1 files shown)


Emerentius:
I thought on this a bit more. I found it puzzling that bitwise inversion would actually make it work.
`-n = !n + 1` (two's complement) or put another way, `!n = -n - 1`. It just shifts the centuries by 1 and in that case the first modulo 4 check never passes.

So why did I think it passes all the tests? Because all the centuries tests check for `false` which this happens to return and `cargo test -- --ignored` doesn't run all tests, only the ignored ones. The very first test would have detected this.

We can ignore this PR but we should get a way to just run all tests in one command.

petertseng:
For this reason, I have always been forced to run `cargo test && cargo test -- --ignored`. I continue to look for any better solutions. Let me know of any.

Emerentius:
I found this relevant issue: https://github.com/rust-lang/rust/issues/50363
A quick search through the source of libtest makes me believe that adding an `--all` flag shouldn't be difficult. Compiling rustc will be the biggest hurdle for me, but I'll try to tackle this.

PR #635, opened 2018-08-23
clock: Add test case for comparing 24:00 and 00:00

felix91gr:
This test helped me find a bug in my solution, specifically where I did the modulo operation.

I had this:

```rust
    fn mod_n(a: i32, n: i32) -> i32 {
        if n <= 0 {
            panic!("Tried mod with negative base: {:?}", n);
        }

        if a > 0 {
            a % n
        }
        else {
            a % n + n
        }
    }
```

And it should've been this:

```rust
    fn mod_n(a: i32, n: i32) -> i32 {
        if n <= 0 {
            panic!("Tried mod with negative base: {:?}", n);
        }

        if a >= 0 {
            a % n
        }
        else {
            a % n + n
        }
    }
```

"0 hours" was treated like a negative amount, but the solution still passed all the tests. Anyway, I hope this helps, and lemme know if I should change something!

### DIFF
--- a/exercises/clock/tests/clock.rs
+++ b/exercises/clock/tests/clock.rs
@@ -328,3 +328,9 @@ fn test_compare_clocks_with_negative_hours_and_minutes() {
 fn test_compare_clocks_with_negative_hours_and_minutes_that_wrap() {
     assert_eq!(Clock::new(18, 7), Clock::new(-54, -11513))
 }
+
+#[test]
+#[ignore]
+fn test_compare_full_clock_and_zeroed_clock() {
+    assert_eq!(Clock::new(24, 0), Clock::new(0, 0))
+}


felix91gr:
CI failed misteriously:

image

Any idea of why it did?

ZapAnton:
Considering that every other build is successful, it is probably just Travis configuration fault. It does that from time to time.

exercises/clock/tests/clock.rs
```diff
@@ -328,3 +328,9 @@ fn test_compare_clocks_with_negative_hours_and_minutes() {
 fn test_compare_clocks_with_negative_hours_and_minutes_that_wrap() {
     assert_eq!(Clock::new(18, 7), Clock::new(-54, -11513))
 }
+
+#[test]
+#[ignore]
+fn test_compare_full_day_and_no_day() {
+    assert_eq!(Clock::new(24, 00), Clock::new(0, 0))
```
petertseng: I think for consistency, it would be good if the `00` in this line were to change to `0`, OR the `0` in this line change to `00`. Right?

petertseng:
I will prefix the PR title to contain `clock:` in its name. I request to please do so for the commit message as well, so that it is possible to see in log which exercise is affected, without having to resort to flags like `-p` or `--stat`

exercises/clock/tests/clock.rs
felix91gr: Indeed, thank you! :)

felix91gr:
Thanks for the suggestions, I've applied them all (I think). I'm gonna take a look at the canonical data now so that this can be upstreamed :)

coriolinus:
Now that the problem specifications have been updated with this case, we should update the upstream version in `Cargo.toml`. Otherwise, LGTM.

felix91gr:
Wait, I think I should re-do this PR regenerating the test case from the specifications, now that I merged it [here](https://github.com/exercism/problem-specifications/pull/1306). I prefer that, just so the naming and format are consistent with the upstream's. How can I do that?

coriolinus:
Unfortunately, the process for updating an exercise from its problem specifications is mostly manual. One experimental workflow which may help semi-automate this:

- move the `clock` directory to make a backup
- run `bin/init_exercise.py --dont-update-config` to regenerate from specifications
- overwrite all the generated code _except_ `tests/clock.rs` with the existing code
- update `tests/clock.rs` appropriately to get things working again

However, it may be simpler to just update by hand.

felix91gr:
I've updated the naming manually. Now this is the only thing left:

> we should update the upstream version in `Cargo.toml`

Which `Cargo.toml` should I update? The one for `clock`? And what version should I list? The version of `canonical-data.json`? I ask because we're listing currently the `1.0.0`, and that file is already at `2.3.0` c:

coriolinus:
That's the correct file, but I didn't realize that we'd gotten so far
behind.

If you're willing, it would be very helpful if you could just go through
that file and ensure we have a test for every case there. If we do, then we
can claim to support that version of the canonical data. If not, then we
need to add tests before updating the version. The intent here is to ensure
that our test suite is a superset of the tests in the canonical data for
the declared version.

Given that this is non-trivial work, you may of course forego it and leave
the Cargo.toml file alone. I will not prevent merging this PR on the basis
of an outdated version claim.

felix91gr:
Yeah, I'd prefer to leave it for a later PR, just because of modularization. I'll open a new issue to have the pending update documented.

coriolinus:
That's fine. Thanks for all the work you've put into this!

PR #637, opened 2018-08-25
Update README template and regenerate READMEs

Emerentius:
Uploading an incomplete solution no longer gives access to other people's submissions.


Emerentius:
Actually, it may be worth it to wait and see what the general solution to this will be.
https://github.com/exercism/exercism/issues/4076 is the relevant issue

Emerentius:
I'll monitor the "upstream" decision and close this for now.

Issue #629, opened 2018-08-21
accumulate: can/should it test for mapping to a different type?


petertseng:
Traditionally `map` would be `(a -> b) -> [a] -> [b]`, but here we only have `(a -> a) -> [a] -> [a]` (and in fact we only have `i32` anyway). Can/should it test for mapping to a different type?

Note that this would subsume https://github.com/exercism/rust/issues/628 ; you'd not be able to mutate the input if it were a different type, so this could make both proposed changes at once

Emerentius:
I think this would be a good change. We can ratchet it up through the tests step by step, for example like this:  

1. `i32 -> i32`
Requires generics, at least `Fn(i32) -> i32`

2. `i64 -> i64`
Generic input / output, requires at least `Fn(T) -> T, T: Copy`
but user may still do `for &x in input.iter() { ... }`

3. `&mut i32 -> &mut i32` => `Fn(T) -> T`
Or some other Non-Clone type. Force moving.
`for x in input { .. }`

4. `String -> Vec`    => `Fn(T) -> U`
Differentiate input/output

5. `FnMut`, closure with mutation.
`Fn` is a needlessly high expectation for a function. It just needs to be repeatedly callable.

This simple exercise is hiding a fair amount of complexity if it's fully generalized: Generics, the Fn traits and their differences, `IntoIterator` and moving in general. Non-copy types have come up before but the moving concept may not be fully understood, yet.

Maybe this is too much at once at that point.

coriolinus:
Agree that it can and should test mapping to a new type.

> Maybe this is too much at once?

No, I really like that test ratchet, @Emerentius.

coriolinus:
Implemented in https://github.com/exercism/rust/issues/629.

PR #645, opened 2018-08-31
accumulate: expand tests to require general solutions

Emerentius:
Implements the test scheme I proposed in #629 which enforces that solutions are generally applicable and minimally restrictive.
Because ignored tests still cause compile failures, I had to comment them out entirely. I've used a multi-line comment so that including more tests can be done by just moving 2 characters. This is similar to how the C++ track organizes their tests and it's also quite fitting here, because it further encourages going through the tests in order.
An alternative would be to add a series of cargo features and conditionally compile the tests but that would be more work for students and a new concept to learn.

### DIFF
--- a/config.json
+++ b/config.json
@@ -481,7 +481,9 @@
       "unlocked_by": "luhn",
       "difficulty": 4,
       "topics": [
-        "function_pointer"
+        "generics",
+        "higher_order_functions",
+        "move_semantics"
       ]
     },
     {
--- a/exercises/accumulate/.meta/hints.md
+++ b/exercises/accumulate/.meta/hints.md
@@ -1,7 +1,14 @@
 ## Hints
 
-It may help to look at the [Fn trait](https://doc.rust-lang.org/std/ops/trait.Fn.html).
+It may help to look at the Fn\* traits:
+[Fn](https://doc.rust-lang.org/std/ops/trait.Fn.html),
+[FnMut](https://doc.rust-lang.org/std/ops/trait.Fn.html) and
+[FnOnce](https://doc.rust-lang.org/std/ops/trait.Fn.html).
 
 Help with passing a closure into a function may be found in
 the ["closures as input parameters" section](https://doc.rust-lang.org/stable/rust-by-example/fn/closures/input_parameters.html) of
 [Rust by Example](https://doc.rust-lang.org/stable/rust-by-example/).
+
+The tests for this exercise will cause compile time errors,
+if your function signature does not fit them, even when they're not run.
+You may want to comment some tests out and generalize your solution piece by piece.
--- a/exercises/accumulate/README.md
+++ b/exercises/accumulate/README.md
@@ -27,11 +27,18 @@ Solve this one yourself using other basic tools instead.
 
 ## Hints
 
-It may help to look at the [Fn trait](https://doc.rust-lang.org/std/ops/trait.Fn.html).
+It may help to look at the Fn\* traits:
+[Fn](https://doc.rust-lang.org/std/ops/trait.Fn.html),
+[FnMut](https://doc.rust-lang.org/std/ops/trait.Fn.html) and
+[FnOnce](https://doc.rust-lang.org/std/ops/trait.Fn.html).
 
 Help with passing a closure into a function may be found in
 the ["closures as input parameters" section](https://doc.rust-lang.org/stable/rust-by-example/fn/closures/input_parameters.html) of
 [Rust by Example](https://doc.rust-lang.org/stable/rust-by-example/).
+
+The tests for this exercise will cause compile time errors,
+if your function signature does not fit them, even when they're not run.
+You may want to comment some tests out and generalize your solution piece by piece.
 
 
 ## Rust Installation
--- a/exercises/accumulate/example.rs
+++ b/exercises/accumulate/example.rs
@@ -1,9 +1,10 @@
-pub fn map<F>(mut values: Vec<i32>, f: F) -> Vec<i32>
+pub fn map<F, T, U>(values: Vec<T>, mut f: F) -> Vec<U>
 where
-    F: Fn(i32) -> i32,
+    F: FnMut(T) -> U,
 {
-    for val in &mut values {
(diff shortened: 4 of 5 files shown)


config.json
```diff
@@ -481,7 +481,7 @@
       "unlocked_by": "luhn",
       "difficulty": 4,
       "topics": [
-        "function_pointer"
+        "generics", "higher_order_functions", "move_semantics"
```
coriolinus: Because you've changed config.json, you'll need to rerun `configlet fmt` in order for Travis to succeed.

exercises/accumulate/src/lib.rs
```diff
@@ -1,4 +1,9 @@
+// Note: The tests for this exercise are not #[ignore]d,
+//       but commented out with a multi-line comment
+         /* this is a multi-line style comment */
```
coriolinus: I'm not a fan of commenting out the tests instead of using `#[ignore]`, and neither is Travis. Beyond aesthetic reasons, using comments instead of attributes defeats common idioms such as `cargo test -- --ignored`. This exercise already has [`.meta/ALLOWED_TO_NOT_COMPILE`](https://github.com/exercism/rust/blob/master/exercises/accumulate/.meta/ALLOWED_TO_NOT_COMPILE) set, which should prevent Travis from complaining about the compile failures.

exercises/accumulate/src/lib.rs
```diff
@@ -1,4 +1,9 @@
...
+//       To activate a test, move the comment starter "/*" below it.
+
 /// What should the type of _function be?
-pub fn map(input: Vec<i32>, _function: ???) -> Vec<i32> {
+pub fn map(input: Vec<i32>, function: ???) -> Vec<i32> {
```
coriolinus: Eliminating the leading underscore from `_function` will cause the compiler to complain about an unused variable. Despite the fact that this exercise is allowed to not compile, I'd prefer to minimize the number of compiler errors the student gets when they first download the exercise.

exercises/accumulate/src/lib.rs
Emerentius: Ignored tests are still type checked. If I didn't comment them out, a student would be confronted with compile errors until they already had the final, fully general function signature.

coriolinus: Yes, that's true. It's also more or less unavoidable when the point of the exercise is to figure out how to write a generalized generic function signature. 

Reading the compiler errors and fixing them is an essential skill for anyone studying a language with a compiler. We try to arrange things to minimize the compiler errors a student is initially presented with, but IMO it's a good thing that there are a few exercises for which they just have to figure out what's going on based on what the compiler tells them.

Emerentius: Right, but need it be >100 lines of compiler errors at once? The compiler doesn't stop at the first function. You can't even write a minimal map implementation for i32 -> i32 before having written `fn map<T, U, F: FnMut(T) -> U>(input: Vec<T>, mut function: F)`  
It's a bit overwhelming on the command line.

If we don't comment the tests out, the student needs to have that idea on their own or advancing can become very frustrating.
It's better if you use an editor that shows you the compile error sources directly in the source, but even that is probably very tedious if you aren't already familiar with generics and the closure traits. I may be wrong there, I'm too far removed from the newbie experience by now.

Side note on in-editor errors: I'm not sure why, but this exercise drives RLS crazy. It's permanently pegged at 100% in VS Code.

I just try-solved it again with all tests active from the start and the compiler told me at some point that it needed Debug bounds. That was a misdiagnosis, the correct solution makes them unnecessary.  I'll add a final test with a struct that has only the bare minimum of traits.

coriolinus: > need it be >100 lines of compiler errors at once?

Compilers do that sometimes. It may seem overwhelming at first, but with practice, one learns to just take them one at a time. Getting that comfort level requires exposure to the phenomenon. What's worse: hitting that for the first time here, in a mentored environment, or hitting it at some point in the future in their own code?

I wouldn't object to some text in the hints file that if the student finds the number of errors overwhelming, they can try temporarily commenting out later tests to reduce the scope of what needs to be implemented. I just don't think we should block-comment large sections ourselves.

exercises/accumulate/src/lib.rs
Emerentius: I'll add it back, although it barely changes anything what with all the other compiler errors.

exercises/accumulate/src/lib.rs
Emerentius: > It may seem overwhelming at first, but with practice, one learns to just take them one at a time. Getting that comfort level requires exposure to the phenomenon.

That's true to some extent, but I'm not sure how valuable it really is as a teaching moment as opposed to just annoying. Students are also often expected on exercism to reverse engineer the interface and behaviour from the tests. I personally am mostly annoyed with that. It's not difficult, but tedious.

I can add a note to the hint and `#[ignore]` the tests again.

If I'm reading the error right, travis enforces that only 1 test is not ignored, is that right? For this exercise that's barely any different to having them activated by default. As soon as the compile errors are gone and the first two tests pass, all the others probably do, too.

coriolinus: I actually completely agree that it's annoying to have to reverse-engineer the interface and behavior from the tests. We finally got to the point where we can programmatically verify that all exercises have a non-empty `src/lib.rs` file [literally yesterday](https://github.com/exercism/rust/pull/616); it just took a long time to cover all the existing exercises. 

Still, I think we're in pretty good shape as far as that goes. As of right now, there are 88 exercises (`fd -td -d1 . exercises | wc -l`), of which only 7 don't compile when the student first downloads them (`fd -H ALLOWED_TO_NOT_COMPILE exercises | wc -l`). 

You're right that travis verifies that exactly 1 test is not ignored. By convention, the non-ignored test should be the first. Even if it barely makes a difference, it's better to avoid using magic travis-appeasing files when it is possible to just do the right thing.

Emerentius:
Incorporated the feedback and kept it separate for easier review. I'll squash them when the time to merge comes.

PR #648, opened 2018-08-31
dot-dsl: Add get_node and get_attr methods to required signatures

### LINKED ISSUE #640: dot-dsl: testcases pass for trivial Node/Edge impls
Testcases that test if the builder correctly adds attributes to nodes/edges use the same builder to create the expected cases.

Example:
```rust
    let nodes = vec![Node::new("a").with_attrs(&[("color", "green")])];
    let graph = Graph::new().with_nodes(&nodes);

    assert_eq!(
        graph.nodes,
        vec![Node::new("a").with_attrs(&[("color", "green")])]
    );
```

This allows students to "cheat" with trivial implementations of `Node` and `Edge`. Something like this passes all tests:

```rust
#[derive(Debug, Clone, Eq, PartialEq)]
pub struct Node;

impl Node{
    pub fn new(from: &str) -> Self {
        Node
    }

    pub fn with_attrs(mut self, attrs: &[(&str, &str)]) -> Self {
        self
    }
}
```

I don't think this is intended, is it?

coriolinus:
Closes #640. Doesn't test edges or edge attributes, but hopefully
once students have implemented enough to pass this test, they'll
implement edges properly also.

### DIFF
--- a/exercises/dot-dsl/example.rs
+++ b/exercises/dot-dsl/example.rs
@@ -38,6 +38,10 @@ pub mod graph {
 
             self
         }
+
+        pub fn get_node(&self, name: &str) -> Option<&Node> {
+            self.nodes.iter().filter(|n| n.name == name).nth(0)
+        }
     }
 
     pub mod graph_items {
@@ -75,7 +79,7 @@ pub mod graph {
 
             #[derive(Clone, PartialEq, Debug)]
             pub struct Node {
-                name: String,
+                pub name: String,
 
                 attrs: HashMap<String, String>,
             }
@@ -94,6 +98,10 @@ pub mod graph {
                     });
 
                     self
+                }
+
+                pub fn get_attr(&self, name: &str) -> Option<&str> {
+                    self.attrs.get(name).map(|v| v.as_ref())
                 }
             }
 
--- a/exercises/dot-dsl/tests/dot-dsl.rs
+++ b/exercises/dot-dsl/tests/dot-dsl.rs
@@ -124,3 +124,22 @@ fn test_graph_with_attributes() {
 
     assert_eq!(graph.attrs, expected_attrs);
 }
+
+#[test]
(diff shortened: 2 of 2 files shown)


exercises/dot-dsl/example.rs
```diff
@@ -75,7 +79,7 @@ pub mod graph {
 
             #[derive(Clone, PartialEq, Debug)]
             pub struct Node {
-                name: String,
+                pub name: String,
```
petertseng: note that this `pub` is needed the `get_node` needs it.

exercises/dot-dsl/tests/dot-dsl.rs
```diff
@@ -2,9 +2,9 @@
 extern crate maplit;
 extern crate dot_dsl;
 
+use dot_dsl::graph::Graph;
```
petertseng: I don't think it was necessary to move it as a part of this PR.  I request some explanation be added before merging.

coriolinus: You're right, sorry, this wasn't a necessary change. It wasn't intentional: my dev environment is set up to automatically order use statements alphabetically. 

It being unintentional does not mean that it is not potentially useful. I have a preference at strength 2/10 to keep imports ordered in this way, because in larger projects this means that it's easier to scan the list of imports in a predictable way. 

Does this explanation suffice, or would you prefer that I undo this change before merging?

petertseng: Well, if it's not too much trouble, putting it in a separate commit with something similar (easier to scan if it's in a predictable order) will cause me to have no objections. My preference is only 1/10 even if I did object (if no change is made before merging)

coriolinus:
I have broken out the moved import into a new commit. I'll merge once Travis passes.

PR #649, opened 2018-09-01
macros: test macro without HashMap in scope

### LINKED ISSUE #600: “Macros” exercise: chance of a non-compiling code
Consider a solution from a student:

```rust
#[macro_export]
macro_rules! hashmap {
    () => {
        HashMap::new()
    };

    ($($key:expr => $value:expr),+) => {
        {
            let mut hm = HashMap::new();

            $( hm.insert($key, $value); )*

            hm
        }
    };

    ($($key:expr => $value:expr),+,) => {
        hashmap!($($key => $value),*)
    };
}
```

While it passes the tests, the code will fail to compile if `HashMap` is not imported to a scope first. ([Playground](https://play.rust-lang.org/?gist=4a8a4266752b95b8a2a3f48917953656&version=stable&mode=debug&edition=2015))

I think we need to tweak tests to teach students about such case.

Would you like me to fill a PR?

Emerentius:
fixes #600

I put a test fn in a new module that doesn't import HashMap, so that if a student uses `HashMap::new()` in the macro, the compiler will throw one and only one compile error.

### DIFF
--- a/exercises/macros/tests/macros.rs
+++ b/exercises/macros/tests/macros.rs
@@ -60,3 +60,11 @@ fn test_nested() {
         expected
     );
 }
+
+mod test {
+    #[test]
+    #[ignore]
+    fn type_not_in_scope() {
+        let expected: ::std::collections::HashMap<(), ()> = hashmap!();
+    }
+}


Emerentius:
Beta build timed out

PR #651, opened 2018-09-01
inappropriate usize

### LINKED ISSUE #601: Inappropriate use of usize
While going through the exercises, it occured to me that a couple of them use `usize` inappropriately. A good example is `difference_of_sums`.

```rust
pub fn difference(n: usize) -> usize {
    square_of_sum(n) - sum_of_squares(n)
}
```
`usize` is the pointer-sized integer type and should not be used for general numeric computation. Here, a fixed size of u64 is probably a good default bet. We should educate its use only when actually indexing into memory, like vec and slice (`series` is a good example here).

I have found at least one other such exercise, raindrops.

Other exercises take a different strategy, for example `nth prime` (passes u32).

coriolinus:
Closes #601.

My intent is to create one commit per exercise. False positives (such as all-your-base, binary-search, etc.) are simply checked off in that issue once they have been inspected. True positives are checked off once a commit exists in this PR.

### DIFF
--- a/exercises/book-store/example.rs
+++ b/exercises/book-store/example.rs
@@ -5,9 +5,9 @@ use std::collections::{BTreeSet, HashSet};
 use std::hash::{Hash, Hasher};
 use std::mem;
 
-type Book = usize;
+type Book = u32;
 type GroupedBasket = Vec<Group>;
-type Price = usize;
+type Price = u32;
 const BOOK_PRICE: Price = 800;
 
 #[derive(Debug, Clone, PartialEq, Eq)]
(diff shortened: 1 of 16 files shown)


coriolinus:
I assert that I have reviewed all exercises listed [here](https://github.com/exercism/rust/issues/601#issuecomment-417819745) and have determined that all remaining uses of `usize` and `isize` are appropriate and idiomatic.

coriolinus:
Would you like me to wait a few days before merging to give time to check
for completeness, or are you satisfied with this PR as-is?

petertseng:
Since the cost of being incomplete is so low (we'll just make another PR!) I say it makes sense not to wait.

PR #653, opened 2018-09-04
Add exercise for doubly linked lists with unsafe

Emerentius:
*This PR isn't finished, but I'd like to get some feedback already. Would this be a good addition at all? Are any major design changes necessary?*

This adds a new exercise that is specifically about the nitty gritty details of unsafe. There is a `linked-list` exercise in `problem-specifications`, but this is not supposed to be it (I will have to rename it).  

It's expanded in two ways:
1. The exercise requires the ability to insert or remove elements from any point in `O(1)`
2. implemented with `unsafe {}`

Doubly linked lists are infamous for tripping up newcomers to Rust who choose it as their first trial project and are hit with the full force of the borrow checker before they had a chance to learn about it.

Because they are an owning collection, students will have to learn about how to manually manage memory via `Box::into_raw` and `Box::from_raw` and expose a safe interface. Because it's an owning collection, it also touches on `Send`, `Sync` and covariance over the contained items (these should probably behind a feature flag). There's also some interaction with drop check, but I am yet to understand exactly what the rules are.

I've tried to keep the exercise simple, yet not too far removed from actual use cases by including the cursor. Without it, you may as well use `VecDeque`. There aren't actually that many cases where `unsafe` needs to be explicitly written, but a lot of potential for use-after-free, double-free, accidental leaks, etc, if you don't manage the pointers correctly.  
Overall, it came out a little long. I'd love for the required code to be reduced to <200 lines.


config.json
```diff
@@ -1050,6 +1050,18 @@
       "topics": null,
       "deprecated": true
     },
+    {
+      "uuid": "34738b66-28bb-41bc-85c6-68fcf4a09bc2",
+      "slug": "linked-list",
+      "core": false,
```
coriolinus: One of the following must be true: `core` is not false, or `unlocked_by` is not null. Travis won't pass otherwise. My instinct is that this should be a core exercise toward the end of the track: everyone could benefit from some `unsafe` practice.

exercises/linked-list/.meta/description.md
```diff
@@ -0,0 +1,20 @@
...
+The doubly linked list is a fundamental data structure in computer science,
+often used in the implementation of other data structures. They're
+pervasive in functional programming languages, such as Clojure, Erlang,
+or Haskell, but far less common in imperative languages such as Ruby or
+Python.
+
+Each node in a doubly linked list contains data and pointers to the next
+and previous node, if they exist.
+
+New nodes can be efficiently added at any point in the list, if one already has
```
coriolinus: It may be worth mentioning practical applications, such as the [dancing links algorithm](https://en.wikipedia.org/wiki/Dancing_Links).

exercises/linked-list/Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[package]
+name = "linked-list"
+version = "0.1.0"
+authors = ["Emerentius <emerentius@arcor.de>"]
```
coriolinus: Strip out the authors line, please; that's a track policy we came to some time ago. I can track down the discussion if you like.

exercises/linked-list/src/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+impl<T> Drop for LinkedList<T> {
+    fn drop(&mut self) {
+        unimplemented!()
+    }
+}
+
+// the cursor is expected to act as if it is at the position of a node
+impl<'a, T: 'a> Cursor<'a, T> {
+    /// Take a mutable reference to the current element
+    pub fn peek_mut(&mut self) -> Option<&mut T> {
```
coriolinus: The existence of `peek_mut` implies `peek`, which is however not included in this API. I think that this is to save implementation time for the student; the mutable case is the more interesting case. I approve of this decision.

exercises/linked-list/src/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+        unimplemented!()
+    }
+
+    /// Move one position backward and return a reference to the new position
+    pub fn prev(&mut self) -> Option<&mut T> {
+        unimplemented!()
+    }
+
+    /// Remove the element at the current node from the list and return it.
+    pub fn take(&mut self) -> Option<T> {
```
coriolinus: What is the position of the cursor after `take`? That is, if I have the list `[A B C]`, and the cursor is at `B`, after `take`, where should the cursor be? This should be clearly specified in the comment.

coriolinus: Policy options:

- cursor always moves forwards — what about when we `take` the last element in the list?
- cursor always moves backwards — what about when we `take` the first element in the list?
- cursor moves forwards, except when it is at the last element, in which case it moves backwards — what about when we `take` the final element in the list, leaving no elements remaining? 
- `take` consumes `self`, meaning we don't have to worry about the position of the cursor because the cursor dies when `take` is called

Of these options, I think I prefer the last.

If any of the first three options is chosen, there should be a test which exercises the complicated case.

exercises/linked-list/tests/linked-list.rs
```diff
@@ -0,0 +1,242 @@
+extern crate linked_list;
+use linked_list::*;
+
+// Add your tests here
```
coriolinus: remove this line

exercises/linked-list/tests/linked-list.rs
```diff
@@ -0,0 +1,242 @@
...
+    }
+}
+
+#[test]
+fn is_generic() {
+    struct Foo;
+    LinkedList::<Foo>::new();
+}
+
+fn clone_list<T: Clone>(list: &LinkedList<T>) -> LinkedList<T> {
```
coriolinus: Instead of implementing this function, what do you think of requiring students to implement [`FromIterator`](https://doc.rust-lang.org/std/iter/trait.FromIterator.html)? Then calls to `clone_list` could be replaced with `dest: LinkedList<_> = source.iter().cloned().collect();`.

exercises/linked-list/tests/linked-list.rs
```diff
@@ -0,0 +1,242 @@
...
+    let mut new_list = LinkedList::new();
+    for element in list.iter().cloned() {
+        new_list.push_back(element);
+    }
+    new_list
+}
+
+// implement some additional functionality using the required interface
+// to limit the amount of code the student has to oversee
+trait ListExt<T> {
```
coriolinus: I very much like the idea of using this extension trait to reduce the students' workload. However, we should probably standardize on a single nomenclature pair for the list's ends: either front/back or head/tail. 

I personally conceptualize lists with the head at the front and the tail at the back. It is clear from these implementations that you conceptualize them with the head at the back and the tail at the front. Both conceptualizations are valid, but both will be confusing to people who use the other conceptualization. 

I have a mild (~2/10) preference for using the front/back nomenclature throughout instead of head/tail.

exercises/linked-list/src/lib.rs
Emerentius: The reason I left it out is that there's no use case for `peek`. The cursor is locked in place anyway and all other methods require `&mut`.

exercises/linked-list/src/lib.rs
```diff
@@ -0,0 +1,74 @@
+pub struct LinkedList<T>(std::marker::PhantomData<T>);
+
+pub struct Cursor<'a, T: 'a>(std::marker::PhantomData<&'a mut T>);
+
+pub struct Iter<'a, T: 'a>(std::marker::PhantomData<&'a T>);
```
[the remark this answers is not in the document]
Emerentius: An iter returns different lifetimes. A cursor is locked in place while peeking, the iterator can return references to all elements in the list. You could probably add `Iterator` for `Cursor`, but it's mixing concerns a bit.

exercises/linked-list/src/lib.rs
Emerentius: It's number 3. I was thinking in terms of indices like with a vec. The cursor stays in place and everything after is backshifted. If it's at the end, it moves to stay at the back. The cursor has an optional pointer precisely for the case of an empty list.  
There are valid use cases for `take and move left` or `take and move right` to minimize pointer chasing, but only one is required for generality. The other one can be gotten by stepping right after.  
My cursor here is conceptually different from the one in the `linked-list` crate. That one is located between elements and can be both before and after the list. This one is always conceptually on a node, which is why it moves backwards to follow the list when it's already at the end.

The last option precludes removal of several nodes. Traversing a list is costly, so that is not possible for a cursor.

exercises/linked-list/tests/linked-list.rs
Emerentius: You're totally right. I took over the head/tail nomenclature from `std::collections::{LinkedList, VecDeque}` and at times I was somewhat confused which was where. There's no benefit to keeping two names around. I want to keep the `{push,pop}_{back, front}` interfaces consistent with std so I'll start using those everywhere.

exercises/linked-list/tests/linked-list.rs
Emerentius: It would some more code to an already large exercise, but I'm not opposed. It's not hard to implement and it would make building arbitrary lists in the tests very clean.

exercises/linked-list/src/lib.rs
coriolinus: Oh--I think I'd misread the way `Cursor` operated; I thought it mutably traversed the list.

If `Cursor` was changed to mutably traverse the list, and if `peek` were added, then it would be very simple to `impl Iterator for Cursor`. These changes may be worth making in order to reduce the total amount of implementation work required of the student; they wouldn't have to implement both `Cursor` and `Iter`. What do you think of this?

exercises/linked-list/src/lib.rs
coriolinus: That makes sense, but if a cursor can shift in response to `take`, then that argues in favor of a single cursor which mutably shifts over the list instead of an immutable cursor per node. If we actually wanted the latter, then there's a fifth option which makes more sense than any of these: the cursor retains its pointer to the current element, which becomes the first entry in a new list.

exercises/linked-list/tests/linked-list.rs
coriolinus: I like it because it's a useful trait that I don't think we require of students elsewhere.

exercises/linked-list/src/lib.rs
coriolinus: I disagree. Let's work that example.

```rust
let first: &str = <Cursor as Iterator>::next(&mut cursor);
```

I'm pretty sure that `first` has to be `Option<&i32>` in this case. That's minor, though.

```rust
let first = <Cursor as Iterator>::next(&mut cursor);
```

`first` must equal `Some(&0)`, by the definition of the `Iterator` trait. It shouldn't borrow anything directly from `cursor`.

```rust
cursor.prev()
```

This should return `None`; the cursor is already on the first element. Let's pretend it said `cursor.next()` here, though: that advances the cursor and returns `Some(&mut 1)`, which reference is immediately destroyed.

```rust
cursor.take();
```

This returns `1` (immediately dropped) and removes its node. The `0` node should point forward to `2`, and the `2` node should point back at `0`. Nothing is dangling.

Emerentius: > the cursor is already on the first element

This can't work. The iterator must progress on `next`, so it's now on the 2nd element.

coriolinus: The `next` methods have different semantics. `Iterator::next`, called for the first time, returns the first element. `LinkedList::next`, called for the first time, advances the cursor and then returns the 2nd element.

Emerentius: But both have to advance. What good is a Cursor iterator if it stays in place?

coriolinus: I see. You're right. 

The upshot is that the borrow checker cannot allow `cursor.prev()` while `first` exists: there may not be a mutable reference and an immutable reference to the same data simultaneously. If the borrow checker forbids this code, then we still have a safe interface. 

An alternate solution would have the cursor's `prev` and `next` methods return `&T`, not `&mut T`. This should be fine, as we can still get a mutable reference by calling `peek_mut` if we need it. In that case the borrow checker should allow this code, and we still have a safe interface.

Emerentius: There really is no way to do this without hacking weird state handling into the cursor. A structure that can mutate any node at any time and an iterator over references are incompatible interfaces.

coriolinus: I think you're right. Our options are fundamentally either to restrict `Cursor` to immutability, or implement `Iter` as a separate struct.

exercises/linked-list/tests/linked-list.rs
Emerentius: An alternative to an extension trait is to put these methods and their implementation directly into `impl<T> LinkedList<T>` in `lib.rs`. That way the student can also use them for `FromIterator` or `Extend`

coriolinus: That's not a bad idea, though it might be surprising for a student to find some methods pre-implemented for them. If you do this, please add a comment explaining that these methods were trivial to write, but useful, so we implemented them for free to give a leg up.

config.json
Emerentius: I still fear we'd overload the mentors with such an exercise. It's already long and unsafe should get extra care. There's smaller stuff to introduce some unsafe, like, say, implementing a version of `std::vec::Drain`.
I've set it to be unlocked by forth for now, but that's just for a travis test run.

coriolinus:
You've caught me at an inopportune moment: I'm about to head on vacation, so can't promise to review this within the next two weeks or so. 

I will comment that while this exercise requires a fair amount of implementation, at least per the examples, it's not the most verbose exercise in the track:

```sh
$ fd example.rs --exec wc -l {} \; | sort -nr | head -n3
Thu Sep 13 09:46:04 DST 2018
342 exercises/poker/example.rs
245 exercises/react/example.rs
187 exercises/decimal/example.rs
```

Therefore, I wouldn't worry too much about the length.

exercises/linked-list/src/pre_implemented.rs
```diff
@@ -0,0 +1,66 @@
+use {LinkedList, Cursor};
+
+// ============================================================================
```
coriolinus: If you use `//!` style, this comment will be applied to the entire module when students run rustdoc.

Emerentius:
No rush. I still have  to figure out what the deal with dropck is and overhaul the tests.

coriolinus:
Looks like filtering the repo on student download is a blacklist system, not a whitelist. What this means is that `pre_implemented.rs` will properly be copied to student machines.

ignore_pattern

coriolinus:
The reason Travis is failing is that the README is other than it expects. Try regenerating that?

coriolinus:
Of interest to the cursor implementation: https://github.com/4e554c4c/rfcs/blob/cursors/text/0000-linked-list-cursors.md

Are you still working on this, @Emerentius? I still think this will be a cool addition to the track, but I haven't seen much movement recently.

Emerentius:
There's one safety related issue that I have been blocked on because I don't yet understand it: the interaction with dropck.
In several places it is mentioned that if you have an owning raw pointer, it's important to use `PhantomData` to show this ownership to dropck.

I haven't yet been able to understand what problems can potentially arise from not doing so and whether the danger is always present or whether you have to opt-in to it somehow.
If the unsafety always exists, then it must be part of the exercise and we should somehow test for it. It would be a great example of the subtleties that you encounter with unsafe. If it's opt-in with an obvious potential for unsafety (e.g. from the [dropck eyepatch](https://github.com/rust-lang/rfcs/blob/master/text/1327-dropck-param-eyepatch.md)), then we could delegate that to later and make it part of the advanced tests like Send/Sync and the lifetime subtyping.

The nomicon doesn't have an example demonstrating the unsafety. [This issue](https://github.com/rust-lang-nursery/nomicon/issues/17) says that the chapter is outdated anyway and I don't know where the authoritative definition is located right now.

coriolinus:
My understanding is that the dropck concern is not relevant to this exercise, because we don't have any structs which own raw pointers. We likewise don't require or request that students implement structs which do.

I'd love to get someone experienced with unsafe code to give a firm answer, but I don't have contact with many people who are experienced with unsafe code. @petertseng do you have thoughts on this?

In conclusion: I'm about 80% confident that there is no dropck concern for this exercise, and therefore you should consider yourself unblocked.

petertseng:
I will caution that I actually don't fit the description of "experienced with `unsafe` code", but when the time is right I would attempt to solve this exercise from the POV of a student and report what I see

Emerentius:
I can't reproduce the travis failure locally

Emerentius:
@petertseng I've changed everything I wanted to. If anything major is amiss, I'm lacking the fresh perspective. If you could try your hand at it, that would be cool.

petertseng:
I wanted to but I found myself unprepared to do `unsafe`. So it kept getting pushed back to a day when I can learn that.

coriolinus:
After _far_ too long, I've made the time to perform a student implementation of this exercise. You can find the sum of my efforts [here](https://github.com/coriolinus/exercism_rust/compare/f88233c...coriolinus:double-linked-list-student-implementation).

https://github.com/coriolinus/exercism_rust/commit/668e14690e4510484701ff7cc669b326ac3e1dc7 fixes some deprecation warnings; rustfmt made some additional changes which I assume are just style updates. This commit should probably be merged into this branch. 

I took notes as I went through and implemented this; the full file is [here](https://github.com/coriolinus/exercism_rust/blob/24d36c66522b42ca89e6fd98730ec04563036e5b/exercises/doubly-linked-list/student%20notes.md), but don't take any of the early stuff to heart: I was more or less just logging my thoughts as I went, and I ended up fixing a lot of the stuff the early notes complain about. The highlights/final form are as follows:

- it took a very long time to go from starting the exercise to get even the first test passing; I think it ended up being over two hours, even though I'd (a few months ago) done the necessary research to know that I wanted to orient my code around the `NNMut<T>` type (an alias for `Option<NonNull<*mut Node<T>>>`). 
- however, the more tests passed, the less time it took to implement enough that subsequent tests passed
- I was overall very happy with the test suite; it took me through a very reasonable progression of complexity.
- The biggest issue I had, initially, was figuring out what `Drop` was supposed to do; the code wouldn't compile without it, but particularly at first, I had no idea what I wanted it to do. In the end, I just lifted the implementation from [the standard library](https://doc.rust-lang.org/src/alloc/collections/linked_list.rs.html#806-810). 

    This led to implementing several methods earlier than I otherwise would have needed to, which contributed to the sense that the first steps were the hardest for this exercise. 

    I have not yet tested it, but I'd consider removing the `Drop` implementation from the stub file entirely. This will cause student implementations to leak memory on each test run for the duration of the test runner, which is acceptable; as soon as the tester exits, all that memory will be freed by the OS anyway. I think that this is acceptable, as there exist several leak tests, and Step 4 explicitly instructs students to implement `Drop`. 
- Though the README instructs us to minimize duplication, my efforts in that direction didn't pan out. They're encapsulated in https://github.com/coriolinus/exercism_rust/commit/a45f346d4d317b10ae537af69bc04308d4f30b93, which I immediately reverted because it caused several tests to fail. For a savings of two lines of code, I didn't think it was worth the debug hassle. 
- The bottom line: I felt like I learned a lot about unsafe code from implementing this exercise from the student perspective; I have a better understanding of what `unsafe` does and how to use it properly. That was the goal of the exercise, and I believe that it was successful. 

I apologize that it took me so long to get to this, but this really was the first opportunity I've had. I believe that these are the only items waiting before merge:

- cherry-pick https://github.com/coriolinus/exercism_rust/commit/668e14690e4510484701ff7cc669b326ac3e1dc7  or otherwise apply equivalent changes
- potentially remove the `Drop` trait impl from `src/lib.rs`, if you agree that it is a good idea

coriolinus:
If there are no comments before then, I intend to do the following on Wednesday the 7th:

- cherry-pick https://github.com/coriolinus/exercism_rust/commit/668e14690e4510484701ff7cc669b326ac3e1dc7 into this PR
- remove the `Drop` trait impl from `src/lib.rs`
- fix any Travis errors produced
- merge this PR

While this has become a low-velocity issue, it is in nobody's interest that it _remain_ one.

Emerentius:
`unimplemented!()` in the Drop impl was a mistake, yep. I'd leave the Drop impl but replace the `unimplemented!()` with a comment, but removing it entirely is also fine.

I also noticed, that the phase 1 tests start off with `push_back()` and `pop_front()`. In the interest of a gradual progression, I'd like to split that one and change the order to:
* empty (keep as is)
* back
  - push / pop single element (new from split)
  - push / pop multiple elements
* front
  - push / pop single element (new from split)
  - push / pop multiple elements
* front + back tests (keep as is)

Emerentius:
More importantly, reading through your solution reminded me that making certain that you're getting unsafe right is *damn hard*. The reasoning necessary to make sure you haven't made any mistakes should be as simple as possible and therefore also as local as possible. All invariants should be clearly documented and every unsafe block should have a comment explaining why it is safe.

My own example is missing that so I'll add it.

coriolinus:
I like that plan for easing the progression in phase 1. I agree that it'll make things easier for the students.

IMO it makes more sense to remove the Drop trait entirely than to have a blank implementation. We explicitly tell them to implement `Drop`; students at this level should be comfortable looking up a trait definition and doing the right thing to implement it. 

Given that you're still active on this @Emerentius, I won't rush to merge this before you tell me you're done. The Wednesday plan was a contingency in case you couldn't be reached. I don't mean to suddenly impose a tight deadline on you!

Emerentius:
I've done all the changes, rebased and regenerated the README.

Adding the invariant documentation helped me clean up the example a lot. There were several functions that were unsafe but not marked as such.

In your solution, I noticed that you implemented `Send` and `Sync` without `T: Send` | `T: Sync` bounds. That is unsound and I've added tests against them.
Tip: Getting a reference from a raw pointer is a common operation and you can write `unsafe { &*ptr }` instead of `Box::leak( unsafe { Box::from_raw(ptr) }`.

coriolinus:
Hah, good; those extra tests would have helped me! That's a good tip, as well.

I think this looks great. I'm still going to wait until Wednesday to merge, just to ensure that we've left enough time for other maintainer comments, but at this point, my intent is to merge on Wednesday.

PR #654, opened 2018-09-05
fetch-configlet: rewrite in Rust

coriolinus:
This transforms the file from an 32-line bash script into 165 lines
of Rust and TOML. Compilation takes several minutes, and the output
executable is several megabytes large.

However, there are advantages as well, such as Windows support. This
has been demonstrated to work properly under WSL, and it should also
work properly under msys2 assuming one has the dependencies (make and
gcc) installed.

Additionally, this makes it possible for other Rust tools to hook into
the library and fetch configlet without needing to drop down to the
command line.

Other features not present in the shell script:

- optional debug output
- automatically locates the bin folder, allowing it to be run from
  anywhere inside the xrust repo
- doesn't depend on presence of git, curl, tar, or any other existing tools:
  makes setting up msys2 relatively easy
- nicer error messages


coriolinus:
Having thought this over, this is a pretty serious case of using the wrong tool for the job. The bash script was faster and better in almost every circumstance. Retracting the PR.

PR #656, opened 2018-09-09
fix misspellings

AndrewKvalheim:
Noticed another typo while doing the exercises so I ran a spell check.


AndrewKvalheim:
Whoops, thanks. The only test I ran was `check-exercises.sh`.

coriolinus:
Thanks @AndrewKvalheim for doing this work, and particularly for following through and writing the upstream PR as well!

PR #657, opened 2018-09-15
Add test for pig latin rule 4

ccouzens:
There was no test for

```
**Rule 4**: If a word contains a "y" after a consonant cluster or as the
second letter if a two letter word it makes a vowel sound (e.g. "rhythm"
-> "ythmrhay", "my" -> "ymay").
```

### DIFF
--- a/exercises/pig-latin/tests/pig-latin.rs
+++ b/exercises/pig-latin/tests/pig-latin.rs
@@ -115,6 +115,12 @@ fn test_word_beginning_with_xr() {
 
 #[test]
 #[ignore]
+fn test_y_is_treated_like_a_vowel_at_the_end_of_a_consonant_cluster() {
+    assert_eq!(pl::translate("rhythm"), "ythmrhay");
+}
+
+#[test]
+#[ignore]
 fn test_a_whole_phrase() {
     assert_eq!(pl::translate("quick fast run"), "ickquay astfay unray");
 }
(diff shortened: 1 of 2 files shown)


ccouzens:
Also, it's in the canonical test cases https://github.com/exercism/problem-specifications/blob/master/exercises/pig-latin/canonical-data.json#L180-L187

exercises/pig-latin/tests/pig-latin.rs
```diff
@@ -113,6 +113,12 @@ fn test_word_beginning_with_xr() {
     assert_eq!(pl::translate("xray"), "xrayay");
 }
 
+#[test]
+#[ignore]
+fn test_word_word_beginning_with_y_and_a_preceding_consonant() {
```
ZapAnton: A small nitpick: perhaps you meant to name the test function `test_word_beginning_with_y_and_a_preceding_consonant`, without an extra `_word_`?

Also I believe the test naming policy on this track is that you take a description from the `canonical-data.json` and replace all spaces ` ` with the `_`, so the name would be `test_y_is_treated_like_a_vowel_at_the_end_of_a_ consonant_cluster` (Note that it is just my comment and you do not have to rename your function unless the maintainer says so)

ccouzens: @ZapAnton both good points. I'll rename it when I'm next at my computer.

PR #660, opened 2018-09-18
Rewrite bin/init_exercise.py in Rust

### LINKED ISSUE #641: Replace init_exercise.py
The exercise init script isn't great. It gets the job done for creating an exercise from scratch, but that's not actually the most common desired use case. The most common desired use case is to update an exercise's tests based on updates to the problem specifications.

Also, the thing is in python, and this is the Rust track. We have a legitimate case to RIIR here.

## Required Features

- written in Rust (stable preferred, but not required--people hacking on this repo can install nightly if necessary)
- can just update `config.json`
- can just create an exercise without touching `config.json`
- can just update an exercise without touching `config.json`
- when creating or updating an exercise: if `problem-specifications` repo is available locally, updates that repo against the current head of master. If it is not available locally, clones it. Either way, uses the current master branch of that repo to get problem specifications. 
- when updating an exercise: keeps a copy of existing `test_*_property` implementations instead of regenerating. This may be done on a purely textual basis. If these implementations do not exist, they are generated as per the "new exercise" case to ease test upgrades.
- when updating an exercise: keeps a copy of existing tests. Tests from the canonical data may be overwritten, but tests not overwritten should be retained in existing form. This helps ensure that optional tests and track-specific tests are not destroyed when updating.

ZapAnton:
The purpose of this PR is to replace the functionality of the existing `bin/init_exersice.py` script with the Rust crate `exercise` and to add additional functionality as per #641.

It is opened as `In Progress` in order for maintainers to track any undesirable decisions/functionality in the new crate, since it is a medium-complexity project, so disagreements may arise.

In addition to the future suggestions the current work plan looks like this:

### Major Tasks
- Reach an agreement on the new crate's API/commands
- Implement `configure` subcommand
- Implement `update` subcommand
- Update the docs to mention the usage of the new crate

### Minor Tasks
- Add any missing comments/fields from the `canonical-data.json` to the generated test suite
- Improve the crate's functions documentation
- Prettify the generated test suite when the `maplit` crate is used. Currently it is an unreadable mess and `rustfmt` does not help
- General refactoring. I am not the most prominent rustcean out there, so my solutions for the problem may not be the most optimal

Note that as per https://github.com/exercism/rust/issues/641#issuecomment-418722003 the crate name is changed from `init_exercise` to `exercise`, so the prefix in the commit messages is changed from `init_exercise:` to `exercise:` accordingly

If this PR is merged, closes #641


ZapAnton:
Questions regarding the `update` subcommand:

As I understand it, the desirable algorithm is to get the latest `canonical-data.json` for the exercise and to apply any diffs to the existing test suite:

- If a new `case`/`property` is added to the `canonical-data`, add the appropriate functions to the existing test suite - no problems here
- If the existing `case` is updated, add a new function to the test suite and do not touch the existing function. Question: How to track the updates to the `cases`?  We cannot use the existing test suite, as the person who implements the exercise is not bound to follow the format of the generated test suite and may use the format of their own. Should we track the changes via git?
- If the existing `case` is deleted from the `canonical-data.json`, what should be done with the existing test functions?

coriolinus:
I think this tool should never delete a test case. It can't easily know if
a case not represented in the canonical data was deleted from the canonical
data, or was a custom case implemented for the Rust track which has never
appeared in the canonical data.

For updating existing test cases, I'd favor a very simple--even
stupid--strategy:

1. capture the bytes of the existing test case in a buffer: the literal
text of the test
2. compare them to the bytes of the generated test case
3. if the generated test case differs from the existing test case, update
the name of the generated case with some suffix and then emit them both.

Users of this script will then have the implicit option: keep all the
cases, and accept some repetition, or manually remove duplicate cases.
That's on them.

Both of these preferences are consistent with the general philosophy that
test suites should be write-only.

bin/build_exercise_crate.sh
```diff
@@ -0,0 +1,26 @@
...
+(
+	cd "$EXERCISE_CRATE_PATH"
+
+	echo "Building exercise crate"
+
+	cargo build --release
+
+	echo "Copying exercise crate from $EXERCISE_CRATE_PATH/target/release/exercise into $BIN_DIR_PATH"
+
+	cp "$EXERCISE_CRATE_PATH/target/release/exercise" "$BIN_DIR_PATH"
```
coriolinus: This will fail on a Windows machine if the user is running this through cygwin: the binary in that case would be `exercise.exe`.

exercise/src/utils.rs
```diff
@@ -11,3 +14,46 @@ pub fn get_track_root() -> String {
 
     track_root.trim().to_string()
 }
+
+pub fn run_configlate_command(command: &str, args: &[&str]) {
```
coriolinus: Typo: "configlet", not "configlate".

exercise/src/cmd/generate.rs
```diff
@@ -234,8 +228,12 @@ fn generate_tests_from_canonical_data(
         .join("tests")
         .join(format!("{}.rs", exercise_name));
 
-    let tests_content =
-        fs::read_to_string(&tests_path).expect("Failed to read existing tests content.");
+    let tests_content = utils::get_tests_content(exercise_name).unwrap_or_else(|_| {
```
coriolinus: `.unwrap_or_else(|_| panic!(x))` is exactly equivalent to `.expect(x)` for all `x`.

ZapAnton: It is actually panic!(x, y), where y is `exercise_name`:
```rust
let tests_content = utils::get_tests_content(exercise_name).unwrap_or_else(|_| {                                                                                                      
         panic!(                                                                                                                                                                           
             "Failed to get the content of the test suite for the '{}' exercise. Aborting.",                                                                                               
             exercise_name                                                                                                                                                                 
         )                                                                                                                                                                                 
}); 
```

coriolinus: Ah, I see. I still think `.expect(format!(...))` is a bit more concise, but it's no big deal either way.

ZapAnton: I  was actually using that approach, but then `clippy` complained about it, so I decided to be use the idiomatic approach instead:

```shell
238 |       let tests_content = utils::get_tests_content(exercise_name).expect(&format!(                                                                                                      
    |  _________________________________________________________________^                                                                                                                     
239 | |         "Failed to get the content of the test suite for the '{}' exercise. Aborting.",                                                                                               
240 | |         exercise_name                                                                                                                                                                 
241 | |     ));                                                                                                                                                                               
    | |______^ help: try this: `unwrap_or_else(|_| panic!("Failed to get the content of the test suite for the '{}' exercise. Aborting.", exercise_name))`                                    
    |                                                                                                                                                                                         
    = note: #[warn(expect_fun_call)] on by default                                                                                                                                            
    = help: for further information visit https://rust-lang-nursery.github.io/rust-clippy/v0.0.212/index.html#expect_fun_call
```

coriolinus: Huh. In that case, prefer clippy over my opinions.

ZapAnton:
Finished all the major tasks, so I think I will remove WIP from the title for the proper review. 

Sorry it took that long, my free time was very limited.

@coriolinus, if you have time, could you give it another look?

coriolinus:
Yes, thanks for all this work! It's a kind of huge PR, so it's going to
take a little while for me to get through it all.

First question, before review: does it have its own test suite?

Second question, before review: would you be willing to add a script to
Travis which ensures that if anything in the utility's path changes, the
thing still builds without error and (if it has tests) its tests all pass?

ZapAnton:
I was thinking about the test suite, but I was not sure, since the target of the utility (the exercises) will change over time (for instance, the missing exercise will be added so I can not use it for testing the `generate` command, since it will break the existing exercise). If you think that the test suite is necessary, I will add it.

And yes I am willing to add a check for Travis, although I am not sure I understand  what you mean by changing anything in the utility's path. Could you elaborate please?

coriolinus:
Note from usage: if an exercise already exists in config.json, `exercise
configure` should default to retaining the existing value.

re: test suite: before looking at the code, I can't say what tests might be
appropriate, if any. I don't want to give a blanket requirement for tests
if they would necessarily be BS.

re: travis check: if $TRACK_ROOT/exercise changes, then there has been an
update to the utility. In that case, it should get checked in travis.
However, this can take a few minutes, and most often, we're not updating
this utility. Therefore, if nothing in $TRACK_ROOT/exercise has changed,
then we shouldn't bother rebuilding it. Make sense? Only build it if it
changes.

coriolinus:
More notes from usage:

- `update` should also update `Cargo.toml` with the appropriate new version string
- `update` inserts new `process_x_case` at the end of the file. I'd prefer it to be inserted at the beginning, before any tests. 
- `update` inserts a `//NEW` line, which I am inclined against keeping. This is somewhat debatable, though. 
- We need special handling for structured canonical data: in the case of the `input` map, it should _not_ represent it as a hashmap, but unpack it into the `input` tuple sent to `process_X_case`. Second special rule: when the tuple sent to `input` contains exactly one item, it doesn't need to be a tuple, but can be a bare item. That is, using `abbreviate` as an example, the generated code should not be:
    ```rust
    process_abbreviate_case(
        {
            let mut hm = ::std::collections::HashMap::new();
            hm.insert("phrase", "GNU Image Manipulation Program");
            hm
        },
        "GIMP",
    );
    ```
    but instead:
    ```rust
    process_abbreviate_case(
        "GNU Image Manipulation Program",
        "GIMP",
    );
    ````
    I'm aware that this is replicating a bug present in the python implementation. 
- when running `update` on `abbreviate`, for some reason, one test was inserted after the `process_abbreviate_case` function. The tests should not be split. 
- when generating `process_X_case` functions, the current implementation causes the compiler to complain about unused variables. We should, instead of generating comments describing what it might look like, just generate the example directly. I.e. using the `abbreviate` case:
    ```rust
    fn process_abbreviate_case<I, O>(input: I, expected: O) {
        assert_eq!(
            student_abbreviate_func(input), // TODO: what is the actual name of student_abbreviate_func?
            expected
        );
    }
    ```
    I am aware that the current implementation replicates the behavior of the Python implementation.
- when generating `process_X_case` functions, we should still generate documentation for the implementer about the expected input items. This is an area where the Rust implementation fails to retain a feature from the Python implementation. It's not so useful right now, because it tended to generate comments like 
    ```rust
    // expected input format: (input, )
    ````

    However, once we implement `input` struct unpacking, it should say:

    ```rust
    // expected input format: (phrase, )
   ````

    And once we implement unpacking single-item tuples, it should say:

    ```rust
    // expected input format: phrase
    ```

coriolinus:
More notes from usage:

- Please remove your name and email from `exercise/Cargo.toml`. We're assigning copyright to Exercism anyway, and it's easier to encourage a broad range of people to come in and help out if nobody appears to be claiming utilities such as these personally.
- `exercise configure` should have a space between the last "e" and the opening parenthesis; currently "...exercise(1, 4, ..."
- `exercise configure` should only accept difficulties equal to or greater than the exercise which unlocks a given exercise
- `exercise configure` should give the user the chance to restart the positioning binary search. That is, after emitting `X is placed between Y and Z exercises`, it should say "Is this correct? (y/N)", and restart the search on N.

I think those are about all the usage notes I'm going to come up with. Tomorrow I'll start looking over the actual code. 

So far, of these notes, there are only a few which I consider critical, which would block merging:

- [ ] if an exercise already exists in config.json, `exercise configure` should default to retaining the existing value.
- [ ] when using exercise generate and an exercise already exists, the program panics.
- [ ] `update` should also update `Cargo.toml` with the appropriate new version string

All the other items can be turned into issues and dealt with after this is merged. This is a big enough program that I don't want to block it until it's perfect.

Thanks for putting this whole thing together! It's clearly been a lot of work.

README.md
```diff
@@ -64,7 +64,7 @@ Please see the documentation about [adding new exercises](https://github.com/exe
 
 Note that:
 
-- The simplest way to generate a project template is to run `bin/init_exercise.py`. You'll need a Python installation >= 3.5 in order to run this script, but it will automate most of the following points for you.
+- The simplest way to generate, update or configure an exercise is to use the [exercise](https://github.com/exercism/rust/tree/master/exercise) utility provided in this repository. To compile the utility you can use the [bin/build_exercise_crate.sh](https://github.com/exercism/rust/tree/master/bin/build_exercise_crate.sh) script or, if the script does not work for you, use the `cargo build --realease` command in the `exercise/` directory and then copy the `exercise` binary from the `exercise/target/release/` directory into the `bin/` directory. Use `bin/exercise --help` to learn about the existing commands and their possible usage.
```
coriolinus: Typo: `--realease` -> `--release`

bin/build_exercise_crate.sh
```diff
@@ -0,0 +1,26 @@
...
+(
+	cd "$EXERCISE_CRATE_PATH"
+
+	echo "Building exercise crate"
+
+	cargo build --release
+
+	RELEASE_PATH="$EXERCISE_CRATE_PATH/target/release/exercise"
+
+	if [ -f "$RELEASE_PATH/exercise.exe" ]; then
```
coriolinus: This will never be true. It needs to be written like
```suggestion
	if [ -f "$RELEASE_PATH.exe" ]; then
```

README.md
coriolinus:
```suggestion
- The simplest way to generate, update or configure an exercise is to use the [exercise](https://github.com/exercism/rust/tree/master/exercise) utility provided in this repository. To compile the utility you can use the [bin/build_exercise_crate.sh](https://github.com/exercism/rust/tree/master/bin/build_exercise_crate.sh) script or, if the script does not work for you, use the `cargo build --release` command in the `exercise/` directory and then copy the `exercise` binary from the `exercise/target/release/` directory into the `bin/` directory. Use `bin/exercise --help` to learn about the existing commands and their possible usage.
```

ZapAnton:
Some questions regarding usage notes:

1) > update inserts a //NEW line, which I am inclined against keeping. This is somewhat debatable, though.

I have included the line as per your comment:

> if the generated test case differs from the existing test case, update
the name of the generated case with _some suffix_ and then emit them both.

The line will change to `//UPDATED` if the test case was already present in the test suite.
Is this behavior undesirable?

2) > update inserts new process_x_case at the end of the file. I'd prefer it to be inserted at the beginning, before any tests.

By 'before any tests' do you mean the tests generated with the `update` subcommand, or the already present test suite?
Because if it is the later, I think it is not the best behavior, since we ungroup the changes made to the test suite and scatter them over the whole file, which could be inconvenient for the contributor who applies the changes. And, since there is no standard format for the test suite, it will be hard to guess where to insert the function, which could breaks the existing formatting and is also inconvenient .

3) > when generating process_X_case functions, the current implementation causes the compiler to complain about unused variables. We should, instead of generating comments describing what it might look like, just generate the example directly. I.e. using the abbreviate case:

Since we will be implementing the `process_x_function` and the current track policy is to always have a non-empty stub file, should we also generate stub template?

4) If we will be tracking the changes made to the utility via Travis script should some sort of versioning be included?

5) When generating or updating the test suite, the utility tries to invoke `rustfmt` on the modified file, which includes the search of the `rustfmt` using the `which` command and, if `rustfmt` is found, applying  the formatting. Because of the `which` command usage, this process is Unix-specifix and will not work on Windows. Would it be enough, if I replace the `which` command with searching the `PATH` environment variable?

ZapAnton:
Additional note about the test suite format:

In the `update` subcommand, I expect the existing test cases to have names like `test_description_from_the_canonical_data`.
Since there is no standard for formatting and naming the test cases, not all exercises adhere to the above rule.
For example if we try to invoke the `update` subcommand on the `leap` exercise we will end up with the whole test suite copied.
Should I left this as it is, and if not should some sort of test naming policy be applied to the track?

ZapAnton:
The Travis script has been added and it seems like it works, but it takes about 5 minutes to complete, which is quite a lot.
Should I change the `cargo build --release` command with the `cargo check` command in this script?

coriolinus:
- re: inserting `//NEW`: I see where the confusion came from. I spoke unclearly. My intent was this:
    - if the generated name for a particular test is `test_description_from_the_canonical_data`, and there already exists a test named `test_description_from_the_canonical_data`, then check if the text of the test is equal to the newly generated test text:
        - if the text of the test is equal to the generated test text, then nothing has changed, so we leave the old test in place and do not write the new test
        - if the text of the test is _not_ equal to the generated test text, then something has changed. We can't expect an automatic tool to know which to keep, so we should keep the existing test, change the name by appending a suffix, and insert the generated test. The format I expect is `test_description_from_the_canonical_data_N`, where `N` is the lowest integer > 1 which does not generate a name collision
    - if the generated name for a particular test is `test_description_from_the_canonical_data`, and there already exists a test named `description_in_other_words`, an automatic tool can't be expected to notice that the test exists already, so simply retain the existing test and also insert the generated test. The tool user or the PR reviewer are the ones who should catch those cases.
- given the rules above, we need neither `//NEW` nor `//UPDATED`.
- re: test file layout. In my opinion, the test file is easiest for a human to read and understand if it contains the following sections in sequence:
    - test module doccomment
    - imports
    - `process_X_case` implementations
    - tests

    I understand that maintaining this format will break up the inserted lines into multiple chunks. That's fine with me. There are plenty of `diff` tools which make the work of finding insertions quite easy. 

    From the `exercise` tool's point of view, the rules above require parsing the test file to determine which tests exist, and what bytes they contain, anyway. Given all that information, I think it should not be difficult to insert new `process_X_case` implementations at the end of the appropriate section.
- I do understand that parsing the existing test file is a non-trivial task; implementing these layout and naming rules will not be required to merge this PR. After merging this PR, I'll create a bunch of issues for items which didn't make it into the PR; at that point, I'll probably start work on implementing the rules above myself. 
- re: generating a stub template. We could think about that for future work, but I think it's beyond the scope of this initial PR. It would be tricky to get right.
- re: versioning the script: I don't know exactly what you mean
- re: searching the PATH instead of using `which`: yes, that's a good idea. I'd be happy to see it in this PR, but it could also be an issue after merging this. I don't think that windows support is a big enough issue for this to block merging this PR.
- re: replacing `cargo build --release` with `cargo check` in travis script: yes, that's a good idea

ZapAnton:
> re: versioning the script: I don't know exactly what you mean

What I wanted to say is that from the start the `exercise` crate had the `0.1.0` version. If we are including the script to check for crate's modifications, should we somehow track and change the crate's version?

coriolinus:
I see. No, I think it's probably best to manually update the crate's version. I'd be very reluctant to give Travis push access to this repo.

coriolinus:
Works for me.

util/exercise/src/cmd/configure.rs
```diff
@@ -0,0 +1,306 @@
+use serde_json::{self, Value};
+use std::{
+    fs,
+    io::{stdin, stdout, Write},
+    path::Path,
+};
+use utils;
+use uuid::Uuid;
+
+fn get_user_input(prompt: &str) -> String {
```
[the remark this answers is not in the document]
ZapAnton: I may have undestood you incorrectly, but the default value will not help in this case, since in the y/N case I do not explicitly check for the N and in the cases with `is_empty` checks I deal with default values of different types (`String`, `u64`, `Vec<String>`) and I don't want to include excess convertions to the `String`.
If I am missing a point here, could you elaborate more on this, please?

ZapAnton:
The removal of the `bin/init_exercise.py` triggered the brach conflict. Should I rebase against master?

ZapAnton:
Ignoring the conflict with the Python script, I think I have addressed all the recent points.

@coriolinus if the last batch of commits does not contain any shorcomings and you are fine with moving the remaining usage and formatting notes into separate issues, I think this PR is ready to be merged.

PR #666, opened 2018-10-04
Add paasio exercise

coriolinus:
This exercise should serve as a gentle introduction to the `io::Read`
and `io::Write` traits, as well as trait implementation in general.

Becuase it's meant to come fairly early in the tree, there is very
little actual implementation required of the student.

The most complex part is the tests, which are generated by macros.
Even so, they're still pretty easy to work with: students just need
to find and remove the `#[ignore]` attributes or run `cargo test -- --ignored`,
exactly as normal.


exercises/paasio/tests/paasio.rs
```diff
@@ -0,0 +1,189 @@
...
+            $(#[$attr])*
+            #[test]
+            fn test_read_chunks() {
+                let data = $input;
+                let size = $len(&data);
+                let mut reader = ReadStats::new(data);
+
+                let mut buffer = [0_u8; CHUNK_SIZE];
+                let mut chunks_read = 0;
+                while reader.read(&mut buffer[..]).expect("read must succeed") > 0 {
```
petertseng: although it might be cumbersome to do so, maybe would be helpful if the error would indicate which read failed, by incorporating chunks_read? Unsure

exercises/paasio/tests/paasio.rs
```diff
@@ -0,0 +1,189 @@
...
+            $(#[$attr])*
+            #[test]
+            fn test_read_buffered_chunks() {
+                let data = $input;
+                let size = $len(&data);
+                let mut reader = BufReader::new(ReadStats::new(data));
+
+                let mut buffer = [0_u8; CHUNK_SIZE];
+                let mut chunks_read = 0;
+                while reader.read(&mut buffer[..]).expect("read must succeed") > 0 {
```
petertseng: > although it might be cumbersome to do so, maybe would be helpful if the error would indicate which read failed, by incorporating chunks_read? Unsure


same

exercises/paasio/tests/paasio.rs
```diff
@@ -0,0 +1,189 @@
...
+            const CHUNK_SIZE: usize = 2;
+            $(#[$attr])*
+            #[test]
+            fn test_write_passthrough() {
+                let data = $input;
+                let size = $len(&data);
+                let mut writer = WriteStats::new(Vec::with_capacity(size));
+                let written = writer.write(data);
+                assert!(written.is_ok());
+                assert_eq!(size, written.unwrap());
```
petertseng: here, too, I would wonder about just combining this line and the above

exercises/paasio/tests/paasio.rs
```diff
@@ -0,0 +1,189 @@
...
+            fn test_read_passthrough() {
+                let data = $input;
+                let size = $len(&data);
+                let mut reader = ReadStats::new(data);
+
+                let mut buffer = Vec::with_capacity(size);
+                let qty_read = reader.read_to_end(&mut buffer);
+
+                assert!(qty_read.is_ok());
+                assert_eq!(size, qty_read.unwrap() );
```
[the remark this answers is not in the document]
coriolinus: I agree with you, that would be nicer. So I tried it, and got this funky error:

```rust
   Compiling paasio v0.0.0 (file:///mnt/d/Users/coriolinus/Documents/Projects/exercism.io/xrust/exercises/paasio)
error[E0277]: can't compare `std::io::Error` with `std::io::Error`
   --> tests/paasio.rs:161:1
    |
161 |    test_read!(read_string (
    |   _^
    |  |_|
    | ||
162 | ||     "Twas brillig, and the slithy toves/Did gyre and gimble in the wabe:/All mimsy were the borogoves,/And the mome raths outgrabe.".as_bytes(),
163 | ||     |d: &[u8]| d.len()
164 | || ));
    | ||   ^
    | ||___|
    | |____no implementation for `std::io::Error == std::io::Error`
    |      in this macro invocation
    |
    = help: the trait `std::cmp::PartialEq` is not implemented for `std::io::Error`
    = note: required because of the requirements on the impl of `std::cmp::PartialEq` for `std::result::Result<usize, std::io::Error>`
    = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)
```

Apparently `std::io::Error` doesn't implement `PartialEq`, which makes that impossible. Oh well.

coriolinus:
> My best guess was that `get_bytes` and `get_reads` should instead be `bytes` and `reads`, but `get_ref` should stay `get_ref`.

After reading those links, I concur. Thank you, by the way, for finding those; I wasn't really aware of the style guide's position on this.

Renaming `get_reads` -> `reads` and `get_writes` -> `writes` was smooth, but I ran into some startling compiler errors when I attempted to rename `get_bytes` to `bytes`. Copying from the commit message, because I'm about to squash the commit in question and I think it's worth preserving:

> Turns out that the implementation was shadowing or being shadowed (not sure which) by the default implementation of [`Read.bytes()`](https://doc.rust-lang.org/std/io/trait.Read.html#method.bytes), which has a different signature and much different semantics.
> 
> For consistency, I named the method bytes_through, so that it applies equally to both Readers and Writers.

PR #667, opened 2018-10-04
Add scale generator

coriolinus:
Add the `scale-generator` exercise.

This is a bit more complex, but most of the work is simply in finding a good model for the problem domain. The problem is not conceptually difficult.


coriolinus:
This is, I believe, ready; WIP plugin is incorrectly marking it as still WIP.

petertseng:
> incorrectly marking it as still WIP

I see, so you propose it would be an improvement to the WIP app if it would only consider the last commit message in determining WIP status, rather than all of the commit messages. Upon reflection, that seems like a reasonable proposal, and it could be a good idea to submit the idea to the app's maintainers.

coriolinus:
It's already been suggested there (https://github.com/wip/app/issues/107); the author seemed uninterested, but did say that in the next major WIP release would enable manual overrides.

exercises/scale-generator/tests/scale-generator.rs
```diff
@@ -0,0 +1,179 @@
+//! Tests for scale-generator
+//!
+//! Generated by [script][script] using [canonical data][canonical-data]
+//!
+//! [script]: https://github.com/exercism/rust/blob/master/bin/init_exercise.py
+//! [canonical-data]: https://raw.githubusercontent.com/exercism/problem-specifications/master/exercises/scale-generator/canonical_data.json
```
petertseng: oh, whoops, is the generator emitting canonical_data.json as the filename? Should be canonical-data.json

exercises/scale-generator/tests/scale-generator.rs
```diff
@@ -0,0 +1,179 @@
...
+
+extern crate scale_generator;
+use scale_generator::*;
+
+/// Process a single test case for the property `chromatic`
```
petertseng: Hmm, I guess it's a bit unfortunate that both of these are `pitches` in the JSON file.

Also, potential confusion for students by seeing "property `chromatic`" here and "`pitches` property" two lines down? I'd consider making these two the same.

Same comment for `interval`

exercises/scale-generator/tests/scale-generator.rs
coriolinus: Oh, good catch. Created https://github.com/exercism/rust/issues/669 for this.

coriolinus:
> If you were given the opportunity to do so, would you have preferred to be able to write the example solution as multiple files?

Yes, I would have. 350 lines isn't enormous for a source file, but it's big enough that it's not perfectly convenient to use anymore. Three files of ~120 lines each would have been much more convenient to work with.

petertseng:
> > If you were given the opportunity to do so, would you have preferred to be able to write the example solution as multiple files?
>
> Yes, I would have. 350 lines isn't enormous for a source file, but it's big enough that it's not perfectly convenient to use anymore. Three files of ~120 lines each would have been much more convenient to work with.

I see. Let's keep it in mind and consider a change if it becomes the case that multiple exercises need this.

exercises/scale-generator/tests/scale-generator.rs
fersss: https://www.fuckbook.com/es/account/premium?

Issue #672, opened 2018-10-17
reverse_string exercise setup with deprecated graphemes feature


jackyzhen:
The graphemes feature has been [deprecated since 1.3.0](https://users.rust-lang.org/t/where-did-str-graphemes-go/7817). Most suggestions point to the [unicode-segmentation crate](https://crates.io/crates/unicode-segmentation) instead.

coriolinus:
[exercise link](https://github.com/exercism/rust/tree/master/exercises/reverse-string)

At no point does this exercise use, encourage use of, or mention the deprecated `str.graphemes()` function. You're misunderstanding what the feature flag does. We declare a `grapheme` feature in `Cargo.toml`, and then have an extra test which may be optionally run by testing with that feature flag. This is described in the README. 

Closing as not a bug.

PR #673, opened 2018-10-17
Remove deprecated grapheme dep from reverse_string

jackyzhen:
Fixes https://github.com/exercism/rust/issues/672.

### DIFF
--- a/exercises/reverse-string/Cargo.toml
+++ b/exercises/reverse-string/Cargo.toml
@@ -2,7 +2,5 @@
 name = "reverse_string"
 version = "1.0.0"
 
-[features]
-grapheme = []
-
 [dependencies]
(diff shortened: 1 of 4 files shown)


coriolinus:
Will not merge: the behavior this removes is a feature, not a bug.

I might consider merging a PR which _only_ added a link to the `unicode-segmentation` crate to the README, but there's a debate there: is pedagogy better served by linking directly to the relevant crate, or by giving students a starting point and encouraging them to google?

jackyzhen:
Doh! Apologies for misunderstanding.

coriolinus:
No apology necessary! I'm grateful that you saw what you thought was a bug, and took action not only to report it but to attempt a solution. That kind of initiative is much appreciated; it's just that in this particular case it was misdirected.

PR #675, opened 2018-10-20
say: fix testcase

lutostag:
The test_max_u64 testcase was not actually testing u64::max_value()
but instead was testing i64::max_value(), for clarity fixed the testcase
to use the correct value.

Works with implementation from https://github.com/lutostag/exercism/tree/master/rust/say


lutostag:
Changed as recommended. Thanks for the feedback.

PR #689, opened 2018-10-21
paasio: Add ReadStats.get_ref test

ccouzens:
The tests don't use it. The description in the readme doesn't make it a
requirement.

As such, I think it's better removed.

The alternative would be to write a test, or modify an existing test so
that it is used. If others are less keen on this being removed, I'll do
that.

### DIFF
--- a/exercises/paasio/tests/paasio.rs
+++ b/exercises/paasio/tests/paasio.rs
@@ -187,3 +187,14 @@ read_file(
     ::std::fs::File::open("README.md").expect("readme must be present"),
     |f: &::std::fs::File| f.metadata().expect("metadata must be present").len() as usize
 ));
+
+#[test]
+#[ignore]
+fn read_stats_by_ref_returns_wrapped_reader() {
+    use paasio::ReadStats;
+
+    let input =
+        "Why, sometimes I've believed as many as six impossible things before breakfast".as_bytes();
+    let reader = ReadStats::new(input);
+    assert_eq!(reader.get_ref(), &input);
+}


ZapAnton:
As per [Exercism pull request guidelines](https://github.com/exercism/docs/blob/master/contributing/pull-request-guidelines.md)  each commit message should have a name of the modified exercise in it.

Please consider renaming your commit message as `paasio: ReadStats.get_ref method` (if this commit was the last on your local machine, then you can simply run `git commit --amend -m "paasio: ReadStats.get_ref method"` and then push to the remote with the `--force` flag)

ZapAnton:
As per [Exercism pull request guidelines](https://github.com/exercism/docs/blob/master/contributing/pull-request-guidelines.md)  each commit message should have a name of the modified exercise in it.

Please consider renaming your commit message as `paasio: ReadStats.get_ref method` (if this commit was the last on your local machine, then you can simply run `git commit --amend -m "paasio: ReadStats.get_ref method"` and then push to the remote with the `--force` flag)

ZapAnton:
As per [Exercism pull request guidelines](https://github.com/exercism/docs/blob/master/contributing/pull-request-guidelines.md)  each commit message should have a name of the modified exercise in it.

Please consider renaming your commit message as `paasio: ReadStats.get_ref method` (if this commit was the last on your local machine, then you can simply run `git commit --amend -m "paasio: ReadStats.get_ref method"` and then push to the remote with the `--force` flag)

ZapAnton:
As per [Exercism pull request guidelines](https://github.com/exercism/docs/blob/master/contributing/pull-request-guidelines.md)  each commit message should have a name of the modified exercise in it.

Please consider renaming your commit message as `paasio: ReadStats.get_ref method` (if this commit was the last on your local machine, then you can simply run `git commit --amend -m "paasio: ReadStats.get_ref method"` and then push to the remote with the `--force` flag)

ZapAnton:
As per [Exercism pull request guidelines](https://github.com/exercism/docs/blob/master/contributing/pull-request-guidelines.md)  each commit message should have a name of the modified exercise in it.

Please consider renaming your commit message as `paasio: ReadStats.get_ref method` (if this commit was the last on your local machine, then you can simply run `git commit --amend -m "paasio: ReadStats.get_ref method"` and then push to the remote with the `--force` flag)

coriolinus:
I am inclined against merging this as it stands.

You're correct that the tests don't use it, and the description in the readme doesn't make it a requirement. However, I would strongly prefer to have you add a test which needs it, rather than removing the method.

One of the overarching goals of Exercism is to demonstrate what good library design looks like. Good library design for a wrapper type like this allows users a way to access the interior type.

coriolinus:
I am inclined against merging this as it stands.

You're correct that the tests don't use it, and the description in the readme doesn't make it a requirement. However, I would strongly prefer to have you add a test which needs it, rather than removing the method.

One of the overarching goals of Exercism is to demonstrate what good library design looks like. Good library design for a wrapper type like this allows users a way to access the interior type.

coriolinus:
> The alternative would be to write a test, or modify an existing test so
that it is used. If others are less keen on this being removed, I'll do
that.

I would strongly prefer this alternative to the one implemented in this PR. 

One of the goals of Exercism is to have people write what is fundamentally library code in every exercise. It's good software architecture for every feature to be implemented within a library, and for executables to be as thin as possible while handling necessary I/O. Good library design, in turn, implies that a wrapper type such as we implement in `paasio` should provide a means by which the wrapped type may be borrowed. 

We want to demonstrate good library design and good architecture so that students can develop an intuition about what "feels good" in a library, because eventually they should be authoring their own. When they do write their own library, they shouldn't leave out methods just because they didn't happen to need them for their own project at the moment.

coriolinus:
Not sure if I have or have not multi-posted here. If so, I blame Github.

ccouzens:
Hi,

Thanks for both pieces of feedback.

I'll action them the next time I have a chance.

exercises/paasio/tests/paasio.rs
```diff
@@ -187,3 +187,14 @@ read_file(
...
+
+#[test]
+#[ignore]
+fn read_stats_by_ref_returns_wrapped_reader() {
+    use paasio::ReadStats;
+
+    let input =
+        "Why, sometimes I've believed as many as six impossible things before breakfast".as_bytes();
+    let reader = ReadStats::new(input);
+    assert_eq!(reader.get_ref(), &input);
```
coriolinus: This compares equality, but not identity. You may want to try something like this (note that I haven't tested this yet):
```suggestion
    assert_eq!(reader.get_ref(), &input);
    assert_eq!(reader.get_ref() as *const _, &input as *const _); // test that the returned reader is the wrapped reader
```

ccouzens: Comparing casts to `*const` failed with the error:

```
---- read_stats_by_ref_returns_wrapped_reader stdout ----
thread 'read_stats_by_ref_returns_wrapped_reader' panicked at 'assertion failed: `(left == right)`
  left: `0x7f0bd21fc600`,
 right: `0x7f0bd21fc5f0`', tests/paasio.rs:198:5
```

I had better luck with

```rust
assert_eq!(reader.get_ref().as_ptr(), input.as_ptr()); // test that the returned reader is the wrapped reader
```

But it is very hard to test that the test is effective.

Even when I use 2 different ways of constructing the slices such as:

```rust
    let input = &['W' as u8];
    let reader = ReadStats::new("W".as_bytes());
    assert_eq!(reader.get_ref().as_ptr(), input.as_ptr()); // test that the returned reader is the wrapped reader
```

it passes. I'm guessing the rust compiler works out that both slices have the same content and collapses them both into the same memory.

To answer your questions:

* I don't know if it's a good idea, but I don't think it's necessary. `get_ref` returning a reference to the original wrapped reader is the easiest solution to this test. Doing anything else would require a lot of effort from the person solving the exercise, since `get_ref` is a generic method and can return many different types. Tests can almost always be bypassed if someone puts in enough effort to do so.
* I'm not sure. It's not syntax I'm familiar with. Casting to `*const` is what I found when googling "rust compare identity". But it didn't pass on my laptop.
* Both casting to `*const` and using `as_ptr` behaved the same for me with and without `--release` specified.

PR #688, opened 2018-10-20
check-exercises: Rewritten script to compile only modified exercises

ZapAnton:
Closes #674

PR #691, opened 2018-10-21
ensure-stubs-compile: Rewritten the script to target only modified exercises

ZapAnton:
A follow-up for #688, that targets another time-consuming Travis script - `_test/ensure-stubs-compile.sh`

### DIFF
--- a/_test/ensure-stubs-compile.sh
+++ b/_test/ensure-stubs-compile.sh
@@ -2,9 +2,25 @@
 
 repo=$(cd "$(dirname "$0")/.." && pwd)
 
+current_branch="$(git rev-parse --abbrev-ref HEAD)"
+
+if [ "$current_branch" != "master" ]; then
+	changed_exercises="$(git diff --name-only master | grep "exercises/" | cut -d '/' -f -2)"
+else
+	changed_exercises=exercises/*
+fi
+
+if [ -z "$changed_exercises" ]; then
+	echo "No exercise was modified. The script is aborted."
+
+	exit 0
+fi
+
 broken=""
 
-for dir in $repo/exercises/*/; do
+for dir in $changed_exercises; do
+  dir="$repo/$dir"
+
   exercise=$(basename "$dir")
 
   # If src/lib.rs contains any non-comment line that contains any non-spaces,
--- a/_test/ensure-stubs-compile.sh
+++ b/_test/ensure-stubs-compile.sh
@@ -5,9 +5,9 @@ repo=$(cd "$(dirname "$0")/.." && pwd)
 current_branch="$(git rev-parse --abbrev-ref HEAD)"
 
 if [ "$current_branch" != "master" ]; then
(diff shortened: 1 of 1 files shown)


_test/ensure-stubs-compile.sh
```diff
@@ -2,9 +2,25 @@
 
 repo=$(cd "$(dirname "$0")/.." && pwd)
 
+current_branch="$(git rev-parse --abbrev-ref HEAD)"
+
+if [ "$current_branch" != "master" ]; then
+	changed_exercises="$(git diff --name-only master | grep "exercises/" | cut -d '/' -f -2)"
+else
+	changed_exercises=exercises/*
```
coriolinus: This only works from the repo root. 
```suggestion
	changed_exercises=$repo/exercises/*
```

coriolinus: No. 

When `$PWD == $repo`, `changed_exercises` is formed from the glob-expansion of `$repo/exercises/*`. If, however, `$PWD == $repo/bin`, `changed_exercises` will attempt to expand the glob `$repo/bin/exercises/*`; it will find nothing matching that pattern, and form an empty list.

As far as Travis is concerned, this is fine; it will always run the scripts from the repo root. However, humans are less consistent. I often run the `_tests/*.sh` scripts manually before submitting a PR, so I can have confidence that my PR will pass Travis before actually submitting, because they finish faster locally. I'm not always in the repo root when I run them, so it's better if they're independent of `$PWD`.

ZapAnton: Then how about replacing the `repo` variable value with git repo root via `git rev-parse --show-toplevel`? This way every path could start from the track root, regardless where the script was invoked

coriolinus: No, the value of `$repo` is fine. The current method will stop working if the script is moved out of `bin/`, but it works regardless of where the script is invoked from. It's fine to change the method for finding `$repo`, but it won't address this issue.

The problem stems from the fact that right now `changed_exercises` attempts to look up the `exercises` subdirectory, which is an unqualified relative path. That is sensitive to `$PWD`. The suggestion for fixing it is to make it relative to `$repo`, which is an absolute path; this eliminates the sensitivity.

ZapAnton: This fix would require two additional steps then:

- remove `dir="$repo/$dir"` line, since this would copy the $`repo` value twice
- modify the `changed_exercises="$(git diff --name-only master | grep "exercises/" | cut -d '/' -f -2)" ` to include `$repo` value in each located path. Maybe you have some suggestion for this one? I am not very proficient with the `sed`/`awk` magic

coriolinus: Yes, these modifications will be necessary.

I'll take a look at this later, but I can't make any promises for how soon.

petertseng:
FYI, this seems like a quite good idea and I'm going to consider it for other track I maintain. Thanks for bringing up the idea and seeing it to completion!

PR #693, opened 2018-10-26
Update instruction in gigasecond test to match src

houhoulis:


houhoulis:
You are right, it is adding redundancy.  Perhaps it's better to remove that line of the instruction than it is to fix it.

The motivation for the PR is that the instruction is incorrect, and it misled me.  Before this PR, the "following two lines" include `use chrono::{TimeZone, Utc};`, which is not what's needed in the solution.

coriolinus:
Ah, I finally understand what you were driving at. Adding links here for future reference. The first two lines in the stub file:

https://github.com/exercism/rust/blob/8d31ba9c2d335fb1d2323170a6a8ef2f76e8eb3b/exercises/gigasecond/src/lib.rs#L1-L2

The description from the test file:

https://github.com/exercism/rust/blob/8d31ba9c2d335fb1d2323170a6a8ef2f76e8eb3b/exercises/gigasecond/tests/gigasecond.rs#L13-L17

The note to students in the tests was written before we wrote stubs for all exercises. It is now both redundant and out of date. The way forward is to remove that comment from the tests file.

It looks at first glance as through the import of `chrono::TimeZone` is not actually used in the tests file. If you're willing, please try removing that as well; if everything passes without it, we're better off avoiding orphan imports.

houhoulis:
The tests don't run if I change `use chrono::{TimeZone, Utc};` to `use chrono::Utc;`

The test file needs `TimeZone`, and the stub needs `DateTime`.

coriolinus:
Great, thanks for checking that. In that case, we just need to remove the
comment; no changes to the actual imports will be necessary.

PR #694, opened 2018-10-26
make exercise util less panicky

### LINKED ISSUE #677: exercise crate is panicky
```sh
$ rg 'panic!|unwrap|expect' -trust -oN --no-filename util/exercise/ | rg -v '^$' | wc -l
Sat Oct 20 13:31:27 DST 2018
110
```

It would be more idiomatic if, instead of panics, errors propagated up to the `main` function where they can be handled appropriately. 

- refactor `util/exercise` crate such that functions which currently panic return `Option` or `Result` instead, as appropriate, and errors are propagated upwards
- refactor `fn main()` to return `-> Result<(), Error`
- consider using the `failure` crate to ease this refactor work

coriolinus:
Closes #677.

Work so far involved adding return values to almost every function in the codebase,
defining a custom Result type for convenience, a tiny refactor to move the
util stuff into a proper library; even with all that, it's only the start.

Next steps:

- [x] clean up `expect(...)`
- [x] clean up explicit `panic!(...)`
- [x] define a proper custom error enum instead of depending on `failure::Error`
- [x] write utility macro to get rid of the repeated `.ok_or(format_err!(...))` calls
- [x] go over diff and re-insert context where appropriate


coriolinus:
This should now be ready for review.

util/exercise/Cargo.toml
```diff
@@ -1,12 +1,14 @@
 [package]
 name = "exercise"
-version = "1.0.0"
+version = "1.1.0"
```
ZapAnton: For future reference, why did you bumped the minor version and not the build version (e.g. `1.0.1`)? 
I thought the minor version is updated when a new  feature is added, which is not a case here (I may be wrong, since proper error-handling could be considered a feature)

ZapAnton:
If you don't mind, could you describe which of the [following patterns](https://boats.gitlab.io/failure/guidance.html) (if any) was used in this PR?
Asking, because I was also going to open a PR to clear the crate of the unnecessary panics, but took some time to learn about the `failure` crate (you obviously acted faster on the problem), so now I am interested what practice of error-handling do you prefer to see on this track (for future reference basically).

util/exercise/Cargo.toml
coriolinus: I consider proper error-handling a feature.

coriolinus:
The major pattern used was a custom Fail enum, defined here: 

https://github.com/exercism/rust/blob/3d729e92c618eadd296136ab4ba4a8a29a96c7f6/util/exercise/src/errors.rs#L7-L38

However, we also use strings as errors (which end up as the `Failure` case of the `Error` enum). It isn't great, but some of the one-off variants didn't seem worth promoting to proper failure cases at the time. Example:

https://github.com/exercism/rust/blob/3d729e92c618eadd296136ab4ba4a8a29a96c7f6/util/exercise/src/cmd/configure.rs#L76-L78

I do not consider myself a master of the `failure` crate. There are definitely new patterns that I could learn, and other places where I could use it. For example, when looking through the `failure` docs after removing the WIP tag, I discovered a [`bail!()` macro](https://docs.rs/failure/0.1.3/failure/macro.bail.html) which could simplify some of the code in this PR. 

I submitted as is because I figured that the primary objective was to use error handling instead of panic/unwrap/expect, and if I grepped correctly, this PR accomplishes that goal. Further refinement can happen in the future.

coriolinus:
Given that they has received no maintainer reviews in over a week, I'm going to merge this and https://github.com/exercism/rust/pull/699 as soon as they both pass Travis.

For the future I will probably be a bit more aggressive about merging my own PRs; it seems like the exercise util is (understandably) of lower priority for other track maintainers than PRs which are directly visible to students.

util/exercise/src/lib.rs
```diff
@@ -17,16 +33,59 @@ lazy_static! {
...
 }
 
-pub fn run_configlet_command(command: &str, args: &[&str]) {
+#[macro_export]
+macro_rules! val_as {
+    ($value:expr, $as:ident) => {
+        $value
+            .$as()
+            .ok_or_else(|| $crate::errors::Error::SchemaTypeError {
+                file: "config.json".to_string(),
```
ZapAnton: Since the `val_as!` is called both when parsing canonical-data and config.json, perhaps the hard-coded `file: "config.json".to_string()` line should be changed here?

coriolinus: I added cases to the macro to make that an overrideable default in https://github.com/exercism/rust/pull/700/commits/e3c726e4275336c4f3458a8fa133954f81ca1885; it just hasn't yet been merged.

PR #697, opened 2018-10-27
added ability to fetch configlet from within exercise util

### LINKED ISSUE #678: Give exercise util the ability to fetch configlet within rust
https://github.com/exercism/rust/blob/45599b887b7a49801da20ac9c1f63a460c5cec4c/util/exercise/src/utils.rs#L41-L48

The exercise script should be able to fetch configlet in a platform-independent manner.

Library code which implements this feature exists in [`coriolinus:riir-fetch-configlet`](https://github.com/coriolinus/exercism_rust/blob/riir-fetch-configlet/bin/fetch-configlet/src/lib.rs). 

- Copy this code into an appropriate place in the exercise util
- replace the call to the bash script with a call to `download()`

mh84:
This will take care of and close #678

### DIFF
--- a/util/exercise/src/cmd/configure.rs
+++ b/util/exercise/src/cmd/configure.rs
@@ -302,8 +302,7 @@ pub fn configure_exercise(exercise_name: &str) {
     fs::write(
         &config_path,
         serde_json::to_string_pretty(&config_content).unwrap(),
-    )
-    .expect("Failed to write the updated track configuration to the config.json file");
+    ).expect("Failed to write the updated track configuration to the config.json file");
 
     println!("Formatting the config.json file via 'bin/configlet fmt'");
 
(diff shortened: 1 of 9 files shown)


util/exercise/src/cmd/configure.rs
```diff
@@ -302,7 +302,8 @@ pub fn configure_exercise(exercise_name: &str) {
     fs::write(
         &config_path,
         serde_json::to_string_pretty(&config_content).unwrap(),
-    ).expect("Failed to write the updated track configuration to the config.json file");
+    )
+    .expect("Failed to write the updated track configuration to the config.json file");
```
coriolinus: This (and the similar related changes throughout this PR) doesn't look like what `rustfmt` generates, and in any case, it's unrelated to the change which is the main thrust of the PR. Please undo these. Simply running `rustfmt` should accomplish that.

PR #700, opened 2018-10-28
WIP: use templates in exercise util

### LINKED ISSUE #698: exercise util: use templates instead of manually modifying files
Currently many operations in this utility operate at the file level, directly appending strings and bytes. We'll get better results with higher-level abstractions: never directly appending to `Config.toml`, for example, but using a toml-parsing crate instead. Likewise, writing the tests file will be much easier using a template system like [`handlebars`](https://github.com/sunng87/handlebars-rust). Note that this mention of handlebars isn't a requirement to use that particular crate, just an example of a plausible templating engine.

It is not necessary to implement this before #679 or #686, but those issues will probably be easier to fix after the fix for this has been implemented.

coriolinus:
Closes #698. ~~Depends on https://github.com/exercism/rust/pull/699; do not merge until~~ that PR is merged.


coriolinus:
Closing because I haven't had time to work on this in ages, and it's gotten more than a little stale. Maybe someone will pick up the effort in the future.

PR #701, opened 2018-10-31
 exercise: Added fetch_configlet subcommand

ZapAnton:
A follow up for #697. 

Thought it would make sense for users to have a way to use the Rust implementation of the `fetch_configlet`


util/exercise/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "exercise"
-version = "1.0.0"
+version = "1.2.0"
```
coriolinus: Let's increment the version sequentially. If this is merged before #694, then this gets to be v1.1.0.

ZapAnton:
What do you think should be done with the original shell script? Should the script and this subcommand co-exist or should the script be removed?

coriolinus:
They should coexist. The reason https://github.com/exercism/rust/pull/654 was closed without merge was that the bash script just executed quickly, whereas the rust version took minutes to compile. For something which needs to run on every Travis execution, it didn't make sense to spend those minutes.

PR #704, opened 2018-11-04
wordy: simplify example solution

petertseng:
It seems that some elements of the example solution can be removed, so I propose some.

Includes the commit from #703 because that is a part of some changes made.

I intend that I will be the one to merge this after review, but if I should be unavailable for that, I recommend not squashing on merge.

### DIFF
--- a/exercises/wordy/example.rs
+++ b/exercises/wordy/example.rs
@@ -19,9 +19,9 @@ impl <'a> Token<'a> {
     }
 }
 
-pub fn answer(c: &str) -> Option<isize> {
+pub fn answer(c: &str) -> Option<i32> {
     let mut t = tokens(c);
-    let mut result: isize = 0;
+    let mut result: i32 = 0;
     let mut opr = "plus";
 
     if t.len() <= 1 {
@@ -36,7 +36,7 @@ pub fn answer(c: &str) -> Option<isize> {
     }
 }
 
-fn evaluate(r: isize, operator: &str, operand: isize) -> isize {
+fn evaluate(r: i32, operator: &str, operand: i32) -> i32 {
     match operator {
         "plus" => r + operand,
         "minus" => r - operand,
@@ -46,7 +46,7 @@ fn evaluate(r: isize, operator: &str, operand: isize) -> isize {
     }
 }
 
-fn operand(t: &Token) -> isize {
+fn operand(t: &Token) -> i32 {
     t.value.parse().unwrap()
 }
 
--- a/exercises/wordy/example.rs
+++ b/exercises/wordy/example.rs
@@ -1,4 +1,3 @@
-#[derive(Debug, PartialEq)]
 struct Token<'a> {
     value: &'a str,
 }
--- a/exercises/wordy/example.rs
+++ b/exercises/wordy/example.rs
@@ -1,9 +1,9 @@
 #[derive(Debug, PartialEq)]
-struct Token {
-    value: String,
+struct Token<'a> {
+    value: &'a str,
 }
 
-impl Token {
+impl <'a> Token<'a> {
     fn is_valid(&self) -> bool {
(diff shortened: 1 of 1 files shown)


petertseng:
aside: in the similar vein of https://github.com/exercism/rust/issues/601 (inappropriate usize), I think wordy is an inappropriate `isize`

petertseng:
at least it's limited to the example; the stub is already using i32

exercises/wordy/example.rs
```diff
@@ -18,74 +12,57 @@ impl Token {
...
+    let mut t = tokens(c);
+    let mut result: i32 = 0;
+    let mut opr = "plus";
 
-impl WordProblem {
-    fn new(c: &str) -> Self {
-        WordProblem {
-            command: String::from(c),
+    if t.len() <= 1 {
+        None
```
coriolinus: I'm surprised at this behavior. Not saying you're wrong: this keeps compatibility with the previous implementation. All the same, I'd expect `answer("42") == Some(42)`. Do you think we should change this behavior and add a test case?

exercises/wordy/example.rs
```diff
@@ -18,74 +12,57 @@ impl Token {
...
+fn evaluate(mut r: i32, operator: &str, operand: i32) -> i32 {
+    if operator == "plus" {
+        r += operand;
+    } else if operator == "minus" {
+        r -= operand;
+    } else if operator == "multiplied" {
+        r *= operand;
+    } else if operator == "divided" {
+        r /= operand;
     }
```
coriolinus: Would this `if-else` chain be better expressed as `match`?

exercises/wordy/example.rs
```diff
@@ -18,74 +12,57 @@ impl Token {
...
+    r
+}
 
-    fn operand(&self, t: &Token) -> isize {
-        t.value.parse().unwrap()
-    }
+fn operand(t: &Token) -> i32 {
+    t.value.parse().unwrap()
```
coriolinus: The `unwrap` here is also surprising: `answer("plus 1 2")` will panic, instead of returning `None`. This is another potential improvement we could make to this exercise, though it might not fit into this PR.

(In an ideal world, it might return `Err(wordy::Error::ReversePolishNotation)`, but that's probably not a change we actually want to make to this exercise.)

exercises/wordy/example.rs
petertseng: I like your suggestion since it reduces the repetitiveness of the `operator ==` bits.

petertseng: I think the `match` was not possible while `String` was being used, so it appears this PR was necessary to unlock your suggestion

exercises/wordy/example.rs
petertseng: Although I would consider the string "42" to not be a fully-formed word problem, one could imagine "What is 42?" to be a valid problem and have an answer of 42. It depends on whether a valid word problem is required to have at least one operation in it.

My default mode of operation is not to be too interested and therefore to be content with leaving "What is 42?" as undefined behaviour, but at this current moment I am also reasonably able to start a discussion in problem-specifications and see where it goes.

exercises/wordy/example.rs
coriolinus: No, it's possible with a string; you just [need to use `as_ref`](https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=0e5e751616dea178d53e12c495fe397a).

exercises/wordy/example.rs
petertseng: One might also imagine testing that `"1 plus plus 2"` results in `None`. While usually I'm not so interested, I do have a suspicion that without such cases, it'd be possible to get away with parsers that are a bit too fragile. The parser in this example is indeed a little too permissive.

PR #706, opened 2018-11-04
[WIP] Added checker crate

ZapAnton:
This PR adds a new util crate - `checker`.

The motivation behind the crate is to RIIR the existing bash scripts (mainly from the `_test/` directory), that contain non-trivial logic.

The possible advantages:
- It is easier to maintain complex Rust code than complex Bash code
- Some repeating functionality from different scripts (for instance which exercise was modified) is located in the single crate
- The checks become cross-platform 

Possible scripts to be included in the crate:

- `ensure-stubs-compile.sh`
- `check-exercises.sh` 
- `check-stubs.sh`
- `ensure-lib-src-rs-exist.sh`
- `check-exercise-crate.sh`

The crate is being developed without any external dependencies to avoid the problems of the #654 PR. As the result, the crate builds in about 2 seconds, but has no nice feature like `lazy_static!`, args parsing with `clap` or error handling with `failure`. If the PR is approved, the external crates could be added in the future. 

Since it is unknown if this type of util crate is desirable, it is designed very roughly, without any arg parsing or proper error handling and with only `ensure-stubs-compile.sh` functionality present. This could all be further developed if the maintainers are agreed that the crate is needed. Therefore the `In Progress` status of the PR.


util/checker/src/cmd/check_stubs_compile.rs
```diff
@@ -0,0 +1,169 @@
+use std::{fs, path::PathBuf, process::Command};
+
+fn make_reserve_copies(modified_exercise: &PathBuf) {
```
coriolinus: `PathBuf` and `Path` have the same relationship that `String` and `str` do. It generally makes sense to write a function to take `&Path` unless you need to modify the original, for the same reason that you'd generally take `&str` instead of `&String`.

util/checker/src/cmd/check_stubs_compile.rs
```diff
@@ -0,0 +1,169 @@
...
+    let stub_path = modified_exercise.join("src").join("lib.rs");
+
+    let stub_path_copy = modified_exercise.join("src").join("lib.rs.orig");
```
coriolinus: Why specify `*_copy` variables manually instead of just using [`with_extension`](https://doc.rust-lang.org/std/path/struct.Path.html#method.with_extension)?

util/checker/src/cmd/check_stubs_compile.rs
```diff
@@ -0,0 +1,169 @@
...
+            fs::copy(
+                &entry_path,
+                &tests_path_copy.join(entry_path.file_name().unwrap()),
+            ).unwrap_or_else(|_| panic!("Failed to make a reserve copy for {:?}", entry_path));
+        }
+    }
+}
+
+fn remove_copies(modified_exercise: &PathBuf) {
+    let stub_path = modified_exercise.join("src").join("lib.rs");
```
coriolinus: Given that you have to produce the mapping of originals and copies in a few places, it seems like it may be worthwhile to extract out a common function which generates a list of tuples, which you could iterate over.

util/checker/src/cmd/check_stubs_compile.rs
```diff
@@ -0,0 +1,169 @@
...
+        panic!(
+            "Failed to rename {:?} to {:?}: {}",
+            &tests_path_copy, &tests_path, error
+        )
+    });
+}
+
+fn add_deny_warning_flag(file_path: &PathBuf) {
+    let file_content = fs::read_to_string(&file_path)
+        .unwrap_or_else(|error| panic!("Failed to read file {:?}: {}", &file_path, error));
```
coriolinus: It would be nice if this utility used proper error handling from the beginning instead of needing a refactor. The good news is that a custom error enum should be pretty easy for this use case, and you don't actually need the `Failure::Error` trait to use it.

coriolinus: You could probably get pretty far, for now, if your custom error type had a single variant which wrapped `io::Result`.

coriolinus: Oh, I see down below you've actually done that. Might as well make use of that here, then!

util/checker/src/cmd/check_stubs_compile.rs
ZapAnton: The where some times, when rustc complained that it does not know the size of &Path at compile time, so it cannot be used

coriolinus: Are you sure? `Path` isn't sized, but `&Path` should be fine as a function argument.

util/checker/src/cmd/check_stubs_compile.rs
ZapAnton: Was just working on it. 
I only fear that my approach of using `map_err` to construct custom Error from the `io::Error` introduces too much boilerplate. Maybe there is a better way?

coriolinus: Well, that's where the `Error` trait would come in handy. Probably the simplest way to do this without external crates:

- [x] define your custom `Error` enum
- [ ] impl `From<E>` for each error type such as `io::Error` that you wish to wrap
- [ ] handle errors like `.map_err(|e| e.into())?`

The only way to get much shorter is to write a little macro.

ZapAnton: Sorry, should have specified the problem.

I have made a wrapper around `io::Error` like this: `IOCommandError(String, io::Error)`, where `String` is the failed command. 
So, in order to construct the above wrapper, I have to use code like that:

```rust
map_err(|io_error| {                                                                                                                     
            OSInteractionError::IOCommandError(                                                                                                   
                "git rev-parse --abbrev-ref HEAD".to_string(),                                                                                    
                io_error,                                                                                                                         
            )
```    
which, I fear, is repetitive and quite large when used a lot.

I guess I will just write a macro for it later.

coriolinus: Ah. Yeah, in that case, the macro is probably the right way to go. Might as well write it now, it'll prevent you from needing to write all those big `map_err` clauses manually.

ZapAnton:
Disregarding the macro problem, what do you feel should be done with this PR for now?

Should replacing the original `_test/ensure-stubs-compile.sh` with the new crate and adding building and checking scripts for the crate be enough?

Or do you want to see more functionality in the crate?

util/checker/src/cmd/check_stubs_compile.rs
coriolinus: You should look into the [`$crate`](https://doc.rust-lang.org/1.15.0/book/macros.html#the-variable-crate) pseudo-variable; it's meant to solve that problem. It's also weirdly-poorly documented these days; the language team is trying to get away from it.

coriolinus:
Before replacing anything, we need to benchmark. I'm willing to stipulate that this Rust code will be faster than the Bash equivalent. I'm less certain that the total runtime plus compile time will be less than the total Bash execution time. 

Travis execution runtime is a constant pain point in this repository, so if the total execution time increases by running this crate, I'd be disinclined to actually replace the scripts. 

[Hyperfine](https://github.com/sharkdp/hyperfine) output on a local machine showing `cargo run` timing with an empty `target` directory would be convincing, particularly if you also included hyperfine output showing the runtime of the equivalent bash scripts. Even better than that would be adjusting the travis config so we could see the timing on Travis' servers, but that's kind of a pain.

ZapAnton:
I did a first series of benchmarks. 

I decided to check worst-case scenario - when every existing exercise is modified 

For this I have used the following script:

```shell
   1   │ #!/bin/bash
   2   │ 
   3   │ for dir in exercises/*; do
   4   │         (
   5   │                 cd "$dir"
   6   │ 
   7   │                 cargo clean
   8   │ 
   9   │                 echo "//TEST" >> src/lib.rs
  10   │         )
  11   │ done
```

The  `hyperfine -i '_test/ensure-stubs-compile.sh'` command yielded the following result:
```text
Benchmark #1: _test/ensure-stubs-compile.sh
  Time (mean ± σ):     83.472 s ± 13.595 s    [User: 96.237 s, System: 20.733 s]
  Range (min … max):   73.638 s … 120.630 s
```

The `hyperfine -i 'target/release/checker'` command yielded the following result (note that this was run in the `checker` directory, where `cargo clean` and `cargo build --release` were run. The build took about 3 seconds):

```text
Benchmark #1: target/release/checker
  Time (mean ± σ):      1.372 s ±  0.585 s    [User: 991.0 ms, System: 475.1 ms]
  Range (min … max):    1.137 s …  3.036 s
 
  Warning: The first benchmarking run for this command was significantly slower than 
the rest (3.036 s). 
This could be caused by (filesystem) caches that were not filled until after the first run. 
You should consider using the '--warmup' option to fill those caches before the actual benchmark.
Alternatively, use the '--prepare' option to clear the caches before each timing run.
```

The `checker`  benchmark does not look realistic to me. 
May be something wrong with running the `cargo` command in the crate

ZapAnton:
Tried the `hyperfine -i './checker' --prepare './append.sh'`, where `append.sh` is the above script. Got the following result:

```text
Benchmark #1: ./checker
  Time (mean ± σ):     870.0 ms ±  59.5 ms    [User: 487.2 ms, System: 366.9 ms]
  Range (min … max):   811.7 ms … 1001.3 ms
 
  Warning: Ignoring non-zero exit code.
```

coriolinus:
The purpose of requesting these benchmarks is to capture total time including compilation. That means I'd be interested in a run like this: `hyperfine -i 'cargo run' --prepare './append.sh' --prepare 'rm -rf target'`.

ZapAnton:
Result for `hyperfine -i 'cargo run --release' --prepare 'rm -rf target; ./append.sh'` command:

```text
Benchmark #1: cargo run --release
  Time (mean ± σ):      8.668 s ±  1.164 s    [User: 11.211 s, System: 1.083 s]
  Range (min … max):    7.199 s … 10.639 s
```

coriolinus:
Ok, so including compilation time, you're showing a 10x performance improvement by using Rust over Bash. This is surprising, to put it mildly: I'd been convinced that the bulk of the execution time was in actually checking every stub, not in the looping code.

This seems like it opens the way forward for further development, but let's ensure that everything is actually doing what's expected.

ZapAnton:
Final, more realistic benchmarks:

`hyperfine -i 'cargo run --release' --prepare 'rm -rf target; ./append.sh'`:

```text
Benchmark #1: cargo run --release
  Time (mean ± σ):     140.631 s ±  8.617 s    [User: 199.557 s, System: 25.767 s]
  Range (min … max):   126.729 s … 149.167 s
```

`hyperfine -i '_test/ensure-stubs-compile.sh' --prepare './append.sh'`:

```text
Benchmark #1: _test/ensure-stubs-compile.sh
  Time (mean ± σ):     126.906 s ± 11.586 s    [User: 180.146 s, System: 24.388 s]
  Range (min … max):   113.959 s … 146.967 s
```

ZapAnton:
Seems like there is no obvious advantages in terms of speed for the Rust version.

Two point in it's favor though:
- 80% of time there will be only 1 exercise changed and checked
- We could try to parallelize the Rust version

coriolinus:
Agree that as of right no there's no reason to use the Rust version. 

I'd be interested to see the benchmarks when only one exercise is changed; I actually kind of expect Rust to come out relatively worse, in that scenario. Right now using rust only adds about 10% to the runtime; in the case where only one exercise is checked, I'd expect compilation to represent a larger fraction of the total time. 

Parallelization is interesting, though even finding the number of CPUs requires an external crate. That will slow down compilation. Also, as you point out, the vast majority of the time, only a single crate will need to be checked; that will eliminate any benefit it would provide. 

If you want to keep working on this, by all means, feel free. However, for a resource as constrained as Travis is, I am not willing to merge a RIIR unless performance is, at least, equivalent. That may be an intractable problem in the case of these scripts.

ZapAnton:
Made benchmarks for a single modified exercise.

The following prepare script was used:

```shell
   1   │ #!/bin/bash
   2   │ 
   3   │ track_root="$(git rev-parse --show-toplevel)"
   4   │ 
   5   │ (
   6   │         cd $track_root/exercises/grep
   7   │ 
   8   │         cargo clean
   9   │ 
  10   │         echo "//TEST" >> src/lib.rs
  11   │ )
``` 

`hyperfine -i '_test/ensure-stubs-compile.sh' --prepare './apend_single.sh'`:

```text
Benchmark #1: _test/ensure-stubs-compile.sh
  Time (mean ± σ):     20.606 s ±  1.725 s    [User: 43.554 s, System: 2.117 s]
  Range (min … max):   19.955 s … 25.510 s
```
`hyperfine -i 'cargo run --release' --prepare 'rm -rf target; ./apend_single.sh'`:

```text
Benchmark #1: cargo run --release
  Time (mean ± σ):     23.038 s ±  0.131 s    [User: 49.646 s, System: 2.319 s]
  Range (min … max):   22.889 s … 23.297 s
```

Not that bad for the Rust version actually.

ZapAnton:
I also tried several some naive methods to reduce compile time, like using `lto`, disabling panics and debug symbols, using system allocator and running without `--release` flag, but got no significant results.
Also, if I understand [this post](https://blog.mozilla.org/nnethercote/2018/11/06/how-to-speed-up-the-rust-compiler-in-2018-nll-edition/) correctly, in the future versions there will be an optimized borrow-checker, so maybe we could win some compile time  in Rust 2018.

For now I guess there are two options to try to beat long compile times: refactor the source code, to reduce compile time (I am not sure, but the introduction of macros actually doubled the compile time), or try to tweak Travis in some ways, so he could used the cached version of the crate (don't know if possible).

If the above suggestions are not sound, then the PR could closed for now - there is nothing left to do in this direction.

A shame though. Spending half an hour deciphering Bash syntax or trying to guess what `grep -v '^//' $dir/src/lib.rs | grep '\S' > /dev/null` does makes you wish for ~a nuclear winter~ a RIIR

ZapAnton:
Wonder if the [sccache](https://github.com/mozilla/sccache) tool could be somehow integrated to cache the compiled version of the crate?

PR #707, opened 2018-11-08
 ensure-stubs-compile: Removed the redundant empty stub check

ZapAnton:
Since it is a track policy to not have empty stub files, the check is no longer needed


coriolinus:
Disagree: this check is the easiest way to enforce the policy.

PR #710, opened 2018-11-12
gigasecond:change chrono crate link

andy5995:
When I used https://docs.rs/chrono/0.4.0/chrono/, the paragraph under
Duration has 3 dead links. Suggest changing this link to
https://crates.io/crates/chrono where the links under the Duration
section work.

### DIFF
--- a/exercises/gigasecond/README.md
+++ b/exercises/gigasecond/README.md
@@ -4,7 +4,7 @@ Calculate the moment when someone has lived for 10^9 seconds.
 
 A gigasecond is 10^9 (1,000,000,000) seconds.
 
-If you're unsure what operations you can perform on `DateTime<Utc>` take a look at the [chrono crate](https://crates.io/crates/chrono) which is listed as a dependency in the `Cargo.toml` file for this exercise.
+If you're unsure what operations you can perform on `DateTime<Utc>` take a look at the [chrono crate](https://docs.rs/chrono) which is listed as a dependency in the `Cargo.toml` file for this exercise.
 
 
 ## Rust Installation
--- a/exercises/gigasecond/README.md
+++ b/exercises/gigasecond/README.md
@@ -4,7 +4,7 @@ Calculate the moment when someone has lived for 10^9 seconds.
 
 A gigasecond is 10^9 (1,000,000,000) seconds.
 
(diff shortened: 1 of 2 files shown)


exercises/gigasecond/README.md
```diff
@@ -4,7 +4,7 @@ Calculate the moment when someone has lived for 10^9 seconds.
 
 A gigasecond is 10^9 (1,000,000,000) seconds.
 
-If you're unsure what operations you can perform on `DateTime<Utc>` take a look at the [chrono crate](https://docs.rs/chrono/0.4.0/chrono/) which is listed as a dependency in the `Cargo.toml` file for this exercise.
+If you're unsure what operations you can perform on `DateTime<Utc>` take a look at the [chrono crate](https://crates.io/crates/chrono) which is listed as a dependency in the `Cargo.toml` file for this exercise.
```
coriolinus: I think that it's better to point directly at the documentation than to crates.io. The issue was that we'd previously been pointing to a version of the documentation that has since become outdated.

The suggested link will automatically redirect to the most current version of the documentation, which should keep links fresh. 
```suggestion
If you're unsure what operations you can perform on `DateTime<Utc>` take a look at the [chrono crate](https://docs.rs/chrono) which is listed as a dependency in the `Cargo.toml` file for this exercise.
```

andy5995:
image

Because of that error, after changing .hints, I manually updated the README.

ZapAnton:
Sorry, forgot to mention that.

In order for `configlet` to regenerate the READMEs, the copy of the [problem-specifications repo](https://github.com/exercism/problem-specifications) should be present in the file system, on the same level as this track repo.

So something like this should work:
```shell
cd ~/src
git clone https://github.com/exercism/problem-specifications
cd excercism-rust
bin/configlet generate . -o gigasecond
```
The above commands are not needed to pass the Travis check (it has already passed), but may be usefull in the future

andy5995:
>The above commands are not needed to pass the Travis check (it has already passed), but may be usefull in the future

ok, got it!

andy5995:
I'm having a heckuva time with this exercise. First the problem with the dead duration links, then the stated goal, which I found vague:

>Calculate the moment when someone has lived for 10^9 seconds.

But examining lib.rs and the tests cleared that up.

Now this: https://users.rust-lang.org/t/using-std-duration/22218

Anyway.. these comments are just feedback in case it's useful. I'll hopefully be able to finish this exercise after I get some more answers in my forum post.

PR #713, opened 2018-11-12
 pythagorean-triplet: Updated the test suite to the version 1.0.0

### LINKED ISSUE #712: Change pythagorean triplet function signature to remove confusion.
Right now Pythagorean triplet takes no arguments and expects students to encode in the function implementation assumptions about the solution they are trying to find, that the sum is 1000, and that it is a Pythagorean triplet. This is verified by a single test case that checks that the output is a specific number wrapped in an Option.

This specific problem doesn't justify the use of an Option type, students have to imagine that the problem could be changed and they could need to tweak their implementation such that it would have to search for a different sum that it could fail to find, which would justify returning None.

This leaves us in a situation where solutions that end in `unreachable!()` rather than returning `None` are technically valid, and students very frequently return `Some(0)` instead of None because there is never a negative test case, there cannot be with the current design.

My proposal is that we adjust the parameters so that the sum is passed as an argument. The documentation is updated to say, for example, there is one valid Pythagorean triplet that has a  perimeter of `1000`, rather than focusing the exercise on this specific Pythagorean triplet. And test cases are added for cases where the solution SHOULD fail to find a triplet.

Side issue: I'm not sure if there are any perimeter values at which there could be multiple pythagorean triplets, I'm guessing no. Also I expect there are certain problematic pythagorean triplets out there that might cause some solutions, such as those that rely on floating point math, to fail. Also I expect there are some that might look like valid triangles to incomplete solutions but actually be incorrect, particularly solutions where students use signed integers and don't correctly bound their iteration. I plan on setting up some tests to iterate over a large range of sums attempt to find problematic cases that might catch more implementation errors that are currently sneaking into review.

ZapAnton:
Updates the ` pythagorean-triplet` according to the latest canonical-data.

The canonical-data introduced the `sum` input parameter, therefore the stub template is updated alongside with the test suite and the example solution.

Closes #712

### DIFF
--- a/exercises/pythagorean-triplet/example.rs
+++ b/exercises/pythagorean-triplet/example.rs
@@ -1,11 +1,19 @@
-pub fn find() -> Option<u32> {
-    for a in 1..1000 {
-        for b in (a + 1)..(1000 - a) {
-            let c = 1000 - (a + b);
+pub fn find(sum: u32) -> Option<Vec<[u32; 3]>> {
+    let mut triplets = vec![];
+
+    for a in 1..sum {
+        for b in (a + 1)..(sum - a) {
+            let c = sum - (a + b);
+
             if a * a + b * b == c * c {
-                return Some(a * b * c);
+                triplets.push([a, b, c]);
             }
         }
     }
-    None
+
+    if triplets.is_empty() {
+        None
+    } else {
+        Some(triplets)
+    }
 }
--- a/exercises/pythagorean-triplet/example.rs
+++ b/exercises/pythagorean-triplet/example.rs
@@ -1,12 +1,14 @@
-pub fn find(sum: u32) -> Vec<[u32; 3]> {
-    let mut triplets = vec![];
+use std::collections::HashSet;
+
+pub fn find(sum: u32) -> HashSet<[u32; 3]> {
+    let mut triplets = HashSet::new();
 
     for a in 1..sum {
         for b in (a + 1)..(sum - a) {
             let c = sum - (a + b);
 
             if a * a + b * b == c * c {
-                triplets.push([a, b, c]);
+                triplets.insert([a, b, c]);
             }
         }
     }
--- a/exercises/pythagorean-triplet/example.rs
+++ b/exercises/pythagorean-triplet/example.rs
@@ -1,4 +1,4 @@
-pub fn find(sum: u32) -> Option<Vec<[u32; 3]>> {
+pub fn find(sum: u32) -> Vec<[u32; 3]> {
     let mut triplets = vec![];
 
     for a in 1..sum {
@@ -11,9 +11,5 @@ pub fn find(sum: u32) -> Option<Vec<[u32; 3]>> {
         }
     }
 
-    if triplets.is_empty() {
-        None
-    } else {
-        Some(triplets)
-    }
+    triplets
 }
--- a/exercises/pythagorean-triplet/tests/pythagorean-triplet.rs
+++ b/exercises/pythagorean-triplet/tests/pythagorean-triplet.rs
@@ -2,47 +2,47 @@ extern crate pythagorean_triplet;
 
 use pythagorean_triplet::find;
 
-fn process_tripletswithsum_case(sum: u32, expected: Option<Vec<[u32; 3]>>) {
-    if let Some(expected_triplets) = expected {
-        let triplets = find(sum).unwrap();
+fn process_tripletswithsum_case(sum: u32, expected: Vec<[u32; 3]>) {
+    if !expected.is_empty() {
+        let triplets = find(sum);
 
         assert!(
-            expected_triplets
+            expected
                 .iter()
                 .all(|expected_triplet| triplets.contains(expected_triplet))
         );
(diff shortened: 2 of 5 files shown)


ZapAnton:
I may be reading the Travis logs wrong, but for some reason the exercise is checked four times. Should probably investigate this.

exercises/pythagorean-triplet/example.rs
```diff
@@ -1,11 +1,19 @@
-pub fn find() -> Option<u32> {
-    for a in 1..1000 {
-        for b in (a + 1)..(1000 - a) {
-            let c = 1000 - (a + b);
+pub fn find(sum: u32) -> Option<Vec<[u32; 3]>> {
```
petertseng: I strongly believe Option is not necessary. If it's empty, empty vector is a perfectly reasonable return value and is not an error (such that we might use Option for)

petertseng: If this suggestion to remove Option is accepted, don't forget to also remove `option` from `pythagorean-triplets` config.json

ZapAnton: I was under the impression that using empty Vec falls under 'assigning special meaning to the basic types' (like in this [comment](https://github.com/exercism/rust/pull/548#issuecomment-398094945)).

Besides, isn't Option used to convey the absence of a value, not only for errors? Would be strange not to make use of a special type in this case.

Also, Vec could be changed for some other type in the future (like HashSet or some custom struct), but the Option API would remain

petertseng: > I was under the impression that using empty Vec falls under 'assigning special meaning to the basic types'

I'm not under this impression. The empty vector is a valid value that inhabits the `Vec<[u32; 3]>` type, just as any other value of that type does. It is not special in any way.

> Besides, isn't Option used to convey the absence of a value

Sure.

> Would be strange not to make use of a special type in this case.

It is strange to not be using a valid value of the type `Vec<[u32; 3]>`. Under no circumstance will this code ever return `Some(vec![])`, which means using an `Option` is a waste.

> Also, Vec could be changed for some other type in the future (like HashSet or some custom struct), but the Option API would remain

True, but we are not talking about the future, we are talking about this PR as it is proposed. Since `Vec<_>` is inhabitable by an empty vector, we should make use of that fact.

petertseng: > It is not special in any way.

Oops, forgot to mention why that makes it different from `""`. In `two-fer`, we assigned special meaning to `""` by saying that `""` must be special-cased by the implemented function (the user of the passed-in argument) in such a way that `""` is equivalent to `"you"` rather than itself.

In this exercise, one who wishes to examine the empty vector need not do any special casing. Operations such as `len` and iterating over it all act as they should. Therefore, no special meaning is needed and none should be assigned.

ZapAnton: Got it. Thanks for the explanation

coriolinus:
Approve of the work done so far, but a little bit remains:

- remove "option" from exercise topics in `config.json` at https://github.com/exercism/rust/blob/4704fa4e1829ef3249820b0ccaa8b15ddcdbabff/config.json#L154

I wanted to add "update README" to that list, but apparently the problem-specifications README is also out of date and misleading.

petertseng:
my comment appears to have disappeared. one second while I remake it.

exercises/pythagorean-triplet/tests/pythagorean-triplet.rs
```diff
@@ -1,6 +1,79 @@
...
+use pythagorean_triplet::find;
+
+fn process_tripletswithsum_case(sum: u32, expected: Vec<[u32; 3]>) {
+    if !expected.is_empty() {
+        let triplets = find(sum);
+
+        assert!(
+            expected
+                .iter()
+                .all(|expected_triplet| triplets.contains(expected_triplet))
```
petertseng: This line IS necessary, but I think another check is necessary. Otherwise, an impl that adds `[0, 0, 0]` to any non-empty vec will also pass the tests, since it doesn't check that every element of `triplets` should be there.

Try some of the ways used before. Perhaps check the `len` as in https://github.com/exercism/rust/blob/master/exercises/allergies/tests/allergies.rs#L5-L21 . Or use a HashSet as in https://github.com/exercism/rust/blob/master/exercises/anagram/tests/anagram.rs#L6-L12

petertseng: or sorting the input, as in https://github.com/exercism/rust/blob/master/exercises/saddle-points/tests/saddle-points.rs#L6-L10

There appear to be many exercises where we use an ordered data structure but don't actually care about the order, and we solved it in a different way each time, that is quite interesting.

ZapAnton: If the `HashSet` approach is used, should the stub function also return `HashSet<String>`?

petertseng: It must, otherwise it would not compile

PR #717, opened 2018-11-13
leap:add a hint

andy5995:
This struck me as a good hint to add. I've used C but couldn't remember
how to test for divisibity because I've never done much with math
functions.


ZapAnton:
For more clarity, perhaps a [link](https://doc.rust-lang.org/std/ops/trait.Rem.html) for arithmetic remainder should be provided? (just my suggestion, you are free to not add this, if the maintainers will agree with your version)

andy5995:
>I was a little surprised to find % referred to as the "arithmetic remainder operator"; I was taught it as the "modulo operator", which is a little faster to say.

[This thread](https://users.rust-lang.org/t/modulus-operator/22224) may be of interest to you.

PR #720, opened 2018-11-13
exercise crate: Made configure subcommand handle null 'unlocked_by' field

### LINKED ISSUE #719: exercise util: Make the val_as! macro accept null as valid input
Discovered the issue when tried to use the configure subcommand on the `pythagorean-triplet` exercise.

The `unlocked_by` field is set to `null` in the `config.json`, which makes this line to return error:

```rust
get!(existing_config, "unlocked_by", as_str)
```
The error is returned, because `as_str` cannot parse `null` and instead invokes the `.ok_or_else` function.

Also the error message could be better, when `val_as!` is called from `get!`. The current message is not very helpful:
```text
Error: SchemaTypeError { file: "config.json", field: "$ name", as_type: "str" }
```

ZapAnton:
This PR adds the handling of the `null` field, read both from `config.json` and user input, by the `configure` subcommand.

Closes #719

### DIFF
--- a/util/exercise/src/cmd/configure.rs
+++ b/util/exercise/src/cmd/configure.rs
@@ -34,7 +34,7 @@ fn get_user_config(exercise_name: &str, config_content: &Value) -> Result<Value>
             if let Value::String(s) = get!(existing_config, "unlocked_by") {
                 s
             } else {
-                "none"
+                "null"
             }
         } else {
             "hello-world"
--- a/util/exercise/src/cmd/configure.rs
+++ b/util/exercise/src/cmd/configure.rs
@@ -29,8 +29,6 @@ fn get_user_config(exercise_name: &str, config_content: &Value) -> Result<Value>
         Uuid::new_v4().to_hyphenated().to_string()
     };
 
-    let core = false;
-
     let unlocked_by: Option<String> = loop {
         let default_value = if let Some(existing_config) = existing_config {
             get!(existing_config, "unlocked_by")
@@ -52,7 +50,7 @@ fn get_user_config(exercise_name: &str, config_content: &Value) -> Result<Value>
         if user_input.is_empty() {
             break default_value;
         } else if user_input == "null" {
-            break None::<String>
+            break None::<String>;
         } else if !get!(config_content, "exercises", as_array)
             .iter()
             .any(|exercise| exercise["slug"] == user_input)
@@ -64,6 +62,8 @@ fn get_user_config(exercise_name: &str, config_content: &Value) -> Result<Value>
             break Some(user_input);
         };
     };
+
+    let core = unlocked_by.is_none();
 
     let difficulty = loop {
         let unlocked_by_difficulty = match unlocked_by {
--- a/util/exercise/src/cmd/configure.rs
+++ b/util/exercise/src/cmd/configure.rs
@@ -31,20 +31,28 @@ fn get_user_config(exercise_name: &str, config_content: &Value) -> Result<Value>
 
     let core = false;
 
-    let unlocked_by = loop {
+    let unlocked_by: Option<String> = loop {
         let default_value = if let Some(existing_config) = existing_config {
-            get!(existing_config, "unlocked_by", as_str)
+            get!(existing_config, "unlocked_by")
+                .as_str()
+                .map(|s| s.to_string())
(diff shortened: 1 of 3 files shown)


util/exercise/src/cmd/configure.rs
```diff
@@ -29,22 +29,28 @@ fn get_user_config(exercise_name: &str, config_content: &Value) -> Result<Value>
...
 
-    let core = false;
-
-    let unlocked_by = loop {
+    let unlocked_by: Option<String> = loop {
         let default_value = if let Some(existing_config) = existing_config {
-            get!(existing_config, "unlocked_by", as_str)
+            get!(existing_config, "unlocked_by")
+                .as_str()
+                .map(|s| s.to_string())
```
coriolinus: Instead of converting `default_value` to `Option<String>`, which requires a bunch of expansions below to do the right thing, what about restructuring the code like this:

```rust
let default_value = if let Some(existing_config) = existing_config {
    if let Some(Value::String(ref s)) = get!(existing_config, "unlocked_by") {
        s
    } else {
        "none"
    }
} else {
    "hello-world"
}
```

I haven't run that snippet past the compiler, so there may need to be some small changes, but I think what you end up with will be more readable and more usable than keeping this current setup.

util/exercise/src/cmd/configure.rs
```diff
@@ -53,19 +59,28 @@ fn get_user_config(exercise_name: &str, config_content: &Value) -> Result<Value>
...
-            .ok_or(format_err!(
-                "exercise '{}' not found in config",
-                unlocked_by
-            ))?;
-        let unlocked_by_difficulty = get!(unlocked_by, "difficulty", as_u64);
+        let unlocked_by_difficulty = match unlocked_by {
+            Some(ref unlocked_by) => {
+                let unlocked_by_exercise = get!(config_content, "exercises", as_array)
+                    .iter()
+                    .find(|exercise| exercise["slug"] == unlocked_by.as_str())
```
coriolinus: We search for an exercise several times in this codebase. Do you think it might be worth extracting a function like 

```rust
fn find_exercise<F>(config_content: Value, finder: F) -> Option<Value> where F: FnOnce(Value) -> bool {
    get!(config_content, "exercises", as_array)
                    .iter()
                    .find(finder)
}
```

Then this expression, and similar ones like it elsewhere, could be compacted into one line:

```rust
let unlocked_by_exercise = find_exercise(config_content, |exercise| exercise["slug"] == unlocked_by.as_str())
```

util/exercise/src/cmd/configure.rs
```diff
@@ -29,22 +29,28 @@ fn get_user_config(exercise_name: &str, config_content: &Value) -> Result<Value>
...
-    let unlocked_by = loop {
+    let unlocked_by: Option<String> = loop {
         let default_value = if let Some(existing_config) = existing_config {
-            get!(existing_config, "unlocked_by", as_str)
+            get!(existing_config, "unlocked_by")
+                .as_str()
+                .map(|s| s.to_string())
         } else {
-            "hello-world"
+            Some("hello-world".to_string())
```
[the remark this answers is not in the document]
ZapAnton: Simply removing the two `to_string()` would not work - the `unlocked_by` variable expects `Option<String>`. If I change it to the `Option<&str>` (along with some additional changes), I am blocked by the line 62 `break Some(&user_input);`, since `user_input` does not live long enough to make a borrow (does not leave outside the loop). 
Maybe you have some sort of suggestion for the problem? I have met a lot of moments, where I need an additional allocation, just so that the returned value is `String` and not `&str`, and I feel like this approach is not very idiomatic (given how many crates claim to be zero-copy)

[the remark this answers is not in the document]
coriolinus: https://github.com/ZapAnton/rust/pull/1: if you merge it, it will update this PR

util/exercise/src/cmd/configure.rs
coriolinus: I tried it and it added more code than it saved, so I'm reverting this suggestion.

ZapAnton: Just tried it myself - same result

ZapAnton:
All done

PR #721, opened 2018-11-13
 exercise crate: Added update-readme flag for the update subcommand

ZapAnton:
Added the optional `update-readme` / `-r` flag for the `update` subcommand. 

Decided to turn this into a flag and not into an unconditional functionality, because users may not always want to update the README (maybe their `problem-specifications` is outdated)


ZapAnton:
I was trying to use the 'explicit is better than implicit' approach here - I would not want to update the README if I was not sure that `problem-specifications` is up-to-date. But since such cases would be caught by Travis your way is more practical.

If this is OK, the flag would be renamed into `r` / `dont-update-readme`

ZapAnton:
Done

PR #724, opened 2018-11-14
saddle-points: Add test for all saddle points

mtodor:
It would be nice to cover case, when all points are actually saddle points.


petertseng:
yeah that's interesting. I even think it's worth submitting to https://github.com/exercism/problem-specifications/blob/master/exercises/saddle-points/canonical-data.json .

PR #727, opened 2018-11-17
do not merge: see what happens if we force checking everything

petertseng:


petertseng:
That means that the weekly cron is not running anything. If it were, it would have the same result.

PR #728, opened 2018-11-17
clock: Add test for exactly -60 minutes.

tmccombs:
I ran into a solution that would give the wrong solution if the minutes were exactly a multiple of 60 and negative.

### DIFF
--- a/exercises/clock/Cargo.toml
+++ b/exercises/clock/Cargo.toml
@@ -1,5 +1,5 @@
 [package]
 name = "clock"
-version = "2.3.0"
+version = "2.4.0"
 
 [dependencies]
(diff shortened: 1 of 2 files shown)


tmccombs:
will do

tmccombs:
Corresponding PR for problem specification: https://github.com/exercism/problem-specifications/pull/1400

petertseng:
Would you be so kind as to change https://github.com/exercism/rust/blob/86e22ca550f897d33016107c7d95d03279827a71/exercises/clock/Cargo.toml#L3 to 2.4.0 now that it got assigned a version in problem-specifications?

Optional: Note that this test was placed in a different position than in problem-specifications. I'm assuming this current order was OK, but I'd also like to throw out there the possibility of putting it in the same place as it was in problem-specifications. That decision can just be based on whether either position is a more logical progression of tests; no obligation to use the same order as that in problem-specifications, though it makes things a bit easier

PR #731, opened 2018-11-17
More clippy (examples only)

petertseng:


petertseng:
These all affect examples only, so review is not considered urgent (not student-facing).

PR #732, opened 2018-11-17
pythagorean-triplet example performance

petertseng:
Not student-facing, prioritise accordingly lower as necessary.

It's possible to get some improvements in pythagorean-triplet performance, which I have found helpful when testing locally.

Two commits are included, with each successive commit going farther down the "more performant but harder to understand" path. Reviewers should decide which point on the path is an appropriate point for Travis and locally running.

(Though I've implicitly made my position clear by having submitted this PR)


petertseng:
"all the time" isn't quite true with the advent of https://github.com/exercism/rust/pull/688, but is still somewhat important

It's also true that each individual exercise is only a small portion of the total, but I suppose it's good to take all the wins we can; the small wins could add up.

PR #736, opened 2018-11-18
wordy 1.4.0

petertseng:


exercises/wordy/example.rs
```diff
@@ -1,65 +1,56 @@
...
+    let words = c.trim_end_matches('?').split_whitespace().map(|word| {
+        if let Ok(i) = word.parse::<i32>() {
+            Token::Number(i)
+        } else {
+            Token::NonNumber(word)
         }
-        result = evaluate(result, opr, operand(&t.remove(0)));
-        Some(result)
+    }).collect::<Vec<_>>();
+    if words.len() < 3 {
```
petertseng: there is no test case that exercises this condition. I am considering `What is?` as wordy 1.5.0

exercises/wordy/example.rs
```diff
@@ -1,65 +1,56 @@
...
+        [
+            Token::NonNumber("What"),
+            Token::NonNumber("is"),
+            Token::Number(i),
+        ] => i,
+        _ => return None,
+    };
+    let mut words = words.split_at(3).1;
+    while !words.is_empty() {
+        let tmp = apply_op(result, words)?;
```
petertseng: I would have liked to write `(result, words) = apply_op(result, words)?;` or something, but that is not valid. A shame.

exercises/wordy/example.rs
```diff
@@ -1,65 +1,56 @@
...
-    let mut result: i32 = 0;
-    let mut opr = "plus";
-
-    if t.len() <= 1 {
-        None
-    } else {
-        while t.len() > 1 {
-            result = evaluate(result, opr, operand(&t.remove(0)));
-            opr = operator(&t.remove(0));
+    if !c.ends_with('?') {
```
petertseng: no test case tests this line of code

exercises/wordy/example.rs
```diff
@@ -1,65 +1,56 @@
...
+        return None;
+    }
+    let words = c.trim_end_matches('?').split_whitespace().map(|word| {
```
petertseng: note that if there were a test case along the lines of `What is 1 + 2???` then that might raise questions as to whether that case is acceptable. This code would accept it.

exercises/wordy/example.rs
```diff
@@ -1,65 +1,56 @@
...
+        if let Ok(i) = word.parse::<i32>() {
+            Token::Number(i)
+        } else {
+            Token::NonNumber(word)
         }
-        result = evaluate(result, opr, operand(&t.remove(0)));
-        Some(result)
+    }).collect::<Vec<_>>();
+    if words.len() < 3 {
+        return None;
```
petertseng: I wish I didn't have to do this separately, but `words[0..3]` will panic if `words.len() < 3`. Maybe something with `iter().take(3)`, but then I hae to `collect` as well and it doesn't seem worth it.

exercises/wordy/example.rs
coriolinus: I see no reason to keep it, then. This implements a parser stricter than the tests and readme require.

exercises/wordy/example.rs
petertseng: that was pointed out in https://github.com/exercism/rust/pull/665#discussion_r222796492 . I wonder if I should also consider using that function that takes `&mut`. Note though that there still remains the possibility that the function will cause an early `return None` (currently done by the `?`), so there's a bit more to it

exercises/wordy/example.rs
coriolinus: I think `words[..3]` might not panic, being an unbounded range. Might be worth looking into, at least.

exercises/wordy/example.rs
coriolinus: Yeah, destructuring assignment would be a really nice usability improvement for the language.

exercises/wordy/example.rs
petertseng: Interesting. Looks like it still does though. `thread 'reject_empty' panicked at 'index 3 out of range for slice of length 0', libcore/slice/mod.rs:1932:5`

PR #740, opened 2018-11-18
bracket-push 1.5.0

petertseng:
Note that I decided to keep original authors, because all of the value being added in the upstream commits was in saying what the test case was, so the fact that I happened to be the person performing the translation to a given track doesn't mean that I have added any value here

### DIFF
--- a/exercises/bracket-push/Cargo.toml
+++ b/exercises/bracket-push/Cargo.toml
@@ -1,3 +1,3 @@
 [package]
 name = "bracket-push"
-version = "1.4.0"
+version = "1.5.0"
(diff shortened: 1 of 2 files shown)


coriolinus:
For what it's worth, I disagree: you have added value by 

1. Searching for exercises with updates upstream
1. Performing the work of updating these exercises
1. Creating a pull request

This work is not negligible; you would be well within your rights to claim authorship of these commits.

petertseng:
That's fair. I'm content because `git` will still recognise me as the Committer (though `git` only shows that with `--pretty=fuller` or similar). I suppose it may lead to some interesting representations in https://github.com/exercism/rust/graphs/contributors , but I don't make any decisions based on anything on that page anyway.

PR #741, opened 2018-11-18
declare compliance with problem-specifications where doing so is a no-op

petertseng:


petertseng:
miseed queen-attack. doing that too.

petertseng:
add bowling to the list of ones I wanted to do. two-bucket and alphametics are maybes

after 48 hours, assume all bets are off and make no assumptions that I will do anything more

PR #750, opened 2018-11-19
forth 1.5.0

petertseng:
Note that 1.6.0 and 1.7.0 are out, but require some changes to the example solution that I've not made yet. I might make them at some point, but no guarantees.


petertseng:
I'm going to have to put Forth 1.6.0 in its own PR because there are changes to the example.rs that I'd have to discuss. So this PR will only go as far as 1.5.0

PR #752, opened 2018-11-19
 grep: Updated the exercise to 1.2.0 version

ZapAnton:
Adds 4 new test cases plus clippy suggestions.

Relevant PR: https://github.com/exercism/problem-specifications/pull/1356

### DIFF
--- a/exercises/grep/Cargo.toml
+++ b/exercises/grep/Cargo.toml
@@ -1,6 +1,6 @@
-[package]
-name = "grep"
-version = "1.1.0"
-
 [dependencies]
 failure = "0.1.1"
+
+[package]
+name = "grep"
+version = "1.2.0"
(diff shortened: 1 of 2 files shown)


petertseng:
For PRs that follow problem-specifications, I would appreciate if you would link to the relevant problem-specifications PRs

ZapAnton:
Could you elaborate a little in what form do you want a link - in the comment, in the commit message or merging the original commit like in 19b185a?

petertseng:
Personally I think it would be good if they would be put in the commit message, because I think being able to find the source of something seems like a useful thing to keep in git history.

As for what form I'd like them, full URL (`https://github.com/exercism/problem-specifications/pull/1234`) seems fine, but I wonder if you have other ideas

ZapAnton:
No other ideas, although there could be a problem with full URL, in case the linked repo is deprecated/renamed, but such case is exotic and is easily solved by the person, who looks up the link

coriolinus:
I'm going to give this a day to give other maintainers the chance to chime in, but if nobody objects, I'm going to merge this tomorrow.

PR #753, opened 2018-11-19
 leap: Updated exercise to 1.4.0 version

ZapAnton:
This PR syncs the `leap` exercise with the `problem-specifications` repo.

It introduces the following changes:

- Adds all the cases from canonical data. The cases that were present in the suite were not present in the canonical version, so I added new cases in a `test_description_in_lowercase` format. The old cases were not touched (except the one that checked the 1996 year - it was present in canonical data, so I replaced it with the formatted case). Perhaps some old test cases could be deleted?

- Replaced `i32` argument type with `u64` - no negative years are expected, so I thought it would make sense

Relevent PR: https://github.com/exercism/problem-specifications/pull/1359

### DIFF
--- a/exercises/leap/Cargo.toml
+++ b/exercises/leap/Cargo.toml
@@ -1,3 +1,3 @@
 [package]
 name = "leap"
-version = "1.0.0"
+version = "1.4.0"
--- a/exercises/leap/example.rs
+++ b/exercises/leap/example.rs
@@ -1,4 +1,4 @@
-pub fn is_leap_year(year: i32) -> bool {
+pub fn is_leap_year(year: u64) -> bool {
     let has_factor = |n| year % n == 0;
     has_factor(4) && (!has_factor(100) || has_factor(400))
 }
--- a/exercises/leap/src/lib.rs
+++ b/exercises/leap/src/lib.rs
@@ -1,3 +1,3 @@
-pub fn is_leap_year(year: i32) -> bool {
+pub fn is_leap_year(year: u64) -> bool {
     unimplemented!("true if {} is a leap year", year)
 }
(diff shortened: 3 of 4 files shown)


petertseng:
For this update to `leap`, note that there are multiple PRs that fall within the definition of "relevant PR" since this one spans multiple versions 1.0.0 through 1.4.0. I usuallly would use something like https://github.com/exercism/problem-specifications/pulls?q=is%3Apr+leap+is%3Aclosed to find them, or `git log exercises/leap/canonical-data.json`. In this case:

* `https://github.com/exercism/problem-specifications/pull/1112` https://github.com/exercism/problem-specifications/pull/1112
* `https://github.com/exercism/problem-specifications/pull/971` https://github.com/exercism/problem-specifications/pull/971
* `https://github.com/exercism/problem-specifications/pull/955` https://github.com/exercism/problem-specifications/pull/955

ZapAnton:
Still not sure about the old, track-specific tests. Will they not overload the students?

coriolinus:
In this respect my approach differs from Exercism's official policy: I personally always immediately remove all `#[ignore]` tags before beginning a new exercise, and write code which solves all tests simultaneously. It produces the same result as approaching the problem incrementally, while ensuring that I don't start heading down architectural routes which turn out not to work with the requirements of later tests.

In a professional context, assuming a well-managed project, coders deal with test suites comprising thousands of tests. If a dozen fail because of a change you made, you just have to deal with it. I therefore believe that we should encourage students not to be afraid of a bunch of test failures, but to embrace them: each one tells you about something your code needs to do. 

To answer your question for myself, then, I'd say no, students won't be overwhelmed: correct code will pass all tests anyway; incorrect code can always use more examples of what outputs should correspond to what inputs. 

To answer your question speaking ex officio as the track maintainer, I'd say that by maintaining a track policy of preferring to add tests without removing them, we are simultaneously following good project management practices and encouraging students to view tests as a bulk product for code improvement, not as individual hurdles which must be cleared.

PR #754, opened 2018-11-20
Fix pythagorean-triplet readme, in accordance to the updated tests

fudanchii:
This is as described in https://github.com/exercism/problem-specifications/pull/1395

The current README will really confuse non math-savvy students, as it asked to find `a*b*c`, while the tests actually check for the triplets.

Change it to asks for the triplet explicitly. Give the answer for N = 1000 as it's also spoiled in the test file ;)


fudanchii:
woops, didn't know that it checks for problem-specifications. closing this and will wait for https://github.com/exercism/problem-specifications/pull/1395 instead 🙏

PR #761, opened 2018-11-20
reverse-string: Updated the exercise to the 1.1.0 version

ZapAnton:
Relevant PR:
- https://github.com/exercism/problem-specifications/pull/1162

No new test cases, just renamed the old case and removed the comments
generated by the `bin/init_exercise.py` script


coriolinus:
Why did you remove the doccomments? I can understand removing the "example implementation goes here" stuff, but the rest of them seemed still to be relevant.

ZapAnton:
```rust
//! Tests for reverse-string
//!
//! Generated by [script][script] using [canonical data][canonical-data]
//!
//! [script]: https://github.com/exercism/rust/blob/master/bin/init_exercise.py
//! [canonical-data]: https://raw.githubusercontent.com/exercism/problem-specifications/master/exercises/reverse-string/canonical_data.json
```
The script link does not exist and I thought replacing it with the `exercise` crate would be misleading.

The comments above the test cases just repeat the name of the cases, therefore relevant only if the documentation is generated 

`/// Process a single test case for the property reverse` could be returned I guess

What would you like me to do?

ZapAnton:
Done

PR #762, opened 2018-11-20
 hamming: Updated the exercise to the 2.1.1 version

ZapAnton:
Relevant PRs:
    - https://github.com/exercism/problem-specifications/issues/625
    - https://github.com/exercism/problem-specifications/pull/844
    - https://github.com/exercism/problem-specifications/pull/845
    - https://github.com/exercism/problem-specifications/pull/875
    - https://github.com/exercism/problem-specifications/pull/953
    - https://github.com/exercism/problem-specifications/pull/1129
    - https://github.com/exercism/problem-specifications/pull/1389

Basically adds every test case from the current canonical-data.json  without deleting the old ones.

### DIFF
--- a/exercises/hamming/tests/hamming.rs
+++ b/exercises/hamming/tests/hamming.rs
@@ -21,31 +21,31 @@ fn test_no_difference_between_identical_strands() {
 #[test]
 #[ignore]
 fn test_complete_hamming_distance_in_small_strand() {
-    assert_eq!(hamming::hamming_distance("ACT", "GGA"), Some(3));
+    process_distance_case(["ACT", "GGA"], Some(3));
 }
 
 #[test]
 #[ignore]
 fn test_small_hamming_distance_in_the_middle_somewhere() {
-    assert_eq!(hamming::hamming_distance("GGACG", "GGTCG"), Some(1));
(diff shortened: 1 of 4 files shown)


exercises/hamming/tests/hamming.rs
```diff
@@ -1,14 +1,21 @@
 extern crate hamming;
 
+fn process_distance_case(strand_pair: [&str; 2], expected_distance: Option<usize>) {
```
coriolinus: I'm curious why you wrote this as `[&str; 2]` instead of `(&str, &str)`. I tend to think of function arguments as tuple-like instead of array-like, so the latter would have seemed more natural to me.

I suspect that the two forms are equivalent in memory, and this does work; you certainly don't need to change it. I'm just curious about your design process.

ZapAnton: Guess I decided to go with the array, because the types of the variables are similar (perhaps a little C-like design).

If the strands where represented with different types, e.g. `LeftStrand` / `RightStrand`, then I would surely used the tuple.

petertseng:
Since this also updates the README, https://github.com/exercism/problem-specifications/pull/1360 is now also relevant.

PR #768, opened 2018-12-09
exercise README: Added the sections about 'rustfmt' and 'clippy'

ZapAnton:
As of Rust `1.30` and `1.31` the `rustfmt` and `clippy` tools have been standardized. 

I thought this is a good opportunity to expose more students to these tools, by adding the corresponding sections in the README.

Did not regenerate the actual READMEs to not clutter the PR, and to provide readability for the maintainers. If the changes are approved, the regeneration would be the closing commit.

### DIFF
--- a/config/exercise_readme.go.tmpl
+++ b/config/exercise_readme.go.tmpl
@@ -46,52 +46,17 @@ To learn more about Rust tests refer to the [online test documentation][rust-tes
 Make sure to read the [Modules](https://doc.rust-lang.org/book/ch07-02-modules-and-use-to-control-scope-and-privacy.html) chapter if you
 haven't already, it will help you with organizing your files.
 
-## Writing well-formatted code
+## Further improvements
 
-When you are solving the exercise, you are free to choose any coding format you want.
-However when you are writing a real-world application or a library, your source code will
-be read by other people, not just you. To solve a problem when different people choose
-different formats for a single project, the developers set a standard coding format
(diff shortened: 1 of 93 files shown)


coriolinus:
Agree--I think this documentation is good, but it's better suited for a single central location in the track repo than in every exercise's README.

ZapAnton:
Will do.
The only caveat I have is that it would be quite easy to skip additional sections, once the Rust installation is complete. Would you mind if I moved the current sections to the `INSTALLATION.md` and left a small note in the exercise README (something like 'You can use this command to run `clippy`/`rustfmt`. If you haven't installed it, here is the link to the `INSTALLATION.md`')?

coriolinus:
Sure, that seems reasonable.

ZapAnton:
Done.

For some reason Travis build hasn't failed, although the READMEs were not regenerated.
Perhaps `_test/ensure-readmes-are-updated.sh` could be adjusted to catch such cases?

ZapAnton:
Since there was no more comments about the contents of the additional section,
I went ahead and regenerated the READMEs (also resolved the merge conflict).

ZapAnton:
The Travis check fails, because of the two reasons:

- `paasio` tests do not pass
- READMEs for the `difference-of-squares` and the `grains` are out-of-date

Both of these reasons are not related to the PR, so maybe they could be ignored?

PR #770, opened 2018-12-09
Fixed the links to the Rust Book

ZapAnton:
The current links, that contain '2018-edition' in them, would lead to the page with the following message:

```text
The 2018 edition of the book is no longer distributed with Rust's documentation.

If you came here via a link or web search, you may want to check out the current version of the book instead.

If you have an internet connection, you can find a copy distributed with Rust 1.30.
```

Plus the `Modules` chapter changed its URL.

This PR fixes the above problem with links

### DIFF
--- a/config/exercise_readme.go.tmpl
+++ b/config/exercise_readme.go.tmpl
@@ -43,7 +43,7 @@ $ cargo test some_test -- --ignored
 
 To learn more about Rust tests refer to the [online test documentation][rust-tests]
 
-Make sure to read the [Modules](https://doc.rust-lang.org/book/2018-edition/ch07-00-modules.html) chapter if you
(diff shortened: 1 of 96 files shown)


ZapAnton:
> I was planning to Approve this PR when I got a chance to see that:
> new links in template and docs/ are suitable replacements for what they replaced
> (CI will check the generated result)

Sorry, @petertseng, I could not understand the wording in this one. Do you suggest to leave the changes  only in template/docs and to remove the regeneration of the README for existing exercises?

petertseng:
> Do you suggest to leave the changes only in template/docs and to remove the regeneration of the README for the existing exercises?

Hmm, I will make a suggestion when I have a chance, but I predict that when I do, it will be to leave everything as is.

PR #769, opened 2018-12-09
track: Migrated the existing exercises to the Rust 2018 edition

ZapAnton:
This PR shows what simple migration to the 2018 edition would look, when `cargo fix` is applied to every exercise on the track.

Since the changes were nor discussed nor approved by the maintainers , this PR serves as a demonstration of the migration, and should be closed, if the changes are not desirable.

A script, that was used for the migration:

```bash
#!/usr/bin/env bash                                                                                                                               
                                                                                                                                                  
for dir in exercises/*/; do                                                                                                                       
    (                                                                                                                                             
        cd $dir                                                                                                                                   
                                                                                                                                                  
        sed -i '2iedition = "2018"' Cargo.toml                                                                                                    
                                                                                                                                                  
        cargo fix --edition-idioms --allow-dirty                                                                                                  
                                                                                                                                                  
        cargo fmt                                                                                                                                 
                                                                                                                                                  
        cargo clean                                                                                                                               
    )                                                                                                                                             
done    
```

PR #773, opened 2018-12-10
two-fer, hexadecimal, nucleotide-codons: Removed the deprecated exercises

ZapAnton:
As per [this comment](https://github.com/exercism/rust/pull/769#issuecomment-445758356)


coriolinus:
I plan to merge on Wednesday if nobody objects. I do not object if other maintainers merge this sooner.

coriolinus:
Travis fails due to configlet lint:

```
0.28s$ ./bin/configlet lint .
-> An exercise with slug 'nucleotide-codons' is referenced in config.json, but no implementation was found.
-> An exercise with slug 'two-fer' is referenced in config.json, but no implementation was found.
-> An exercise with slug 'hexadecimal' is referenced in config.json, but no implementation was found.
The command "./bin/configlet lint ." exited with 1.
```

While we could fix this by removing the deprecated exercises from `config.json`, I believe there to be value in retaining the deprecated exercises in `config.json`: this lets the maintainers see quickly in the future that these exercises used to be in the track, and were removed. I have filed a configlet issue requesting lint changes for this case.

I will therefore override Travis and merge this (finally!) anyway.

petertseng:
Regarding whether it is safe to remove the code for deprecated exercises: My read is that it is safe. I show my work:

One may use the search "add back deleted, deprecated exercises"[1] to see the following quote
[1] https://github.com/search?q=org%3Aexercism+%22Add+back+deleted%2C+deprecated+exercises%22&type=Issues

> When migrating the data from v1 to v2, we will lose all the solutions submitted to deprecated exercises, unless they exist in the track configuration.
> 
> **We can delete them from the track again after we launch v2**, if we want to.

(emphasis mine)

Note that `git log -p -S deprecated` shows that the most recent deprecation is June 2018, so only students who have been around for a long time will have seen the deprecated exercises. This means impact should be small.

The only value in potentially doing so is if a student can choose to still download it and complete anyway, but we have better alternatives for them:

* nucleotide-codons -> protein-translation
* hexadecimal -> all-your-base
* two-fer -> any exercise that actually does Option properly

coriolinus:
Per new documentation in https://github.com/exercism/docs/pull/126, this was _not_ actually a safe change. I'd also believed that it was safe, but it's proven to be a problem. I will revert this.

PR #774, opened 2018-12-10
Shorten topics

coriolinus:
Addresses https://github.com/exercism/exercism/issues/4536

Arbitrarily chose 25 as the char limit; renamed all topics over that len

### DIFF
--- a/config.json
+++ b/config.json
@@ -246,7 +246,7 @@
       "unlocked_by": "saddle-points",
       "difficulty": 4,
       "topics": [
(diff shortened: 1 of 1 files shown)


coriolinus:
To verify that no topic exceeds 25 chars:

```sh
jq '.exercises[] | .topics[] | if (. | length) > 25 then . else empty end' config.json
```

iHiD:
Brilliant. Thank you :)

PR #777, opened 2018-12-16
hexadecimal, nucleotide-codons: Add stubs indicating deprecation

petertseng:
Needed because https://github.com/exercism/rust/pull/707, and the track
has a policy to not have empty stub files.

hexadecimal: Copied from the example.
nucleotide-codons: Copied from protein-translation.

Since the build on master is broken without this, I regret to inform
that this will be merged as soon as Travis CI passes, but post-merge
review is of course welcome.


petertseng:
lucky triple 7

PR #779, opened 2018-12-17
forth: Prepare for 2018 by not using `Term` and using `?` not `try!`

petertseng:
Prep for https://github.com/exercism/rust/pull/769 .

Blocking https://github.com/exercism/rust/pull/769 and all other otherwise-active maintainers declared inactivity, so I'm merging when tests pass.

### DIFF
--- a/exercises/forth/example.rs
+++ b/exercises/forth/example.rs
@@ -2,8 +2,6 @@ use std::collections::HashMap;
 use std::collections::LinkedList;
 use std::str::FromStr;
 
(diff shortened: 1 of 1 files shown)


petertseng:
I understand that I could have just done `use crate::Term` instead (as instructed by https://rust-lang-nursery.github.io/edition-guide/rust-2018/module-system/path-clarity.html), but I wasn't happy enough with the readability/clarity tradeoff.

PR #784, opened 2019-01-14
exercise crate: Removed the dependency on the openssl

ZapAnton:
This PR removes the dependency on the `openssl` by disabling the default feature of the `reqwest` and `git2` crates. This frees the users of the crate from configuring the opensll library on their systems.

The `default-tls` feature of the `reqwest` was replaced with the `rustls-tsl` feature. Apart from removing dependency on the `openssl` this reduced the number of the compiling crates by 4, thus saving some compilation time.

The disabling of the default features of the `git2` crate introduced additional crates to compile, but the crate is only used to find the `bin` directory in the repository. The `exercise` crate already has this functionality, which means that `git2` crate could be removed in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ Please see the documentation about [adding new exercises](https://github.com/exe
 
 Note that:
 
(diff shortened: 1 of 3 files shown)


ofek:
@TheDarkula @petertseng @iHiD How does this look?

petertseng:
not my area of knowledge, not able to say how it looks

TheDarkula:
@ofek That's awesome that they replaced it with rustls :)

There are also new versions of other crates available. Failure and failure_derive are at 0.1.5, flate2 is at 1.0.6, lazy_static is at 1.2.0, serde_json is at 1.0.38, tar is at 0.4.20, toml is at 0.4.10, uuid is at 0.7.2, and reqwest is at 0.9.9.
Do you want to tag on another commit and update those in `Cargo.toml` as well?

Also, if you feel that we don't need the `git2` dependency, would you prefer to just remove it now?

ZapAnton:
@TheDarkula , if possible I would like to have changes, that are not related, in the separate PRs.

So if this is OK with you, I will update the create versions in this PR, and remove the `git2` dependency in the other.

TheDarkula:
@ZapAnton Perfect :)

ZapAnton:
@TheDarkula done

TheDarkula:
@ZapAnton Thanks so much!

PR #792, opened 2019-02-17
exercise crate: Dropped the dependency on the 'git2' crate

ZapAnton:
As mentioned in #784


TheDarkula:
@ZapAnton Thanks again :)

PR #793, opened 2019-02-19
exercise crate: Migrated the crate to the Rust 2018 edition

ZapAnton:
Via `cargo fix --all --edition-idioms --allow-dirty`

### DIFF
--- a/util/exercise/src/lib.rs
+++ b/util/exercise/src/lib.rs
@@ -1,13 +1,12 @@
 #[macro_use]
 extern crate failure;
-extern crate failure_derive;
-extern crate flate2;
+
 #[macro_use]
 extern crate lazy_static;
-extern crate reqwest;
-extern crate serde_json;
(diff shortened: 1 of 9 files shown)


util/exercise/src/lib.rs
```diff
@@ -1,13 +1,12 @@
 #[macro_use]
 extern crate failure;
-extern crate failure_derive;
-extern crate flate2;
+
 #[macro_use]
 extern crate lazy_static;
```
coriolinus: I believe that the `#[macro_use]` idiom is now deprecated, and the correct idiom should now be 

```rust
use lazy_static::lazy_static;
```

This won't be touched by autofixers, because there's no straightforward way for them to enumerate the list of macros imported by this code from that crate. 

This applies also to the `failure` crate above, and possibly others which didn't happen to show up in the diff I'm seeing.

PR #796, opened 2019-02-21
Update link to macros chapter in Rust book

bantic:
Thanks for this exercise!
The old link went to a 404 page. I believe this is the correct link — it's the link to the macros chapter of the current book. The old link looked like it was pointing at the appending of a previous version of the book; I think that the macro info must have been promoted to its own chapter in the new book.


bantic:
@coriolinus Thanks for clarifying that! I skimmed the original repo README and missed the part about the exercise readme being held in the file in `.meta`. I wasn't sure how to run `configlet generate` so I've hand-edited the meta file as well and force-pushed.

PR #797, opened 2019-02-24
Added Rayon, parallelised code where possible, updated dependencies.

TheDarkula:
@coriolinus @ZapAnton See what you think of this :)

### DIFF
--- a/util/exercise/src/cmd/configure.rs
+++ b/util/exercise/src/cmd/configure.rs
@@ -1,5 +1,6 @@
 use exercise::{self, get, get_mut, val_as, Result};
 use failure::format_err;
+use rayon::iter::{IndexedParallelIterator, IntoParallelRefIterator, ParallelIterator};
 use serde_json::{self, json, Value};
 use std::{
     fs,
@@ -21,8 +22,8 @@ fn get_user_input(prompt: &str) -> Result<String> {
 
 fn get_user_config(exercise_name: &str, config_content: &Value) -> Result<Value> {
     let existing_config = get!(config_content, "exercises", as_array)
-        .iter()
-        .find(|exercise| exercise["slug"] == exercise_name);
+        .par_iter()
+        .find_any(|exercise| exercise["slug"] == exercise_name);
 
     let uuid = if let Some(existing_config) = existing_config {
         get!(existing_config, "uuid", as_str).to_string()
@@ -51,7 +52,7 @@ fn get_user_config(exercise_name: &str, config_content: &Value) -> Result<Value>
         } else if user_input == "null" {
             break None;
         } else if !get!(config_content, "exercises", as_array)
-            .iter()
+            .par_iter()
             .any(|exercise| exercise["slug"] == user_input)
         {
             println!("{} is not an existing exercise slug", user_input);
@@ -67,8 +68,8 @@ fn get_user_config(exercise_name: &str, config_content: &Value) -> Result<Value>
         let unlocked_by_difficulty = match unlocked_by {
             Some(ref unlocked_by) => {
                 let unlocked_by_exercise = get!(config_content, "exercises", as_array)
-                    .iter()
-                    .find(|exercise| exercise["slug"] == unlocked_by.as_str())
+                    .par_iter()
+                    .find_any(|exercise| exercise["slug"] == unlocked_by.as_str())
                     .ok_or(format_err!(
                         "exercise '{}' not found in config",
                         unlocked_by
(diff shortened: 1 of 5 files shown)


util/exercise/src/cmd/configure.rs
```diff
@@ -21,8 +22,8 @@ fn get_user_input(prompt: &str) -> Result<String> {
 
 fn get_user_config(exercise_name: &str, config_content: &Value) -> Result<Value> {
     let existing_config = get!(config_content, "exercises", as_array)
-        .iter()
-        .find(|exercise| exercise["slug"] == exercise_name);
+        .par_iter()
```
coriolinus: This doesn't feel like a case where parallelization is actually helpful: we have ~100 exercises in this list, each of which will be processed near-instantaneously.

Benchmarking this would tell us for sure, but my intuition is that Rayon won't actually use more than a single thread in this case, meaning that we pay the parallelization initialization overhead for no reason, meaning that this change actually slows things down.

util/exercise/src/cmd/update.rs
```diff
@@ -85,7 +86,7 @@ fn apply_diffs(exercise_name: &str, diffs: &HashSet<String>, tests_content: &str
         "{}\n{}",
         tests_content,
         diffs
-            .iter()
+            .par_iter()
```
coriolinus: This scales with `diffs`, which is bounded to be less than or equal to the number of tests for an exercise. Most exercises have few tests:

```sh
$ ls exercises/*/tests/*.rs | xargs rg -Fc '[test]' | awk 'BEGIN {FS=":"; OFS=","} {print $2, $1}' | sort -rn | head -n 5
Sun Feb 24 15:18:07 CET 2019
50,exercises/clock/tests/clock.rs
43,exercises/decimal/tests/decimal.rs
37,exercises/custom-set/tests/custom-set.rs
29,exercises/forth/tests/forth.rs
28,exercises/poker/tests/poker.rs
```

Therefore, for the same reasons as above, it's better to leave this single-threaded.

util/exercise/src/lib.rs
```diff
@@ -197,10 +196,11 @@ fn into_literal(item: &Value, use_maplit: bool) -> Result<String> {
         String(s) => format!("\"{}\"", s),
         Number(_) | Bool(_) => format!("{}", item),
         Array(vs) => {
-            let mut items = Vec::with_capacity(vs.len());
-            for im in vs.iter() {
-                items.push(into_literal(im, use_maplit)?);
-            }
+            let items: Vec<string::String> = vs.par_iter()
```
coriolinus: It is not impossible that we may see some benefit from parallelization here, but I'd want to see a benchmark to really be convinced.

util/exercise/src/lib.rs
```diff
@@ -235,7 +235,7 @@ pub fn generate_test_function(case: &Value, use_maplit: bool) -> Result<String>
         use Value::*;
         match comments {
             Array(cs) => cs
-                .iter()
+                .par_iter()
```
coriolinus: We're iterating over the number of comments for a single test, here. This will not be a helpful speedup.

util/exercise/src/lib.rs
TheDarkula: What's the best way to benchmark this function? I gather that `&Value` is from `serde_json`. What data should I input to the function?

coriolinus: I don't recommend using native Cargo benchmarks because you end up jumping through a ton of hoops just to get them working right on stable rust, plus you have to have a nightly install available as well to actually run the benches. It's not really worth it. [Criterion](https://crates.io/crates/criterion) makes the process much simpler, and it's pretty well documented.

`into_literal` gets called when transforming a JSON type into a Rust literal. There are three basic tactics you could use to generate `Value`s to bench in this function:

- make up a bunch of JSON by hand
- use a property-based testing library like [quickcheck](https://github.com/BurntSushi/quickcheck) to generate a bunch of cases automatically
- go to the actual data and extract a fair sample of cases

I recommend the third. Since we know that all of our input data comes from the `canonical-data.json` files found in the [problem-specifications repo](https://github.com/exercism/problem-specifications), and since we only care about arrays for this benchmark, we can extract a fair sample of arrays with the following command line:

```sh
$ ls exercises/*/canonical-data.json | 
  xargs jq -c '.. | .cases?[]? | .input[] | arrays' 2>/dev/null | 
  sort -R | 
  head 
```

The output from that command can be transformed pretty trivially into a list of strings, which the benchmark can parse into JSON. Once you've constructed your input vector of JSON `Value`s, all the bencher needs to do is transform each using `into_literal`. Don't forget to [`black_box`](https://docs.rs/criterion/0.2.10/criterion/fn.black_box.html) the output so the compiler doesn't optimize it away!

coriolinus: (If you can be tempted to include a non-random selection in your sample: one of the runs when I was developing that command line came up with a really fun array example for a recursive transformer: `[null,[[[null]]],null,null,[[null,null],null],null]`.)

coriolinus:
@TheDarkula do you have any benchmark results? I'd like to either move forward with this issue, or close it.

PR #799, opened 2019-02-27
passio: test fix when reading odd number of bytes

shenek:
The error becomes visible when README.md (which is read during the test) has odd number of bytes.

This situation occured with last README.md update 5f23d6af5a2fa3898cd46e2d3c85b5c72d355f6e


TheDarkula:
@shenek Great catch!
I tested this with the previous `README.md` to verify that it works with an even number of bytes as well.

Thanks for the pull request :)

PR #800, opened 2019-03-01
clock: Modified the add_minutes stub function to take self by reference

ZapAnton:
There is no need to consume the original value of Clock, when making a new Clock via the `add_minutes` method.

### DIFF
--- a/exercises/clock/src/lib.rs
+++ b/exercises/clock/src/lib.rs
@@ -9,7 +9,7 @@ impl Clock {
         );
     }
 
-    pub fn add_minutes(self, minutes: i32) -> Self {
+    pub fn add_minutes(&self, minutes: i32) -> Self {
         unimplemented!("Add {} minutes to existing Clock time", minutes);
     }
 }


coriolinus:
It appears that the semantics of `add_minutes` are underspecified, given that the example has yet another signature and things all still work:

https://github.com/exercism/rust/blob/4471bcb412ee9c2a916ff7499577f33fe12f8bf2/exercises/clock/example.rs#L31

Let's consider some possible signatures for `add_minutes` and the semantics they imply:

### `add_minutes(self, minutes: i32) -> Self`

This is the current default. It is modeled on [`std::ops::Add`](https://doc.rust-lang.org/std/ops/trait.Add.html). This is a combining operation: it consumes two values to produce a new value. However, this is invisible most of the time, because it is mostly implemented on types which are `Copy`. 

### `add_minutes(&self, minutes: i32) -> Self`

This is the proposed change. It does not consume `self`, so non-`Copy` clocks retain their values.

- It adds the teaching point that we do not need to consume `self` to produce useful new values
- It removes the teaching point that adding `#[derive(Clone, Copy)]` can make your type much easier to use

### `add_minutes(&mut self, minutes: i32) -> Self`

This is an odd one: it implies that adding minutes to a clock both modifies the original and also returns a copy, possibly for ease of function chaining? I'd generally recommend avoiding this signature style; we should probably remove it from the example.

### discussion

Of those three contenders, I assert that only the first two are serious options. We can, and probably should, add a test requiring that the third option is disallowed: `add_minutes` should not modify the original, but produce a new value.

Because for most implementations it will be utterly trivial to derive `Copy`, I cannot think of a straightforward test requiring only one of the first two options. This suggests to me that in some ways, it doesn't matter.

Of the first two options, both have reasonable arguments for preferring them. I can think of two arguments which, to me, imply that we should not make this change. 

- Inertia: we should not make changes simply for the fun of making changes. They should have a definite purpose outweighing the reason for the initial choice.
- Learning goals: it appears to me that the main point of this exercise is to use and implement the `Display` trait; everything else about the exercise is absolutely trivial. If trait implementation is the point of the exercise, I think it's reasonable to prefer the function signature which encourages students to discover the power of `#[derive]`. 

I am receptive to counterarguments, but for now, my inclination is to close without merge. I will not yet impose a time limit on this.

jonhoo:
I am strongly in favor of making the receiver `&self`:

 - It means that students don't have to _also_ use this exercise to learn about move-vs-copy semantics.
 - It avoids getting students into the habit of writing `self` receivers, which are _almost_ never what you want in real applications.
 - It removes another hurdle for students in figuring out how to write an idiomatic solution to this. I have seen _so_ many solutions that use a weird mix of `self`, `&self`, **and `&mut self`** in Clock thus far, just of their own volition. Starting them off with just `&self` will probably lead to less confusion around that.

Taking a step back, I actually found the decision to model Clock after `Add` quite odd in the first place. It's pretty rare that you end up writing code along the lines of `&self -> Self` in normal Rust code; it's basically _only_ when you are writing code that mutates a `Copy` type, and that's not all that common. Even then, such an API is ripe for mis-use, and often requires things like a `#[must_use]` annotation. It is far more common to have a signature of `&mut self -> ()` for methods that mutate an object (more along the lines of `AddAssign`). I would have expected us to start with the common case of `&mut self`, and maybe in a later exercise cover `&self -> Self`, `self`, and `Copy`, which in my mind are actually _more advanced_ use-cases.

PR #801, opened 2019-03-01
README: Added the 'Submitting the solution' section

ZapAnton:
Very often, when using an external crate, the students do not submit the `Cargo.toml` file, which slows the review process. Plus some students try to submit the test suite file.

This PR adds a section to the exercise README, which explains what files need to be submitted.

The READMEs are not regenerated to not clutter the PR. The regeneration will be in the closing commit, in case the PR is approved.

### DIFF
--- a/config/exercise_readme.go.tmpl
+++ b/config/exercise_readme.go.tmpl
@@ -62,6 +62,10 @@ To see, if your solution contains some common ineffective use cases, inside the 
 cargo clippy --all-targets
 ```
 
+## Submitting the solution
+
+Generally you should submit all files in which you implemented your solution (`src/lib.rs` in most cases). If you are using any external crates, please consider submitting the `Cargo.toml` file. This will make the review process faster and clearer.
+
 ## Feedback, Issues, Pull Requests
 
 The [exercism/rust](https://github.com/exercism/rust) repository on GitHub is the home for all of the Rust exercises. If you have feedback about an exercise, or want to help implement new exercises, head over there and create an issue. Members of the rust track team are happy to help!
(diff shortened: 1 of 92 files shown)


TheDarkula:
@ZapAnton I completely agree with this :)
I feel that it might be better to say:
```
`Cargo.toml` file so your mentor knows what external dependencies your solution requires. This will make the review process faster and clearer.
```

ZapAnton:
I am afraid I disagree with the change. The problem is not that the mentor doesn't know _what_ 
 external dependencies the solution uses (he can see it in the solution source, and sometimes the usage of the crate is suggested by the mentor, like the `Rayon` crate for example), but which _version_ of the crate is used, so the mentor can clearly reproduce the solution on their machine

config/exercise_readme.go.tmpl
```diff
@@ -62,6 +62,10 @@ To see, if your solution contains some common ineffective use cases, inside the
 cargo clippy --all-targets
 ```
 
+## Submitting the solution
+
+Generally you want to submit the files in which you have implemented your solution (`src/lib.rs` in most cases). If you are using any external crates, please consider submitting the `Cargo.toml` file in order for the mentor to learn which version of the library you are using. This will make the review process faster and clearer.
```
coriolinus: Both reasons for submitting `Cargo.toml` are valid. I therefore propose the King Solomon solution:

```suggestion
Generally you should submit all files in which you implemented your solution (`src/lib.rs` in most cases). If you are using any external crates, please consider submitting the `Cargo.toml` file. This will make the review process faster and clearer.
```

ZapAnton:
I have rewritten the solution according to the @coriolinus suggestion and regenerated the README.
Travis check fails because of the deprecated methods usage warning, which are not related to the PR

PR #806, opened 2019-03-23
Promoted "Reverse String" to be the first core exercise after "Hello World"

TheDarkula:
In the process of reworking the Rust track through the [Track Anatomy Project](https://exercism.io/blog/track-anatomy-project), this promotes the side exercise "Reverse String" to be our first core exercise after "Hello World". This is due to it being a "Level 1" exercise, which means that it covers only basic datatypes and only one basic principle. This is still a work in progress and not the final track :)

### DIFF
--- a/config.json
+++ b/config.json
@@ -52,18 +52,6 @@
       ]
     },
     {
-      "slug": "reverse-string",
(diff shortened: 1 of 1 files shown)


TheDarkula:
Great!
It's to keep in line with the order of core, then side exercises unlocked by that core.

PR #807, opened 2019-03-23
Updated example.rs notes for 'Reverse String'.

TheDarkula:
Since Reverse String is now our first real core exercise, this updates the corresponding `example.rs`.

### DIFF
--- a/exercises/reverse-string/example.rs
+++ b/exercises/reverse-string/example.rs
@@ -1,5 +1,3 @@
 pub fn reverse(input: &str) -> String {
-    let mut output = String::with_capacity(input.len());
-    output.extend(input.chars().rev());
-    output
+    input.chars().rev().collect()
 }


exercises/reverse-string/example.rs
```diff
@@ -1,5 +1,3 @@
 pub fn reverse(input: &str) -> String {
-    let mut output = String::with_capacity(input.len());
-    output.extend(input.chars().rev());
-    output
+    input.chars().rev().collect::<String>()
```
ZapAnton: The `::<String>` turbofish should not be necessary

TheDarkula: I'm aware it's not "necessary"; I feel it's best to teach students from the beginning to be explicit.

ZapAnton: Makes sense,  but I suspect that not many student would (or even should) refer to this repo in order to find the solution.
Besides if the automatic mentoring is ever introduced to the Rust track, I think the example solution would be a go to example for the automatic tool, therefore it should contain as little elements as possible

coriolinus: The return type of the function is explicit. I won't say that it's better to let the type inferencer do its job implicitly in _every_ case where it's capable of doing so, as there are times when making things explicit makes it easier to understand what's going on. I will say that in this case, things seem clear enough to me that I'd prefer to leave things implicit.

petertseng:
If I consider the *sole* purpose of example.rs as "make sure Travis CI passes, and therefore that we haven't provided impossible tests to the student", then we are under no obligation to ever update any example.rs (unless we add new tests). It is not for student consumption. Therefore, this PR is not necessitated by `reverse-string`'s status as a core exercise. It may be necessitated by any other reasons we decide below.

It is clear from the history of this repo that we implicitly do not consider Travis CI the *sole* purpose of example.rs. For example, we have fixed clippy warnings and made changes for shorter (and hopefully clearer) ways to express concepts, such as using `insert_or` instead of pattern-matching on `map.entry(k)`.

So, what do we consider the purposes of example.rs to be? To what standard will we want to hold example.rs?

Code that we would personally write, we were personally to try to solve the problems stated?
Code that is easy to maintain and extend? After all, example.rs will need to change when new tests are added.
Other standards?

coriolinus:
> So, what do we consider the purposes of example.rs to be? To what standard will we want to hold example.rs?
> 
> Code that we would personally write, we were personally to try to solve the problems stated?
> Code that is easy to maintain and extend? After all, example.rs will need to change when new tests are added.

I would hope that the two standards proposed are actually the same standard most of the time. 

I suppose the standard I personally have been applying is that I typically ignore `example.rs` while it's working, but when I must edit it, I do so in a way which "looks nice". This is a very fuzzy standard, but in general, I want things to seem idiomatic, not generate clippy lints, and operate in a straightforward fashion.

The change in this PR looks nice, with the exception of the turbofish. It is possibly a little bit less efficient than the code it replaces, but for this use case, I'm willing to accept a tiny performance hit in exchange for clarity and idiomaticity. The turbofish isn't nice because it's boilerplate: extra code for the programmer to read and understand, which doesn't contribute to the functionality. 

We have options. I will enumerate them, and list the people who support each who have so far commented on this PR, in my understanding. If I misplace you, please speak up and correct me.

- Accept the PR as-is: @TheDarkula 
- Accept the PR without the turbofish: @ZapAnton, @petertseng, @coriolinus.
- Reject the PR: nobody. 

I think we should just pick one of these options. There's been over a month of no activity on this. I do not propose that this is a democracy, but I do propose that the "wisdom of the crowds" can be a useful guideline for determining the correct option when there is disagreement and no clear solution.

petertseng:
I'm executing the option "Accept the PR without the turbofish" now.

Normally I have no problem taking no action on a PR and waiting for the submitter to take an appropriate action, since leaving a PR in the queue doesn't interfere with my workflow in any way, but in this case since the PR was made on a branch in Exercism's copy of the repo, there is a strong incentive to remove that branch.

PR #808, opened 2019-03-28
Promoted "Gigasecond" to be a core exercise.

TheDarkula:
In the process of reworking the Rust track through the [Track Anatomy Project](https://exercism.io/blog/track-anatomy-project), this promotes the side exercise "Gigasecond" to be a core exercise after "Reverse String". This is due to it being a "Level 1" exercise, which means that it covers only basic datatypes and principles. This is still a work in progress and not the final track :)

### DIFF
--- a/config.json
+++ b/config.json
@@ -20,16 +20,6 @@
       ]
     },
     {
-      "slug": "gigasecond",
-      "uuid": "f880b1ef-8f66-41f3-bb89-f39a4ed592a2",
-      "core": false,
-      "unlocked_by": "hello-world",
-      "difficulty": 1,
-      "topics": [
-        "crates"
-      ]
-    },
-    {
       "slug": "leap",
       "uuid": "ef52f576-9c33-4cc1-a018-803ace8897f6",
       "core": false,
@@ -161,6 +151,16 @@
         "iterator",
         "str",
         "string"
+      ]
+    },
+    {
+      "slug": "gigasecond",
+      "uuid": "f880b1ef-8f66-41f3-bb89-f39a4ed592a2",
+      "core": true,
+      "unlocked_by": null,
+      "difficulty": 1,
+      "topics": [
+        "crates"
       ]
     },
     {


coriolinus:
I don't understand the motivation here. 

My understanding of the intended flow is that there is a graduated series of exercises of increasing difficulty. The tree of exercises should be short and broad, kind of a shrub. If I understand correctly, each core exercise should introduce some new concept. 

We already have an initial exercise demoing basic datatypes and principles, so shouldn't the next core exercise require more from the student than that? If students want to keep refining and expanding their expertise with concepts they've already covered, this is exactly what the side exercises are for.

kytrinyx:
@coriolinus This is part of the Track Anatomy project. For more on the background there, please see the issue that I've just opened here: https://github.com/exercism/rust/issues/809

This PR is part of Round 2. In this Round we're looking at the start of the track, adding exercises that bridge the gap to the harder exercises, and that are easy to mentor (ideally, mentoring on the earliest exercises will largely consist of copying, pasting, and tweaking snippets of text that address common concerns.

In this case, the gap between the first and the second exercise (Bracket Push) is significant. The choice for `Gigasecond` here is entirely pragmatic: it's already implemented as a side exercise, so we have it available, and it's easier than Bracket Push.

kytrinyx:
> We already have an initial exercise demoing basic datatypes and principles, so shouldn't the next core exercise require more from the student than that? If students want to keep refining and expanding their expertise with concepts they've already covered, this is exactly what the side exercises are for.

I forgot to address this above.

One of the things that we will be looking for feedback on is whether or not there are common patterns to the errors that people make on Gigasecond, when it is promoted to a core exercise. If (mostly) everyone does the exercise perfectly the first time through, then we likely need to move it back to a side exercise. If, however, there are typically things to address during mentoring, then it most likely introduces at least one concept that was not introduced in the prior exercise, and may be a good choice to keep in the core of the track.

petertseng:
Please remember to delete branches after a PR is merged. This is necessary because this is the exercism copy of the repo, and any branches get copied to forks. It is discourteous to confuse contributors with these branches and force them to ask "what are these branches for, do I need to keep them on my fork, do I delete them, what's the deal here".

I personally choose to avoid this problem by putting all branches on my own fork instead of exercism's copy.

petertseng:
As I will mention in the next PR, there was good rationale in the PR description.

**This rationale should go in the commit message as well**, so that it is available from `git log`.

PR #810, opened 2019-03-29
Promoted "Bob" to be a core exercise.

TheDarkula:
This pull pequest is part of the [Track Anatomy Project](https://exercism.io/blog/track-anatomy-project). For more on the background there, please see [issue #809](https://github.com/exercism/rust/issues/809) opened by @kytrinyx.

This is part of "Round 2", where we're looking at the start of the track and adding exercises that bridge the gap to the harder exercises and that are easy to mentor (ideally, mentoring on the earliest exercises will largely consist of copying, pasting, and tweaking snippets of text that address common concerns).

In this case, the gap between the second ("Gigasecond") and third exercise ("Bracket Push") is significant. The choice for "Bob" here is entirely pragmatic: it's already implemented as a side exercise, so we have it available, and it's easier than "Bracket Push".

This is still a work in progress and not the final track :)

### DIFF
--- a/config.json
+++ b/config.json
@@ -54,17 +54,6 @@
       ]
     },
     {
-      "slug": "bob",
-      "uuid": "38ef1802-2730-4f94-bafe-d2cd6b3e7f95",
-      "core": false,
-      "unlocked_by": "hello-world",
-      "difficulty": 1,
-      "topics": [
-        "chars",
-        "string_functions"
-      ]
-    },
-    {
       "slug": "beer-song",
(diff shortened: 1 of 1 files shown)


petertseng:
There is rationale given for this in the PR description.

**This rationale should go in the commit message as well**, so that it is available from `git log`. Note that PRs consisting of a single commit, such as this one, commit message is automatically entered into the PR description, so this does not force the typing of the rationale a second time.

For this PR that has already been created, the commit should be amended.

TheDarkula:
@petertseng Done :)

PR #816, opened 2019-04-08
tournament: rework the 'a_draw_is_one_point_each' test to handle any ordering

samcday:
When getting the tests to pass one by one, the final test is to sort teams by
score and then fallback to alphabetical by team name when scores match. Since
the earlier draw test also results in two teams with one score, it kinda ruins
the flow by requiring solution to be sorting by team name already.

---

Also interestingly enough, in my case I chose to sort the items unstably (for speeeeed), which meant that the test would intermittently alternate between passing and failing. I'm almost torn about fixing this edge-case because while it might vex some people, it could also lead many of them to learning that `unstable_sort` deliberately randomizes the tiebreaker, which could send them down all sorts of interesting rabbit holes ;)

### DIFF
--- a/exercises/tournament/tests/tournament.rs
+++ b/exercises/tournament/tests/tournament.rs
@@ -46,18 +46,6 @@ fn a_different_team_can_win() {
 
 #[test]
 #[ignore]
-fn a_draw_is_one_point_each() {
-    let input = "Allegoric Alaskans;Blithering Badgers;draw";
-    let expected = "".to_string()
-        + "Team                           | MP |  W |  D |  L |  P\n"
-        + "Allegoric Alaskans             |  1 |  0 |  1 |  0 |  1\n"
-        + "Blithering Badgers             |  1 |  0 |  1 |  0 |  1";
-
-    assert_eq!(tournament::tally(&input), expected);
-}
-
-#[test]
-#[ignore]
 fn there_can_be_more_than_one_match() {
     let input = "Allegoric Alaskans;Blithering Badgers;win\n".to_string()
         + "Allegoric Alaskans;Blithering Badgers;win";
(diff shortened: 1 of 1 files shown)


coriolinus:
Good spot, noticing this. It's an interesting note about the flow. I agree that we can improve the flow. I am not convinced yet that the approach you chose to fix the flow is the best one. While searching for substrings does ensure that the ordering doesn't matter for this test, it means this test is unlike any of the others, and allows unbounded results.

What about just adding a proper win to the Alaskans, so that the ordering is guaranteed while still testing the draw behavior?

exercises/tournament/tests/tournament.rs
```diff
@@ -44,12 +44,13 @@ fn a_different_team_can_win() {
 #[test]
 #[ignore]
 fn a_draw_is_one_point_each() {
-    let input = "Allegoric Alaskans;Blithering Badgers;draw";
-    let expected = "Team                           | MP |  W |  D |  L |  P\n".to_string()
-        + "Allegoric Alaskans             |  1 |  0 |  1 |  0 |  1\n"
-        + "Blithering Badgers             |  1 |  0 |  1 |  0 |  1";
+    let input = "Allegoric Alaskans;Blithering Badgers;loss\n".to_string()
+        + "Allegoric Alaskans;Blithering Badgers;win";
```
coriolinus: Also you've taken out the `draw` input and replaced it (incorrectly) with one win apiece.

samcday: Oops!

samcday:
@coriolinus good call, I pushed up a new attempt. I moved `a_draw_is_one_point_each` down one position so it comes after the `there_can_be_more_than_one_match` since that should obviously be the first test to check if multiple matches are supported.

PR #817, opened 2019-04-09
dot-dsl: Use zip instead of enumerate + map-with-indexing

petertseng:
Suggested by clippy:
https://travis-ci.org/exercism/rust/builds/517554492
warning: an inclusive range would be more readable
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#range_plus_one

Note that this `&attributes[i..=i]` is saying "one-element slice (the
ith element of `attributes`)". An equivalent construction would be
`&[attributes[i]]`. The latter would copy `attributes[i]`, so we'll
avoid by continuing to slice.

---

I ask reviewers to pay careful attention to the factual accuracy of the last paragraph and also judge whether this is a desirable thing to continue to do, as opposed to the alternative construction presented.

### DIFF
--- a/exercises/dot-dsl/tests/dot-dsl.rs
+++ b/exercises/dot-dsl/tests/dot-dsl.rs
@@ -133,7 +133,7 @@ fn test_graph_stores_attributes() {
         &['a', 'b', 'c']
             .iter()
             .enumerate()
-            .map(|(i, n)| Node::new(&n.to_string()).with_attrs(&attributes[i..i + 1]))
+            .map(|(i, n)| Node::new(&n.to_string()).with_attrs(&attributes[i..=i]))
             .collect::<Vec<_>>(),
     );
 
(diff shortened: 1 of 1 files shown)


coriolinus:
> Note that this `&attributes[i..=i]` is saying "one-element slice (the ith element of attributes)". An equivalent construction would be `&[attributes[i]]`. The latter would copy `attributes[i]`, so we'll
avoid by continuing to slice.

`&attributes[i..=i]` is fundamentally a referencing operation. `attributes[i]` is a _move_ operation, except in the special case that the members of `attributes` implement `Copy`. I think that is the case here, in which case you were probably assuming the context of this test case, in which case we don't disagree.

In either case, `&attributes[i..=i]` only produces equivalent values to `&[attributes[i]]` when the contained type is `Copy`, and the produced slices have distinct identities. For that reason, I think it is not safe to assume that the two are identical. 

In the context of this particular case, it looks like `.zip` would be a more idiomatic replacement if we're contemplating rewriting the test. I would prefer that over retaining either slice-based mechanism.

petertseng:
(That was an accidental close)

That seems sensible. Let's try this implementation with `zip`, for reviewers' consideration.

PR #819, opened 2019-04-14
matching-brackets: Rename to bracket-push; _test: Exclude deleted files

petertseng:
This rename was proposed in:
https://github.com/exercism/problem-specifications/pull/1501

The rationale for the name is:

* to name the exercise by its story, not by what it potentially teaches
* to avoid unnecessarily biasing the solution space


coriolinus:
Sounds like a reasonable policy. I suspect that deletion will be pretty
rare, at least compared to deprecation, which is already uncommon.

petertseng:
I decided to file an issue anyway, and just close it immediately. In the issue I also noted the point that I include *exercise rename* in the list of operations that involve deletion.
Perhaps one day we will make a mistake in a rename and that will motivate a closer look at the scripts.

In light of the decision, I would say that Travis CI is in as good of a state as it will ever be for this pull request, then.

petertseng:
Since I think there's a quick win, I've instead executed the change described, and propose it for inclusion. Since it's quite hard to test it without an accompanying delete, I've included it in this PR.

In light of this (there are two commits that do completely different things), it should go without saying that these commits absolutely cannot be squashed.

PR #822, opened 2019-04-24
Update copyright in license

kytrinyx:
This ensures that the year is up to date, and
that the copyright is assigned to the not-for-profit
Exercism, rather than the (now-defunct) Delaware
Corporation that we had in place as a temporary
measure for legal protection until we could
figure out how to structure things properly.

Ref: https://github.com/exercism/exercism/issues/4823

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2017 Exercism, Inc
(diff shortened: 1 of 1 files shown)


TheDarkula:
@kytrinyx Perfect :)
Would you like me to merge this?

Issue #626, opened 2018-08-20
nth-prime tests use the wrong indexes


mathw:
https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.nth

"Like most indexing operations, the count starts from zero, so nth(0) returns the first value, nth(1) the second, and so on."

However, the tests for nth-prime expect to find the first prime after calling `nth(1)`, the second at `nth(2)` etc. For example:

```
#[test]
fn test_first_prime() {
    assert_eq!(np::nth(1), Some(2));
}
```

I could hack my solution to conform to this expectation, but that would involve putting a fudge in my `Iterator` implementation which I shouldn't have to do.

Changing this would break every solution currently submitted though...

petertseng:
While this exercise is not *obligated* to adhere to the same interface semantics as Iterator, because of the similarity in signature, it would be *surprising* if it did not, and we'd like to promote good habits and lessen surprise. So it could make sense to 0-index here. A clarifying comment in either the tests or README could help.

I wonder if https://github.com/exercism/problem-specifications/blob/master/exercises/nth-prime/canonical-data.json should change. But since some languages may 1-index and others 0-index, I think either way some languages will have to make a decision on what to do.

> Changing this would break every solution currently submitted though...

True, but if everyone agrees the current tests are broken, then not changing this would break all future solutions, so the change has to be made.

Now, the website *claims* to handle test suite changes well, by remembering which test suite each individual student downloaded (I acquired this information from https://github.com/exercism/exercism/issues/4266). I can't speak to how clearly it shows the differing versions when viewing others' solutions as either a student or a mentor, since I have not attempted to do either of these things, but it looks like there is discussion to make that right.

PR #823, opened 2019-04-28
nth-prime: tests do not match problem specifications

imbolc:
In the exercise description 13 is 6th number while in tests it's 5th, I changed it to be consistent with python version

### DIFF
--- a/exercises/nth-prime/.meta/hints.md
+++ b/exercises/nth-prime/.meta/hints.md
@@ -1,0 +1,1 @@
+Remember that while people commonly count with 1-based indexing (i.e. "the 6th prime is 13"), many programming languages, including Rust, use 0-based indexing (i.e. `primes[5] == 13`). Use 0-based indexing for your implementation.
--- a/exercises/nth-prime/README.md
+++ b/exercises/nth-prime/README.md
@@ -8,6 +8,9 @@ the 6th prime is 13.
 If your language provides methods in the standard library to deal with prime
 numbers, pretend they don't exist and implement them yourself.
 
+Remember that while people commonly count with 1-based indexing (i.e. "the 6th prime is 13"), many programming languages, including Rust, use 0-based indexing (i.e. `primes[5] == 13`). Use 0-based indexing for your implementation.
+
+
 ## Rust Installation
 
 Refer to the [exercism help page][help-page] for Rust installation and learning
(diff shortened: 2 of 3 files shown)


coriolinus:
@imbolc thanks for contributing this work! Travis has rejected it because, while you updated the tests, you didn't also edit [`example.rs`](https://github.com/exercism/rust/blob/b39d0309aa7358f052b82ce5df592c1570cee26b/exercises/nth-prime/example.rs), which contains an example implementation. 

Once the example implementation is back in sync with the tests, all should be good for this PR!

petertseng:
Since we have both a reason to use zero-indexed numbers (https://github.com/exercism/rust/issues/626) and a reason to use one-indexed numbers (problem-specifications), let's be sure we have weighed the two against each other and make a choice we are good with for the current time (by that I mean we can be open to change our minds in the future too)

petertseng:
I would like to explicitly note that it does not matter to me which decision we make, so I will not recommend either decision over the other.

coriolinus:
Oh, yeah, I'd forgotten about that. Looking at the two points of view, it appears to me that

- you're right, it doesn't really matter if we index by 0 or 1
- if we merge this PR, we will confuse students who expect Iterator-ish semantics
- if we do not merge this PR, we will confuse students who have read the problem specification and noted the discrepancy

My proposal for a path forward: 
- discard these changes and retain 0-based indexing
- add a note in `.meta/hints.md` explicitly stating that we use 0-based indexing, and a short note why.
- regenerate the README so it includes that hint

@imbolc are you willing to move forward with that plan?

imbolc:
In my experience I just googled a prime numbers algorithm, converted it to rust and then after tests broke, I started to look for where I messed up with indexes converting this algorithm :) I mean it's hard to predict different kinds of uncertainty students minds can have related to new language and math and I'd like to see description as precise as possible. Also description and tests probably should be consistent between languages, so changes will invalidate community solutions (for rust if we change description and for other languages if we change tests).

coriolinus:
In the interest of making forward progress, I have implemented the changes I suggested in this branch. However, this disqualifies me from reviewing this branch. @petertseng you already looked at this; are you willing to review these changes?

coriolinus:
Note: Travis fails erroneously, and will continue to do so until https://github.com/exercism/configlet/issues/159 is fixed.

coriolinus:
WRT configlet, I have opened an issue and PR which will close the issue if approved: https://github.com/exercism/configlet/pull/160.

PR #829, opened 2019-05-04
Wordy: test for bonus condition

ccouzens:
The wordy Readme has the following:

> ## Bonus — Exponentials
> 
> If you'd like, handle exponentials.
> 
> > What is 2 raised to the 5th power?
> 
> 32

I've added tests for this, hidden behind the `exponentials` feature flag

### DIFF
--- a/exercises/wordy/example.rs
+++ b/exercises/wordy/example.rs
@@ -9,9 +9,9 @@ fn apply_op<'a, 'b>(num1: i32, words: &'a [Token<'b>]) -> Option<(i32, &'a [Toke
         Token::Number(_) => true,
         Token::NonNumber(_) => false,
     })?;
-    let (op_and_num, remainder) = words.split_at(number_pos + 1);
+    let (op_and_num, mut remainder) = words.split_at(number_pos + 1);
     let (op, num2) = op_and_num.split_at(number_pos);
-    let num2 = match num2 {
+    let &num2 = match num2 {
         [Token::Number(i)] => i,
         _ => unreachable!(
             "We split at a Number above, so num2 is surely a single-element slice w/ a number"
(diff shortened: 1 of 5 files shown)


ccouzens:
Thank you for giving feedback so quickly :)

> why do you prefer assert_eq over assert? It looks like most of the tests you replaced in that commit are of the form assert!(answer(command).is_none());, which to my eye is better than assert_eq!(None, answer(command));

I made than change in a separate commit and would be happy to rebase it out of this pull request.

The reason I preferred it, is when working through this exercise as a student, when my solution is wrong I get more information with `assert_eq!`.

To use this test as an example:

```rust
fn reject_problem_missing_an_operand() {
    let command = "What is 1 plus?";
-    assert!(answer(command).is_none());
+    assert_eq!(None, answer(command));
} 
```

With `assert!(answer(command).is_none());` I might see that my solution is wrong, but nothing more.
With `assert_eq!(None, answer(command));` I might see my solution is `Some(1)` and I can better see what changes I might need to make to fix it.

> IMO the current wording in hints.md is slightly too informal. Suggest dropping the clause concerning the "last couple tests", and just instructing the students to use the --features argument. They should know that they also need to remove the #[ignore] tag by this point.

I've made the change suggested dd7eace9cede5c07eaadad9174440ecdc5b8e544

Thank you,

Chris

exercises/wordy/example.rs
```diff
@@ -34,6 +42,11 @@ pub fn answer(c: &str) -> Option<i32> {
         .map(|word| {
             if let Ok(i) = word.parse::<i32>() {
                 Token::Number(i)
+            } else if let Some(i) = word
+                .get(..word.len() - 2)
+                .and_then(|word| word.parse::<i32>().ok())
+            {
+                Token::Number(i)
```
petertseng: if it were necessary to distinguish between these and the `Number`s above, I would have suggested `Token::Ordinal(i)`. However, since it is not necessary, this suggestion should instead be ignored.

ccouzens:
I'm getting the following Travis CI build failure

```
0.31s$ ./bin/configlet lint .
-> An exercise with slug 'two-fer' is referenced in config.json, but no implementation was found.
-> An exercise with slug 'hexadecimal' is referenced in config.json, but no implementation was found.
-> An exercise with slug 'nucleotide-codons' is referenced in config.json, but no implementation was found.
The command "./bin/configlet lint ." exited with 1.
```

It looks like you're on-top of it already :)

https://github.com/exercism/configlet/pull/160

PR #830, opened 2019-05-08
Forego formerly deprecated exercises

coriolinus:
The rust track had several deprecated exercises. As they were removed
from the list of exercises available to students some time ago,
https://github.com/exercism/rust/pull/773 removed the actual exercises,
in order to clean things up. This produced a CI failure, but it was
overridden, because it was a configlet bug, not a real problem.

Overriding the CI failure was a mistake: I hadn't considered that
this would cause _every_ CI job to fail from that point forward.

I opened https://github.com/exercism/configlet/pull/160 fixing the
issue with configlet, but it has not yet been reviewed, and there
will be an unknowable amount of time before it is reviewed and
approved. In order to make configlet stop complaining and restore
CI functionality to the track, this PR moves the removed exercises
into the foregone list, which doesn't suffer the configlet bug.

### DIFF
--- a/config.json
+++ b/config.json
@@ -4,8 +4,11 @@
   "blurb": "Rust is a compiled programming language designed for speed, concurrency, and memory safety. Rust programs can run almost anywhere, from low-power embedded devices to web servers.",
   "foregone": [
     "binary",
+    "hexadecimal",
+    "nucleotide-codons",
     "octal",
-    "trinary"
+    "trinary",
+    "two-fer"
   ],
   "exercises": [
     {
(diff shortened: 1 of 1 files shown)


coriolinus:
Whether or not we put them back, I think, depends on the exact semantics of "deprecated" and "foregone".  I have an idea of what those statuses mean, but it's an empirical one, based on what I've seen. I have not seen any documentation clarifying the terms. 

I think "foregone" means that exercism as a whole has decided that the exercise was actually a bad idea and has completely removed support. I think "deprecated" means that the exercise was formerly implemented for the track, but then removed as a bad idea for the track. It may still be appropriate for other tracks. 

If that understanding of the statuses is correct, then once configlet is updated appropriately, we should revert this PR. If that understanding is incorrect, then we should reevaluate given the new information.

petertseng:
Here are my conceptions, which are also entirely empirical rather than based on any documentation. It seems there is a need for some documentation.

`Foregone` is a track-local decision: the track maintainers decide the exercise should not be implemented for the track.

`Deprecated` is also a track-local designation, which is similar to foregone in that we don't want students to do the exercise... but they were implemented at some point in time, and kept for... historical reasons? I have no idea.

The only info I find in docs are:

* https://github.com/exercism/docs/blob/master/language-tracks/configuration/configlet.md
* https://github.com/exercism/docs/blob/master/language-tracks/configuration/track.md

coriolinus:
Haha, agree that some more documentation would be useful. 

Whether or not `foregone` exercises are a track-local or external decision, I think there is value in distinguishing them from the `deprecated` list. Ideally, I would want the following UX:

- `foregone` exercises never appear in the exercise list
- `deprecated` exercises are not available for students who have not already submitted solutions, but students who have already completed them may still see their solutions and any comments, mentoring, etc on them.

I have no idea if the exercism website actually implements that UX.

If it is an exercism goal to present that UX, then we should revert this PR once configlet properly supports it. If not, then it doesn't really matter, so we may as well retain the status quo.

PR #831, opened 2019-05-08
Adhere to convention assert_eq!(expected, actual)

marcoesposito1988:
First of all, thank you very much for your work on the Rust exercise track! I am finally productive with this language. I would be glad to contribute back, and I stumbled on a particular annoyance.

While the Rust documentation for assert_eq! does not mention this, it is common to consider the first argument the "expected" value (see [this thread on StackOverflow](https://stackoverflow.com/questions/26102865/assertequals-what-is-actual-and-what-is-expected)).

This convention is followed in some exercises but not in all of them. This can lead to very confusing results with IDEs that support the testing framework (e.g. CLion), and report the "expected" and the "actual" values to the user... reversed.

This pull request fixes this across all exercises.

### DIFF
--- a/exercises/accumulate/tests/accumulate.rs
+++ b/exercises/accumulate/tests/accumulate.rs
@@ -10,7 +10,7 @@ fn square(x: i32) -> i32 {
 fn func_single() {
     let input = vec![2];
     let expected = vec![4];
-    assert_eq!(map(input, square), expected);
+    assert_eq!(expected, map(input, square));
 }
 
 #[test]
@@ -18,7 +18,7 @@ fn func_single() {
 fn func_multi() {
     let input = vec![2, 3, 4, 5];
     let expected = vec![4, 9, 16, 25];
-    assert_eq!(map(input, square), expected);
+    assert_eq!(expected, map(input, square));
 }
 
 #[test]
@@ -26,7 +26,7 @@ fn func_multi() {
 fn closure() {
     let input = vec![2, 3, 4, 5];
     let expected = vec![4, 9, 16, 25];
-    assert_eq!(map(input, |x| x * x), expected);
+    assert_eq!(expected, map(input, |x| x * x));
 }
 
 #[test]
@@ -34,7 +34,7 @@ fn closure() {
 fn closure_floats() {
     let input = vec![2.0, 3.0, 4.0, 5.0];
     let expected = vec![4.0, 9.0, 16.0, 25.0];
-    assert_eq!(map(input, |x| x * x), expected);
+    assert_eq!(expected, map(input, |x| x * x));
 }
 
 #[test]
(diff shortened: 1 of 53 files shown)


petertseng:
I agree that consistency has advantages. I think many can agree with this. An example advantage is reducing mental overhead of having to figure out which is which every time.

I predict that because there is no automatic way to enforce the argument order, there is the risk that things will drift away as time goes on. This shouldn't prevent an interested contributor form submitting a pull request (as is done with this pull request), but is a caution that either 1) someone must volunteer to provide ongoing review/maintenance effort to see that future additional exercises/tests have the order desired (to which I explicitly do not volunteer) or 2) there will be an inevitable risk that they will be out of order

> While the Rust documentation for assert_eq! does not mention this, it is common to consider the first argument the "expected" value

To see what convention is expected, I think the only reasonable place to look is in fact the Rust docs. Since https://doc.rust-lang.org/std/macro.assert_eq.html is silent on the issue, perhaps we look elsewhere, such as https://doc.rust-lang.org/std/vec/struct.Vec.html#examples-16 . In https://doc.rust-lang.org/std/vec/struct.Vec.html#examples-16 we see that consistently `assert_eq(actual, expected)` is used. So they use the mnemonic *right* (opposite of left) is *right* (correct).

> This can lead to very confusing results with IDEs that support the testing framework (e.g. CLion), and report the "expected" and the "actual" values to the user... reversed.

If there are any IDEs that make any assumptions about which of `left`/`right` is `expected`/`actual`, I would suggest that any such IDE should stop making such an assumption when dealing with `assert_eq!` as implemented in the Rust language. Unless the Rust docs explicitly mention it, it is not possible to make such an assumption, since it appears that different projects can choose differently.

For this project, I would say the track has the power to decide on which, if any, convention it would like to use. I recommend that if a convention is chosen, it should be documented in the README or other suitable location. I will not express a preference for either convention over the other, so interested parties will have to do that.

coriolinus:
Strongly agree with @petertseng 's comments: there is no automatic way to enforce this standard without active human effort, and I am not personally interested in contributing the ongoing effort required. 

If we pick a track convention, I'd prefer to pick the "rusty" convention, which appears to be `(actual, expected)`. However, given that neither of the two most active maintainers are interested in maintaining a convention, it seems to me that we should not establish one in the first place.

That leaves the question of the disposition of this PR. It seems to me we have the following options:

- If we decide to establish the convention `(expected, actual)`, then this PR should also update the `README` appropriately.
- If we decide to establish the convention `(actual, expected)`, then this PR should update the README, revert the existing changes, and (if @marcoesposito1988 is willing) also implement that convention across all exercises.
- If we decide to establish no official convention, we could accept this PR on the grounds that it does no harm, or we could reject it on the grounds that it adds unnecessary chaff to the commit history. 

My personal inclination would be to establish no official convention, but accept this PR to acknowledge the work that @marcoesposito1988 has put into it. This is a low-strength preference; I could be persuaded into any of the other options.

Also, I note that I would not yet accept it in its current state: it's broken `gigasecond` due to missing trailing commas.

marcoesposito1988:
If there is no interest in maintaining a convention (which in my opinion would be a great advantage for the users, removing the useless fatigue of having to check the assert_eq! call every time), then it makes no sense to merge these changes.

Personally, I think that the mistake was made at the Rust level, since the absence of a convention (ideally the most popular one) has no advantage but considerable drawbacks.

I will open an issue on the JetBrains Rust plugin repo, so that they maybe do not show the values as "expected" and "actual", but rather show the call site of assert_eq!.

Thanks for taking the time to review the pull request!

PR #832, opened 2019-05-10
Revert removing deprecated exercises

coriolinus:
Per new documentation in https://github.com/exercism/docs/pull/126, it is actually _not_ safe to remove deprecated exercises. As such, #773 should not have been approved. Without #773, we don't need #830. 

This reverts both of those PRs.


exercises/hexadecimal/src/lib.rs
```diff
@@ -1 +1,33 @@
+fn parse_hex_digit(c: char) -> Option<i64> {
+    match c {
```
petertseng: although it does not hurt to have it since it's deprecated, this lib.rs doesn't necessarily need to have a complete implementation (and we would want it not to have one if it were non-deprecated)

exercises/nucleotide-codons/src/lib.rs
```diff
@@ -1 +1,46 @@
...
+
+pub fn parse<'a>(pairs: Vec<(&'a str, &'a str)>) -> CodonInfo<'a> {
+    CodonInfo {
+        actual_codons: pairs.into_iter().collect(),
+    }
+}
+
+impl<'a> CodonInfo<'a> {
+    pub fn name_for(&self, codon: &str) -> Result<&'a str, &'static str> {
+        if codon.len() != 3 {
```
petertseng: > although it does not hurt to have it since it's deprecated, this lib.rs doesn't necessarily need to have a complete implementation (and we would want it not to have one if it were non-deprecated)

same

exercises/two-fer/src/lib.rs
```diff
@@ -1,3 +1,6 @@
 pub fn twofer(name: &str) -> String {
-    unimplemented!("One for {}, one for me.", name);
+    match name {
```
petertseng: > although it does not hurt to have it since it's deprecated, this lib.rs doesn't necessarily need to have a complete implementation (and we would want it not to have one if it were non-deprecated)

same

coriolinus:
I copied the full implementations over the three stub `src/lib.rs` implementations because there were some compilation errors, and I didn't believe it was worth the effort of figuring out and fixing the real problems there. 

As you've already approved this (thank you!), and Travis now passes, I will now merge this.

petertseng:
Given that the documentation *claims* that students who have not already started a deprecated exercise will not be able to do so, it is OK to have stubs that already provide a solution.

For the purposes of our track upholding its documented obligations, it is irrelevant that the aforementioned claim is false (I can currently download hexadecimal and submit an implementation to it); that issue must be solved elsewhere, so this track need not take specific action on that.

PR #833, opened 2019-05-10
exercise crate: Refactored the fetch_configlet.rs file

ZapAnton:
This PR refactors the `fetch_configlet.rs`  in two steps:
- Move the`fetch_configlet.rs` file into the `exercise/src/cmd` directory, since `fetch_configlet` is a valid crate subcommand on par with the `update`, `generate`, etc.
- Refactor the fetching algorithm, in particular the way the download url for the latest release is generated. Currently the download url is taken and parsed from the HTTP response header, which (IMHO) is somewhat unreliable. Instead in this PR the download url is taken from the official [Github API](https://developer.github.com/v3/repos/releases/#get-the-latest-release), which only requires a JSON parsing. Plus the overall size of the file is decreased from the 73 LOC to the 54 LOC, which could be considered an improvement.


ZapAnton:
@coriolinus, lets merge it then. These changes do not affect the workflow of the track and therefore should be safe.

PR #837, opened 2019-05-30
anagram: Added 2 hints

Dimkar3000:
This PR's adds some hints for the problem that to my opinion are needed. There are no indications other than the tests failing that the solution is case insensitive. Another thing that I found out by trial and error is that the word itself cannot be part of the solution. 

Those 2 should be better explained. So added hints for both of them

### DIFF
--- a/exercises/anagram/.meta/hints.md
+++ b/exercises/anagram/.meta/hints.md
@@ -2,4 +2,4 @@
 
 The solution is case insensitive, which means `"WOrd"` is the same as `"word"` or `"woRd"`. It may help to take a peek at the [std library](https://doc.rust-lang.org/std/index.html) for functions that can convert between them.
 
-The solution cannot contain the word itself. Basically a word is always an anagram of it's self but for this exercise we don't care. This mean, given `"hello"` and the list `["hello", "olleh"]` the answear is `["olleh"]`
+The solution cannot contain the input word. A word is always an anagram of itself, which means it is not an interesting result. Given `"hello"` and the list `["hello", "olleh"]` the answer is `["olleh"]`.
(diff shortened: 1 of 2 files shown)


exercises/anagram/.meta/hints.md
```diff
@@ -0,0 +1,5 @@
+# Hints
+
+The solution is case insensitive, that means `"WOrd"` is the same with `"word"` or `"woRd"`. It may help to take a peak in the [std library](https://doc.rust-lang.org/std/index.html) for functions that can convert between them.
```
coriolinus: Fixing some typos:

```suggestion
The solution is case insensitive, which means `"WOrd"` is the same as `"word"` or `"woRd"`. It may help to take a peek at the [std library](https://doc.rust-lang.org/std/index.html) for functions that can convert between them.
```

If you merge this, don't forget to regenerate README.md

exercises/anagram/.meta/hints.md
```diff
@@ -0,0 +1,5 @@
...
+
+The solution cannot contain the word itself. Basically a word is always an anagram of it's self but for this exercise we don't care. This mean, given `"hello"` and the list `["hello", "olleh"]` the answear is `["olleh"]`
```
coriolinus: More typos, grammar, and style:

```suggestion
The solution cannot contain the input word. A word is always an anagram of itself, which means it is not an interesting result. Given `"hello"` and the list `["hello", "olleh"]` the answer is `["olleh"]`.
```

exercises/anagram/.meta/hints.md
coriolinus: I think that if we include a link, it should be to a more useful page than the docs homepage. Suggest https://doc.rust-lang.org/std/primitive.str.html. I would also accept removing the link here.

Dimkar3000: from this link, the solution is one search away. I thing its good experience searching for what you want. Maybe I should add "search for char" or something similar

Dimkar3000:
Ok I fixed everything you pointed out. If there is anything else wrong let me know but I thing it is ready to merge

coriolinus:
Sorry I didn't merge this when I planned to; I straight up forgot.

PR #840, opened 2019-06-25
exercise_readme: fix broken link to modules docs

stringparser:
The original link was renamed.

### DIFF
--- a/config/exercise_readme.go.tmpl
+++ b/config/exercise_readme.go.tmpl
@@ -43,7 +43,7 @@ $ cargo test some_test -- --ignored
 
 To learn more about Rust tests refer to the [online test documentation][rust-tests]
 
-Make sure to read the [Modules](https://doc.rust-lang.org/book/ch07-02-modules-and-use-to-control-scope-and-privacy.html) chapter if you
+Make sure to read the [Modules][modules] chapter if you
 haven't already, it will help you with organizing your files.
 
 ## Further improvements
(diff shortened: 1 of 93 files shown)


coriolinus:
That's a good spot and a good fix, but the fix is applied in the wrong place.

Rust's README files are all based on a [template file](https://github.com/exercism/rust/blob/master/config/exercise_readme.go.tmpl), which contains that link. What you'll need to do is update that template, then use configlet (`bin/fetch-configlet`) to regenerate the documentation for every exercise, like:

```sh
bin/configlet generate . 
```

That will update everything in a way that will make Travis happy.

stringparser:
I see! Updated the template and fetched the configlet. 

There is some error shown by `bin/configlet generate .` though

```sh
 ➜  rust (master) ✔ bin/configlet generate .
-> path not found: /Users/stringparser/code/problem-specifications
```

coriolinus:
Oops, forgot configlet didn't do that for you. You need to fetch the problem specifications repo also. Simplest way, from within the rust repo:

```sh
git clone git@github.com:exercism/problem-specifications.git ../problem-specifications
```

stringparser:
Ok, done. Let's see what travis says 👍 👎

petertseng:
guess we were only out of date on anagram and bob. we really should put those into their own commits though. did that. please do not squash when merging

petertseng:
since this is no longer about just hello-world only, hello-world should be removed from commit message and PR title

Issue #841, opened 2019-07-20
palindrome-products code doesn't match spec


BartMassey:
The `palindrome-products` exercise specifies returning both the factors and the product. However, the skeleton and tests depend on

```rust
pub type Palindrome = u64;
```

Probably would be better to fix the skeleton and tests to use

```rust
/// Factors of the palindromic number.
pub struct Palindrome([u32;2]);
```

This assumes that `min` and `max` are passed in as `u32` to guarantee that the product of two of them will fit in a `u64`, which should be done regardless.

Then change the result type to `(Vec<Palindrome>, Vec<Palindrome>)` and the `min()` and `max()` functions can be removed.

I'm too lazy to do this big refactor now. The other alternative would be to change the problem spec to match the current code…

Also, the spec should be changed to say "[…] which are the product of a pair of numbers within that range." This is clearly what is intended. Also note that the problem as specified would give both `(9, 1)` and `(1, 9)` as factors of 9, which is a little redundant but probably OK.

PR #847, opened 2019-07-22
palindrome products: refactor

coriolinus:
Closes https://github.com/exercism/rust/issues/841. Attn @bartmassey. 

Recommend squash-merging when appropriate. 

- Eliminates the trivial `min` and `max` methods
- Removes requirement to accumulate all palindromes
- Simplifies naming
- Updates the exercise problem set to 1.2.0
- Updates the tests to match the description


coriolinus:
In what way do you believe that this doesn't match the description? The only thing I can see that's different at all is the function's inputs, which take a pair of arguments instead of `[T; 2]`; that's not so much a change as a differing convention between the mathematical notation used in the description and the programming notation used within the actual program.

BartMassey:
Maybe I'm just confused. The description I am looking at says:

> Your solution should return the largest and smallest palindromes, along with the
factors of each within the range. If the largest or smallest palindrome has more
than one pair of factors within the range, then return all the pairs.

So the result signature should presumably be something like

```rust
pub struct Palindromic {
    pub product: u64,
    pub factors: Vec<(u64, u64)>,
}

pub fn palindrome_products(min: u64, max: u64) -> Option<(Palindromic, Palindromic)> {
    …
```

to match the description?

coriolinus:
Oh! You're right. Thanks for pointing that out!

I'll update this, probably tonight.

coriolinus:
The example is one possible solution, but it is not the only possible solution.

Students can write their own `Palindrome` struct which may or may not _store_ the value. The requirement is that the `Palindrome` be able to _produce_ the value, and that requirement is enforced by means of the struct's `value` method. I assert that this is a better exercise implementation than forcing students to use the exact structure from the problem specifications.

Because students can implement their own `Palindrome` struct, we can't guarantee that it trivially wraps the `factors` vector. Therefore, I believe that it's better to retain the option.

PR #851, opened 2019-08-08
Fix 'long literal lacking separators' clippy warning & use separator in 5 digit integers in exercises' tests

eddyp:


coriolinus:
Let me amend my statement: I intend to merge this next wednesday, provided that at some point before then you run `configlet generate . --only raindrops` so that Travis passes.

eddyp:
> Let me amend my statement: I intend to merge this next wednesday, provided that at some point before then you run `configlet generate . --only raindrops` so that Travis passes.

Done and they pass :)

petertseng:
Would you be able to do the favour of including https://github.com/exercism/problem-specifications/pull/1565 in the commit message so that it is possible to understand why raindrops was changed? thank you

coriolinus:
Added the link to the problem-specifications PR in the squash-merge message.

petertseng:
That's too bad that it was squashed. It would have been better not to have been.

eddyp:
> > Added the link to the problem-specifications PR in the squash-merge message.

> That's too bad that it was squashed. It would have been better not to have been.

I think I had left the option of maintainers to be able to amend the commits, I was under the impression this meant that squash-merging wasn't necessary. Am I wrong?

(BTW, I also generally dislike squash-merging, especially in cases like these where there are different and logically separate changes.)

coriolinus:
Yup, this was on me. It was possible for me to edit the commit messages
without squashing, but squash-merged instead for simplicity. I didn't think
it would be a big deal.

eddyp:
> Yup, this was on me. It was possible for me to edit the commit messages without squashing, but squash-merged instead for simplicity. I didn't think it would be a big deal.

Ah, OK. Sorry if you thought I was hammering in "you made a mistake" message, I just wanted to be sure a) I left the option open and b) I made the right thing(tm) when I proposed the fixes, so I know to correct in the future, if needed.

petertseng:
I agree that it is not a big deal, but I mentioned it anyway to build good habits for times when it actually would be a big deal.

If it's such a big deal to me that I would make a complaint about it, then I should have made my positions clear beforehand, otherwise it's unreasonable to expect others to know them. I know I failed to do so.

For this PR I would hope that it would have two commits:

* `Fix 'long literal lacking separators' clippy warning`, containing the changes from the submitted commit of the same title plus `Do separations for 5 digit integers, too` (since the latter commit is the same sort of change and doesn't need to stand alone
* `raindrops: simplify README`

Over time I have seen my views are not universal, so not a hill I need to die on or get angry about. But still something I talk about to try to promote.

PR #855, opened 2019-08-13
all-your-base: Fixing test: Empty slice corresponds to zero

CGMossa:
In the template code, it says
```
> /// Notes:
> ///  * The empty slice ( "[]" ) is equal to the number 0.
```
But this test assumes that the empty slice should yield an empty slice, and not zero as noted here.

### DIFF
--- a/exercises/all-your-base/tests/all-your-base.rs
+++ b/exercises/all-your-base/tests/all-your-base.rs
@@ -109,7 +109,7 @@ fn empty_list() {
     let input_base = 2;
     let input_digits = &[];
     let output_base = 10;
-    let output_digits = vec![];
+    let output_digits = vec![0];
     assert_eq!(
         ayb::convert(input_digits, input_base, output_base),
         Ok(output_digits)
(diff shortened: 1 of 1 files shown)


coriolinus:
Thank you @CGMossa for making this change! I agree, this improves the exercise.

You will also need to update `example.rs` so that the example demonstrates this new behavior. Once you have done so, Travis will pass. Once that is done, I will approve this.

petertseng:
In addition to the necessary code changes to example.rs, in src/lib.rs the sentence stating `Never output leading 0 digits.` will have to be changed to something like "Render the value 0 as [0]. Never output leading 0 digits in any other case.", otherwise it would no longer be consistent with the tests.

workingjubilee:
Looks like this would still be good with those followup changes, no merge conflicts.

coriolinus:
Closing due to inactivity.

CGMossa:
Sorry, I have completely forgotten about this. If there is something you know I can do or something, please send me a message.

workingjubilee:
Hello, while I am leaving this closed for now, all you need to do is fix the problems cited by the other maintainer's comments and then this PR can be accepted!

> You will also need to update `example.rs`...

> In addition to the necessary code changes to example.rs, in src/lib.rs...

PR #857, opened 2019-08-19
doubly-linked-list: test drops of large lists

Emerentius:
I've been watching submissions of the new exercise and the very first one (from an independent mode user) uses a node struct like this:

```rust
struct Link<T> {
    value : T,
    next : Option<Box<Link::<T>>>,
    prev : *mut Link::<T>,
}
```

with no separate `Drop` implementation. This is a bad idea because `Box` recursively calls the drop of its contained type. That causes 2 function calls per element in the list, leading to a stackoverflow on large lists.

This PR adds a test provoking exactly that.

### DIFF
--- a/exercises/doubly-linked-list/tests/doubly-linked-list.rs
+++ b/exercises/doubly-linked-list/tests/doubly-linked-list.rs
@@ -249,6 +249,12 @@ fn drop_no_double_frees() {
     assert_eq!(counter.get(), N);
 }
 
+#[test]
+#[ignore]
+fn drop_large_list() {
+    drop((0..2_000_000).collect::<LinkedList<i32>>());
+}
+
 // ———————————————————————————————————————————————————————————
 // Tests for Step 5 (advanced): covariance and Send/Sync
 // ———————————————————————————————————————————————————————————


coriolinus:
I think I don't entirely understand the problem that this solves, yet.

You're saying that the naive default `Drop` behavior is logically correct, but suboptimal because of stack overflows. It's also clear that a cursor-based `Drop` implementation would avoid that problem. 

- > That causes 2 function calls per element in the list,

    I only see one: the recursive `Drop` call. Which one am I missing?

- Why manually call drop, instead of just letting the list fall out of scope?

- Could you link me to the student's implementation? I'd like to test the test, and ensure that it fails for that implementation.

Emerentius:
>You're saying that the naive default Drop behavior is logically correct, but suboptimal because of stack overflows. It's also clear that a cursor-based Drop implementation would avoid that problem.

Yes, although it doesn't have to be cursor-based. The important part is looping instead of recursing. 

Secondly, and much less importantly, I find the design of using one `Box` and one raw pointer inelegant. This solution only reminded me of it, I've seen it before elsewhere. It's an asymmetric implementation for an otherwise perfectly symmetric datastructure. And the only advantage it gives you is its flawed `Drop`.  
I thought about adding this as a point in the mentor notes, but why bother when we can show this limitation automatically with a test. It distracts from the larger issue of memory unsafety.

>* That causes 2 function calls per element in the list,
>>I only see one: the recursive Drop call. Which one am I missing?

Thinking about this, I'm not actually sure how the drop calls are implemented. My thought was this: Box drops the node, node drops the next box. Two steps for a cycle, It doesn't have to be 2 actual calls though and it doesn't seem to be from what I can see on godbolt.

>Why manually call drop, instead of just letting the list fall out of scope?

`Iterator::collect` is marked `must_use` and would warn if we don't do something with it.

>Could you link me to the student's implementation? I'd like to test the test, and ensure that it fails for that implementation.

https://exercism.io/tracks/rust/exercises/doubly-linked-list/solutions/6c8856c67da8483c8add725210e6b6a2

PR #860, opened 2019-08-25
tournament: use default centering for format! parameters

stevensonmt:
The current tests expect that centering odd len strings in an even len column width to place the string to the right (sort of rounding up). The default behavior of the format! macro when passed `{: ^4}` to indicate center alignment padded by spaces for width of 4 is instead to "round down" and place the odd len string to the left with more space to the right. Since the table format appears to be favoring a centered alignment and an even len column width, consider changing the tests to expect the default behavior of the format! macro for center alignment.

### DIFF
--- a/exercises/tournament/tests/tournament.rs
+++ b/exercises/tournament/tests/tournament.rs
@@ -3,103 +3,103 @@ use tournament;
 #[test]
 fn just_the_header_if_no_input() {
     let input = "";
-    let expected = "Team                           | MP |  W |  D |  L |  P";
+    let expected = "Team                            | MP | W  | D  | L  | P";
 
     assert_eq!(tournament::tally(&input), expected);
 }
 
 #[test]
-#[ignore]
+// #[ignore]
 fn a_win_is_three_points_a_loss_is_zero_points() {
     let input = "Allegoric Alaskans;Blithering Badgers;win";
     let expected = "".to_string()
-        + "Team                           | MP |  W |  D |  L |  P\n"
-        + "Allegoric Alaskans             |  1 |  1 |  0 |  0 |  3\n"
-        + "Blithering Badgers             |  1 |  0 |  0 |  1 |  0";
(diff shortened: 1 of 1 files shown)


coriolinus:
A key assumption motivating this PR is that the numbers are meant to be centered. I assert that they are meant to be right-aligned. There is no text in the exercise which explicitly mentions the alignment.

`example.rs` demonstrates that using default alignment, which is right-aligned for numbers, produces proper results. 

I am inclined to close this PR without merge. I will do so not earlier than Wednesday and not later than Friday unless convincing counterarguments come up in the meantime.

exercises/tournament/tests/tournament.rs
```diff
@@ -3,103 +3,103 @@ use tournament;
...
-#[ignore]
+// #[ignore]
 fn a_win_is_three_points_a_loss_is_zero_points() {
     let input = "Allegoric Alaskans;Blithering Badgers;win";
     let expected = "".to_string()
-        + "Team                           | MP |  W |  D |  L |  P\n"
-        + "Allegoric Alaskans             |  1 |  1 |  0 |  0 |  3\n"
-        + "Blithering Badgers             |  1 |  0 |  0 |  1 |  0";
+        + "Team                            | MP | W  | D  | L  | P\n"
+        + "Allegoric Alaskans              | 1  | 1  | 0  | 0  | 3\n"
```
petertseng: I will focus on one column.

Before:

```
before: |  1 |
after : | 1  | 
```

I think the previous (Before) version looks better, for the reason: imagine for a moment what happens if a two-digit number needs to be entered into this column. We'll disregard for a moment that a two-digit number never appears in any output in this exercise. If it were present, it would would have:

```
before: |  1 |
after : | 1  | 
2digit: | 10 | 
```

the `before` version seems to make more sense, since it is easier to scan the columns and understand that `10` is of a larger order of magnitude than `1`. The `after` version makes that harder to see.

PR #862, opened 2019-08-30
travis: reject invalid UUIDs

petertseng:
Want to prevent any invalid UUIDs from being entered.

Want to put this in configlet lint, but can't:
exercism/configlet#99
exercism/configlet#168
So it will go in individual tracks' .travis.yml for now.

It appears that the site will accept pretty much arbitrary strings as
UUIDs for now, but we want to make less work for ourselves when valid
UUIDs are required.


petertseng:
It shall be done (moved into a script)

petertseng:
I have pretty high confidence in the tests, enough that I would consider just merging it now, but I'll leave it open for about 16 hours in case additional comments pop up.

PR #863, opened 2019-08-30
simple-linked-list: test_peek_returns_head_element tests head element

### LINKED ISSUE #834: simple-linked-list: test_peek_returns_head_element doesn't actually test what it claims to
The `peek` of simple linked list is defined to return head.

However, it will also work if the user has created a stack, and peek returns the *last* element.
(In general, simple linked list's definition is a little confusing - pop and peek typically return the same element, but pop here is defined to return tail, and peek to peek at head).

This test should test with 2 different elements, or a new test should be added.

https://github.com/exercism/rust/blob/da21df6e8c8f7406ebe0c4ae2b68631b84fa27b1/exercises/simple-linked-list/tests/simple-linked-list.rs#L42-L50

ocstl:
This PR aims to close #834, making sure that the `peek` method always references the head element and doesn't remove it (like `pop`).

### DIFF
--- a/exercises/simple-linked-list/tests/simple-linked-list.rs
+++ b/exercises/simple-linked-list/tests/simple-linked-list.rs
@@ -30,7 +30,7 @@ fn test_pop_decrements_length() {
 
 #[test]
 #[ignore]
-fn test_pop_returns_last_added_element() {
+fn test_pop_returns_head_element_and_removes_it() {
     let mut list: SimpleLinkedList<u32> = SimpleLinkedList::new();
     list.push(1);
     list.push(2);
@@ -41,7 +41,7 @@ fn test_pop_returns_last_added_element() {
 
 #[test]
 #[ignore]
-fn test_peek_returns_head_element() {
+fn test_peek_returns_reference_to_head_element_but_does_not_remove_it() {
     let mut list: SimpleLinkedList<u32> = SimpleLinkedList::new();
     assert_eq!(list.peek(), None, "No element should be contained in list");
     list.push(2);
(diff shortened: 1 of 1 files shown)


ocstl:
Good point on linking to the issue.

I'm open to splitting it, I just haven't found a clean way to do so.

petertseng:
I wonder if some names could be changed to avoid that confusion. Notice that this `peek` test is the only time that "head" is referred to, and "tail" is not referred to at all. So perhaps `peek` could use the same terminology as previous tests. Something like:

```
test_pop_removes_last_added_element_and_removes_it
test_peek_returns_last_added_element_but_does_not_remove
```

coriolinus:
I think those name changes are a good idea. If incorporated into this PR, changing those names should be a separate commit.

ocstl:
> I wonder if some names could be changed to avoid that confusion. Notice that this `peek` test is the only time that "head" is referred to, and "tail" is not referred to at all. So perhaps `peek` could use the same terminology as previous tests. Something like:
> 
> ```
> test_pop_removes_last_added_element_and_removes_it
> test_peek_returns_last_added_element_but_does_not_remove
> ```

I think you mean `test_pop_returns...`, but I get your point and the name change seems like a good idea. I'm not entirely happy with the second name though, since the new head element following a `pop` is not the last added element, but the next-to-last added element. And so on.

On the other hand, the implementation hints clearly mention `head`:
> The head field points to the first element (Node) of this linked list.

Plus, "head" and "tail" are pretty much basic nomenclature for linked lists. So, I feel it makes more sense to use "head element" in the context then "last_added_element". For example:

> test_pop_returns_head_element_and_removes_it
> test_peek_returns_reference_to_head_element_but_does_not_remove_it

petertseng:
Yes, thanks for correcting my mistakes.

ocstl:
Thanks to you both for reviewing it and offering suggestions.

PR #865, opened 2019-09-02
Add test for single macro argument

ssomers:
I was mystified that replacing `$(...),*` with `$(...),+` in my macro rule seemed to work fine.


ssomers:
Thank you, for creating the site. It looks very neat and proved quite useful.

PR #874, opened 2019-09-24
macros: Update README.md

DarthStrom:


coriolinus:
Good catch, thanks for spotting and contributing this @DarthStrom! 

Unfortunately, CI won't pass it yet, because our README files are generated from separate sources; you've edited the output, but not the input. You'll also need to edit [`.meta/description.md`](https://github.com/exercism/rust/blob/2d6018e1b1dd3e15bc5bdc13ae877769785fae94/exercises/macros/.meta/description.md)` in sync with this edit to make Travis happy, which will make it possible for me to merge.

Instead of editing two files, it would also be possible to just edit `description.md` and then run `bin/configlet generate . --only macros`, which will regenerate the README from the inputs. Up to you which approach you prefer.

DarthStrom:
Bah, was hoping it would be as easy as updating the text through github's edit mechanism.  Ah well, I'll update the input file as well.

coriolinus:
Thanks! This looks good to me. As it's a documentation change, I'll merge now without waiting for other maintainer input.

PR #876, opened 2019-09-25
Move bob later into the track

iHiD:
Bob is extremely painful to mentor due to the large amount of permutations of solutions. We're therefore moving it to later in tracks across Exercism, or removing it from core altogether. This PR moves it a few notches down for now. We can analyse the data and determine if it needs to go further down or be removed from core altogether at a later stage.

(cc @kytrinyx)

### DIFF
--- a/config.json
+++ b/config.json
@@ -153,27 +153,6 @@
       ]
     },
     {
-      "slug": "bob",
-      "uuid": "38ef1802-2730-4f94-bafe-d2cd6b3e7f95",
-      "core": true,
-      "unlocked_by": null,
-      "difficulty": 1,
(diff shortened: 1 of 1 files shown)


coriolinus:
The core exercise immediately preceding `bob` in this PR is `sublist`, a difficulty-7 exercise. IMO students who have successfully completed `sublist` are likely to get much out of `bob`. I would therefore prefer to just make `bob` into a side exercise.

iHiD:
@coriolinus OK - that's fine. I think the whole core is going to need a lot of much simpler exercises in it, but we can come to that at a later date. I'll open a new PR.

PR #877, opened 2019-09-25
Demotes bob to a side exercise

iHiD:
Demotes bob to a side exercise.

See https://github.com/exercism/rust/pull/876 for rationale.


petertseng:
please don't forget to delete branches

PR #879, opened 2019-09-28
triangle: Add invalid floating-point tests

jalovatt:
For completeness' sake, and just in case someone comes up with a crazy solution that would pass everything else but break at this point, I've added a pair of tests for invalid triangles to the bonus floating-point section.


petertseng:
Thank you for the contribution and sorry I "slightly" exceeded 72 hours (by 24 hours).

A reminder for all that when https://github.com/exercism/problem-specifications/pull/1598 is evaluated, if it results in significant changes to these tests, consider whether to reflect them here. For the time being, it looks like there would not be a significant change.

PR #884, opened 2019-10-04
raindrops: pick topics from problem-specifications

azul:
Using topics from https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt

* `conditionals` - flow control depending on wether numbers have a factor or not
* `strings` - this is all about composing a string
* `type_conversion` - need to handle `&str` and `String` strings
* `variables` - building up the string is easiest when usign a variable.

### DIFF
--- a/config.json
+++ b/config.json
@@ -37,8 +37,10 @@
       "unlocked_by": "hello-world",
       "difficulty": 1,
       "topics": [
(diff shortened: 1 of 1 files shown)


petertseng:
Great, thank you.

Issue #679, opened 2018-10-20
exercise util should use the same template 


coriolinus:
When using `exercise generate`, the generated file is different depending on whether or not canonical data exists for that exercise:

- if an exercise has canonical data, the tests module has a nice explanatory doccomment
- if an exercise does not have canonical data, the tests module does not have the module doccomment 

Both cases should use the same basic template, with the doccomment.

dvermd:
I have a PR for this one.
There seems to be some redundancy in the template as in `[utility][utility]` and `[canonical data][canonical data]`.
Should I had another commit to dedup these ?

PR #885, opened 2019-10-07
util/generate: Use the same template for all generations

dvermd:
This PR also do not repeat the test name creation.

Resolves: #679


util/exercise/src/cmd/generate.rs
```diff
@@ -246,7 +230,33 @@ pub fn generate_exercise(exercise_name: &str, use_maplit: bool) -> Result<()> {
...
+                )
+                .into_bytes(),
+            )?;
+            if use_maplit {
+                test_file.write_all(b"#[macro_use]\nextern crate maplit;\n")?;
+            }
+
+            test_file.write_all(
+                &format!(
+                    "extern crate {escaped_exercise_name};\n\
```
coriolinus: As this track now uses the 2018 edition by default, we can remove the `extern crate` line.

dvermd: I'll remove that.
Following this advice, what about also changing 
```
#[macro_use]
extern crate maplit;
```
to 
`use maplit::*;`

coriolinus:
It's not a bad idea to remove the extern crate, but we shouldn't be
encouraging people to use global imports. I think

use maplit::maplit;

will do what we need, but that'll need to be tested.

dvermd:
A quick example from maplit shows that `use maplit::maplit;` is not working and one should write `use maplit::hashmap;` if the `hashmap!` macro is to be used.

Right now, there's 5 macros in the maplit crate. I might import them all explicitly and then `cargo build` will warn about unused imports.
I'm not happy with this but this is the best I can think right now.

Another point : should I remove one of the "[utility][utility]" in the template ? I didn't manage to find documentation on a specific behaviour related to "[]" in doc comments.

coriolinus:
use maplit::hashmap;

is correct. That was my mistake earlier. I don't think the generator uses
any other literals, so we should not import the others.

You should not remove the "[utility][utility]"; that's a markdown external
hyperlink, which inlines the link which appears lower in the comment.

PR #804, opened 2019-03-05
Reordered exercises for Round 1 of track anatomy.

TheDarkula:

Issue #524, opened 2018-05-03
Authors in Cargo.toml


christianpoveda:
Several exercises (20 of them) have an author in their Cargo.toml files. Of course no one submits its Cargo.toml file but I think it would be better if that field was omitted. The exercises are:

```
accumulate/Cargo.toml
4:authors = ["sacherjj <sacherjj@gmail.com>"]

binary-search/Cargo.toml
4:authors = ["<marco.stahl@mailbox.org>"]

book-store/Cargo.toml
4:authors = ["Peter Goodspeed-Niklaus <peter.r.goodspeedniklaus@gmail.com>"]

crypto-square/Cargo.toml
4:authors = ["Peter Goodspeed-Niklaus <peter.r.goodspeedniklaus@gmail.com>"]

clock/Cargo.toml
4:authors = ["sacherjj <sacherjj@gmail.com>"]

decimal/Cargo.toml
4:authors = ["Peter Goodspeed-Niklaus <peter.r.goodspeedniklaus@gmail.com>"]

macros/Cargo.toml
4:authors = ["Peter Goodspeed-Niklaus <peter.r.goodspeedniklaus@gmail.com>"]

nth-prime/Cargo.toml
4:authors = ["sacherjj <sacherjj@gmail.com>"]

nucleotide-codons/Cargo.toml
4:authors = ["Peter Minten <peter@pminten.nl>"]

palindrome-products/Cargo.toml
4:authors = ["Kirill Meng <kirill.meng@hotmail.de>"]

pig-latin/Cargo.toml
4:authors = ["sacherjj <sacherjj@gmail.com>"]

poker/Cargo.toml
4:authors = ["Peter Goodspeed-Niklaus <peter.r.goodspeedniklaus@gmail.com>"]

prime-factors/Cargo.toml
4:authors = ["sacherjj <sacherjj@gmail.com>"]

proverb/Cargo.toml
4:authors = ["sacherjj <sacherjj@gmail.com>"]

protein-translation/Cargo.toml
4:authors = ["Peter Minten <peter@pminten.nl>"]

pythagorean-triplet/Cargo.toml
4:authors = ["sacherjj <sacherjj@gmail.com>"]

reverse-string/Cargo.toml
4:authors = ["Peter Goodspeed-Niklaus <peter.r.goodspeedniklaus@gmail.com>"]

run-length-encoding/Cargo.toml
4:authors = ["zombiefungus <divmermarlav@gmail.com>"]

rotational-cipher/Cargo.toml
[...]

PR #886, opened 2019-10-08
Port High Scores

filalex77:
This change is [Reviewable](https://reviewable.io/reviews/exercism/rust/886)


exercises/high-scores/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[dependencies]
+
+[package]
+authors = ["Oleksii Filonenko <brightone@protonmail.com>"]
```
coriolinus: Please remove this line; policy since https://github.com/exercism/rust/issues/524 has been to leave it out.

exercises/high-scores/example.rs
```diff
@@ -0,0 +1,30 @@
+#[derive(Debug)]
+pub struct HighScores {
+    scores: Vec<u32>,
+}
+
+impl HighScores {
+    pub fn new(scores: &[u32]) -> Self {
+        Self {
+            scores: scores.to_vec(),
```
coriolinus: Instead of converting the scores input to an owned vector, this exercise could serve as a simple introduction to lifetimes by retaining a reference to the input slice. I think that direction would improve the quality of this exercise.

Necessary changes to implement that:

- update this example
- add a `.meta/hints.md` file recommending that students retain a reference to the input, and regenerate the readme
- update the topics in `config.json` to include `lifetimes`

exercises/high-scores/example.rs
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    pub fn personal_best(&self) -> u32 {
+        *self.scores.iter().max().unwrap()
+    }
+
+    pub fn personal_top_three(&self) -> Vec<u32> {
+        let mut sorted = self.scores.clone();
+        sorted.sort();
+        sorted.iter().rev().take(3).cloned().collect()
```
coriolinus: We can't avoid the first instance of cloning the list, since we otherwise have no way to sort it. Still, we can at least avoid the second clone/collect cycle.

```suggestion
        sorted.reverse();
        sorted.truncate(3);
        sorted
```

exercises/high-scores/src/lib.rs
```diff
@@ -0,0 +1,27 @@
+#[derive(Debug)]
+pub struct HighScores;
+
+impl HighScores {
+    pub fn new(_: &[u32]) -> Self {
```
coriolinus: You refer to the `scores` variable below, but you haven't set it yet.

```suggestion
    pub fn new(scores: &[u32]) -> Self {
```

exercises/high-scores/example.rs
```diff
@@ -0,0 +1,30 @@
...
+            scores: scores.to_vec(),
+        }
+    }
+
+    pub fn scores(&self) -> &[u32] {
+        &self.scores
+    }
+
+    pub fn latest(&self) -> u32 {
+        *self.scores.last().unwrap()
```
coriolinus: We should not encourage students to use `unwrap` unconditionally like this. We should change the API to make the struct safer.

There are at least two possible ways to change this API to improve its safety:

- change the return type of `latest` and `personal_best` to `Option<u32>`
- change the return type of `new` to `Option<Self>`, returning None when the input list is empty

I do not have a preference between those two options. However, we should definitely add an appropriate test in which the input slice is empty to ensure that the code doesn't panic, as it would with `unwrap`.

exercises/high-scores/Cargo.toml
filalex77: Okay, sure. `exercise generate` put it there for me. Maybe it should get removed automatically?

exercises/high-scores/example.rs
filalex77: I did it with lifetimes first, and then reverted. On the second look, it's probably not that difficult with this exercise. I'll update to using references.

exercises/high-scores/example.rs
filalex77: Truncate is coll, didn't know about it. Thanks!

exercises/high-scores/src/lib.rs
filalex77: Oops, saw the "use more placeholders" recommendation and forgot that it's used. Good catch!

exercises/high-scores/example.rs
filalex77: I think the first variant is a bit clearer. And the test is a good idea.

exercises/high-scores/Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[package]
+edition = "2018"
+name = "high-scores"
+version = "4.0.0"
```
petertseng: just confirming that I looked at this and it matches the corresponding problem-specifications version :+1: . I haven't done a full review yet, but if I don't get to it in time I trust other maintainers' judgment, so no need to wait for me.

exercises/high-scores/tests/high-scores.rs
```diff
@@ -11,14 +11,21 @@ fn test_list_of_scores() {
 #[ignore]
 fn test_latest_score() {
     let high_scores = HighScores::new(&[100, 0, 90, 30]);
-    assert_eq!(high_scores.latest(), 30);
+    assert_eq!(high_scores.latest(), Some(30));
 }
 
 #[test]
 #[ignore]
-fn test_personal_best() {
```
coriolinus: I am surprised to see that you removed `test_personal_best` entirely. While I entirely approve of the addition of `test_latest_score_empty`, `test_personal_best_empty`, and `test_personal_top_three_empty`, I think we should not remove this canonical test case.

filalex77: I'm pretty sure it was by accident. Thanks a lot!

config.json
```diff
@@ -1123,6 +1123,18 @@
       "difficulty": 0,
       "topics": null,
       "deprecated": true
+    },
+    {
+      "slug": "high-scores",
```
petertseng: I know that we currently have no exercise ordering check in the tests right now because https://github.com/exercism/rust/pull/804#issuecomment-470197811 . But maybe it makes sense to move high-scores above the deprecated exercise `hexadecimal`?

This change is optional - feel free to merge without having made this change without seeking approval from me on whether it is OK to not make the change

coriolinus: Good point! I hadn't actually looked at where in the configuration file this had been added. 

I agree, this exercise should fall in the region after the exercise which unlocks it (which should not be null) and before the subsequent core exercise.

PR #889, opened 2019-10-10
custom-set: remove Debug trait bound from stub

### LINKED ISSUE #631: What trait bounds are appropriate for stubs?
As we add stubs, we should ask what trait bounds are appropriate.

Decision made in https://github.com/exercism/rust/pull/497 (simple-linked-list https://github.com/exercism/rust/pull/497/commits/c5f7a433c5e954aa78162bfcab14eab96510bc4f): Not to add `Debug`.

Should `Debug` remain in `custom-set`? https://github.com/exercism/rust/pull/605#discussion_r211114280

Not to put `Debug` in `circular-buffer`: https://github.com/exercism/rust/pull/608#discussion_r210513499

Discussion in `sublist`: Be minimally restrictive: https://github.com/exercism/rust/pull/610#discussion_r211059332

Summary: **I think we are indeed trying to be consistent about being minimally restrictive.** Please chime in if you think we should do differently, or for any further input on the topic.

Current trait bounds (the `grep` pattern may not be the best, feel free to suggest a better)

```
$ git grep '<[^>]:.*>' exercises/*/src/lib.rs 
exercises/custom-set/src/lib.rs:impl<T: Debug> CustomSet<T> {
exercises/react/src/lib.rs:impl<T: Copy + PartialEq> Reactor<T> {
exercises/react/src/lib.rs:    pub fn create_compute<F: Fn(&[T]) -> T>(
exercises/react/src/lib.rs:    pub fn add_callback<F: FnMut(T) -> ()>(
exercises/simple-linked-list/src/lib.rs:impl<T: Clone> SimpleLinkedList<T> {
exercises/simple-linked-list/src/lib.rs:impl<'a, T: Clone> From<&'a [T]> for SimpleLinkedList<T> {
```

If there is no further input, I think the way to close this issue is to remove `Debug` from `custom-set`. You will then need to remove references to variables in `unimplemented!` messages, so will have to replace them with underscored names.

coriolinus:
Closes #631.

### DIFF
--- a/exercises/custom-set/src/lib.rs
+++ b/exercises/custom-set/src/lib.rs
@@ -1,4 +1,3 @@
-use std::fmt::Debug;
 use std::marker::PhantomData;
 
 #[derive(Debug, PartialEq)]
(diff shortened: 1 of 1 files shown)


coriolinus:
IMO the descriptive messages aren't adding much; in my view, they're a hack to use the variable more than useful text on their own. That's why I removed them here.

Issue #529, opened 2018-05-08
Standardize topics in config.json


coriolinus:
The new `configlet fmt` goes some way toward standardizing the format of topics in `config.json`, but that just reveals the extent to which they could be further refined. In ~80 current exercises, we have 111 distinct topics:

```sh
$ jq '.exercises[] | .topics[]?' config.json --raw-output | sort -u | wc -l
111
```

I won't paste the whole list here, but several categories which we may wish to consider removing / merging into others:

- 9 entries which end in `_optional` (`| grep option`)
- 1 entry which ends in `_probably` (`| grep prob`)
- 35 entries which appear to be pluralizations of other entries in the list (`| grep -P s$`)
- `goofy_bowling_logic`
- 70 entries which appear only once in the file (`sort | uniq -c | grep -P '\b1\b'`)

coriolinus:
Related: what exactly are these topics actually _for_? Are they presented to the student at all, in any way? Are there plans to do so? If not, could we just eliminate all of them?

petertseng:
> Are they presented to the student at all, in any way?

Not in v1, but in https://v2.exercism.io/tracks/rust I can see the topics.

In the face of doubts that v2 will ever become live (this judgment is pure speculation based on the level of activity publicly observable via https://github.com/exercism/prototype and is not meant as a personal judgment against any contributor to v2 nor their dedication and motivation toward getting it live), I'll give an answer that might possible be applicable to v1 as well: Perhaps seeing them in config.json helps maintainers decide where to place exercises?

petertseng:
The current documentation on topics is https://github.com/exercism/docs/blob/master/language-tracks/configuration/exercises.md#topics. If I understand history correctly, the concept of Exercism having topics at all stemmed from this very track, in https://github.com/exercism/rust/issues/127 .

I haven't put much thought into them since I have not yet been convinced it is a fruitful use of my time. Left to my own devices, the only topics I would use are probably things like option and result, where the decision on whether to include it is automatically deducible based on its presence in the test file or stub. Then again, one may argue that if it's automatically deducible, why bother manually recording it

coriolinus:
One option would be to write a little utility which guaranteed that every
topic chosen appeared in
https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt .
This would relieve us of the burden of deciding what's a legitimate topic,
and what isn't.

coriolinus:
A canonical topics list in the problem specifications repo would be a
[...]

PR #891, opened 2019-10-16
topics: standardize `strings` topic

azul:
There were quite a few topics that were string related
and not listed in the [topics list](https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt).

* Replace all of them with `strings` which is in the list.
* Add `parsing` if the topic was `string_parsing`.

### DIFF
--- a/config.json
+++ b/config.json
@@ -65,7 +65,7 @@
       "topics": [
         "case",
         "loop",
-        "string_concatenation",
+        "strings",
         "vector_optional"
       ]
     },
@@ -141,8 +141,8 @@
       "difficulty": 1,
       "topics": [
         "iterator",
-        "str",
-        "string"
+        "str_vs_string",
+        "strings"
       ]
     },
     {
@@ -175,7 +175,7 @@
       "difficulty": 1,
       "topics": [
         "chars",
-        "string_functions"
+        "strings"
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


azul:
working towards #529

config.json
```diff
@@ -308,7 +307,7 @@
       "topics": [
         "match",
         "result",
-        "str_vs_string",
```
coriolinus: I think we should not remove the `str_vs_string` topic from any exercises. While that's not in the canonical topics list, it's an important distinction that is unique to Rust; it's useful to retain even if we also add the `strings` topic.

azul:
I made the changes @coriolinus requested and pushed again.

config.json
```diff
@@ -636,7 +640,7 @@
         "chars",
         "iterators",
         "primitive_types",
-        "str_vs_string"
```
coriolinus: You're still removing `str_vs_string` here, which I think should remain.

config.json
```diff
@@ -648,7 +652,7 @@
       "topics": [
         "ascii",
         "chars",
-        "str_vs_string"
```
coriolinus: Here also.

azul:
@coriolinus thanks for pointing them out. Should be fixed now.

PR #892, opened 2019-10-19
add links to readme describing where new contributors can find pending issues

coriolinus:
The issues tab is fairly well-known, but the track status board is both more obscure and more useful for potential contributors looking for a way to helpfully contribute which doesn't require too much effort.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,8 @@ We welcome pull requests of all kinds. No contribution is too small.
 
 We encourage contributions that provide fixes and improvements to existing exercises. Please note that this track's exercises must conform to the Exercism-wide standards described in the [documentation](https://github.com/exercism/docs/tree/master/language-tracks/exercises). If you're unsure about how to make a change, then go ahead and open a GitHub issue, and we'll discuss it.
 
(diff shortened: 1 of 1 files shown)


coriolinus:
I don't object to that policy; I don't feel strongly either way about that. I wouldn't object if you wanted to write a CI script which enforced that policy. 

If I were to write such a script, I think I'd do it like this:

- for each `Cargo.toml` in the changed files in the PR
- if the version has changed 
- grep the commit messages in the PR for links to pulls in the `problem-specifications` repo
- fail if no such link appears

Such a script wouldn't be an exhaustive proof that all appropriate PRs have been linked, but I suspect that if we force people to find at least one link, they'll do the right thing and put in all the links.

ocstl:
I think it's a good idea, but, as @petertseng rightly pointed out, this could become burdensome for some updates.

I feel it makes a lot of sense to at least point to the most recent PR in the `problem-specifications` repo at the time of the update instead of (or rather, in addition to) the current file, to make it less confusing if the problem-specifications have since changed. It would make it easier to find the relevant PRs as well.

PR #894, opened 2019-10-19
exercise tool: update dependencies

coriolinus:
This slightly improves compilation time, as the total number of \
dependencies is reduced.


coriolinus:
I built the tool before and after the update, and looked at the text portion of the progress bar.

PR #897, opened 2019-10-24
affine-cipher: Add exercise

ktomsic:
This adds the affine-cipher exercise from [the problem specifications repo](https://github.com/exercism/problem-specifications/tree/master/exercises/affine-cipher).


exercises/affine-cipher/src/lib.rs
```diff
@@ -0,0 +1,18 @@
+/// While the problem description indicates a return status of 1 should be returned on errors,
+/// it is much more common to return a `Result`, so we provide an error type for the result here.
```
coriolinus: Your reasoning here is precisely correct, but you should be aware of https://github.com/exercism/problem-specifications/pull/1614. At your discretion, you can choose to wait for that issue to be resolved before moving forward with this PR; just let me know. If you'd prefer to get this merged as quickly as possible, that is acceptable.

exercises/affine-cipher/tests/affine-cipher.rs
```diff
@@ -0,0 +1,152 @@
...
+        "swxtj npvyk lruol iejdc blaxk swxmh qzglf"
+    )
+}
+
+#[test]
+#[ignore]
+fn encode_with_a_not_coprime_to_m() {
+    let expected_error = AffineCipherError::NotCoprime(6);
+    match encode("This is a test.", 6, 17) {
+        Err(AffineCipherError::NotCoprime(6)) => (),
```
coriolinus:
```suggestion
        Err(expected_error) => (),
```

Note that this will require `#[derive(Clone,Copy)]` on `AffineCipherError`.

config.json
```diff
@@ -572,6 +572,21 @@
...
+      "slug": "affine-cipher",
+      "uuid": "2a1dcf38-ec05-4b24-a2e2-2e5b3595f3f0",
+      "core": false,
+      "unlocked_by": "atbash-cipher",
+      "difficulty": 4,
+      "topics": [
+        "ascii",
+        "chars",
+        "iterators",
+        "primitive_types",
```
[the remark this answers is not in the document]
ktomsic: I actually just copied the tags from the other ciphers. "rotational-cipher" has "primitive_types", too. I can remove it if you'd like.

Although it does exercise a bit of conversion between primitives. Encoding / Decoding individual `char`s entails converting to an integral type and back, and if you choose too small of an integral type, you have to worry about overflow.

I'll let you make the call though.

exercises/affine-cipher/src/lib.rs
ktomsic: I can wait, and I'll update the phrasing when the specification's description.md gets updated.

exercises/affine-cipher/tests/affine-cipher.rs
```diff
@@ -0,0 +1,152 @@
...
+        encode("The quick brown fox jumps over the lazy dog.", 17, 33).unwrap(),
+        "swxtj npvyk lruol iejdc blaxk swxmh qzglf"
+    )
+}
+
+#[test]
+#[ignore]
+fn encode_with_a_not_coprime_to_m() {
+    let expected_error = AffineCipherError::NotCoprime(6);
+    match encode("This is a test.", 6, 17) {
```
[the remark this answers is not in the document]
ktomsic: Thanks!

exercises/affine-cipher/tests/affine-cipher.rs
ktomsic: Will do.

config.json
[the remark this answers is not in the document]
coriolinus: Might as well leave it; that justification seems sound to me.

exercises/affine-cipher/tests/affine-cipher.rs
ktomsic: The suggested approach doesn't quite work -- it binds a local `expected_error` variable to whatever `Err()` contains rather than matching against the outer `expected_error` -- but it did get me thinking about this a little differently.

While we can't pattern match on variables, we *can* pattern match on constants (provided they implement `Eq` and `PartialEq`), so I changed `expected_error` into a `const` and added the appropriate derives.

Hopefully that's a little bit closer to what you're looking for.

exercises/affine-cipher/tests/affine-cipher.rs
```diff
@@ -0,0 +1,146 @@
+use affine_cipher::*;
+
+#[test]
+fn encode_yes() {
+    assert_eq!(encode("yes", 5, 7).unwrap(), "xbt")
```
petertseng: There was once upon a time in https://github.com/exercism/rust/issues/338 where we thought all assertions of the form `assert_eq!(a.unwrap(), b)` should instead be written as `assert_eq(a, Ok(b))` because it gave better test outputs when the implementation said `Err` when it should instead give `Ok`.

However, I have just given this a try, and here is a comparison of the two:

```
---- tests::dont_unwrap_and_expect_ok_string stdout ----
thread 'tests::dont_unwrap_and_expect_ok_string' panicked at 'assertion failed: `(left == right)`
  left: `Err(NotCoprime(2))`,
 right: `Ok("hello")`', src/lib.rs:16:9

---- tests::do_unwrap_and_expect_string stdout ----
thread 'tests::do_unwrap_and_expect_string' panicked at 'called `Result::unwrap()` on an `Err` value: NotCoprime(2)', src/libcore/result.rs:1084:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

I deem this sufficiently helpful (shows what was attempted to be unwrapped) that we no longer need to consider https://github.com/exercism/rust/issues/338. In fact this has always been the case (it had always shown the `Err` that was `unwrap`ped, even at the time of filing https://github.com/exercism/rust/issues/338), so it's simply that this track has generally gotten better at having good `Error`s, and this caused me to change my mind on the issue.

I think the former is still slightly more helpful (shows what was actually expected) and should be used in situations where it is possible without inconvenience, but note that since this is a `Result<String, AffineCipherError>`, we don't seem to just be able to write `Ok("xbt")`, but instead would have to write `Ok("xbt".to_string())` or something. I deem that enough of an inconvenience so as to say it need not happen here. I leave it as a comment in case anyone feels differently.

exercises/affine-cipher/tests/affine-cipher.rs
```diff
@@ -0,0 +1,146 @@
...
+        decode("vszzm    cly   yd cg    qdp", 15, 16).unwrap(),
+        "jollygreengiant"
+    )
+}
+
+#[test]
+#[ignore]
+fn decode_with_a_not_coprime_to_m() {
+    const EXPECTED_ERROR: AffineCipherError = AffineCipherError::NotCoprime(13);
+    match decode("Test", 13, 11) {
```
petertseng: although the canonical data has `5` in place of `11` here, since this is meant to error anyway, I have no reason to prefer one value over the other. So I accept this difference.

exercises/affine-cipher/tests/affine-cipher.rs
coriolinus: Yes, it is; that looks great! Sorry for having lead you astray with the variable binding.

PR #898, opened 2019-10-24
acronym: Update to 1.7.0

ktomsic:
Acronym 1.7.0 has several new test cases and removed a couple of others.
Of particular note is the new `apostrophes()` test case, which broke the
original example implementation.

### DIFF
--- a/exercises/acronym/tests/acronym.rs
+++ b/exercises/acronym/tests/acronym.rs
@@ -31,24 +31,57 @@ fn punctuation() {
 
 #[test]
 #[ignore]
-fn all_caps_words() {
+fn all_caps_word() {
+    assert_eq!(acronym::abbreviate("GNU Image Manipulation Program"), "GIMP");
+}
+
+#[test]
+#[ignore]
+fn all_caps_word_with_punctuation() {
     assert_eq!(acronym::abbreviate("PHP: Hypertext Preprocessor"), "PHP");
 }
 
 #[test]
 #[ignore]
-fn non_acronym_all_caps_word() {
+fn punctuation_without_whitespace() {
     assert_eq!(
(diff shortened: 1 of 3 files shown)


exercises/acronym/tests/acronym.rs
```diff
@@ -19,36 +19,57 @@ fn lowercase_words() {
...
-        acronym::abbreviate("GNU Image Manipulation Program"),
-        "GIMP"
+        acronym::abbreviate("Rolling On The Floor Laughing So Hard That My Dogs Came Over And Licked Me"),
+        "ROTFLSHTMDCOALM"
     );
 }
 
 #[test]
 #[ignore]
-fn hyphenated() {
```
coriolinus: please also restore this test case

exercises/acronym/tests/acronym.rs
```diff
@@ -19,36 +19,57 @@ fn lowercase_words() {
 
 #[test]
 #[ignore]
-fn camelcase() {
```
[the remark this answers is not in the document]
ktomsic: Done!

exercises/acronym/tests/acronym.rs
ktomsic: Not sure if you meant a different test case, but `hyphenate()` was just renamed to `punctuation_without_whitespace()`. They both check that "Complementary metal-oxide semiconductor" becomes "CMOS".

coriolinus: Oh, oops! I missed that, sorry. Given a bunch of cases, I just grepped through the changes; I didn't think to grep for the actual acronym. I agree, consistency in name with the canonical data makes more sense.

PR #900, opened 2019-10-25
Clarify Rust indices from 0 in saddle-points and regenerate READMEs

### LINKED ISSUE #791: saddle-points: Reconcile 1-based indices in README with 0-based indices in tests
https://github.com/exercism/problem-specifications/pull/1429 which was applied to this track in https://github.com/exercism/rust/pull/789 uses one-based indices in the https://github.com/exercism/rust/blob/master/exercises/saddle-points/README.md. The track still uses zero-based indices in the https://github.com/exercism/rust/tree/master/exercises/saddle-points/tests . This inconsistency is a potential source of student confusion.

Various solutions:

1. Change the tests to be one-indexed.
2. Maintain our own README with zero-based indices. You do this by placing the README we want in `.meta/description.md`. Not recommended.
    * We cannot just use hints.md since that only allows additive changes. This is a modification of existing text, which hints.md does not support.
3. Keep both the README and tests as is, but insert a statement somewhere telling students to be aware of the discrepancy.
4. Suggest to problem-specifications that no, really, they should go back to using zero-based indices.
5. Your suggestion here.

I do not recommend 2 (maintain our own README) so if you want to go this route you have to convince me why we should be in the business of maintaining our own READMEs.

I have no specific preferences or recommendations between any of the other options.

workingjubilee:
Closes #791 

Of the 5 choices in that issue, results of the discussion were (effectively):
"Do not do 2."
"Do 2 or 3 to be consistent with Rust."
"Do 3 for now since it is easy, then 1 later."
"1 would be possible using a little more code."

As option 2 effectively has a negative vote, and 3 has two votes for, modifying hints.md comes out as the maintainer-preferred solution. This PR implements that and regenerates the README appropriately. Or rather, READMEs. While doing so, I discovered two other READMEs were dated and their latest version was more readable, so I included them in this PR.

### DIFF
--- a/exercises/saddle-points/.meta/hints.md
+++ b/exercises/saddle-points/.meta/hints.md
@@ -1,3 +1,11 @@
+## Rust Indices Start At 0
+
+By convention, ordered sequences of values in Rust have their contents numbered
+("indexed") starting from 0. This applies regardless of what the rest of the
+exercise description in this README says, such as references to indices that
+start at 1, so you will have to subtract 1 to translate those index numbers
+to Rust index numbers.
+
 ## Efficiency Notice
 
 This exercise uses a _vector of vectors_ to store the content of matrices. While
--- a/exercises/saddle-points/README.md
+++ b/exercises/saddle-points/README.md
@@ -28,6 +28,14 @@ The matrix can have a different number of rows and columns (Non square).
 Note that you may find other definitions of matrix saddle points online,
 but the tests for this exercise follow the above unambiguous definition.
 
+## Rust Indices Start At 0
+
+By convention, ordered sequences of values in Rust have their contents numbered
+("indexed") starting from 0. This applies regardless of what the rest of the
+exercise description in this README says, such as references to indices that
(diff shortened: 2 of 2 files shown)


coriolinus:
Hello @workingjubilee, thanks for contributing this!

Travis has failed because it believes that the `saddle-points` README is out of date. You can see the failure for yourself by clicking the "Details" button to the right of the Travis status indicator, then clicking through to the first failed suite. 

Would you please re-pull your copy of the `problem-specifications` repo and then regenerate the READMEs again? That should resolve the Travis issue unless something particularly weird is going on.

workingjubilee:
Whoops! I failed to regenerate again with configlet after I made a teensy tiny change in-repo, actually. Fixed.

petertseng:
EBWODP: Reword "I am down" -> "I am good"

workingjubilee:
'kay! Looking around at the other exercise READMEs that got touched, it seems like they might need more updates anyways, so I'm just going to withdraw that change from this PR and then PR in some full updates. 💖

PR #901, opened 2019-10-28
atbash-cipher: update to canonical data v1.2.0

ocstl:
Version 1.2.0 adds two tests: "decode with too many spaces" and "decode with no spaces".

Relevant PRs:
 - 1.1.0 -> 1.2.0: exercism/problem-specifications#1277

Of possible relevance is exercism/problem-specifications#529, which removed the "encode ignores non ascii" test from the specs, but that test is still present in the Rust track tests. I feel it would be best to remove it, as it doesn't really add much for this particular exercise (compared to, say, "reverse-string"), but I thought I'd check others' opinion first.

### DIFF
--- a/exercises/atbash-cipher/Cargo.toml
+++ b/exercises/atbash-cipher/Cargo.toml
@@ -1,4 +1,4 @@
 [package]
 edition = "2018"
 name = "atbash-cipher"
-version = "1.1.0"
+version = "1.2.0"
--- a/exercises/atbash-cipher/tests/atbash-cipher.rs
+++ b/exercises/atbash-cipher/tests/atbash-cipher.rs
@@ -55,12 +55,6 @@ fn test_encode_all_the_letters() {
 
 #[test]
 #[ignore]
-fn test_encode_ignores_non_ascii() {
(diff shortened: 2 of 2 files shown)


ocstl:
Thanks for the review @petertseng !

Considering #902 , I would say it's probably easier to merge #902 first, seeing as how I can then update with the new tests so they conform to the new format, rather than have a mishmash of the two formats. Unless someone has a better idea, I suggest we close this PR, and I'll submit a new one once #902 is merged.

petertseng:
EBWODP: Clarify that "in situations where significant discussion has occurred in an old PR" does not apply to current situation, so the preference of the submitter is accepted.

ocstl:
Ah, you're right. I've just spent a couple of years dealing with no VCS, so my thinking is still crooked. :confounded:

petertseng:
OK, ready to rebase on #902

ocstl:
Sorry, I accidentally merged when making the changes. All better now. :sweat_smile: 

I removed the "test_encode_ignores_non_ascii" as well, since it was somewhat hard to track with no version number attached.

petertseng:
It was 529 (was stated in original PR description), not 599, so I edited the commit message accordingly.

ocstl:
Good catch! Thanks!

PR #902, opened 2019-10-29
Reformat tests in atbash-cipher exersice

vadimkerr:
Now atbash-cipher exercise use `assert_eq!(expected, actual)` format for tests. 
I think it's better to reverse the order, since this format is more natural and is used by most Rust projects I've seen so far.
Also, tests in other exercises use `assert_eq!(actual, expected)` format, so it would improve consistency.

### DIFF
--- a/exercises/atbash-cipher/tests/atbash-cipher.rs
+++ b/exercises/atbash-cipher/tests/atbash-cipher.rs
@@ -2,89 +2,89 @@ use atbash_cipher as cipher;
 
 #[test]
 fn test_encode_yes() {
-    assert_eq!("bvh", cipher::encode("yes"));
+    assert_eq!(cipher::encode("yes"), "bvh");
 }
 
 #[test]
 #[ignore]
 fn test_encode_no() {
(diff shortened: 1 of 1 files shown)


petertseng:
Sure, that seems like something that can be merged. I would Approve now, but I note that one of #901 or #902 will have to take the other into account, so I'm semi-arbitrarily saying that #901 will probably be merged first, then #902 will have to also change the order of the two newly-added tests.

I caution that, just like in https://github.com/exercism/rust/pull/831, I am very unlikely to attempt to enforce any maintenance of this ordering, but that I would probably do these: I would continue to try to make sure that all the tests of any single exercise at least only adhere to one scheme or the other (referred to as being _internally consistent_), even if exercises differ from each other. If there should ever be an edit war for a given exercise (some people think it should be `actual, expected` and others think it will be `expected, actual`), I would arbitrarily pick one convention (probably the convention more commonly seen among the Rust documentation), but I would make no requirement of newly-added exercises (other than that they remain _internally consistent_).

PR #904, opened 2019-11-06
clock: deemphasize suggestion to implement From<Clock> for String

coriolinus:
That suggestion was meant to get students to explore the available
type-conversion traits, not as a serious command that students
have to implement it. This PR makes the hypotheticality of the
suggestion more explicit.


coriolinus:
My impetus for this PR came from https://exercism-team.slack.com/archives/CASLGD18D/p1572979300093200, in which a mentor asks whether it's better to implement `From<Clock> for String` or `From<&Clock>`. I doubt they would have asked unless someone had implemented it.

PR #907, opened 2019-11-14
Unify config.json topics

workingjubilee:
This continues the progress of #529 

Rebased on 2019-11-18T06:43Z
This unifies the various config.json topics for format, plurality, conciseness, and to explicitly identify "type theory" ideas using terms that are familiar to people who discuss them (and thus, searchable in meta-language discussion, as opposed to just in Rust... thus `option_type` and `result_type` instead of just option or result or plural forms of same.

### DIFF
--- a/config.json
+++ b/config.json
@@ -64,9 +64,9 @@
       "difficulty": 1,
       "topics": [
         "case",
-        "loop",
+        "loops",
         "strings",
-        "vector_optional"
+        "vectors"
       ]
     },
     {
@@ -98,7 +98,7 @@
       "unlocked_by": "hello-world",
       "difficulty": 1,
       "topics": [
-        "algorithm",
+        "algorithms",
         "borrowing",
         "math"
       ]
@@ -140,7 +140,7 @@
       "unlocked_by": null,
       "difficulty": 1,
       "topics": [
-        "iterator",
+        "iterators",
         "str_vs_string",
         "strings"
       ]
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -258,8 +258,8 @@
         "entry_api",
         "filter",
         "match",
-        "mutablity",
-        "result"
+        "mutability",
+        "result_type"
```
coriolinus: What's the motivation for renaming `result` to `result_type`?

config.json
```diff
@@ -331,9 +331,9 @@
       "difficulty": 4,
       "topics": [
         "entry_api",
-        "option",
-        "struct",
-        "vec"
+        "option_type",
```
coriolinus: Same question: what's the motivation behind changing `option` to `option_type`?

coriolinus:
Note, however, that if Travis still fails, that counts as an objection for merging purposes. Looks like you just need to rerun `configlet fmt`.

config.json
workingjubilee: 1) Because "results" looks weird,
2) especially since everyone talks about "Result" as in the Result type,
3) technically it's Result<T, Err>,
4) and possibly most important of all, Mr. Wikipedia mentions it: https://en.wikipedia.org/wiki/Result_type

config.json
workingjubilee: ditto: https://en.wikipedia.org/wiki/Option_type

PR #910, opened 2019-12-04
Optimize saddle-points; show some Rust "tricks"

murlakatamenka:
Optimizations:

- row maximum is calculated once per row, not once per matrix element
- no `Vec` allocations for column minimums, using iterators is sufficient

---

Commit also shows some Rust "tricks":

- assign several variables in one `let` statement
- `match` on several items
- use logical or `|` in `match` arms conditions

---

What do you think of the suggested changes?

### DIFF
--- a/exercises/saddle-points/example.rs
+++ b/exercises/saddle-points/example.rs
@@ -1,20 +1,21 @@
 pub fn find_saddle_points(input: &[Vec<u64>]) -> Vec<(usize, usize)> {
-    let (width, height) = match (input.len(), input[0].len()) {
-        (0, _) | (_, 0) => return Vec::new(),
-        (n, k) => (n, k),
+    let height = if !input.is_empty() && !input[0].is_empty() {
+        input[0].len()
+    } else {
+        return Vec::new();
     };
 
     let mut saddle_points = Vec::new();
 
-    for i in 0..width {
-        let row_max = input[i].iter().max().unwrap();
+    let col_mins: Vec<u64> = (0..height)
+        .filter_map(|j| input.iter().map(|row| row[j]).min())
+        .collect();
+
+    for (i, row) in input.iter().enumerate() {
+        let row_max = *row.iter().max().unwrap();
 
         for j in 0..height {
-            let column_min = input.iter().map(|x| x[j]).min().unwrap();
-
-            let value = input[i][j];
-
-            if value == *row_max && value == column_min {
+            if row[j] == row_max && row[j] == col_mins[j] {
                 saddle_points.push((i, j));
             }
         }
--- a/exercises/saddle-points/example.rs
+++ b/exercises/saddle-points/example.rs
@@ -1,23 +1,24 @@
 pub fn find_saddle_points(input: &[Vec<u64>]) -> Vec<(usize, usize)> {
+    let (width, height) = match (input.len(), input[0].len()) {
+        (0, _) | (_, 0) => return Vec::new(),
+        (n, k) => (n, k),
+    };
+
     let mut saddle_points = Vec::new();
 
-    let width = input.len();
-    let height = input[0].len();
-
     for i in 0..width {
-        for j in 0..height {
-            let column = input.iter().map(|x| x[j]).collect::<Vec<u64>>();
-            let row = &input[i];
+        let row_max = input[i].iter().max().unwrap();
 
-            let max = row.iter().max().unwrap();
-            let min = column.iter().min().unwrap();
+        for j in 0..height {
+            let column_min = input.iter().map(|x| x[j]).min().unwrap();
 
             let value = input[i][j];
 
-            if value >= *max && value <= *min {
+            if value == *row_max && value == column_min {
                 saddle_points.push((i, j));
             }
         }
     }
+
     saddle_points
 }


exercises/saddle-points/example.rs
```diff
@@ -1,23 +1,24 @@
...
     for i in 0..width {
-        for j in 0..height {
-            let column = input.iter().map(|x| x[j]).collect::<Vec<u64>>();
-            let row = &input[i];
+        let row_max = input[i].iter().max().unwrap();
 
-            let max = row.iter().max().unwrap();
-            let min = column.iter().min().unwrap();
+        for j in 0..height {
+            let column_min = input.iter().map(|x| x[j]).min().unwrap();
```
coriolinus: You're computing the column min for every `i` and `j`; computing that min is ~O(n), so overall, the column minima scan is O(n^3). It's more efficient to precompute a vector of column minima ahead of time, and then index into it; this can be done with O(n^2) time and O(n) memory, [like this](https://github.com/coriolinus/exercism-rs/blob/master/saddle-points/src/lib.rs):

```rust
            let col_min = (0..min_width)
                .map(|coln| input.iter().map(|row| row[coln]).min().unwrap())
                .collect::<Vec<_>>();
```

This isn't a huge deal, but you do say that this is an optimization, so let's make it optimal.

exercises/saddle-points/example.rs
```diff
@@ -1,23 +1,24 @@
...
 
-    let width = input.len();
-    let height = input[0].len();
+    let mut saddle_points = Vec::new();
 
     for i in 0..width {
-        for j in 0..height {
-            let column = input.iter().map(|x| x[j]).collect::<Vec<u64>>();
-            let row = &input[i];
+        let row_max = input[i].iter().max().unwrap();
```
coriolinus:
```suggestion
        let row_max = *input[i].iter().max().unwrap();
```

May as well dereference the maximum here; that'll make the value comparison in line 17 prettier.

exercises/saddle-points/example.rs
```diff
@@ -1,23 +1,24 @@
 pub fn find_saddle_points(input: &[Vec<u64>]) -> Vec<(usize, usize)> {
-    let mut saddle_points = Vec::new();
+    let (width, height) = match (input.len(), input[0].len()) {
+        (0, _) | (_, 0) => return Vec::new(),
+        (n, k) => (n, k),
+    };
```
coriolinus: This is a bit too clever: it sacrifices some clarity for the sake of brevity, and isn't faster than the straightforward implementation:

```rust
let width = input.len();
if width == 0 {
    return Vec::new();
}
let height = input[0].len();
if height == 0 {
    return Vec::new();
}
```

exercises/saddle-points/example.rs
```diff
@@ -1,23 +1,24 @@
 pub fn find_saddle_points(input: &[Vec<u64>]) -> Vec<(usize, usize)> {
-    let mut saddle_points = Vec::new();
+    let (width, height) = match (input.len(), input[0].len()) {
```
coriolinus: This isn't actually safe, anyway. Try adding this test:

```rust
#[test]
fn empty() {
    let input = vec![];
    let expected: Vec<(usize, usize)> = Vec::new();
    assert_eq!(expected, find_saddle_points(&input));
}
```

When I run that test on this code, I get

```
failures:

---- empty stdout ----
thread 'empty' panicked at 'index out of bounds: the len is 0 but the index is 0', src/lib.rs:2:47
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.


failures:
    empty
```

That's because the `match` implementation isn't clever enough to short-circuit and avoid the second calculation if the first is sufficient to determine a branch, so it unconditionally attempts `input[0]` as well.

exercises/saddle-points/example.rs
murlakatamenka: Yep, that's an improvement that I'm aware of. Re-calculating column minimums for each matrix element can't be efficient. I didn't include it not to diverge too much from the original code.

exercises/saddle-points/example.rs
murlakatamenka: Agree, good point.

exercises/saddle-points/example.rs
murlakatamenka: Yes, my suggestion is less clear, especially for people who come from other languages

Frankly speaking we don't need `width` here at all.

```rust
    let height = if !matrix.is_empty() && !matrix[0].is_empty() {
        matrix[0].len()
    } else {
        return Vec::new();
    };

// calculate column minimums and store them in a Vec

for (i, row) in matrix.iter().enumerate() {
    let row_max = *row.iter().max().unrwap();
    
    for j in 0..height {
        // ...
    }
}
```

exercises/saddle-points/example.rs
murlakatamenka: That's a serious overlook. Then https://github.com/exercism/rust/pull/910#discussion_r354475074 gets even more relevant.

murlakatamenka:
Thanks for the feedback! Glad to see positive attitude towards changes too.

> You are aware that the examples aren't shown to students, and very few of them will come to this repository to check the examples?

Yes, I'm aware. I think I've seen the examples myself by an accident, just chencking out this repo.

exercises/saddle-points/example.rs
coriolinus: Good point. I like that approach better.

murlakatamenka:
Thanks! 

In general, such PR don't bring much because of the expore `example.rs` of exercises get at this point, so shouldn't be concerned much, right?

PR #915, opened 2020-01-11
Fix dot-dsl test

TomPradat:
The `new` method on the Node expects a `&str` not a `&String` everywhere else in the tests.

### DIFF
--- a/exercises/dot-dsl/tests/dot-dsl.rs
+++ b/exercises/dot-dsl/tests/dot-dsl.rs
@@ -127,10 +127,10 @@ fn test_graph_with_attributes() {
 fn test_graph_stores_attributes() {
     let attributes = [("foo", "bar"), ("bat", "baz"), ("bim", "bef")];
     let graph = Graph::new().with_nodes(
-        &['a', 'b', 'c']
+        &["a", "b", "c"]
             .iter()
             .zip(attributes.iter())
-            .map(|(name, &attr)| Node::new(&name.to_string()).with_attrs(&[attr]))
+            .map(|(name, &attr)| Node::new(&name).with_attrs(&[attr]))
             .collect::<Vec<_>>(),
     );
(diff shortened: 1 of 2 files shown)


exercises/dot-dsl/tests/dot-dsl.rs
```diff
@@ -127,10 +127,10 @@ fn test_graph_with_attributes() {
 fn test_graph_stores_attributes() {
     let attributes = [("foo", "bar"), ("bat", "baz"), ("bim", "bef")];
     let graph = Graph::new().with_nodes(
-        &['a', 'b', 'c']
+        &["a", "b", "c"]
             .iter()
             .zip(attributes.iter())
-            .map(|(name, &attr)| Node::new(&name.to_string()).with_attrs(&[attr]))
+            .map(|(name, &attr)| Node::new(&name)).with_attrs(&[attr]))
```
coriolinus: CI is failing because you didn't completely eliminate the function call:
```suggestion
            .map(|(name, &attr)| Node::new(&name).with_attrs(&[attr]))
```

TomPradat:
> Hi @TomPradat, thanks for the contribution! I apologize that it's taken this long to respond to you.
> 
> CI needs to pass in order to consider merging this. Happily, that's a simple modification.
> 
> It's worth mentioning that `Node::new` expects an `&str` in all cases. However, because [`String` implements `Deref<Target=str>`](https://doc.rust-lang.org/std/string/struct.String.html#deref), the expression `&name.to_string()` does in fact resolve to an `&str`.
> 
> I'm inclined to merge this anyway, once CI passes, because there's no point confusing people, and it's more straightforward to just use a `&str` literal than to call `to_string` on a character literal.

I was able to finish the exercise with the actual version of the test, but I agree this was a bite confusing to me at first

PR #916, opened 2020-01-13
bowling: Add a new test

isgj:
While solving this exercise I passed all the tests but I realized that 1 result was wrong. Here I added the test.


petertseng:
That is a good test. It may be a good idea to submit it to https://github.com/exercism/problem-specifications/blob/master/exercises/bowling/canonical-data.json as well. I don't want anyone accusing us of being selfish and refusing to share our tests with others.

PR #923, opened 2020-01-17
Fixes #922

### LINKED ISSUE #922: bob: update to v1.6.0
Per https://github.com/exercism/problem-specifications/pull/1632, the bob exercise is updated to avoid some cultural sensitivity issues. I'd like this to get a relatively high priority for that reason; we don't want to offend those we don't have to.

@lewisclement, @filalex77, are either of you willing to take charge of this one? It feels like a good opportunity to get familiar with the [`util/exercise` tool](https://github.com/exercism/rust/tree/master/util/exercise). 

Procedurally, I think it's helpful if we use the Assignees section in the upper right to take ownership of an issue. I don't intend to ever assign someone other than myself to an issue or PR, and hope you guys have the same intention.

lewisclement:
Get rid of the old tests

### DIFF
--- a/exercises/bob/tests/bob.rs
+++ b/exercises/bob/tests/bob.rs
@@ -23,8 +23,7 @@ process_response_case({let mut hm = ::std::collections::HashMap::new(); hm.inser
 #[ignore]
 /// other whitespace
 fn test_other_whitespace() {
-process_response_case({let mut hm = ::std::collections::HashMap::new(); hm.insert("heyBob", "
- 	"); hm}, "Fine. Be that way!");
+process_response_case({let mut hm = ::std::collections::HashMap::new(); hm.insert("heyBob", "\r\r 	"); hm}, "Fine. Be that way!");
 }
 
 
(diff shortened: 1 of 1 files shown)


petertseng:
since the branch is on the exercism copy of the repo, please don't forget to delete the branch when done

petertseng:
Since the commits don't stand by themselves, I think it would have been better to squash this one. We should only leave commits separate if they can stand by themselves.

lewisclement:
Is it still possible to squash them after a merge?

petertseng:
Only by rewriting history, and for `master` branch I reserve rewriting history for special occasions

petertseng:
Since tone is hard to interpret with only text, I'd like to make a reassurance and clarify my tone - I would say this is my fault for not making these things known. After all, if I'm going to insist others adhere to some rules (rules that seem perfectly reasonable to one person may seem arcane and/or draconic to another person), it doesn't make sense for me to insist that if I never wrote them down in a noticeable location and just relied on tribal knowledge. So, it is just a natural happenstance.

lewisclement:
Well, I do happen to agree with you in this circumstance. I'm still a rookie when it comes to large-scale team projects here on Github. I hoped that my experience with BitBucket would help me here, but the web interfaces are quite different. Maybe a good time for me to exercise some CLI git.

PR #931, opened 2020-03-19
Add new test for clear() to 'Circular Buffer'

### LINKED ISSUE #930: Excercise 'Circular Buffer' doesn't actually verify that clear() clears the buffer
I realize that this is just an academic exercise, and so leaving in such a potential memory and resource leak isn't a big deal, but I think there should still be a simple check to see that the buffer actually gets cleared, otherwise, whatever your `T` is can potentially sit inside the "cleared" buffer and waste memory and occupy other resources for a long time.

Currently, you can pass all tests with an implementation that simply sets however you keep track of your buffer's count/size to 0, while leaving the actual elements still alive within the buffer. To defend against this (IMO invalid) approach, I designed this test case:

```rust
#[test]
fn clear_actually_frees_up_its_elements() {
    let mut buffer = CircularBuffer::new(1);
    let element = Rc::new(());
    assert!(buffer.write(Rc::clone(&element)).is_ok());
    assert_eq!(Rc::strong_count(&element), 2);
    buffer.clear();
    assert_eq!(Rc::strong_count(&element), 1);
}
```
It uses `Rc` to make sure that after a `clear()`, the clone count is reduced, meaning that Rust called `drop()` on it, i.e. it was properly cleared out.

alexschrod:
Verifies that clear() actually clears the buffer. Resolves #930.

### DIFF
--- a/exercises/circular-buffer/example.rs
+++ b/exercises/circular-buffer/example.rs
@@ -55,6 +55,11 @@ impl<T: Default + Clone> CircularBuffer<T> {
     pub fn clear(&mut self) {
         self.start = 0;
         self.end = 0;
+
+        // Clear any values in the buffer
+        for element in self.buffer.iter_mut() {
+            std::mem::take(element);
+        }
     }
 
     pub fn is_empty(&self) -> bool {
--- a/exercises/circular-buffer/tests/circular-buffer.rs
+++ b/exercises/circular-buffer/tests/circular-buffer.rs
@@ -1,4 +1,5 @@
 use circular_buffer::{CircularBuffer, Error};
+use std::rc::Rc;
 
 #[test]
 fn error_on_read_empty_buffer() {
--- a/exercises/circular-buffer/tests/circular-buffer.rs
+++ b/exercises/circular-buffer/tests/circular-buffer.rs
@@ -94,6 +94,17 @@ fn clear_does_nothing_on_empty_buffer() {
 
 #[test]
 #[ignore]
+fn clear_actually_frees_up_its_elements() {
+    let mut buffer = CircularBuffer::new(1);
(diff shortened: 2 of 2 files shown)


exercises/circular-buffer/example.rs
```diff
@@ -55,6 +55,11 @@ impl<T: Default + Clone> CircularBuffer<T> {
     pub fn clear(&mut self) {
         self.start = 0;
         self.end = 0;
+
+        // Clear any values in the buffer
+        for element in self.buffer.iter_mut() {
+            std::mem::take(element);
+        }
```
coriolinus: Why not just [delegate the method](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.clear): 
```suggestion
        self.buffer.clear();
```

alexschrod: We can't do that. If you look at how the `new()`, all of the code relies on the buffer always being populated and being `size + 1`:

```rust
pub fn new(size: usize) -> CircularBuffer<T> {
    CircularBuffer {
        buffer: vec![T::default(); size + 1],
        size: size + 1,
        start: 0,
        end: 0,
    }
}
```

Doing `mem::take()` on each element simply returns them all back to `T::default()`, but leaves the vector layout intact, while doing `self.buffer.clear()` removes all the elements.

Now, arguably, having a `T::default()` in each slot isn't very good practice; in my own solution I went with `Option<T>` instead of `T: Default` so I could store `None` values for unused slots instead of whatever `T`'s default is. This also let me avoid the `T: Default` bound. But now I'm just pontificating. 😉

coriolinus: Right, fair enough. I'd forgotten that the implementation depended on that assumption.

PR #934, opened 2020-04-06
&self -> RustMaintainer

workingjubilee:
After discussing with @efx and @coriolinus I realized I was being very silly by not formally adding myself as a maintainer. Here's to version control.

### DIFF
--- a/config/maintainers.json
+++ b/config/maintainers.json
@@ -110,6 +110,6 @@
       "link_url": null,
       "avatar_url": null,
       "bio": "Builder of rainbow bridges."
-    },
+    }
   ]
 }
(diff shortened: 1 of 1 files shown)


config/maintainers.json
```diff
@@ -100,6 +100,16 @@
...
+    {
+      "github_username": "workingjubilee",
+      "alumnus": false,
+      "show_on_website": false,
+      "name": "Jubilee Young",
+      "link_text": null,
+      "link_url": null,
+      "avatar_url": null,
+      "bio": "Builder of rainbow bridges."
+    },
```
efx, line 113:
```suggestion
    }
```

workingjubilee: I was saying to myself "surely this won't break a JSON parse?" and then I actually tried it.

Issue #936, opened 2020-04-17
exercise: diffie-hellman: missing test for really big primes + example failing


LeBlue:
The diffie-hellman exercise is missing tests for really big prime/input numbers. With a test ```test_public_key_correct_biggest_numbers``` a lot of solutions **including the example** will fail with

```
thread 'test_public_key_correct_biggest_numbers' panicked at 'attempt to multiply with overflow'
```
The tests:
```
#[test]
#[ignore]
fn test_public_key_correct_biggest_numbers() {
    // two biggest 64bit primes
    let p: u64 = 0xF_FFF_FFF_FFF_FFF_FC5;
    let g: u64 = 0xF_FFF_FFF_FFF_FFF_FAC;

    let private_key: u64 = 0xF_FFF_FFF_FFF_FFF_FC3;

    let expected: u64 = 13_281_655_733_070_877_121;

    assert_eq!(public_key(p, g, private_key), expected);
}

#[test]
#[ignore]
fn test_secret_key_correct_biggest_numbers() {
    // two biggest 64bit primes
    let p: u64 = 0xF_FFF_FFF_FFF_FFF_FC5;
    let g: u64 = 0xF_FFF_FFF_FFF_FFF_FAC;

    let private_key_a = 0xF_FFF_FFF_FFF_FFF_FC3;
    let public_key_a = public_key(p, g, private_key_a);

    let expected_a = 13_281_655_733_070_877_121;
    assert_eq!(public_key_a, expected_a);

    let private_key_b = 0xE_FFF_FFF_FFF_FFF_FC0;
    let public_key_b = public_key(p, g, private_key_b);

    let expected_b = 4_340_425_873_327_658_043;
    assert_eq!(public_key_b, expected_b);

    let expected_key = 12_669_955_479_143_291_250;

    let secret_key = secret(p, public_key_b, private_key_a);

    assert_eq!(secret_key, expected_key);

    let secret_key = secret(p, public_key_a, private_key_b);

    assert_eq!(secret_key, expected_key);
}

#[test]
#[ignore]
fn test_secret_key_equals_for_biggest_numbers() {
    // two biggest 64bit primes
    let p: u64 = 0xF_FFF_FFF_FFF_FFF_FC5;
    let g: u64 = 0xF_FFF_FFF_FFF_FFF_FAC;

    let private_key_a = private_key(p);
    let public_key_a = public_key(p, g, private_key_a);

    let private_key_b = private_key(p);
[...]

PR #937, opened 2020-04-17
diffie-hellman: Add test cases for inputs near the type limits

LeBlue:
Fixes https://github.com/exercism/rust/issues/936

* Add tests with inputs near the u64 type limits as a solutions may
  panic 'multiplication overflow' while passing existing tests
* Fix example code to also pass these tests

### DIFF
--- a/exercises/diffie-hellman/.meta/hints.md
+++ b/exercises/diffie-hellman/.meta/hints.md
@@ -1,2 +1,14 @@
 One possible solution for this exercise is to implement your own modular exponentiation function.
 To learn more about it refer to the [following page](https://en.wikipedia.org/wiki/Modular_exponentiation).
+
+## For bonus points
+
+Many implementations do not work for the entire domain of inputs.
+There are additional optional tests defined which help ensure that all valid inputs produce valid results.
+
+To run the bonus tests, remove the `#[ignore]` flag and execute the tests with
+the `big-primes` feature, like this:
+
+```bash
+$ cargo test --features big-primes
+```
--- a/exercises/diffie-hellman/Cargo-example.toml
+++ b/exercises/diffie-hellman/Cargo-example.toml
@@ -5,3 +5,6 @@ version = "0.1.0"
 
 [dependencies]
 rand = "0.3.18"
+
+[features]
+big-primes = []
--- a/exercises/diffie-hellman/tests/diffie-hellman.rs
+++ b/exercises/diffie-hellman/tests/diffie-hellman.rs
@@ -67,6 +67,63 @@ fn test_secret_key_correct_big_numbers() {
     assert_eq!(secret, expected);
 }
 
+// two biggest 64bit primes
+#[cfg(feature = "big-primes")]
+const PRIME_64BIT_1: u64 = 0xFFFF_FFFF_FFFF_FFC5;
+#[cfg(feature = "big-primes")]
+const PRIME_64BIT_2: u64 = 0xFFFF_FFFF_FFFF_FFAC;
+#[cfg(feature = "big-primes")]
+const PRIVATE_KEY_64BIT: u64 = 0xFFFF_FFFF_FFFF_FFC3;
+#[cfg(feature = "big-primes")]
+const PUBLIC_KEY_64BIT: u64 = 0xB851_EB85_1EB8_51C1;
+
+#[test]
+#[ignore]
+#[cfg(feature = "big-primes")]
+fn test_public_key_correct_biggest_numbers() {
+    assert_eq!(
+        public_key(PRIME_64BIT_1, PRIME_64BIT_2, PRIVATE_KEY_64BIT),
+        PUBLIC_KEY_64BIT
+    );
+}
+
+#[test]
+#[ignore]
+#[cfg(feature = "big-primes")]
+fn test_secret_key_correct_biggest_numbers() {
+    let private_key_b = 0xEFFF_FFFF_FFFF_FFC0;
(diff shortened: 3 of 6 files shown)


exercises/diffie-hellman/tests/diffie-hellman.rs
```diff
@@ -67,6 +67,69 @@ fn test_secret_key_correct_big_numbers() {
...
+#[test]
+#[ignore]
+fn test_public_key_correct_biggest_numbers() {
+    // two biggest 64bit primes
+    let p: u64 = 0xF_FFF_FFF_FFF_FFF_FC5;
+    let g: u64 = 0xF_FFF_FFF_FFF_FFF_FAC;
+
+    let private_key: u64 = 0xF_FFF_FFF_FFF_FFF_FC3;
+
+    let expected: u64 = 13_281_655_733_070_877_121;
```
coriolinus, line 79: You define the first two of these values in three places, and the other two in two. May as well extract them all into module-private consts.

exercises/diffie-hellman/tests/diffie-hellman.rs
```diff
@@ -67,6 +67,69 @@ fn test_secret_key_correct_big_numbers() {
     assert_eq!(secret, expected);
 }
 
+#[test]
+#[ignore]
+fn test_public_key_correct_biggest_numbers() {
+    // two biggest 64bit primes
+    let p: u64 = 0xF_FFF_FFF_FFF_FFF_FC5;
```
coriolinus, line 74: Just a style thing: with hexadecimal values, it is unconventional to divide the digits into groups of three, as you do here. It is much more common to group the digits to some power of 2, because every two hex digits corresponds to exactly one byte.

```suggestion
    let p: u64 = 0xFFFF_FFFF_FFFF_FFC5;
```

The same goes, of course, for all the other hex literals.

exercises/diffie-hellman/tests/diffie-hellman.rs
LeBlue: Done

exercises/diffie-hellman/tests/diffie-hellman.rs
LeBlue: Done

LeBlue:
@coriolinus I implemented the suggested changes.

For option
> Retain the example.rs changes, drop the diffie-hellman.rs changes, so that the example is correct but there are no tests verifying the behavior for large inputs.

I split of the example fixes in a separate commit (Do I need to/should I create a new pull request for that?).

The third option
> Keep all these changes, but feature-gate the new tests, adding an appropriate section to the README briefly describing the nature of the feature-gated tests and how to run them. Feature gates neatly get us out of the problem of correctness issues which are out of scope of the main learning objective for an exercise.

is now implemented in this pull request. I am not really satisfied with the section in the README, feel free to suggest some changes.

exercises/diffie-hellman/.meta/hints.md
```diff
@@ -1,2 +1,17 @@
 One possible solution for this exercise is to implement your own modular exponentiation function.
 To learn more about it refer to the [following page](https://en.wikipedia.org/wiki/Modular_exponentiation).
+
+## For bonus points
+
+Did you get the tests passing and the code clean? If you want to, there
+are some additional things you could try.
+
+- Currently your modular exponentiation function will probably not work for
+  the full range of inputs. There are additional tests you can run.
```
coriolinus, line 10:
```suggestion
Many implementations do not work for
the entire domain of inputs. There are additional optional tests 
defined which help ensure that all valid inputs produce valid results.
```

exercises/diffie-hellman/Cargo-example.toml
```diff
@@ -5,3 +5,6 @@ version = "0.1.0"
 
 [dependencies]
 rand = "0.3.18"
+
+[features]
+big-primes = []
```
coriolinus, line 10: Please ensure that all text files end with a newline. I see that the original `Cargo.toml` didn't, which was an error of review on our part.

exercises/diffie-hellman/tests/diffie-hellman.rs
```diff
@@ -67,6 +67,63 @@ fn test_secret_key_correct_big_numbers() {
     assert_eq!(secret, expected);
 }
 
+// two biggest 64bit primes
+#[allow(dead_code)]
```
coriolinus, line 71:
```suggestion
#[cfg(feature = "big-primes")]
```

These consts are only dead code when the appropriate tests are ignored during compilation, so may as well apply the same feature gate to the consts instead of compiling them unconditionally.

exercises/diffie-hellman/Cargo-example.toml
LeBlue: Also noticed that and decided to stick with the original

exercises/diffie-hellman/.meta/hints.md
LeBlue: Included

exercises/diffie-hellman/tests/diffie-hellman.rs
LeBlue: I wondered if there was a better solution

LeBlue:
@coriolinus All suggestions should be included/all fixed now.

patricksjackson:
I'm not a maintainer here, but am on the C/C++ tracks.

I agree that keeping these changes as bonus tests is the best choice here, LGTM :+1:

PR #938, opened 2020-04-24
macros: Add tests to check for HashMap scoping

pedantic79:
This test makes sure that students use `::std::collections::HashMap`
instead of just `std::collections::HashMap`. It is intended
to show that a user can override the intended type.

They receive this error if the wrong scoping is used:

```
---- test_type_override stdout ----
thread 'test_type_override' panicked at 'Do not allow users to override which HashMap is used', tests/macros.rs:81:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### DIFF
--- a/exercises/macros/tests/macros.rs
+++ b/exercises/macros/tests/macros.rs
@@ -73,7 +73,7 @@ fn test_type_override() {
     // The macro should always use std::collections::HashMap and ignore crate::std::collections::HashMap
     mod std {
         pub mod collections {
-            pub struct HashMap();
+            pub struct HashMap;
 
             impl HashMap {
                 #[allow(dead_code)]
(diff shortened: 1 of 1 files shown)


exercises/macros/tests/macros.rs
```diff
@@ -67,6 +67,31 @@ fn test_nested() {
     );
 }
 
+#[test]
+#[ignore]
+fn test_type_override() {
+    // The macro should always use std::collections::HashMap and ignore crate::std::collections::HashMap
+    mod std {
+        pub mod collections {
+            pub struct HashMap();
```
coriolinus, line 76:
```suggestion
            pub struct HashMap;
```

I believe that both declarations properly compile down to a ZST, but this form makes it more obvious.

PR #939, opened 2020-04-26
Update to fix math errors in tests

jgilray:
Fixed math errors in wordy/tests/wordy.rs

### DIFF
--- a/exercises/wordy/tests/wordy.rs
+++ b/exercises/wordy/tests/wordy.rs
@@ -94,7 +94,7 @@ fn multiple_multiplications() {
 #[ignore]
 fn addition_and_multiplication() {
     let command = "What is -3 plus 7 multiplied by -2?";
-    assert_eq!(Some(-8), answer(command));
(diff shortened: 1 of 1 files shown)


workingjubilee:
Thank you for your offered contribution! However, there is actually no error, as this follows a verbal order of operations, as specified in the README:
> Since these are verbal word problems, evaluate the expression from left-to-right, ignoring the typical order of operations.
So this is a duplicate of https://github.com/exercism/problem-specifications/pull/1234 in effect, for which reason I am closing this PR.

jgilray:
Oops... thanks.  I didn't realize that it was that way... maybe it should be fixed as that would be less confusing and lead to a less verbose exercise explanation?

workingjubilee:
I believe the intention is that the current evaluation order actually makes it less frustrating to implement, so you can parse through the words and perform the ops in the same order. I am totally game if you want to explore a more concise exercise explanation in the README, however!

PR #940, opened 2020-04-26
Remove Atbash Cipher from Rust core

workingjubilee:
Per discussion on Slack, Atbash Cipher is a "redundant" exercise in many ways as while the exercise involves string processing, it has many tedious fine details and is not that exciting to implement in Rust. Also learned a bit of `jq`!

If desired, I believe two other changes might be prudent but I wanted to discuss them before I implemented them as I am not sure of their technical details, whether consensus is available, and if they should go in this or another PR.

1. I believe the implementation of Space-Age is considerably simpler than Clock, while they are both similar exercises, and thus they should swap places. They both involve mathematics and `impl` but one is very simple and involves a lot of pregenerated boilerplate to work with. The other involves many stylistic choices, derives, and fine details of mathematics not often engaged with. However, both are fairly valuable, so I would not remove either.

2. I believe the implementation of Luhn is considerably simpler than Acronym, while they are both similar exercises, and thus they should swap places. They both involve string manipulation and iteration but one involves lifetimes in a way the other does not, and is thus automatically harder. However, both are fairly valuable, so I would not remove either.

### DIFF
--- a/config.json
+++ b/config.json
@@ -576,7 +576,7 @@
       "slug": "affine-cipher",
       "uuid": "2a1dcf38-ec05-4b24-a2e2-2e5b3595f3f0",
       "core": false,
-      "unlocked_by": "atbash-cipher",
+      "unlocked_by": "luhn",
       "difficulty": 4,
       "topics": [
         "ascii",
@@ -615,8 +615,8 @@
     {
       "slug": "atbash-cipher",
       "uuid": "53298a14-76a9-4bb9-943a-57c5e79d9cf7",
-      "core": true,
-      "unlocked_by": null,
+      "core": false,
+      "unlocked_by": "luhn",
       "difficulty": 4,
       "topics": [
         "ascii",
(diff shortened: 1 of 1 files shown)


coriolinus:
1. Agree. Swapping them should to into another PR.
2. I don't really have an opinion on that pair, and you have more mentoring experience than I do, so I'm willing to take your word on it. That swap should go into a third PR.

I think you make good points; the general rule is that one PR should accomplish one conceptual operation.

workingjubilee:
Well, OK yes. It's mostly an accident of how the tests work which puts it in an awkward position where if it was easier or harder, it'd be more interesting.

Yeah, the only actual reason I'd shove those changes into the same PR is if it was somehow easier for the fixup scripts to just run all at once (and they'd still each be their own commit).

iHiD:
I'll do this tomorrow. Thanks.

iHiD:
Done and script running. Thank you :)

PR #941, opened 2020-04-26
all-your-base: Fixing test: Empty slice corresponds to zero II

CGMossa:
This is the continuation of #855.

### DIFF
--- a/exercises/all-your-base/example.rs
+++ b/exercises/all-your-base/example.rs
@@ -56,11 +56,6 @@ pub fn convert<P: AsRef<[Digit]>>(
         return Err(Error::InvalidDigit(invalid));
     }
     
-    // no output digits correspond to zero
-    if digits.as_ref().is_empty() {
-        return Ok(vec![0])
-    }
-
     // convert all digits into a single large number
     let mut immediate: Digit = digits
         .as_ref()
--- a/exercises/all-your-base/example.rs
+++ b/exercises/all-your-base/example.rs
@@ -55,6 +55,11 @@ pub fn convert<P: AsRef<[Digit]>>(
     if let Some(&invalid) = digits.as_ref().iter().find(|&num| *num >= from_base) {
         return Err(Error::InvalidDigit(invalid));
     }
+    
+    // no output digits correspond to zero
+    if digits.as_ref().is_empty() {
+        return Ok(vec![0])
+    }
 
     // convert all digits into a single large number
     let mut immediate: Digit = digits
@@ -73,5 +78,9 @@ pub fn convert<P: AsRef<[Digit]>>(
     }
     // fix order of digits
     res.reverse();
-    Ok(res)
+    if res.is_empty() {
+        Ok(vec![0])
+    } else {
+        Ok(res)
+    }
 }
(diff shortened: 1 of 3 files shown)


exercises/all-your-base/src/lib.rs
```diff
@@ -33,7 +33,7 @@ pub enum Error {
 ///
 /// Notes:
 ///  * The empty slice ( "[]" ) is equal to the number 0.
-///  * Never output leading 0 digits. However, your function must be able to
+///  * Never output leading 0 digits, unless the input number is 0, in which the output must be `[0]`. However, your function must be able to
 ///     process input with leading 0 digits.
```
pedantic79, line 37: This line is really wide. 
```suggestion
///  * Never output leading 0 digits, unless the input number is 0, in which the output must be `[0]`.
///    However, your function must be able to process input with leading 0 digits.
```

exercises/all-your-base/example.rs
```diff
@@ -55,6 +55,11 @@ pub fn convert<P: AsRef<[Digit]>>(
     if let Some(&invalid) = digits.as_ref().iter().find(|&num| *num >= from_base) {
         return Err(Error::InvalidDigit(invalid));
     }
+    
+    // no output digits correspond to zero
+    if digits.as_ref().is_empty() {
+        return Ok(vec![0])
+    }
```
coriolinus, line 62: This check is redundant. I understand that it is slightly faster in the case that the digits list is empty, but for the examples, I think DRY is more important than optimization.

exercises/all-your-base/example.rs
```diff
@@ -73,5 +78,9 @@ pub fn convert<P: AsRef<[Digit]>>(
     }
     // fix order of digits
     res.reverse();
-    Ok(res)
+    if res.is_empty() {
+        Ok(vec![0])
+    } else {
+        Ok(res)
+    }
```
coriolinus, line 85: This can be simplified to avoid an allocation:

```suggestion
    if res.is_empty() {
        res.push(0);
    } 
    Ok(res)
```

exercises/all-your-base/src/lib.rs
```diff
@@ -43,4 +43,4 @@ pub fn convert(number: &[u32], from_base: u32, to_base: u32) -> Result<Vec<u32>,
         from_base,
         to_base
     )
-}
+}
```
coriolinus, line 46: Please ensure files end with a trailing newline.

CGMossa: Okay, thank you!

exercises/all-your-base/example.rs
CGMossa: I've removed this check here: 056f7c2.

CGMossa:
Does anyone know how I can run the test on the example file? Right now I am c/p the `example.rs` into `lib.rs` and then run it there.

coriolinus:
> Does anyone know how I can run the test on the example file?

[`bin/test-exercise`](https://github.com/exercism/rust/blob/master/bin/test-exercise) exists to do exactly that.

CGMossa:
Thanks!
I've figured that it must be this:

```bash
./bin/test-exercise exercises/all-your-base/ --release
```

CGMossa:
Thanks!
I've figured that it must be this:

```bash
./bin/test-exercise exercises/all-your-base/ --release
```

PR #933, opened 2020-03-30
macros: Add new tests to check for invalid syntax without trybuild

pedantic79:
- This uses code inspired by buildtry, but this is less flexible.
- These tests contain invalid syntax, and are designed not to compile.
- Student macros should produce an error instead of compile.
- Add note about Rust 1.36 requirement

---

Output of all passing tests like with `example.rs` and this [student's implementation](https://exercism.io/mentor/solutions/80240c6472fd40f4a6ae5924e2bd0f33):
```
running 13 tests
test test::type_not_in_scope ... ok
test test_empty ... ok
test test_single ... ok
test test_no_trailing_comma ... ok
test test_nested ... ok
test test_trailing_comma ... ok
test test_compile_fails_triple_arguments ... ok
test test_compile_fails_two_arrows ... ok
test test_compile_fails_double_commas ... ok
test test_compile_fails_only_arrow ... ok
test test_compile_fails_single_argument ... ok
test test_compile_fails_only_comma ... ok
test test_compile_fails_comma_sep ... ok
```

[Another student](https://exercism.io/mentor/solutions/644c98b29f7041a5a46b26f093b26263) that has a failure sees this:
```
running 13 tests
test test::type_not_in_scope ... ok
test test_empty ... ok
test test_single ... ok
test test_no_trailing_comma ... ok
test test_nested ... ok
test test_trailing_comma ... ok
test test_compile_fails_triple_arguments ... ok
test test_compile_fails_only_arrow ... ok
test test_compile_fails_comma_sep ... ok
test test_compile_fails_only_comma ... ok
test test_compile_fails_double_commas ... FAILED
test test_compile_fails_single_argument ... ok
test test_compile_fails_two_arrows ... ok

failures:

---- test_compile_fails_double_commas stdout ----
thread 'test_compile_fails_double_commas' panicked at 'tests/invalid/double-commas.rs compiled sucessfully, but should not.', tests/macros.rs:138:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

PR #942, opened 2020-04-27
macros: additional tests without HashMap in scope

pedantic79:
- This is a continuation of #649
- Move the tests above the compilation failure tests.
- Add a line to test with parameters. If students submit a macro
  with a specific match for e.g. ` () => {{ ::std::collections::HashMap::new() }};`
  but then not include the same for the default case.

### DIFF
--- a/exercises/macros/tests/macros.rs
+++ b/exercises/macros/tests/macros.rs
@@ -72,7 +72,7 @@ mod test {
     #[test]
     #[ignore]
     fn type_not_in_scope() {
-        let _empty = hashmap!();
+        let _empty: ::std::collections::HashMap<(), ()> = hashmap!();
         let _expected = hashmap!(23=> 623, 34 => 21);
     }
 }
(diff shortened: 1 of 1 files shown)


exercises/macros/tests/macros.rs
```diff
@@ -67,6 +67,16 @@ fn test_nested() {
     );
 }
 
+mod test {
+    use macros::hashmap;
+    #[test]
+    #[ignore]
+    fn type_not_in_scope() {
+        let _empty = hashmap!();
```
coriolinus, line 75: You need a type annotation for the compiler to be able to work with this.

```suggestion
        let _empty: ::std::collections::HashMap<(), ()> = hashmap!();
```

pedantic79: Hmm. I thought I had tested that. I guess I didn't.

pedantic79:
> It's not obvious to me why you moved the test; would you mind explaining the motivation there?

Mostly because I felt it appeared in a weird spot. It was between all the fail compile tests and the `simple_trybuild` module. It's easy to lose track of it as a test. Ideally, those other tests should have been added after this one in #933, instead of sorta before it, and then the module after.

PR #945, opened 2020-04-29
macros: Check recursive solutions use $crate

pedantic79:
- A recursive solution, including the one in `example.rs` should
  use `$crate::` to prefix the recursive macro call. Otherwise,
  renaming the module fails.

### DIFF
--- a/exercises/macros/tests/macros.rs
+++ b/exercises/macros/tests/macros.rs
@@ -68,22 +68,19 @@ fn test_nested() {
 }
 
 mod test {
-    use macros::hashmap;
     #[test]
     #[ignore]
     fn type_not_in_scope() {
+        use macros::hashmap;
+
         let _empty: ::std::collections::HashMap<(), ()> = hashmap!();
         let _expected = hashmap!(23=> 623, 34 => 21);
     }
-}
-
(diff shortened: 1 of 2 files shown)


exercises/macros/tests/macros.rs
```diff
@@ -77,6 +77,16 @@ mod test {
...
 }
 
+mod renamed {
+    use macros as macros_renamed;
+
+    #[test]
+    #[ignore]
+    fn test_renamed_module() {
+        let _expected = macros_renamed::hashmap!(23 => 623, 34 => 21,);
+    }
```
coriolinus, line 87:
```suggestion
    #[test]
    #[ignore]
    fn test_macro_out_of_scope() {
        let _expected = macros::hashmap!(23 => 623, 34 => 21,);
    }
```

Let's simplify: what this test depends on is the `hashmap` macro being out of scope, which is solved by introducing a new module. Renaming the module just confuses things.

pedantic79: I've change the test case, but rather than having two separate modules. I've merged this new test case with the other out of scope one. That required me to move the `use scope::hashmap;` to inside the other testcase.

PR #947, opened 2020-05-08
macros: Assign invalid tests to variable

pedantic79:
- Try to assign to variable, in most cases this is `<_, _>`
- only-arrow and only-comma use `<(), ()>`
- Also enables leading-comma which was accidentally not enabled

### DIFF
--- a/exercises/macros/tests/invalid/Cargo.toml
+++ b/exercises/macros/tests/invalid/Cargo.toml
@@ -41,3 +41,7 @@ path = "triple-arguments.rs"
 [[bin]]
 name = "two-arrows-rs"
 path = "two-arrows.rs"
+
+[[bin]]
+name = "leading-comma-rs"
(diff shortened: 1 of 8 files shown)


coriolinus:
Sorry for the very long delay. Obviously the leading-comma test should be enabled. 

Why assign the output to a variable? It's not obvious to me what the intent behind the change is.

pedantic79:
It's really only for only-arrow. In that case, depending on the pattern used the following code will compile:
```rust
let _hm: HashMap<(), ()> = hashmap!(,);
```

I guess the others don't need to have an assignment.

coriolinus:
Do you mean only-comma?

pedantic79:
sorry yes.

coriolinus:
Given that it's been almost two weeks, I think that we can assume that nobody else will speak up.

PR #948, opened 2020-05-21
macros: Check for additional scoped syntax

pedantic79:
- Had a student who had inverted the logic between trailing-comma
  and non-trailing comma. This change ensures that it works in both
  cases.
- Also added lines between the two similar tests to check for all
  three possible inputs.


pedantic79:
@coriolinus please merge

coriolinus:
Oops, sorry about that!

PR #951, opened 2020-06-01
macros: Additional syntax for hashmap override

pedantic79:
- Some students implement the solution in different ways. This
  ensures that they prefix the namespace to make sure no one
  overrides it.


pedantic79:
Found more students doing weird things that the tests didn't catch.

PR #952, opened 2020-06-05
clock: Add a test case for some edge case

danieljl:
Edge case: Negative minutes that make the hours negative


danieljl:
@coriolinus Is it now safe to merge?

coriolinus:
Yes, sorry for the delay!

PR #953, opened 2020-06-09
protein-translation: Add couple of test to exercise proper decoding

nertpinx:
Length which is not divisible by 3 should not make the test panic, but rather
return None.  That is not the case if it happens after a stop codon.


coriolinus:
Hello, and thanks for the contribution! I'm sorry that nobody else has reviewed this, and I don't have the time to give it proper attention right now. @exercism/rust would one of the other maintainers please look this over?

efx:
I'll pull this down and take a look!

PR #956, opened 2020-06-18
synchronize example with src/lib.rs

efx:
Made plural to match lib/rs.


coriolinus:
It's weird that this got out of sync while keeping CI green, but given that it passes CI, I guess it must be fine.

PR #963, opened 2020-08-11
forth: Fix a bug where user-defined words were deferred

gefjon:
This commit fixes a bug in the definition of `Forth::step_word` where,
to evaluate a user-defined word, it appended that word's definition to
the end of the current code list, which caused that word's definition
to be evaluated only after any remaining code. This commit makes
`step_word` corecursive with `step_term`; to evaluate a user-defined
word, it repeatedly invokes `step_term` on each of the terms in the
definition of that word.

This commit also adds calls to `eprintln` on several error cases, so
that failing tests will log to standard error when and where they fail.

This should fix [#962].

### DIFF
--- a/exercises/forth/example.rs
+++ b/exercises/forth/example.rs
@@ -67,9 +67,8 @@ impl Forth {
 
     fn run(&mut self) -> ForthResult {
         while let Some(term) = self.code.pop_front() {
-            self.step_term(term)?
+            self.step_term(term)?;
         }
-
         Forth::ok()
     }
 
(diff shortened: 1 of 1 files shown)


coriolinus:
Would you merge this branch into #962, please? The code looks plausible, but it would be better to have both commits in one PR, so we can demonstrate that the changes here fix the CI issues there. 

Please don't squash the commits when you merge the two PRs; they'll get squashed automatically when we merge into `master`; until then, it's useful to see the development history.

coriolinus:
Thanks! I'm now going to close this PR in favor of #962. I still intend to merge on Friday, assuming no objections from other maintainers.

PR #968, opened 2020-09-27
poker: example update

elektronaut0815:
Some little updates so the example works with the current versions of Rust and the Counter crate.

### DIFF
--- a/exercises/poker/Cargo-example.toml
+++ b/exercises/poker/Cargo-example.toml
@@ -4,5 +4,4 @@ name = "poker"
 version = "1.0.0"
 
 [dependencies]
-try_opt = "0.1.1"
-counter = "0.1.0"
+counter = "0.5.2"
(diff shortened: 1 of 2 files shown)


coriolinus (commented):
Thanks for contributing this work!

Looks plausible in principle, but of course CI needs to pass before I can approve. Looks like the compiler doesn't like your use of `.collect::<Counter<_>>()` for some reason.

coriolinus:
Oh--it's not really documented well, but the `FromIterator` implementation in `Counter` was added in v0.4.0; the `Cargo-example.toml` file still requires only v0.1.0. Fixing that should fix you up.

elektronaut0815:
Now it's working.

coriolinus (approved):
Looks good, and thanks again!

elektronaut0815:
Thanks for the exercise!

PR #969, opened 2020-09-28
poker: using 'TryFrom' in example

elektronaut0815:
Implementing `std::convert::TryFrom` as this has become the norm.


coriolinus (approved):
Looks good!

It's pretty clear that the Poker example was written a while ago, huh?

elektronaut0815:
Yeah, and also the language is evolving pretty fast.

PR #970, opened 2020-09-29
poker: also using Result for the tested function

elektronaut0815:
I think it's more consistent if Result is used everywhere.

### DIFF
--- a/exercises/poker/example.rs
+++ b/exercises/poker/example.rs
@@ -7,11 +7,11 @@ use counter::Counter;
 ///
 /// Note the type signature: this function should return _the same_ reference to
 /// the winning hand(s) as were passed in, not reconstructed strings which happen to be equal.
-pub fn winning_hands<'a>(hands: &[&'a str]) -> Option<Vec<&'a str>> {
+pub fn winning_hands<'a>(hands: &[&'a str]) -> Result<Vec<&'a str>, &'static str> {
     let mut hands = hands
         .into_iter()
         .map(|source| Hand::try_from(*source))
-        .collect::<Result<Vec<_>, _>>().ok()?;
+        .collect::<Result<Vec<_>, _>>()?;
     hands.sort_by(|a, b| a.partial_cmp(b).unwrap_or(Ordering::Less));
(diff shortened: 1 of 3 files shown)


coriolinus (commented):
Thank you for doing this work, but I do not intend to merge this PR unless you can persuade me to change my mind. Rationale:

- `Option<Vec<_>>` is already questionably idiomatic; an empty vector conveys precisely the same information, namely that there are no winning hands provided. `Result<Vec<_>, _>` doesn't provide any more information than that, because there's still only one circumstance that would trigger the `Err(_)` case; it's just more verbose.
- `Result<_, &'static str>` is an antipattern: if there's only one way a function can fail, it's usually better to return an `Option`, and if there are multiple, then `pub enum Error` is more powerful and flexible. It's fine in things like the example, which most students don't see, but we should absolutely not put that into the public interface that every student must implement.
- `assert!(whatever().is_err());` is an antipattern: it's always better to use `assert_eq!(whatever().unwrap_err(), Error::ExpectedError);`

  (`assert!(whatever().is_none());` is fine, though, because we're not throwing away information about _which_ error has been raised in this case.)

I will basically always approve PRs which only add new tests to the suite, so if you wish to update the new cases to the existing API and discard the rest of the changes, we can get this merged. Otherwise, recommend closing this PR. If there is no further action, then I will close this at some point later than 1 week from now.

elektronaut0815:
Thank you for the comment. Sounds reasonable and is good advice for me on my Rust journey.
Speaking of antipatterns: `Result<_, &'static str>` can be found in _nucleotide-codons_, `assert!(whatever().is_err());` also there and in _grep_ and in _rna-transcription_.

coriolinus:
Following up on the antipattern discussion:

- `nucleotide-codons` is deprecated (see `/config.json`), and isn't shown to students. The only reason it hasn't been deleted entirely is that doing so breaks exercism's website (https://github.com/exercism/rust/pull/832).
- `grep` is a special case where we're teaching the use of application-level helper crates. 
- `rna-transcription` I haven't yet looked at, but will soon.

PR #973, opened 2020-10-02
Grep: replacing deprecated 'failure' with 'thiserror'

elektronaut0815:
As failure [is deprecated](https://github.com/rust-lang-nursery/failure/commit/135e2a3b9af422d9a9dc37ce7c69354c9b36e94b#diff-04c6e90faac2675aa89e2176d2eec7d8), I suggest to switch to [thiserror](https://github.com/dtolnay/thiserror). Also changing the test so it checks the error type.


elektronaut0815:
This PR was superfluous, didn't notice the already existing one by coriolinus.

Issue #974, opened 2020-10-05
Replace Travis CI with Github Actions


coriolinus:
The Rust track has a fairly extensive set of [automatic tests and checks for PRs](https://github.com/exercism/rust/blob/master/.travis.yml), most of which are expressed as scripts in the [`_test`](https://github.com/exercism/rust/tree/master/_test) folder. This all works fine, and has for years now. However, they can be improved.

The biggest expected improvement is in readability: instead of a single red X on a failing commit, requiring someone to dig through the Travis logs to figure out what went wrong, it's relatively simple in Github Actions to break each test into its own independent check, and get a comprehensive list of successes and failures for each commit. 

A related expected improvement is in parallelism: rust compilation can sometimes be slow, and the test suites are sometimes large. If we can perform more checks in parallel, slowdowns should matter less. 

Criteria for acceptance:

- Create appropriate configuration to express all existing tests as independent checks in Github Actions.
- Remove Travis configuration entirely.

dgunay:
I think I'll take a shot at this.

coriolinus:
Closed by #975.

PR #975, opened 2020-10-06
Replace Travis CI with GitHub Actions

dgunay:
Addressing this issue: https://github.com/exercism/rust/issues/974

One note: this only runs the tests on the latest Ubuntu. I saw there is a different process for running the tests on Windows - that would probably require more work to configure correctly. Let me know if anything needs to be changed.


coriolinus:
One thing I don't see in this PR is a replication of the stable/beta/nightly and environment matrix. For example, on Travis, both stable and beta channels run both with and without `DENYWARNINGS=1`, but allows failures for the beta channel with warnings. The nightly channel runs the `check-exercises.sh` script with `BENCHMARKS=1` enabled. I don't see anything in this current configuration selecting a Rust channel at all. Did I miss it?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,53 @@
+name: Run all tests
+
+# Run this workflow every time a new commit is pushed to the repository
+on: push
```
SaschaMann, line 4: This won't work if people open a PR from forks.

```suggestion
on:
  push:
    branches:
    - master
  pull_request:
```

If you want pushes to any branch to trigger a build, remove the filter. The setup above avoids duplicate build in PRs from branches in this repo.

dgunay:
@coriolinus You're right, I did not replicate that. I'll try and implement that.

.github/workflows/tests.yml
```diff
@@ -20,13 +20,14 @@ jobs:
...
-        # For nightly rust, run the check-exercises.sh script with DENYWARNINGS=1
+        # TODO: I don't know how to make it run only the check_exercises.sh script
+        # For nightly rust, run the job with DENYWARNINGS=1
         include:
           - rust: nightly
             deny_warnings: 1
     
     # Allow failures on nightly rust, or beta if DENYWARNINGS is enabled
-    continue-on-error: ${{ matrix.rust == "nightly" || (matrix.rust == "beta" && DENYWARNINGS == 1)}}
+    continue-on-error: ${{ matrix.rust }} == "nightly" || ( ${{ matrix.rust }} == "beta" && DENYWARNINGS == 1 )
```
dgunay, line 30: Having a bit of trouble expressing this condition in a way that pleases GitHub Actions. I get this error message:

```
Error when evaluating 'continue-on-error' for job 'run-all-tests'. (Line: 30, Col: 24): Unexpected value 'stable == "nightly" || ( stable == "beta" && DENYWARNINGS == 1 )'
```

Does anyone know what the correct syntax/placement here is? Hard to test it with my setup because I have to make an entire commit + push to actually run it.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,74 @@
...
+  push:
+    branches:
+    - master
+  pull_request:
+
+jobs:
+  run-all-tests:
+    name: Run all tests
+
+    # TODO: not sure what the best OS to run on is so ubuntu will do for now
```
coriolinus, line 14: Ubuntu is fine; nothing in this track should be OS-specific enough that the host OS actually matters.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,74 @@
...
+    runs-on: ubuntu-latest
+
+    strategy:
+      # Allows running the build multiple times with different configurations
+      matrix:
+        rust: ["stable", "beta"]
+        deny_warnings: ['', 1]
+
+        # TODO: I don't know how to make it run only the check_exercises.sh script
```
coriolinus, line 23: Instead of grouping it in with this strategy, what if you created a new job, which just ran nightly?

Incidentally, the environment for nightly isn't `DENY_WARNINGS=1`; it's `BENCHMARK=1`.

.github/workflows/tests.yml
SaschaMann: I think you're missing quotation marks around `${{ matrix.rust }}`

coriolinus: I don't know for sure, but I suspect that the entire expression has to go inside a `${{ ... }}` group; [docs](https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions). 

At a guess, therefore: 
```suggestion
    continue-on-error: ${{ matrix.rust == 'nightly' || ( matrix.rust == 'beta' && matrix.deny_warnings == '1'  ) }}
```

.github/workflows/tests.yml
```diff
@@ -0,0 +1,74 @@
...
+    
+    # Allow failures on nightly rust, or beta if DENYWARNINGS is enabled
+    continue-on-error: ${{ matrix.rust }} == "nightly" || ( ${{ matrix.rust }} == "beta" && DENYWARNINGS == 1 )
+
+    steps:
+      # Checks out a copy of your repository on the ubuntu-latest machine
+      - name: Checkout code
+        uses: actions/checkout@v2
+        env: 
+          DENYWARNINGS: ${{ matrix.deny_warnings }}
```
coriolinus, line 37: It looks like you're going to need to copy the `env: DENYWARNINGS: ${{ matrix.deny_warnings }}` stuff into each appropriate job step, which should be the main check exercises and ensure stubs compile steps. The rest of the checks shouldn't care about this particular environment variable.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,74 @@
...
+        run: sh ./_test/ensure-stubs-compile.sh
+
+      - name: Count ignores
+        run: sh ./_test/count-ignores.sh
+
+      - name: Fetch configlet
+        run: ./bin/fetch-configlet
+
+      - name: Check configlet format
+        run: ./_test/check-configlet-fmt.sh
```
coriolinus, line 56: There are a few sequential dependencies in here; this one, in particular, requires configlet from the previous "Fetch configlet" step. The [`needs` key](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idneeds) seems designed for this situation.

.github/workflows/tests.yml
coriolinus: Alternately, promoting the environment variable setting to the entire job is probably the better way to go.

.github/workflows/tests.yml
SaschaMann: You only need the `needs` key if you have multiple jobs in a workflow. Steps only run if the previous step succeeded by default. This workflow only has one job with many steps. 

(Splitting them up to different jobs maybe be useful though, for example to run configlet independant of exercise tests.)

coriolinus: I see. In that case, we should probably split this up into a few different jobs which can run in parallel. Something like:

- **compilation job**: ensure stubs compile, check exercises, check-exercise-crate
- **ensuring conventions are followed job**: ensure src/lib.rs files exist, count ignores, check uuids, verify exercise difficulties, check exercises for authors, 
- **configlet job**: fetch configlet, check configlet format, configlet lint, ensure readmes are updated

One nice thing about splitting out the jobs like that is that only the compilation job actually needs to run on a matrix.

coriolinus:
Another thing I just noticed: in several places in `_test`, scripts look for a travis environment variable. A representative example:

https://github.com/exercism/rust/blob/a268661f7bbf9e054848fd911d353b1c8bf754c8/_test/check-exercise-crate.sh#L7

That will have to be dealt with one way or another. At the most basic level, we can re-write the scripts to use a different mechanism to detect whether they're running in a pull request, or we can inject that environment variable if the action was triggered by a pull request, as opposed to just checking the most recent commit of the `master` branch. I have a mild preference for rewriting the scripts to stop referencing Travis, but I suspect that we'd need to inject our own environment variable anyway, so I don't mind leaving the scripts as they are.

dgunay:
Fiddling with the job configuration in a separate branch to avoid spamming this PR - one worrying thing I've noticed is that the scripts (I can see the ones with the old TRAVIS_ env variables at least) fail silently while reporting success in the GitHub Actions UI:

image

coriolinus:
Interesting. I'm having a hard time replicating that error--I don't know what git version they're using, but it appears to be different than mine--but I think we can at least improve error propagation by injecting

```bash
set -e -o pipefail
```

into the affected scripts near the top.

coriolinus:
Also, the error seems to imply that `master` is not in the working tree, which suggests that something in the checkout has failed; that branch should always be available.

dgunay:
I believe this might be what's happening: https://github.community/t/git-ambiguous-argument-master/17832/2

I think changing the script to refer to `origin/master` instead of just `master` might fix it.

SaschaMann:
> I don't know what git version they're using

[git  2.28.0](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md)

dgunay:
> Interesting. I'm having a hard time replicating that error--I don't know what git version they're using, but it appears to be different than mine--but I think we can at least improve error propagation by injecting
> 
> ```shell
> set -e -o pipefail
> ```
> 
> into the affected scripts near the top.

This makes the jobs fail at least, but changing to `origin/master` did not help. I'll try seeing what's wrong.

dgunay:
Added another step to each job to checkout the master branch locally - turns out that wasn't happening when I had a look with `git branch -a`. Will have something new to look at soon.

dgunay:
As far as I am aware, the remaining things left to fix:

1. I'm not sure how to conditionally set `TRAVIS_PULL_REQUEST` based on whether or not the action is being triggered by a pull request.

2. Still not quite sure how to do `continue-on-error` when on the beta toolchain with DENYWARNINGS=1

coriolinus:
I haven't experimented with this, but in the [docs](https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions), I see the following snippet:

```yaml
env:
  my_env_var: ${{ <expression> }}
```

This is tempting, but apparently wrong, because github actions syntax does not support inline conditional expressions. Instead, we should keep looking. For example:

`github.event` | `object` | The full event webhook payload. For more information, see "[Events that trigger workflows.](https://docs.github.com/en/free-pro-team@latest/articles/events-that-trigger-workflows)"
-- | -- | --

Follow through with the [pull request object docs](https://docs.github.com/en/free-pro-team@latest/rest/reference/pulls), the [commands docs](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable), and a bit of [bash variable expansion magic](https://wiki.bash-hackers.org/syntax/pe#use_a_default_value) and it looks like what we need is something like this:

```yaml
steps:
  - name: set TRAVIS_PULL_REQUEST
    env:
      PR_NUMBER: ${{ github.event.pull_request.number }}
    run: echo "TRAVIS_PULL_REQUEST=${PR_NUMBER:-false}" >> $GITHUB_ENV
```

---

Did the suggestion [here](https://github.com/exercism/rust/pull/975#discussion_r500523808) fail? If yes, how?

dgunay:
Okay, I think I've got something that works as intended. Things addressed:

- I made sure to set the selected toolchain as the default during `rustup`
- @coriolinus your suggestion for setting `TRAVIS_PULL_REQUEST` worked nicely
- I got `continue-on-error` to work by copying and pasting it for each step that needed it.

I tested it by altering the `check-exercises.sh` script to fail on `beta` with `DENYWARNINGS=1`. The step 

I'm not super happy about the copying and pasting but perhaps someone with more knowledge of GitHub Actions could make it more DRY.

By the way, `check-exercises.sh` seems to fail right now on `stable` with `DENYWARNINGS=1`. I can reproduce this locally.

dgunay:
Oh, another thing - is `nightly` supposed to be run both with and without `DENYWARNINGS`?

coriolinus:
Great! At this point, it seems that the only things remaining will be to clean up some of the comment-only edits to the test scripts, and actually test the whole thing. 

It is definitely possible to reduce duplication within the source file by using some YAML Anchor and Alias trickery, but I have mixed feelings about that; I've seen the technique used to produce very dense configurations which were almost impossible to understand. I don't oppose trying it out, but I'd reserve the right to ask to revert those changes. I'm sorry to be so vague about the acceptance criteria for this, but I'm finding them hard to verbalize. All I can say is that if the anchor/aliased config is readable and well-documented, I'll probably accept it; if I find it confusing, I'd ask it to be reverted.

Thanks for spotting the DENYWARNINGS failure on stable! I'm working on a fix for that now.

`nightly` does not need `DENYWARNINGS`; we only run nightly at all in order to run the benchmarks.

dgunay:
Sorry for the lack of activity - school kind of took over my life for a bit. Do you have any recommendations for how to test it? Otherwise I guess I was just going to run it manually in these scenarios (third dimension not pictured: with and without failing code):

|   | stable| beta  | nightly  |  
|---|---|---|---|
| direct push |   |   |   |   
| pull request  |   |   |   |   

I do hope there's a good automated way of testing GitHub actions because it seems like a lot of work...

coriolinus:
No worries about lack of activity; this is volunteer work for all of us, and we've all got other things going on in our lives.

I don't think you need to worry about pushing some code which behaves differently according to the stable/beta/nightly dimension; manual verification that those stages have been run is sufficient. Testing direct push to `master` is probably not necessary either: we know that it'll be running the same actions as a pull request, and it'll be straightforward to verify whether or not those steps are complete after merging this PR.

To my mind, what we really want to demonstrate are these cases:

- most recent commit to a PR for which all is good, succeeds
- at least one example for each action which causes that action alone to fail independently. Probably the simplest examples:
  - compilation failure in an exercise's example
  - wrong number of `#[ignore]` lines in a test file
  - README for an exercise has an edit 
  - benchmarking fails for an exercise with a bench

As for how to demonstrate this, I think creating a PR based on this PR targeting your own fork of this repo will cause the actions to run. Create some innocuous commit in that PR to demonstrate the success case, then add / revert commits to demonstrate each of the failure cases. Once that is all done, just link to that separate PR here, and we can inspect to verify that things worked as expected.

coriolinus:
I did https://github.com/exercism/rust/pull/983 to test that these actions work as expected; they succeeded very nicely. I'm going to do a little bit of cleanup now, then approve/merge.

coriolinus (approved):
Thanks for doing all this work @dgunay!

dgunay:
@coriolinus Thank you and @SaschaMann  for the feedback! Glad I could help and learn a bit about GitHub Actions in the process.

PR #988, opened 2020-10-28
Correct lifetimes in anagrams_for signature

BeatButton:
A completed solution will otherwise fail to compile, since the `HashSet` that's returned uses the `&str` references from within `possible_anagrams`.

The proposed signature is also used in [the example solution](https://github.com/exercism/rust/blob/d536862ce3eeaf2fe891e62160bb245f7a8ee79a/exercises/anagram/example.rs#L9).

### DIFF
--- a/exercises/anagram/src/lib.rs
+++ b/exercises/anagram/src/lib.rs
@@ -1,6 +1,6 @@
 use std::collections::HashSet;
 
-pub fn anagrams_for<'a>(word: &str, possible_anagrams: &[&str]) -> HashSet<&'a str> {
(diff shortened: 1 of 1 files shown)


coriolinus:
Prefer to avoid this. One of the goals of this exercise is to teach students what a lifetime error is, and how to fix it. Providing a complete compiling example demonstrates what the lifetimes should look like, which provides much less educational value than figuring it out for oneself.

BeatButton:
Oh, I had assumed that since the exercise instructions made no mention of lifetimes that they weren't a relevant part of the instruction.

coriolinus:
Yeah, the exercise could probably do with a hint about that.

PR #991, opened 2020-10-29
Correct naming conventions in rna-transcription exercise

### LINKED ISSUE #986: Bad naming convention in the nucleotide exercises
Acronyms and Initialisms types should all be UpperCamelCase.  So `DNA` and `RNA` structs should be `Dna` and `Rna`.

Here's the proof: <https://rust-lang.github.io/rfcs/0430-finalizing-naming-conventions.html>

benreyn:
Closes #986

### DIFF
--- a/exercises/rna-transcription/example.rs
+++ b/exercises/rna-transcription/example.rs
@@ -1,4 +1,4 @@
-/// The possible nucleotides in Dna and Rna
+/// The possible nucleotides in DNA and RNA
 #[derive(Debug, Eq, PartialEq, Clone, Copy)]
 pub enum Nucleotide {
     Adenine,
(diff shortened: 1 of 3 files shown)


exercises/rna-transcription/example.rs
```diff
@@ -1,4 +1,4 @@
-/// The possible nucleotides in DNA and RNA
+/// The possible nucleotides in Dna and Rna
```
coriolinus, line 1: Comments are exceptions; normal acronym rules (all-caps) should apply there.

coriolinus (commented):
Please restore comment capitalization. I marked the first instance. Other than that, looks good!

exercises/rna-transcription/example.rs
benreyn: Addressed in 64b1b68. Thanks for the review!

coriolinus (approved):
Looks great, thanks for doing this work!

Issue #992, opened 2020-10-29
PaaS I/O test suit does not account for failing io operations.


meltinglava:
The test suit does not inspect what happens on failed IO operation, neither does the task say anything about it. My suggestion would be that we say that we say that you should only log on sucsessful operations.

coriolinus:
I'd be happy to review a PR adding that documentation.

meltinglava:
I can add the documentation, but the tests does not reflect this. I saw examples of people completing the task with incrementing the amount of reads counters first, then read the bytes, then check for failures and then check update the amount of bytes read. There needs to be a test that makes sure that when an operation fails, the counter and the amounts of bytes read/written as well as the amunts of reads/writs are unchanged.

coriolinus:
That's sensible. That test, plus documentation, would be even better than just the documentation.

ayushmishra2005:
Is someone working on this issue?

coriolinus:
Not that I'm aware of. If you want to give it a shot, go ahead!

senekor:
related [forum post](https://forum.exercism.org/t/rust-paas-i-o-test-cases-for-error-handling/12229)

PR #994, opened 2020-11-05
create new practice exercise: xorcism

coriolinus:
Helps students practice medium-to-advanced lifetimes, generics, and misc. type system trickery. See in particular things like [this](https://github.com/exercism/rust/compare/xorcism?expand=1#diff-ec36775de93da2ecd39a3a8341d24aee142bc8586e02b80c912dcac491b86456R28-R41).

~~Test suite needs expansion and cleanup.~~ Ready for review.


exercises/xorcism/.meta/description.md
```diff
@@ -0,0 +1,69 @@
+Write a streaming adaptor which contains a reference to a key, and bitwise-xors it with arbitrary data.
+
+XOR is a fundamental binary operation: for each bit in the inputs, set the corresponding bit of the
+output to `1` if the input bits are different. If both inputs are `1` or both are `0`, then the corresponding
+output bit is `0`.
+
+XORing a document with a key has been used for cryptography as recently as the early 1900s. The key is repeated
```
petertseng, line 7: This was already obvious to me, but I will also attempt to think of this from the eyes of someone who might not have worked with crypto as much: "The key is repeated as many times as necessary" is a pretty important instruction because it specifies behaviour that is going to be tested. But the first sentence of this paragraph is background context and not important to the problem specification. Might this cause someone to assume the rest of the paragraph is of the same nature (background context) and thus miss the important instruction?

exercises/xorcism/.meta/description.md
```diff
@@ -0,0 +1,69 @@
...
+
+```rust
+fn writer(self, impl Write) -> impl Write
+```
+
+These functions each convert the `Xorcism` struct into a stream adaptor in the appropriate direction.
+They use these traits:
+
+- [`Read`](https://doc.rust-lang.org/std/io/trait.Read.html)
+- [`Write`]https://doc.rust-lang.org/std/io/trait.Write.html
```
petertseng, line 69: It looks like the parentheses are missing here to make this a link. I like to view the rendered file which shows it: https://github.com/exercism/rust/blob/cfd1d8b193650cca9ba260a7038b40876551f923/exercises/xorcism/.meta/description.md

exercises/xorcism/.meta/description.md
```diff
@@ -0,0 +1,69 @@
+Write a streaming adaptor which contains a reference to a key, and bitwise-xors it with arbitrary data.
```
petertseng, line 1: While reading this sentence, I found that https://doc.rust-lang.org/std/iter/trait.Iterator.html contains both "adaptor" and "adapter". Both are valid spellings, but if that Iterator documentation had been consistently using one spelling or the other, I would have suggested using whichever spelling matches the documentation. Since they use both, it seems the choice is ours.

petertseng (commented):
So far, I've reviewed the text and have some food for thought on it. I'd also like to attempt to implement the exercise myself (without having read example.rs) to get a sense of what the student experience might be like. But if I take too long on that, I might say we should not wait for me to try to implement.

exercises/xorcism/.meta/description.md
coriolinus: Good point. Updated in 9aea97b.

exercises/xorcism/.meta/description.md
coriolinus: You're right! Also fixed in 9aea97b.

exercises/xorcism/.meta/description.md
coriolinus: Thanks for checking that. Agree that if the documentation were consistent, we should remain consistent with the documentation; as it stands, I prefer the 'or' variant over 'er'.

coriolinus:
I'm happy to wait as long as it takes for you to attempt to solve this. I wrote the implementation, and then wrote an exercise which fit around it; it would be good to know that _someone_ has solved this exercise as an exercise, before releasing it to students.

I don't remember precisely how long it took, but my impression now is that the original implementation took about five hours to write. It's probably worth budgeting around that much time.

exercises/xorcism/tests/xorcism.rs
```diff
@@ -0,0 +1,303 @@
...
+#[test]
+#[ignore]
+fn statefulness() {
+    // we expect Xorcism to be stateful: at the end of a munging run, the key has rotated.
+    // this means that until the key has completely rotated around, equal inputs will produce
+    // unequal outputs.
+    let key = &[0, 1, 2, 3, 4, 5, 6, 7];
+    let input = &[0b1010_101, 0b0101_0101];
+
+    let mut xs = Xorcism::new(&key);
```
petertseng, line 36: This now makes `&&[{integer}, 8]` - was the second `&` intended?

Of course, with the right type for `new`, no extra handling is required to deal with it, but it's interesting nevertheless.

exercises/xorcism/tests/xorcism.rs
```diff
@@ -0,0 +1,303 @@
...
+}
+
+#[test]
+#[ignore]
+fn statefulness() {
+    // we expect Xorcism to be stateful: at the end of a munging run, the key has rotated.
+    // this means that until the key has completely rotated around, equal inputs will produce
+    // unequal outputs.
+    let key = &[0, 1, 2, 3, 4, 5, 6, 7];
+    let input = &[0b1010_101, 0b0101_0101];
```
petertseng, line 34: Although these two elements have the same value, they are formatted differently and thus may look confusing. Notice how the first one has three elements after the underscore, and the second has four. I don't suppose this is intentional?

exercises/xorcism/.meta/description.md
```diff
@@ -0,0 +1,81 @@
...
+
+When XORing a document with a key, the key is repeated as many times as
+necessary, producing an output document equal in length to the input document.
+
+XORing a document with a key has been used for cryptography as recently as the
+early 1900s. While this is thoroughly obsolete as a method for hiding data, it
+can be surprisingly useful for generating noisy random-seeming data without
+needing the complication of true randomness. It is still used occasionally in
+modern cryptography for certain ciphers: the cipher itself is just a mechanism
+for generating a very random, infinitely long key, which is xor'd with the
```
petertseng, line 16: sometimes in this problem statement, `xor` is lowercased, but sometimes it is all caps. Is there a rule for when one or the other should be used? If not, might want to settle on one, for internal consistency.

exercises/xorcism/tests/xorcism.rs
coriolinus: That was an accident, but as you say, it shouldn't matter given appropriate type parameterization for `new`, so I think I'd prefer to leave it.

exercises/xorcism/tests/xorcism.rs
coriolinus: Good catch! That was an accident. ebfa0e2

exercises/xorcism/.meta/description.md
coriolinus: I know of no rule, so capitalized all instances in dff3350.

exercises/xorcism/tests/xorcism.rs
```diff
@@ -0,0 +1,303 @@
...
+fn statefulness() {
+    // we expect Xorcism to be stateful: at the end of a munging run, the key has rotated.
+    // this means that until the key has completely rotated around, equal inputs will produce
+    // unequal outputs.
+    let key = &[0, 1, 2, 3, 4, 5, 6, 7];
+    let input = &[0b1010_1010, 0b0101_0101];
+
+    let mut xs = Xorcism::new(&key);
+    let out1: Vec<_> = xs.munge(input).collect();
+    let out2: Vec<_> = xs.munge(input).collect();
```
petertseng, line 38: A student who reads only up to this point in the code and hasn't read the later parts may have a question about an ambiguity of multiple calls to munge. They might ask this question:

So, let's say you have two separate in-flight munges:

```rust
let mut iter1 = xs.munge(input);
let mut iter2 = xs.munge(input);
println!("1 has {}", iter1.next().unwrap());
println!("2 has {}", iter2.next().unwrap());
println!("1 has {}", iter1.next().unwrap());
println!("2 has {}", iter2.next().unwrap());
```

Given that the calls to `next()` were interleaved between the two, what should have happened?

1. `iter1` XORs its input with 0 and 1 and `iter2` XORs its input with 2 and 3, just as they would have if the calls were not itnerleaved
2. `iter1` XORs its input with 0 and 2 and `iter2` XORs its input with 1 and 3, reflecting the fact that they were interleaved

Now, behaviour 2 would be extremely difficult to implement (it implies that `out1` and `out2` are mutating some state common to both of them). And, in fact, we test behaviour 1 later on in the round-tripping tests, since we are chaining two iterator adaptors together and the round tripping will only work correctly with behaviour 1.

The only possible concern here, therefore, is if a student does all three of the following:

* reads this test
* doesn't read the round-tripping tests
* somehow misinterprets the test to understand that behaviour 2 is expected.

I cannot predict how likely this scenario will be, but it requires three conditions simultaneously so it could be pretty rare. So I would be happy if the answer here is "take no action for now". However I can also suggest doing one of the following two:

1. Make some sort of comment here making clear that this test isn't testing for behaviour 2. Not sure exactly how to word such a comment, though.
2. Before this test, place a more explicit test of interleaving that makes clear what the right behaviour is

petertseng: EBWODP:

> And, in fact, we test behaviour 1 later on in the round-tripping tests, since we are chaining two iterator adaptors together and the round tripping will only work correctly with behaviour 1.

What I said didn't make sense; it would only be testing behaviour1 if the key length evenly divides the text length. In fact the round-tripping uses two instances of Xorcism.

So we are not testing either behaviour 1 or 2 right now. So we could explicitly say that such behaviour is not being tested at all, explicitly test the behaviour that is actually plausible to implement, or do nothing.

coriolinus: You can't have two separate in-flight munges: `munge` takes `&mut self`.

petertseng: Ooh. That settles things quite nicely, then.

Now, I'm making a guess here that we know this because we know what type `munge` must eventually return (I don't know for sure yet since I haven't completed my implementation, but I know what lifetime I need it to have).

If we start from just the stub signature given to the student:

```rust
pub fn munge<Data>(&mut self, data: Data) -> impl MungeOutput
```

They may not yet know at that point that you cannot have two in-flight munges, and (as I just checked), such a test compiles with the stub signature (though presumably not with the final signature).

So, mentioning that it's not going to be tested could still have value. But leaving things as-is could also be a useful learning experience. I don't feel strongly about it, so perhaps we leave it as-is unless finding out from student feedback that a change is needed.

coriolinus: My inclination in this case is to leave things as-is, changing it if we receive student reports that they're confused on that point.

petertseng:
So I'm done with the main portion of the exercise. My implementation can be examined at https://github.com/petertseng/exercism-rust/tree/xorcism-impl/exercises/xorcism. I have not yet done the `io` portion, and I will be occupied for a few hours so that may be a few days before I can get to that. I also have realised I am probably violating the non-allocation edict of the README because I did a `clone` of an iterator in `munge`. So I'll have to revisit that.

Even though I wrote notes as I went along at https://github.com/petertseng/exercism-rust/blob/xorcism-impl/exercises/xorcism/impl-notes.md, I took out the important parts and reordered them here, so I don't think there's additional value in reading impl-notes.md. But I guess you can if you want.

In terms of whether I would change anything, overall the only thing I found that I would ask about is the third point, so take a look at that one. Everything else is just comments on how it went.

* I feel positively about this exercise. It was instructive in the use of a few traits. I already knew about IntoIterator, but I think I gained a better understanding of AsRef and Borrow. However, I am not yet confident about Sized even after this exercise.
* Your guess of five hours was about correct for me.
* I ended up having `munge` return a concrete type `Munger<'a, Iter, Itm>` rather than an `impl MungeOutput`. I just didn't find it particularly useful to return an `impl` if I only ever return exactly one type. Of course, `Munger` still needs to impl `Iterator` (and also `ExactSizeIterator`, as it turns out). I'd like to ask what you think about making the stub have it that way by default?
* I think it's unfortunate that the compiler doesn't suggest the possibility of `?Sized` in situations where ```the size for values of type `str` cannot be known at compilation time```, but there is not much we can do to change that; that suggestion would need to be made to `rustc`.
* It's also unfortunate we must present the student with so many compiler errors at the start, but I don't think there's a way around it. The student would do well to realise that all of those compiler errors stem from the same cause.
* `munge_in_place` was kind of trivial for the `Cycle<Iter<'a, u8>>`. However `new` and `munge` were instructive. I flailed around unproductively for a while (unproductively guess at some type that looks like it might work, pick one compiler error and fix it, only to cause another one) before taking stock of all the possible types that are provided as arguments to these functions. That allowed me to figure out what trait covers them all. If I had one piece of advice to a student trying to solve this exercise, that would be it.
* I unnecessarily spun my wheels trying to figure out the right traits to use for things like "has a cheap conversion" and "Shouldn't matter whether the byte iterator's values are owned or borrowed." Only upon rereading the README did I realise that everything I needed was in the Useful Traits section. I don't think we can do much about this, since if a student chooses not to read the README and just plow ahead, that's a self-inflicted problem.
* I needed to enforce that the iterator resulting from `IntoIterator` on `munge`'s input is `ExactSizeIterator` and I wasn't sure I was happy with doing that. Turns out, the example does this too, for similar reasons as why I did it, so I think this is inevitable.
* It turns out the interleave test still compiles with my solution. But since I think there could possibly be solutions where it wouldn't compile, I still wouldn't suggest adding it. We already discussed why not.

petertseng:
> I ended up having `munge` return a concrete type `Munger<'a, Iter, Itm>` rather than an `impl MungeOutput`. [...] I'd like to ask what you think about making the stub have it that way by default?

Actually, I think I have to retract this question. the example is returning a `zip`, and I imagine students will possibly want to do that. Easier to provide this `impl MungeOutput` and have them delete it if they decide not to use it than the other way around.

coriolinus:
Incidentally, I believe that you are _not_ violating the no-allocation contract by cloning the `Iter`: you're cloning the iterator itself, which references the data but does not own it. That's cheap, and allowable.

coriolinus:
Incidentally, returning `impl Trait` still constrains you to precisely one output type; the compiler complains otherwise. It's intended for cases where it would be tedious or impossible to name the precise output type. If you need to return a variety of types which satisfy some trait bounds, you need to return `Box<dyn Trait>`.

petertseng (approved):
I finished my `reader` and `writer` implementations and am ready to approve.

They did not take too much additional time (< 30 minutes). `reader` was entirely straightforward  `writer` would theoretically require dealing with the fact that the underlying Write may either error or write not the entire buffer. I'm not wholly satisfied with my solution (write one byte at a time) since I don't know whether it's disadvantageous to make so many calls to `write`. I understand that the tests do not test either of those conditions. I also understand that we're exploring similar possibilities in https://github.com/exercism/rust/issues/992. So as for whether to do the same for this exercise, the arguments would either be "we're already teaching what to do in the face of errors in paasio and therefore don't need to repeat the lesson here" vs "we should always instill the habits of doing the right thing in the face of errors, so we should do it here too". I'll go with whatever you decide for that.

coriolinus:
WRT `writer`: as I see it, the student has three strategies to choose from for their implementation:

- Munge and write one byte at a time, stopping at the first error. Easy to implement and doesn't redo any work, but requires an intermediate `BufWriter` for efficiency if the destination writer has overhead, such as writing to a file or network.
- Munge the entire input slice and call `write_all` on the underlying writer. Easy to implement, but potentially blocks for longer than necessary as multiple calls to the underlying writer are made internally.
- Munge the entire input, but just call the underlying writer's `.write` method. If that returns `Ok(n)` where `n < data.len()-1`, back up the internal state appropriately.

My opinion on testing streaming errors is that if we choose to do so, `paasio` is the right place for it. The intent of adding the streaming features to this exercise was to demo converting a static converter to a streaming one. IMO it's easier to do the right thing for error handling than anything else, given these APIs; therefore, I'm not sure there's much point explicitly testing it in more than one exercise.

PR #996, opened 2020-11-14
Add a test to construct clock with negative 120 minutes

shaaraddalvi:
During my own exercising found an edge case in my code where I handled subtracting 60 minutes but not any of the multiples of 60 minutes, so adding a test hoping it will help others too


coriolinus (changes requested):
Thanks for contributing! Please add a new test instead of editing this one.

coriolinus (approved):
My apologies, I accidentally viewed a single commit instead of the entire diff. This looks good! I'll merge it shortly.

PR #997, opened 2020-11-15
[parallel-letter-frequency]: Clippy says to remove this and all tests still pass

ClashTheBunny:


coriolinus (approved):
Thanks for bringing this exercise more in line with modern standards!

PR #998, opened 2020-11-15
[parallel-letter-frequency]: Clippy says to use resize or static creation

ClashTheBunny:
For cases with identical data, the for loop is not needed


coriolinus (commented):
In this case static creation is the better bet: it avoids mutability entirely, and accomplishes the same thing. It looks like `let v = vec![""; 1000];`. Still, thanks for contributing!

ClashTheBunny:
I'll push an update for this too.

ClashTheBunny:
Switched to static and all tests pass!

coriolinus (approved):

PR #1005, opened 2020-11-17
grep: remove `'static` for statics

petertseng:
clippy:
warning: statics have by default a `'static` lifetime
note: `#[warn(clippy::redundant_static_lifetimes)]` on by default
help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_static_lifetimes

### DIFF
--- a/exercises/grep/tests/grep.rs
+++ b/exercises/grep/tests/grep.rs
@@ -2,7 +2,7 @@ use grep::{grep, Flags};
 
 use std::fs;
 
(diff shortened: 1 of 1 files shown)


coriolinus (approved):
Makes sense, though it might also make sense to actually grep/sed through the repo; something like `s/^((const|static).*&)'static/\1/`.

petertseng:
So, I think this is the only one for tests, but you make a good point that I only did this one exercise. Your suggestion might be applicable to some example solutions. I'll take a look if I have some time, though there's a specific goal I'm looking for right now (Clippy accepts all tests)

Issue #659, opened 2018-09-17
Should rustfmt and clippy checks be added to the CI?


ZapAnton:
The [README for submitting PR](https://github.com/exercism/rust#submitting-a-pull-request) states that the submitted code should be formatted  with `rustfmt`.

Since there is already a [script](https://github.com/exercism/rust/blob/master/bin/format_exercises) that is used to format existing exercises, should it be used in the Travis CI to check the format of the submitted code?

Also as of Rust 1.29 the `cargo clippy` command works on stable. Should it also be used in Travis to ensure that the submitted code is idiomatic?

petertseng:
clippy would be exciting. Might be time to bring back https://github.com/exercism/rust/pull/445 . As of this moment I can't get it to run on the tests, so I guess it'd just be the examples and stubs? That's a bit of a shame. Want to hear solutions.

ZapAnton:
`cargo clippy --all-targets` seems to be working on the whole crate

ZapAnton:
I have run a simple script over the existing exercises:
```shell
   1   │ #!/bin/bash
   2   │ 
   3   │ for dir in exercises/*; do
   4   │         (
   5   │                 cd "$dir"
   6   │ 
   7   │                 cargo clippy --all-targets
   8   │         )
   9   │ done
```

clippy managed to check the `src/lib.rs` and `tests/exercise-name.rs` but could not check `example.rs`.
Perhaps another point in favor of #658?

petertseng:
I have great news to share: Clippy is now being enforced by CI, on the tests only.

Yet to be done is:

* Clippy on the stubs (won't be able to do this for stubs that are allowed to not compile, but can be considered for stubs that do)
* Clippy on the example solutions (pretty low priority IMO, and I'd probably close this issue even if that was not done)
* rustfmt

petertseng:
Regarding example solutions: with the merging of https://github.com/exercism/rust/pull/1009, now we enforce that there are no Clippy *errors* in the example solutions, but we still allow Clippy *warnings*.

petertseng:
Rustfmt has been added to CI. Now I am considering what, if anything, to do about the remaining Clippy warnings in examples and stubs.

[...]

PR #1006, opened 2020-11-17
GitHub Actions: Run Clippy on tests

petertseng:
Supports https://github.com/exercism/rust/issues/659

Note that if we merge this PR, it implies we accept the burden of keeping ourselves in line with Clippy so long as we don't revert it. So we should make sure we want to accept that burden before merging.

Should discuss: Is it possible that an upstream Clippy update will cause previously-succeeding CI to fail on the same commit because of new Clippy lints that have been added? If so, what, if anything, do we want to do about that?

A PR that breaks compilation is included because the behavior of bin/test-exercise changed and it's important to check that it still propagates its exit code.

This will also fail until being rebased on https://github.com/exercism/rust/pull/1003, but that is just as well - it verifies that the Clippy check works correctly.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -99,9 +99,6 @@ jobs:
           PR_NUMBER: ${{ github.event.pull_request.number }}
         run: echo "TRAVIS_PULL_REQUEST=${PR_NUMBER:-false}" >> $GITHUB_ENV
 
-      - name: Install Clippy
-        run: rustup component add clippy
-
       # run scripts as steps
       # TODO: the TRAVIS_PULL_REQUEST variable is a holdover from before the
       # migration to GitHub Actions. The scripts that use it do so in order to
@@ -123,6 +120,35 @@ jobs:
           DENYWARNINGS: ${{ matrix.deny_warnings }}
         run: sh ./_test/check-exercise-crate.sh
         continue-on-error: ${{ matrix.rust == 'beta' && matrix.deny_warnings == '1' }}
+
+  clippy:
+    name: Clippy
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        rust: ["stable", "beta"]
+
+    steps:
+      - name: Checkout master
+        uses: actions/checkout@v2
+        with:
+          ref: master
+
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      # We actually think explicitly installing Clippy isn't necessary,
+      # as it is currently installed by default.
+      # But we'll include this step anyway just in case that changes.
+      - name: Install Clippy
+        run: rustup component add clippy
+
+      - name: Setup toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: ${{ matrix.rust }}
+          default: true
 
       - name: Clippy tests
         env:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -99,6 +99,9 @@ jobs:
           PR_NUMBER: ${{ github.event.pull_request.number }}
         run: echo "TRAVIS_PULL_REQUEST=${PR_NUMBER:-false}" >> $GITHUB_ENV
 
+      - name: Install Clippy
+        run: rustup component add clippy
+
       # run scripts as steps
       # TODO: the TRAVIS_PULL_REQUEST variable is a holdover from before the
       # migration to GitHub Actions. The scripts that use it do so in order to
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yml
```diff
@@ -99,6 +99,9 @@ jobs:
           PR_NUMBER: ${{ github.event.pull_request.number }}
         run: echo "TRAVIS_PULL_REQUEST=${PR_NUMBER:-false}" >> $GITHUB_ENV
 
+      - name: Install Clippy
```
coriolinus, line 102: Right now this is under the `compilation` key, which runs it on all steps in the matrix. I don't think that's necessary. 

Probably the best solution is adding it to its own build job, kind of like the benchmarking on nightly. We might in the future decide that we want to run it on its own matrix, so we can get warnings when new lints are introduced on the beta channel before they hit stable.

petertseng:
> This will also fail until being rebased on #1003, but that is just as well - it verifies that the Clippy check works correctly.

Nope, it doesn't, because of our only running on changed exercises. I'll just push new commits then. But we should still not merge this until #1003 is merged.

_test/check-exercises.sh
```diff
@@ -11,9 +11,9 @@ if [ ! -x "./bin/test-exercise" ]; then
    exit 1
 fi
 
-# In DENYWARNINGS mode, do not set -e so that we run all tests.
+# In DENYWARNINGS or CLIPPY mode, do not set -e so that we run all tests.
 # This allows us to see all warnings.
-if [ -z "$DENYWARNINGS" ]; then
+if [ -z "$DENYWARNINGS$CLIPPY" ]; then
```
coriolinus, line 16:
```suggestion
if [ -z "$DENYWARNINGS" ] || [ -z "$CLIPPY" ]; then
```

Equivalent, but I like the more explicit form better.

petertseng:
> Should discuss: Is it possible that an upstream Clippy update will cause previously-succeeding CI to fail on the same commit because of new Clippy lints that have been added? If so, what, if anything, do we want to do about that?

My assessment is: That is possible, just like how previously-succeeding CI might fail when a new Rust toolchain release gets promoted to beta. We are not doing anything special to deal with that situation either, just fixing them as they come, so I suppose we'd do the same here and not do anything special.

coriolinus (commented):
> Is it possible that an upstream Clippy update will cause previously-succeeding CI to fail on the same commit because of new Clippy lints that have been added?

Yes, and in fact is fairly common. For example, the lint warning about redundant `'static` in `const A_STR: &'static str` was added shortly after it became legal to omit the explicit lifetime from that sort of declaration. The best thing I can think of to mitigate this is to tailor precisely what jobs are run:

- when GHA runs on a PR, it should only run clippy on files modified by that PR
- when GHA runs on `master` branch, it should run clippy on all files

In both cases, we probably want to run it both on `stable` and `beta` rustc so that we get some forewarning when a new lint is going to become a problem. We may wish to configure it that it's not a failing error if clippy reports something on `beta`, but if the build doesn't actually break, I'm not sure we'd actually notice if `beta` started reporting new issues.

_test/check-exercises.sh
petertseng: Sure, I will be more explicit. `-z "$DENYWARNINGS$CLIPPY"` is only true if both are empty, so I think it should be && instead of ||.

petertseng:
> > Is it possible that an upstream Clippy update will cause previously-succeeding CI to fail on the same commit because of new Clippy lints that have been added?
> 
> The best thing I can think of to mitigate this is to tailor precisely what jobs are run:
>
> * when GHA runs on a PR, it should only run clippy on files modified by that PR
> * when GHA runs on master branch, it should run clippy on all files

All right, that seems like a good idea. The "violate clippy by adding return" and subsequent revert give me assurance that the first bullet point has been achieved, and I believe that since it's going through _test/check-exercises.sh that the second bullet point will also be achieved, but 100% certainty on that will only be gained after merging.

petertseng:
If we want either or both of these two new checks "Clippy (stable)" or "Clippy (beta)" to be required, we will need to remember to ask an admin to make them required after this is merged.

petertseng:
... though I see that no checks are currently marked as required, so maybe we don't need to do that and the current situation works just fine for us

.github/workflows/tests.yml
petertseng: OK. I just went ahead and added the matrix. stable and beta.

petertseng: ... which makes me realise and I never actually removed this Install clippy item. and the fact that Clippy still succeeded despite that means that it's being installed by default anyway.

coriolinus (approved):

petertseng:
Oh, here's an idea. Merging this before #1003 would give us a rare opportunity to test the "running on master runs on every exercise" point. I'd still want to be able to merge #1003 right after that gets confirmed, so I'll still wait for approval on #1003.

Squash of this PR is planned.

petertseng:
> running on master runs on every exercise

Nope, didn't happen, because of TRAVIS_PULL_REQUEST. That has convinced me it's time to expunge that. PR coming very soon.

bin/test-exercise
```diff
@@ -82,6 +82,19 @@ done
 # (use subshell so we auto-reset to current pwd after)
 (
    cd $exercise
-   # this is the last command; its exit code is what's passed on
-   time cargo $command "$@"
+   if [ -n "$CLIPPY" ]; then
+     cargo clippy --all-targets --color always "$@" 2>clippy.log
+     # Only consider it a failure if output contains the string in CLIPPY.
+     # For example, if we're only looking in tests, CLIPPY contains tests/
+     if grep -q $CLIPPY clippy.log; then
```
petertseng, line 89: Notice! I have discovered that this has proven error-prone. I will be submitting a PR to correct that in a few hours.

PR #1007, opened 2020-11-17
replace TRAVIS_PULL_REQUEST with GITHUB_EVENT_NAME

petertseng:
When moving off of Travis in
https://github.com/exercism/rust/pull/975, there were some instances of
TRAVIS_PULL_REQUEST left over as a transitionary strategy.

Now we have been hurt by leaving TRAVIS_PULL_REQUEST in:
In https://github.com/exercism/rust/pull/1006 a new GitHub Actions job
was added that forgot to set TRAVIS_PULL_REQUEST and caused incorrect
behaviour.

So let's remove TRAVIS_PULL_REQUEST and instead change it to
GITHUB_EVENT_NAME.

Documentation:
https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -47,12 +47,6 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v2
 
-      # Sets TRAVIS_PULL_REQUEST to false if this is not a pull request.
-      - name: set TRAVIS_PULL_REQUEST
-        env:
-          PR_NUMBER: ${{ github.event.pull_request.number }}
-        run: echo "TRAVIS_PULL_REQUEST=${PR_NUMBER:-false}" >> $GITHUB_ENV
-
       - name: Fetch configlet
         run: ./bin/fetch-configlet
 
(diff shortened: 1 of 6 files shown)


coriolinus (approved):

petertseng:
Apologies, maintainers-admin. my proof commit should have targeted config.json instead. Since I removed the change to config/maintainers.json, your review is not needed. Apologies and thanks for bearing with me.

petertseng:
After this was merged, at this point master's CI fails as desired: https://github.com/exercism/rust/actions/runs/368482577, good.

Time to merge #1003 now.

petertseng:
I suppose the commit message glossed over a few instances of behaviour changes in two scripts. I'm sure they were indeed noticed in review, but I will detail them here for anyone who was not a reviewer and is coming to it after the fact:

1. check-exercise-crate
    * Before: If it's not a pull request, run if exercise crate changed in last 7 days.
    * After: If it's not a pull request, always run it. Reasoning included in comments.
1. check-configlet-fmt
    * Before: If it's not a pull request, run if config.json or config/maintainers.json changed in last 7 days.
    * After: If it's not a pull request, always run it. Reasoning included in comments.

My bad for forgetting to include this in the commit message. One could argue that it's part and parcel of moving off of Travis: the previous "last 7 days" rules assumed that the `master` build runs every 7 days, which was true with Travis is not currently true with GitHub Actions. I think we could set up the master branch to be built every 7 days (https://jasonet.co/posts/scheduled-actions/), but even if we did, I would still have changed the scripts in the way I did, with all the reasoning being the exact same. Regardless of whether it's an integral part of moving off of Travis, it could still have stood a brief mention in the commit message.

PR #1008, opened 2020-11-17
GitHub Actions: Also run weekly

petertseng:
Even if we don't have any activity on master for a while, it is still
valuable to run the CI periodically - this will keep a record of whether
there are any changes caused by new Rust toolchain releases, etc.

All our observed behaviour with our current `push` and `pull_request`
tell us that multiple entries under `on` indicate OR (the workflow runs
on any of these events)
Corroborated by documentation:
https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#on

Documentation on schedule:
https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#schedule

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -6,6 +6,9 @@ on:
     branches:
     - master
   pull_request:
+  schedule:
+    # Weekly.
+    - cron: "0 0 * * 0"
 
 jobs:
   ensure-conventions:


petertseng:
I guess I'm far less worried about this than I was before.

Back in the Travis days, when Exercism administration decided to disable build-on-push for Travis for all of Exercism, I was concerned that this means we'll have no builds for `master` and all our builds would be for PRs only, which (because PRs only build changed exercises) would not be reflective of all exercises.

However, now that we're on GitHub actions and do build on push to master, it's not a concern anymore. I still think it has value because Rust toolchain releases proceed on schedule independently of activity in this repo, but it's just a nice-to-have, not a must-have.

petertseng:
EBWODP: Second paragraph changes to add context to why Exercism admininstration did what they did: I don't mean to paint it as a unilateral decision made without consent. It was a decision reached after discussion with maintainers. Also pointing out that that is why the cron was set up in Travis.

Back in the Travis days, Travis org-wide limits on # of concurrent builts meant that it was necessary for all of Exercism to disable build-on-push for Travis, as the org discussed in https://github.com/exercism/discussions/issues/174. I was concerned that this means we'll have no builds for `master` and all our builds would be for PRs only, which (because PRs only build changed exercises) would not be reflective of all exercises. This is precisely the reason why I set up the weekly cron build in Travis in the first place.

petertseng:
I guess I'm far less worried about this than I was before.

Back in the Travis days, Travis org-wide limits on number of concurrent builds meant that it was necessary for all of Exercism to disable build-on-push for Travis, as the org discussed in https://github.com/exercism/discussions/issues/174. I was concerned that this means we'll have no builds for `master` and all our builds would be for PRs only, which (because PRs only build changed exercises) would not be reflective of all exercises. This is precisely the reason why I set up the weekly cron build in Travis in the first place.

However, now that we're on GitHub Actions and do build on push to master, it's not a concern anymore. I still think it has value because Rust toolchain releases proceed on schedule independently of activity in this repo, but it's just a nice-to-have, not a must-have.

coriolinus (approved):
Agree, it's nice to have, but not essential. Still, as you say: might as well keep the badge updated in case a new release breaks something somehow.

PR #1009, opened 2020-11-18
test-exercise: Detect Clippy failure by exit code, not by grep

petertseng:
The previous grep-based approach was too fragile:
On beta, Clippy isn't even installed.
Therefore, all attempts to run it simply show that Clippy wasn't
installed, without actually running Clippy.
This failure went completely undetected because the grep for `tests/`
isn't going to match that.

It's much less fragile to insert a `deny(clippy:all)` directive and let
the exit code speak.

As a side-effect, now Clippy *errors* for example solutions will cause
the CI to fail, whereas Clippy *warnings* in them will still be
accepted.
We can see this as a value-add, assuming that if Clippy defaults to
treating a particular lint as an error that it really should be fixed.
However this does require that some examples be fixed in this PR.

When merging this PR I plan to keep the commits separate, but will
remove the proof commit and its reversion before merging.


petertseng:
I would like to draw attention to the fact that at this point, https://github.com/exercism/rust/runs/1415913405 says

```
error: 'cargo-clippy' is not installed for the toolchain 'beta-x86_64-unknown-linux-gnu'
To install, run `rustup component add clippy`
```

That leads to the next commit

coriolinus (approved):

exercises/decimal/example.rs
```diff
@@ -117,7 +118,7 @@ macro_rules! auto_impl_decimal_ops {
 
 auto_impl_decimal_ops!(Add, add, |s, o| s + o, |s, _| s);
 auto_impl_decimal_ops!(Sub, sub, |s, o| s - o, |s, _| s);
-auto_impl_decimal_ops!(Mul, mul, |s, o| s * o, |s, o| s + o);
+auto_impl_decimal_ops!(#[allow(clippy::suspicious_arithmetic_impl)] Mul, mul, |s, o| s * o, |s, o| s + o);
```
petertseng, line 121: upon review, since the attribute goes on the `mul`, perhaps it would have been better to write this line as:

```rust
 auto_impl_decimal_ops!(Mul, #[allow(clippy::suspicious_arithmetic_impl)] mul, |s, o| s * o, |s, o| s + o);
```

(with the corresponding change in the macro)

I guess it can be revisited if there's ever a need to have an attribute on both the `impl ` and the `fn`.

PR #1011, opened 2020-11-18
GitHub Actions: Run Clippy on stubs

petertseng:
Fails the build if any stubs have Clippy errors.
Does not fail the build if stubs have Clippy warnings, but does display
them (and there are a number of them).

The author of this PR is not sure we will ever get around to fixing all
the Clippy warnings, but supposes that it is at least beneficial to make
sure we add no new Clippy errors.

1/3 of https://github.com/exercism/rust/issues/659

(By the way, ensure-stubs-compile already had a bad mix of spaces and
tabs and this commit did not help any in making it better. The author's
text editor cannot easily put tabs in shell scripts)

---

Here is the list of warnings emitted from all the stubs (well, each prints twice so you can probably consider each count halved)

```
grep -B1 src/lib.rs gha-clippy-log.txt | sed 's/^20..-..-..T..:..:..........Z //' | grep 'warning\|error' | sort | uniq -c | sort -nr
      4 warning: useless use of `format!`
      2 warning: you should consider adding a `Default` implementation for `SimpleLinkedList<T>`
      2 warning: you should consider adding a `Default` implementation for `School`
      2 warning: you should consider adding a `Default` implementation for `Robot`
      2 warning: you should consider adding a `Default` implementation for `Reactor<T>`
      2 warning: you should consider adding a `Default` implementation for `LinkedList<T>`
      2 warning: you should consider adding a `Default` implementation for `Forth`
      2 warning: you should consider adding a `Default` implementation for `Fizzy<T>`
      2 warning: you should consider adding a `Default` implementation for `BowlingGame`
      2 warning: unneeded unit return type
      2 warning: item `SimpleLinkedList<T>` has a public `len` method but no corresponding `is_empty` method
      2 warning: item `LinkedList<T>` has a public `len` method but no corresponding `is_empty` method
      2 warning: defining a method called `next` on this type; consider implementing the `std::iter::Iterator` trait or choosing a less ambiguous name
```

### DIFF
--- a/_test/ensure-stubs-compile.sh
+++ b/_test/ensure-stubs-compile.sh
@@ -30,21 +30,40 @@ for dir in $changed_exercises; do
 	cp -r $dir/tests $dir/tests.orig
 	cp $dir/src/lib.rs $dir/lib.rs.orig
 
-	# This sed serves two purposes:
-	# First, in Travis CI, we may have already compiled using the example solution.
-	# Edit the src/lib.rs file so that we surely recompile using the stub.
-	# Second, ensures that the stub compiles without warnings.
-	sed -i -e '1i #![deny(warnings)]' "$dir/src/lib.rs"
+  # In CI, we may have already compiled using the example solution.
+  # So we want to touch the src/lib.rs in some way,
+  # so that we surely recompile using the stub.
+  if [ -n "$CLIPPY" ]; then
+    # We don't deny warnings in clippy mode,
+    # just because there are many we don't want to deal with right now.
+    # So in clippy mode it's just a touch.
+    touch "$dir/src/lib.rs"
+  else
+    # In non-clippy mode, we do want to compile without warnings.
+    sed -i -e '1i #![deny(warnings)]' "$dir/src/lib.rs"
+  fi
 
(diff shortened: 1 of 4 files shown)


_test/ensure-stubs-compile.sh
```diff
@@ -30,21 +30,39 @@ for dir in $changed_exercises; do
 	cp -r $dir/tests $dir/tests.orig
 	cp $dir/src/lib.rs $dir/lib.rs.orig
 
-	# This sed serves two purposes:
-	# First, in Travis CI, we may have already compiled using the example solution.
-	# Edit the src/lib.rs file so that we surely recompile using the stub.
-	# Second, ensures that the stub compiles without warnings.
-	sed -i -e '1i #![deny(warnings)]' "$dir/src/lib.rs"
+  if [ -z "$CLIPPY" ]; then
```
petertseng, line 33: Potentially confusing. -z here, but -n below. Use one or the other consistently throughout, otherwise confuses reader.

coriolinus (approved):

coriolinus:
WRT tabs and spaces. I hadn't noticed, but I would not object to a global replacement of tabs with spaces in our script files.

PR #1012, opened 2020-11-18
DO NOT MERGE: Deny clippy warnings in example solutions

petertseng:
The author isn't really interested in prioritising these but wants to
at least take a look at how the situation is.


petertseng:
Well, I hit enter too soon, so I'm going to have to EBWODP the description.

petertseng:
EBWODP:

The author isn't really interested in prioritising these but wants to
at least take a look at how the situation is.

petertseng:
```
grep -B1 src/lib.rs gha-clippy-log.txt | sed 's/^20..-..-..T..:..:..........Z //' | grep 'warning\|error' | sort | uniq -c | sort -nr
     14 error: constants have by default a `'static` lifetime
     10 error: statics have by default a `'static` lifetime
      8 error: `if` chain can be rewritten with `match`
      6 error: Use sort_unstable instead of sort
      6 error: redundant clone
      6 error: casting a character literal to `u8` truncates
      6 error: called `.nth(0)` on a `std::iter::Iterator`, when `.next()` is equivalent
      4 error: writing `&Vec<_>` instead of `&[_]` involves one more reference and cannot be used with non-Vec-based slices.
      4 error: unneeded unit return type
      2 error: you should consider adding a `Default` implementation for `SimpleLinkedList<T>`
      2 error: you should consider adding a `Default` implementation for `School`
      2 error: you should consider adding a `Default` implementation for `LinkedList<T>`
      2 error: you should consider adding a `Default` implementation for `graph::Graph`
      2 error: you should consider adding a `Default` implementation for `Forth`
      2 error: you should consider adding a `Default` implementation for `BowlingGame`
      2 error: using `Result.and_then(|x| Ok(y))`, which is more succinctly expressed as `map(|x| y)`
      2 error: Using `.iter().next()` on an array
      2 error: this `.into_iter()` call is equivalent to `.iter()` and will not move the `slice`
      2 error: this `else { if .. }` block can be collapsed
      2 error: redundant pattern matching, consider using `is_some()`
      2 error: methods called `into_*` usually take self by value; consider choosing a less ambiguous name
      2 error: it looks like the same item is being pushed into this Vec
      2 error: item `SimpleLinkedList<T>` has a public `len` method but no corresponding `is_empty` method
      2 error: item `LinkedList<T>` has a public `len` method but no corresponding `is_empty` method
      2 error: defining a method called `next` on this type; consider implementing the `std::iter::Iterator` trait or choosing a less ambiguous name
      2 error: avoid using `collect()` when not needed
      1 error: this `.into_iter()` call is equivalent to `.iter()` and will not move the `array`
```

Will consider fixing some of these at some point, but don't really find it urgent.

coriolinus:
At this point all lints with more than 2 errors reported should be fixed.

petertseng:
Yes, that is true! Excellent.

```
      2 error: you should use the `starts_with` method
      2 error: you should consider adding a `Default` implementation for `SimpleLinkedList<T>`
      2 error: you should consider adding a `Default` implementation for `School`
      2 error: you should consider adding a `Default` implementation for `Robot`
      2 error: you should consider adding a `Default` implementation for `Reactor<'a, T>`
      2 error: you should consider adding a `Default` implementation for `LinkedList<T>`
      2 error: you should consider adding a `Default` implementation for `graph::Graph`
      2 error: you should consider adding a `Default` implementation for `Forth`
      2 error: you should consider adding a `Default` implementation for `BowlingGame`
      2 error: using `Result.and_then(|x| Ok(y))`, which is more succinctly expressed as `map(|x| y)`
      2 error: Using `.iter().next()` on an array
      2 error: this `.into_iter()` call is equivalent to `.iter()` and will not move the `slice`
      2 error: this `else { if .. }` block can be collapsed
      2 error: redundant pattern matching, consider using `is_some()`
      2 error: methods called `into_*` usually take self by value; consider choosing a less ambiguous name
      2 error: it looks like the same item is being pushed into this Vec
      2 error: item `SimpleLinkedList<T>` has a public `len` method but no corresponding `is_empty` method
      2 error: item `LinkedList<T>` has a public `len` method but no corresponding `is_empty` method
      2 error: defining a method called `next` on this type; consider implementing the `std::iter::Iterator` trait or choosing a less ambiguous name
      2 error: called `filter(p).next()` on an `Iterator`. This is more succinctly expressed by calling `.find(p)` instead.
      2 error: avoid using `collect()` when not needed
      1 error: this `.into_iter()` call is equivalent to `.iter()` and will not move the `array`
```

PR #1014, opened 2020-11-18
rewrite if chains with match in exercise examples

coriolinus:
Helps address #1012.

### DIFF
--- a/exercises/perfect-numbers/example.rs
+++ b/exercises/perfect-numbers/example.rs
@@ -3,12 +3,10 @@ pub fn classify(num: u64) -> Option<Classification> {
         return None;
     }
     let sum: u64 = (1..num).filter(|i| num % i == 0).sum();
-    if sum == num {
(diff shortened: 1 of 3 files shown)


exercises/perfect-numbers/example.rs
```diff
@@ -3,12 +3,10 @@ pub fn classify(num: u64) -> Option<Classification> {
         return None;
     }
     let sum: u64 = (1..num).filter(|i| num % i == 0).sum();
-    if sum == num {
-        Some(Classification::Perfect)
-    } else if sum < num {
-        Some(Classification::Deficient)
-    } else {
-        Some(Classification::Abundant)
+    match sum.cmp(&num) {
```
petertseng, line 6: I notice that putting this whole thing in a `Some` is possible, but right now we are just rewriting `if`s, not doing that.

petertseng (approved):

PR #1017, opened 2020-11-18
use byte literals instead of casting char to u8 in exercise examples

coriolinus:
Helps address #1012.

### DIFF
--- a/exercises/simple-cipher/example.rs
+++ b/exercises/simple-cipher/example.rs
@@ -5,7 +5,7 @@ pub fn encode_random(s: &str) -> (String, String) {
     let mut r = rand::thread_rng();
     let mut key = String::new();
     for _ in 0..100 {
-        key.push(char::from('a' as u8 + r.gen_range(0, 26)));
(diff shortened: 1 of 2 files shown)


exercises/simple-cipher/example.rs
```diff
@@ -34,8 +34,8 @@ fn shift(key: &str, s: &str, dir: i8) -> Option<String> {
     }
     for c in s.chars() {
         let shift = key_arr[i % key_arr.len()] as i8 - 'a' as i8;
-        let n = ((c as i8 - 'a' as i8 + dir * shift) % 26 + 26) % 26;
-        o.push(char::from('a' as u8 + n as u8));
+        let n = ((c as i8 - b'a' as i8 + dir * shift) % 26 + 26) % 26;
```
petertseng, line 37: I don't think Clippy had anything to say about this one, but let's do it anyway for consistency

petertseng (approved):

exercises/simple-cipher/example.rs
coriolinus: Exactly.

PR #1020, opened 2020-11-18
remove unneeded unit return type in exercise examples

coriolinus:
Helps address #1012.


petertseng (approved):
I notice this is one of those where the stub has the same, but I suppose the stub can be dealt with later.

Or you could deal with it now and s/in exercise examples/in react/ for the commit message since react is the only exercise in question

petertseng (approved):

PR #1021, opened 2020-11-18
GitHub Actions: Enforce exercises' formatting; format exercises

petertseng:
Part of #659 

To be discussed before merging: Is there any chance at all that a contributor may have a version of `rustfmt` that is different from that run by GitHub Actions, and therefore causes hardship on them in complying with `rustfmt`?

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -109,6 +109,33 @@ jobs:
         run: sh ./_test/check-exercise-crate.sh
         continue-on-error: ${{ matrix.rust == 'beta' && matrix.deny_warnings == '1' }}
 
+  rustformat:
+    name: Check Rust Formatting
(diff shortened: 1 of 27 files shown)


petertseng:
Proof will be given by the fact that the "Check Rust Formatting" failed before the "format all exercises" commit was added.

coriolinus (approved):
I like this idea. Rustfmt's output has changed in the past for identical inputs, and it's likely to do so again in the future, but we have a convention in this track that everyone should be running the latest stable release anyway; given that rustfmt is distributed alongside Rust itself, I don't think we need to worry too much about contributions from people with outdated compilers.

This brings us one more step in the direction where we can just say "Just merge it once CI goes green" without worrying about having forgotten some edge case.

PR #1029, opened 2020-11-20
doubly-linked-list: Allow should_implement_trait

petertseng:
It's understandable that clippy says the `next` here conflicts with
Iterator's `next`, but I really think that's the best name for it unless
we go with something like `forward` and `back`.

There is one other example of Cursors on the internet:
https://contain-rs.github.io/linked-list/linked_list/struct.Cursor.html
This one also uses `next`, so it seems they did not find a better
solution to this than we could.

Helps address https://github.com/exercism/rust/pull/1011
Helps address https://github.com/exercism/rust/pull/1012

### DIFF
--- a/exercises/doubly-linked-list/src/lib.rs
+++ b/exercises/doubly-linked-list/src/lib.rs
@@ -42,6 +42,7 @@ impl<T> Cursor<'_, T> {
         unimplemented!()
     }
 
+    #[allow(clippy::should_implement_trait)]
     /// Move one position forward (towards the back) and
     /// return a reference to the new position
     pub fn next(&mut self) -> Option<&mut T> {
--- a/exercises/doubly-linked-list/src/lib.rs
+++ b/exercises/doubly-linked-list/src/lib.rs
@@ -42,9 +42,9 @@ impl<T> Cursor<'_, T> {
         unimplemented!()
     }
 
-    #[allow(clippy::should_implement_trait)]
     /// Move one position forward (towards the back) and
     /// return a reference to the new position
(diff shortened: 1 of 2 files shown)


petertseng:
I have changed my mind since saying "I think it'd be weird to distribute this attribute to students". I think it's fine. Maybe it'll get them curious about Clippy.

exercises/doubly-linked-list/src/lib.rs
```diff
@@ -42,6 +42,7 @@ impl<T> Cursor<'_, T> {
         unimplemented!()
     }
 
+    #[allow(clippy::should_implement_trait)]
     /// Move one position forward (towards the back) and
     /// return a reference to the new position
     pub fn next(&mut self) -> Option<&mut T> {
```
coriolinus, line 48:
```suggestion
    /// Move one position forward (towards the back) and
    /// return a reference to the new position
    #[allow(clippy::should_implement_trait)]
    pub fn next(&mut self) -> Option<&mut T> {
```

Purely a style thing, but the convention I prefer is that attributes come after doc-comments. The reasoning is that doc comments can get very long, but there are very rarely more than a few lines of attributes. Putting the doc comment first ensures that attributes are visible when looking at the top of the definition.

coriolinus (approved):
I don't feel strongly one way or the other about whether it's weird to show this annotation to students. We already recommend clippy in every readme file, so they should at least be aware that it's a thing. Once a student has reached the point of implementing doubly-linked list, I think it's unlikely that they'll never have seen anything like this before.

PR #1030, opened 2020-11-20
ensure-stubs-compile: Deny all warnings except new_without_default

petertseng:
We don't find it too useful to have students implement Default.
All other warnings in stubs have been dealt with, and we find it useful
to ensure there are no further ones.

I've determined via https://github.com/petertseng/exercism-rust/actions/runs/373860276 that when #1027, #1028, #1029, and this are merged, there are no more Clippy warnings or errors in stubs.

Closes https://github.com/exercism/rust/issues/659
This is appropriate because:

* stubs and tests are now ensured to have no Clippy warnings or errors
* examples are ensured to have no errors, and we're unconcerned about
  them having warnings since students don't see them
* rustfmt is enforced

### DIFF
--- a/_test/ensure-stubs-compile.sh
+++ b/_test/ensure-stubs-compile.sh
@@ -33,15 +33,8 @@ for dir in $changed_exercises; do
   # In CI, we may have already compiled using the example solution.
   # So we want to touch the src/lib.rs in some way,
   # so that we surely recompile using the stub.
-  if [ -n "$CLIPPY" ]; then
-    # We don't deny warnings in clippy mode,
-    # just because there are many we don't want to deal with right now.
-    # So in clippy mode it's just a touch.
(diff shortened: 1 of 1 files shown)


petertseng:
> We don't find it too useful to have students implement Default.

Well, we should discuss that, since by "we" right now I just mean "I".

coriolinus:
> We don't find it too useful to have students implement Default.

So, I'm a little conflicted about this, because clippy's recommendation is in fact good practice. In general, it's nice to be able to grab a `T::default()` without having to care too much about what that really means. On the other hand, in the context of these exercises, students aren't usually actually creating the types they define; the test suite does that. Making them add `#[derive(Default)]` to their structs is just busywork.

I am therefore convinced that it is not very useful to have students implement `Default`.

coriolinus (approved):

PR #1032, opened 2020-11-20
Fix #961: Change Forth stack return type to slice

### LINKED ISSUE #961: Unidiomatic API in `forth`
https://github.com/exercism/rust/blob/2d163cfc6c02f8e8b6ac2044a860a4c0a3f5da93/exercises/forth/src/lib.rs#L19-L21

Rust is designed to enable APIs to use the privacy of their implementation for soundness and stability. An interpreter's stack is rarely relevant to the public API, and (imo) wouldn't be found in a "good" Rust crate. It's an implementation detail.

I believe the issue is in the test structure: Since the state of the stack is an implementation detail, tests for its state are *unit tests*. Rust handles these using tests in the main crate:

```rust
pub struct Forth;

impl Forth {
    pub fn new() -> Forth {
        unimplemented!()
    }

    fn stack(&self) -> Vec<Value> {
        unimplemented!()
    }

    pub fn eval(&mut self, input: &str) -> ForthResult {
        unimplemented!("result of evaluating '{}'", input)
    }
}

#[test]
fn new_stack_is_empty() {
    assert_eq!(Forth::new().stack(), vec![]);
}
```

I'm also not a fan of this function signature in the exercise. It asks the user to return an owned vector from a reference to their interpreter. This only gives the user the option to clone whatever data structure they're using to represent their stack (which will typically just be `self.stack.clone()`).

As a "hard" task, can students attempting this exercise be expected to have a decent understanding of the language? If they know about existential types and iterators (covered by The Book), I think this API would be far less prescriptive:

```rust
fn stack(&self) -> impl Iterator<Item = Value>;
```

PaulDance:
Closes #961: as discussed, this adjusts the return type to `&[Value]` in order to reduce the amount of cloning required and since it is used for tests only, it does not truly break the type's API.

### DIFF
--- a/exercises/forth/src/lib.rs
+++ b/exercises/forth/src/lib.rs
@@ -16,7 +16,7 @@ impl Forth {
         unimplemented!()
     }
 
-    pub fn stack(&self) -> Vec<Value> {
+    pub fn stack(&self) -> &[Value] {
         unimplemented!()
     }
 


coriolinus (approved):
Haha, I'd imagined it would be a bigger change than that!

PaulDance:
Me too :smile:!

coriolinus:
Actually, I shouldn't have merged quite that quickly. Would you please update `example.rs` as well, as a demonstration that the tests don't need changing? It'll need to be in a new PR, but it's better to be able to show for sure that the new API works.

PaulDance:
Okay, I'll get on it then!

PaulDance:
Wait, actually @coriolinus, the stack used in the example surprisingly is a `LinkedList` instead of a `Vec` so borrowing a slice is impossible... Should we try to refactor the example in order to use a vector instead?

coriolinus:
Yes, please.

PaulDance:
Okay, I just didn't know if it was something we could consider `¯\_(°^°)_/¯`

PR #1035, opened 2020-11-21
decimal: Use unwrap_or; use ? instead of try_opt

petertseng:
https://rust-lang.github.io/rust-clippy/master/index.html#manual_unwrap_or

### DIFF
--- a/exercises/decimal/example.rs
+++ b/exercises/decimal/example.rs
@@ -56,10 +56,7 @@ impl Decimal {
         }
         Some(Decimal::new(
             try_opt!(digits.parse::<BigInt>().ok()),
-            match decimal_index {
-                Some(idx) => idx,
-                None => 0,
-            },
+            decimal_index.unwrap_or_default(),
         ))
     }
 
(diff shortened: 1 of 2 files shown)


exercises/decimal/example.rs
```diff
@@ -56,10 +56,7 @@ impl Decimal {
         }
         Some(Decimal::new(
             try_opt!(digits.parse::<BigInt>().ok()),
-            match decimal_index {
-                Some(idx) => idx,
-                None => 0,
-            },
+            decimal_index.unwrap_or(0),
```
coriolinus, line 59:
```suggestion
            decimal_index.unwrap_or_default(),
```

If we're using monadic combinators, might as well go all the way.

exercises/decimal/example.rs
```diff
@@ -56,10 +56,7 @@ impl Decimal {
         }
         Some(Decimal::new(
             try_opt!(digits.parse::<BigInt>().ok()),
```
coriolinus, line 58: Might be out of scope of this PR, but I suspect that we could also simplify this line:

```suggestion
            digits.parse().ok()?,
```

coriolinus (approved):

PR #1036, opened 2020-11-21
phone-number, scale-generator: use range::contains

petertseng:
I'm honestly not really convinced by the phone-number one but I guess
it's one fewer mention of number_len so it's fine.

https://rust-lang.github.io/rust-clippy/master/index.html#manual_range_contains


coriolinus (approved):
As you say, the phone-number one is a little weird, but might as well resolve the warning.

PR #1042, opened 2021-01-04
space-age: Update README.md

bobahop:
Suggest the use of a macro

### DIFF
--- a/exercises/space-age/README.md
+++ b/exercises/space-age/README.md
@@ -23,6 +23,12 @@ Some Rust topics you may want to read about while solving this problem:
 
 - Traits, both the From trait and implementing your own traits
 - Default method implementations for traits
+- Macros, the use of a macro could reduce the boilerplate for this exercise. For instance, 
+  [a macro can implement a trait for multiple types at once](https://stackoverflow.com/questions/39150216/implementing-a-trait-for-multiple-types-at-once),
+  though it is fine to implement `years_during` in the Planet trait itself. A macro could
+  define both the structs and their implementations. Info to get started with macros can 
+  be found [here](https://doc.rust-lang.org/rust-by-example/macros/syntax.html) and 
+  [here](https://doc.rust-lang.org/1.7.0/book/macros.html).
 
 
 ## Rust Installation


bobahop:
Occasionally a student will say they don't feel they are ready for macros, but often a student appreciatively says something like "I _knew_ there must be a better way to do it! Thanks!" and they submit an iteration using a macro.

coriolinus:
CI is failing because you cannot just update README.md: that's a generated file. Instead, what you should do is:

- write the hint in `.meta/hints.md` (creating the file if necessary)
- get configlet (`rust/bin/fetch-configlet`)
- regenerate the readme: `configlet generate . -o anagram

I'll take a look later, when CI passes.

coriolinus:
That said, I'm not convinced that a macro is actually the most elegant way to do this. These days, I'd lean towards a [const-based implementation](https://github.com/coriolinus/exercism-rs/blob/a5992dc9c9f9954f3aadee5a4d7744f83ab0cdab/space-age/src/lib.rs). You can save a tiny bit of boilerplate by adding a macro, bringing the impl-lines per planet from 3 to 1, but only at the cost of adding a non-trivial recursive macro. I'm not sure that tradeoff is worth it in this case.

iago-lito:
I'd argue that "saving a bit of boilerplate" is not the only benefit of using macros here. Readability of the data counts, and I find the following much more comfortable to parse in [this kind](https://exercism.io/tracks/rust/exercises/space-age/solutions/04d5ad52fe984cc681bc8d79c4bdb1c0) of macro approach:
```rust
declare_planets! {
    Mercury:   0.2408467
      Venus:   0.61519726
      Earth:   1.0
       Mars:   1.8808158
    Jupiter:  11.862615
     Saturn:  29.447498
     Uranus:  84.016846
    Neptune: 164.79132
}
```
In particular, these solutions *can* also use associated consts.
Also, the boilerplate factorized by the macro is rather trivial, which makes it easy to understand and improves readability much. This is a canonical use-case for macros IMHO.

coriolinus:
Good point. That macro is simpler and nicer than I'd expected; I'd imagined for some reason that we'd need a recursive deconstruction of the list, instead of just using the repetition operator. Consider me convinced that it's worth suggesting a macro.

gilescope:
Agreed - all for hinting that a macro would save them time - it's a bit tedious without macros as exercises go. Without macros I could happily live in a solar system with only three planets in.

bobahop:
Okay, I committed the local edits and pushed them to my fork on GH. Not sure how I can create a pull request for just the space-age files, as it did not create a separate branch for today's space-age changes. It looks like it wants to PR today's anagram file changes as well as today's space-age file changes if I do a pull request. I don't want to create a mess. Sorry to be a novice at this.

bobahop:
I'm gonna try to push to yesterday's space-age branch from local...

coriolinus:
You may have pushed the local edits to the wrong fork. Here's how to recover the changes to this one.

- Get the commit hash of your edits:
  ```sh
  $ git log -1
  commit a571f8d9844a669601cc5167b7055c75606c19d9
  ```
  The hash is the long string after "commit"
- Check out this branch. We can see from the top of the PR that this branch is called `patch-1`, so:
  ```sh
  git checkout patch-1
  ```
- Copy that commit into this branch:
  ```sh
  git cherry-pick a571f8d9844a669601cc5167b7055c75606c19d9
  ```
  Where the hash you use is the actual one you got from your commit previously. The branch head automatically updates.
- Push the branch.
  ```sh
  git push
  ```
  This PR will automatically update at that point.

bobahop:
Which didn't work. Sigh...

```
C:\rust_projects\Exercism\rust\exercises\space-age>git push origin branch patch-1
error: src refspec branch does not match any
error: src refspec patch-1 does not match any
error: failed to push some refs to 'https://github.com/bobahop/rust.git'

C:\rust_projects\Exercism\rust>git push origin branch patch-1
error: src refspec branch does not match any
error: src refspec patch-1 does not match any
error: failed to push some refs to 'https://github.com/bobahop/rust.git'
```

bobahop:
I think the problem is that yesterday I edited in the browser and today I edited in my local clone.

```
C:\rust_projects\Exercism\rust\exercises\space-age>git checkout patch-1
error: pathspec 'patch-1' did not match any file(s) known to git
```

bobahop:
Okay, did git fetch and will go through your instructions again...

bobahop:
This is turning into a royal clusterfest because my changes were a bit different from yesterday. I'm gonna delete my fork and redo it the right way from scratch.

bobahop:
Superceded by #1047

PR #1043, opened 2021-01-04
anagram: Update README.md

bobahop:
limit case changes and consider sort_unstable

### DIFF
--- a/exercises/anagram/README.md
+++ b/exercises/anagram/README.md
@@ -15,6 +15,9 @@ The solution cannot contain the input word. A word is always an anagram of itsel
 
 You are going to have to adjust the function signature provided in the stub in order for the lifetimes to work out properly. This is intentional: what's there demonstrates the basics of lifetime syntax, and what's missing teaches how to interpret lifetime-related compiler errors.
 
+Try to limit case changes. Case changes are expensive in terms of time, so it's faster to minimize them.
+
+If sorting, consider [sort_unstable](https://doc.rust-lang.org/std/primitive.slice.html#method.sort_unstable) which is typically faster than stable sorting. When applicable, unstable sorting is preferred because it is generally faster than stable sorting and it doesn't allocate auxiliary memory.
 
 ## Rust Installation
 


bobahop:
Many solutions repeatedly change the case of `word`, and also repeatedly sort it every time it is compared with a possible anagram. Also, often a possible anagram is  case-changed for comparing with the case-changed `word` and then case-changed again for the purposes of sorting. By minimizing the case changes and using `sort_unstable` the solutions are usually two to three times faster.

Another hint would be to test the lengths of `word` and the possible anagram before doing any case changes and sorting of the possible anagram, but I left that out as I considered it might be too much detail up front. But that is another performance tip I often find myself making.

coriolinus:
CI is failing because you cannot just update README.md: that's a generated file. Instead, what you should do is:

- write the hint in `.meta/hints.md` (creating the file if necessary)
- get configlet (`rust/bin/fetch-configlet`)
- regenerate the readme: `configlet generate . -o anagram

I'll take a look later, when CI passes.

gilescope:
Those seem helpful hints to me. 👍

bobahop:
I'm having trouble getting the configlet to work. 

```
C:\rust_projects\Exercism\rust\bin>fetch-configlet
'fetch-configlet' is not recognized as an internal or external command,
operable program or batch file.
```

Do I need to run in a special shell in Windows 10?

bobahop:
Sorry to be so helpless. I'm able to run the downloaded configlet but get a missing path. I find no `problem-specifications` directory when I search the `rust` project.

C:\rust_projects\Exercism\rust\bin>configlet generate . -o anagram
-> path not found: C:\rust_projects\Exercism\rust\problem-specifications

coriolinus:
Oh, yeah, you need a local copy of that repo also. 

```sh
git clone --depth 1 https://github.com/exercism/problem-specifications.git C:\rust_projects\Exercism\rust\problem-specifications
```

bobahop:
Ok, thanks very much! I created a separate PR with the two modified files hint.md and ReadMe.md. So I guess this PR can be deleted/closed.  I didn't want to trouble you about how to modify this one, as it seemed more straightforward to just do another one.

coriolinus:
Superseded by https://github.com/exercism/rust/pull/1045.

PR #1044, opened 2021-01-04
Update forth.rs

bobahop:
Redefine_an_existing_word_with_another_existing_word. Compare with "can_use_different_words_with_the_same_name," which defines a new word with an existing word.


bobahop:
When I first solved Forth I was curious if my solution would pass this test I created, and it did not. So I think it may add something to include this test.

coriolinus (approved):
This is a great addition, and the tests already pass. Thanks for contributing this!

petertseng:
Candidate for inclusion in problem-specifications? https://github.com/exercism/problem-specifications/blob/master/exercises/forth/canonical-data.json

PR #1045, opened 2021-01-05
anagram: Update README.md

bobahop:
Many solutions repeatedly change the case of word, and also repeatedly sort it every time it is compared with a possible anagram. Also, often a possible anagram is case-changed for comparing with the case-changed word and then case-changed again for the purposes of sorting. By minimizing the case changes and using sort_unstable the solutions are usually two to three times faster.

Another hint would be to test the lengths of word and the possible anagram before doing any case changes and sorting of the possible anagram, but I left that out as I considered it might be too much detail up front. But that is another performance tip I often find myself making.


coriolinus (approved):
Looks good! Happy to merge this.

PR #1046, opened 2021-01-05
grade-school: update to fix clippy lints

coriolinus:
Also update the hint about why `grade` doesn't return a reference.

Fixes CI failures on `master` branch.

### DIFF
--- a/exercises/grade-school/example.rs
+++ b/exercises/grade-school/example.rs
@@ -23,7 +23,7 @@ impl School {
         s
     }
 
-    pub fn grade(&self, grade: u32) -> Option<Vec<String>> {
-        self.grades.get(&grade).map(|v| v.to_vec())
+    pub fn grade(&self, grade: u32) -> Vec<String> {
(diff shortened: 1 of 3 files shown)


petertseng (approved):
my approval does not depend on the answer to this question, but I nevertheless put it forth:

what were the clippy lints? I remember the new_without_default when I was counting the clippy lints of the example solutions; were there any others?

The removal of the `Option` around `Vec<String>` is interesting. I suspect the reasoning we agree on here is: Option was useful if there is a difference between "requested valid grade and there are no students" vs "requested invalid grade". Since the tests do not distinguish, there is no need for the return type to distinguish either.

coriolinus:
The most recent lint failure report is [here](https://github.com/exercism/rust/runs/1659291286?check_suite_focus=true). I do agree with your (and presumably clippy's) reasoning about returning `Option<Vec<_>>`: there's no point unless we care about the semantic difference between those cases, and we don't.

PR #1048, opened 2021-01-05
Add HashMap import to invalid test cases for Macros

tjade273:
This solution currently passes all of the tests but is incorrect:

```rust
#[macro_export]
macro_rules! hashmap {
    ( $($key:expr => $val:expr),* $(,)?) => {
        {
            use crate::HashMap;

            let mut map = HashMap::new();
            $(
                map.insert($key, $val);
            )*
            map
        }
    };
}
```

The macro accepts `hashmap!(,)` but because the solution also uses `crate::HashMap`, the invalid cases all fail to compile due to the import even if they would otherwise succeed. This PR prevents this sort of failure by providing the same imports to the negative cases as to the positive cases.

### DIFF
--- a/exercises/macros/tests/invalid/Cargo.toml
+++ b/exercises/macros/tests/invalid/Cargo.toml
@@ -45,3 +45,7 @@ path = "two-arrows.rs"
 [[bin]]
 name = "leading-comma-rs"
 path = "leading-comma.rs"
+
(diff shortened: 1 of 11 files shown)


coriolinus (approved):
Very good catch! Thanks for the contribution.

PR #1049, opened 2021-01-06
Add new missing-argument.rs test for Macros

tjade273:
Another test to catch an invalid student solution. Thanks to @bobahop for catching it.

```rust
#[macro_export]
macro_rules! hashmap {
    ( $($key:expr => $val:expr$(,)?),* ) => {{
        let mut hash_map = ::std::collections::HashMap::new();
        $(
            hash_map.insert($key, $val);
        )*
        hash_map
    }};
}
```


coriolinus (approved):
Great catch @bobahop and thanks for the additional test @tjade273!

PR #1052, opened 2021-01-11
grep: Fix the typo in Blok's poem

k33nice:


coriolinus (approved):
Good catch! Thanks for contributing this fix.

PR #1053, opened 2021-01-17
clock: correct simple typo in test name

kunaltyagi:


petertseng (approved):
that is indeed a typo that should be fixed. Thank you.

The PR and commit are going to be renamed to "clock: correct simple typo in test name", because the exercise should go in the commit message and the commit message's tense should be present imperative. We hope you will adhere to these in your next contribution, until then, thank you for this one!

PR #1084, opened 2021-01-28
rg master --files-with-matches .github/ | xargs sed -i 's/master/main/g'

coriolinus:


coriolinus:
Closing in favor of #1083 which was pushed first.

PR #1099, opened 2021-01-29
finish vec-macro

### LINKED ISSUE #1063: [Rust] Implement new Concept Exercise: vec-macro
This issue describes how to implement the vec concept exercise for the Rust track.

## Getting started

**Please please please read the docs before starting.** Posting PRs without reading these docs will be a lot more frustrating for you during the review cycle, and exhaust Exercism's maintainers' time. So, before diving into the implementation, please read up on the following documents:

- [The features of v3](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md).
- [Rationale for v3](https://github.com/exercism/v3/blob/master/docs/rationale-for-v3.md).
- [What are concept exercise and how they are structured?](https://github.com/exercism/v3/blob/master/docs/features-of-v3.md)

Please also watch the following video:

- [The Anatomy of a Concept Exercise](https://www.youtube.com/watch?v=gkbBqd7hPrA).

## Goal

The vec collection is the most common data container in Rust, and as such has many miscellaneous functions. This should serve as an introduction to common usage of Vecs, and is not intended to cover its usage exhaustively. It should mention that Vec is not an array, but it is not required to treat arrays in detail and focus on Vec should be favored over going down a rabbithole.

## Learning objectives

Student should be able to...
- use `vec!`
- call basic methods on Vec
- recognize that Vec is not an array
- recognize the most obvious cases where Vec may reallocate
- recognize a few good use-cases for Vec

Student may be able to...
- use a Vec as the basis of an Iterator
- understand that Vec<T> is generic over T
- understand details of Vec's relationship with arrays
- take slices out of a Vec

## Out of scope

- complex data structures that may be formed with Vecs
- other std::collections, even VecDeque!
- algorithms for use with Vec, but do use them if they are both basic enough and it seems appropriate for teaching the data structure

## Concepts

- vec
- arrays?
- slices?

## Prerequisites

This exercise should depend only on a basic understanding of functions, types, and variables. They need to know enough to be able to shove something into a Vec, after all.

There are some uncertainties in placement, because this is both a suitable lesson to serve as a stepping stone to these concepts but also it is useful to (re)visit this after they have been understood.
- Is it before or after methods?
- Is it before or after Iterator?
- Is it before or after generics?

The lesson may change accordingly. Including variances is not required: assume the most minimal form. Additional content for possible usage may be included in the `.meta` folder of the exercise.

### Hints

- std::vec: https://doc.rust-lang.org/std/vec/index.html

### After

- Rust by Example (stdlib types): https://doc.rust-lang.org/rust-by-example/std/vec.html#vectors
- The Rust Programming Language (Creating a Vector): https://doc.rust-lang.org/book/ch08-01-vectors.html#creating-a-new-vector

## Representer

This should require no special changes to the representer.

## Analyzer

This should require no special changes to the analyzer.

## Implementing

See our [implementation guide](https://github.com/exercism/v3/blob/master/languages/rust/howto/implementing-a-concept-exercise.md#how-to-implement-a-rust-concept-exercise) for guidance. In addition you may want to make use of the shell script while in the `concept` directory, with the arg vec, e.g. `sh boil_exercise.sh vec`

## Help

If you have any questions while implementing the exercise, please post the questions as comments in this issue.

efx:
Starting work to close #1063.
I intend to create follow up issues for the `vec-collect` and the others mentioned [here](https://github.com/exercism/v3/pull/2918).

### DIFF
--- a/config.json
+++ b/config.json
@@ -101,7 +101,7 @@
           "intro-fn",
           "numbers"
         ],
-        "status": "beta"
+        "status": "active"
       }
     ],
     "practice": [
--- a/config.json
+++ b/config.json
@@ -101,7 +101,7 @@
           "intro-fn",
           "numbers"
         ],
-        "status": "wip"
+        "status": "beta"
       }
     ],
     "practice": [
--- a/exercises/concept/vec-macro/.docs/instructions.md
+++ b/exercises/concept/vec-macro/.docs/instructions.md
@@ -7,7 +7,7 @@ Create a function that creates a buffer of `count` zeroes.
 let my_buffer = create_buffer(5);
 ```
 
-## 1. List the first five elements of the Fibonacci sequence
+## 2. List the first five elements of the Fibonacci sequence
 
 Create a function that returns the first five numbers of the Fibonacci sequence. 
 Its first five elements are `1, 1, 2, 3, 5`
--- a/exercises/concept/vec-macro/.docs/instructions.md
+++ b/exercises/concept/vec-macro/.docs/instructions.md
@@ -1,6 +1,6 @@
 You are going to initialize empty buffers and list the first five numbers, or elements, of the Fibonacci sequence.
 
-The Fibonacci sequence is a set of numbers where the next element is the sum of the prior two. The first two elements are 1 and 2.
+The Fibonacci sequence is a set of numbers where the next element is the sum of the prior two. We start the sequence at one. So the first two elements are 1 and 1.
 
 ## 1. Create a buffer of `count` zeroes.
 
--- a/exercises/concept/vec-macro/.docs/instructions.md
+++ b/exercises/concept/vec-macro/.docs/instructions.md
@@ -7,7 +7,7 @@ Create a function that creates a buffer of `count` zeroes.
 let my_buffer = create_buffer(5);
 ```
 
-## 1. List the first five numbers of the Fibonacci sequence
+## 1. List the first five elements of the Fibonacci sequence
 
 Create a function that returns the first five numbers of the Fibonacci sequence. 
 Its first five elements are `1, 1, 2, 3, 5`
--- a/exercises/concept/vec-macro/.docs/instructions.md
+++ b/exercises/concept/vec-macro/.docs/instructions.md
@@ -1,0 +1,16 @@
+You are going to initialize empty buffers and list the first five numbers of the Fibonacci sequence.
+
+## 1. Create a buffer of `count` zeroes.
+
+Create a function that creates a buffer of `count` zeroes.
+```rust
+let my_buffer = create_buffer(5);
+```
+
+## 1. List the first five numbers of the Fibonacci sequence
+
+Create a function that returns the first five numbers of the Fibonacci sequence. 
+Its first five elements are `1, 1, 2, 3, 5`
+```rust
+let first_five = fibonacci();
+```
--- a/exercises/concept/vec-macro/.docs/instructions.md
+++ b/exercises/concept/vec-macro/.docs/instructions.md
@@ -1,10 +1,13 @@
-You are going to initialize empty buffers and list the first five numbers of the Fibonacci sequence.
+You are going to initialize empty buffers and list the first five numbers, or elements, of the Fibonacci sequence.
+
+The Fibonacci sequence is a set of numbers where the next element is the sum of the prior two.
 
 ## 1. Create a buffer of `count` zeroes.
 
 Create a function that creates a buffer of `count` zeroes.
 ```rust
 let my_buffer = create_buffer(5);
+// [0, 0, 0, 0, 0]
 ```
 
 ## 2. List the first five elements of the Fibonacci sequence
@@ -13,4 +16,5 @@ Create a function that returns the first five numbers of the Fibonacci sequence.
 Its first five elements are `1, 1, 2, 3, 5`
 ```rust
 let first_five = fibonacci();
+// [1, 1, 2, 3, 5]
 ```
--- a/exercises/concept/vec-macro/.docs/instructions.md
+++ b/exercises/concept/vec-macro/.docs/instructions.md
@@ -1,6 +1,6 @@
 You are going to initialize empty buffers and list the first five numbers, or elements, of the Fibonacci sequence.
 
-The Fibonacci sequence is a set of numbers where the next element is the sum of the prior two.
+The Fibonacci sequence is a set of numbers where the next element is the sum of the prior two. The first two elements are 1 and 2.
 
 ## 1. Create a buffer of `count` zeroes.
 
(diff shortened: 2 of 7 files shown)


efx:
@ErikSchierboom  - thanks for taking a look! I'm curious if you could verify I am following the v3 format correctly.

exercises/concept/vec-macro/src/lib.rs
```diff
@@ -1,9 +1,5 @@
-// This stub file contains items which aren't used yet; feel free to remove this module attribute
-// to enable stricter warnings.
```
coriolinus, line 2: IMO we should leave this as documentation for the student.

efx: Ah, I see what you mean. Yes, I'll leave it itn.

exercises/concept/vec-macro/.docs/instructions.md
```diff
@@ -0,0 +1,16 @@
+You are going to initialize empty buffers and list the first five numbers of the Fibonacci sequence.
+
+## 1. Create a buffer of `count` zeroes.
+
+Create a function that creates a buffer of `count` zeroes.
+```rust
+let my_buffer = create_buffer(5);
+```
+
+## 1. List the first five elements of the Fibonacci sequence
```
ErikSchierboom, line 10:
```suggestion
## 2. List the first five elements of the Fibonacci sequence
```

exercises/concept/vec-macro/.docs/instructions.md
```diff
@@ -0,0 +1,16 @@
...
+
+Create a function that returns the first five numbers of the Fibonacci sequence. 
+Its first five elements are `1, 1, 2, 3, 5`
+```rust
+let first_five = fibonacci();
```
ErikSchierboom, line 15: The example should help the student understand what the expected return value looks like. Is there perhaps a way to show the contents of the vector? See [this F# example](https://github.com/exercism/fsharp/blob/main/exercises/concept/bird-watcher/.docs/instructions.md#1-check-what-the-counts-were-last-week).

exercises/concept/vec-macro/.docs/instructions.md
```diff
@@ -0,0 +1,16 @@
+You are going to initialize empty buffers and list the first five numbers of the Fibonacci sequence.
+
+## 1. Create a buffer of `count` zeroes.
+
+Create a function that creates a buffer of `count` zeroes.
+```rust
+let my_buffer = create_buffer(5);
```
ErikSchierboom, line 7: The example should help the student understand what the expected return value looks like. Is there perhaps a way to show the contents of the vector? See [this F# example](https://github.com/exercism/fsharp/blob/main/exercises/concept/bird-watcher/.docs/instructions.md#1-check-what-the-counts-were-last-week).

exercises/concept/vec-macro/.docs/instructions.md
```diff
@@ -0,0 +1,16 @@
+You are going to initialize empty buffers and list the first five numbers of the Fibonacci sequence.
```
ErikSchierboom, line 1: Maybe add a little introduction to the Fibonacci sequence here? It is unlikely that people haven't heard from it, but a reminder would be useful I feel.

ErikSchierboom (approved):
Nice!

exercises/concept/vec-macro/.docs/instructions.md
efx: Agreed. I had it in originally but removed it in my zealous pruning!

exercises/concept/vec-macro/.docs/instructions.md
```diff
@@ -0,0 +1,20 @@
+You are going to initialize empty buffers and list the first five numbers, or elements, of the Fibonacci sequence.
+
+The Fibonacci sequence is a set of numbers where the next element is the sum of the prior two.
```
ErikSchierboom, line 3: You probably also need to define the first two numbers for this definition to be complete :)

efx:
I'm thinking aloud here. 
Decomposing the concept exercise `vec` has made implementation go quite well.
To pass linting, I needed to create a distinct concept called `concepts/vec-macro`.
In Rust, generally speaking vectors and macros are two different concepts. 
So I am mildly concerned that by creating `concept/` entries for distinct pairs of concepts I may be overcomplicating it. That is we may have to do this for an incredibly large number of concept pairs. 

This is early on though. So I will wait and  see if this actually becomes a problem. I feel this is me struggling with the what is an Exercism Concept and what are Rust categories and concepts needed for fluency problem.

coriolinus:
> To pass linting, I needed to create a distinct concept called concepts/vec-macro.
> In Rust, generally speaking vectors and macros are two different concepts.

I think the `vec-macro` concept covers, specifically, the `vec!` macro, not any other aspects of macros in general.

That said, vectors are fundamental and have several different major use cases. I'm hoping that this is true for very few other concepts. 

> I will wait and see if this actually becomes a problem.

👍

ErikSchierboom:
Waiting to see how this turns out seems like a wise approach.

config.json
```diff
@@ -101,7 +101,7 @@
           "intro-fn",
           "numbers"
         ],
-        "status": "wip"
+        "status": "beta"
```
coriolinus, line 104: We could probably yolo this into active status now, but that's not essential.

coriolinus (commented):
I would like to see the stub cleaned up such that these lines aren't necessary: https://github.com/exercism/rust/blob/f4361fa7dac8239368c7b169c1db0ed7cbf5018b/exercises/concept/vec-macro/src/lib.rs#L1-L3

The CI was failing universally without them, but I don't think we should have any beta or active exercises which contain them.

Other than that, this LGTM. I'll approve once the stub works without the annotation.

config.json
efx: I went for it.

efx:
It looks like `ensure-stubs-compile.sh` adds the deny directive. I will explore exempting any active or beta concept exercise from that addition.
If I'm reading it correctly that means some jq filtering of `config.json` here:
https://github.com/exercism/rust/blob/673eb06815571e747f71c810bcda9edc972a8df1/_test/ensure-stubs-compile.sh#L43

coriolinus:
No; the goal is to make CI pass, not to break CI.

Example: to fix

```
error: unused variable: `count`
  --> src/lib.rs:10:22
   |
10 | pub fn create_buffer(count: usize) -> Vec<u8> {
   |                      ^^^^^ help: if this is intentional, prefix it with an underscore: `_count`
   |
```

Either change the stub to

```rust
pub fn create_buffer(count: usize) -> Vec<u8> {
  unimplemented!("create a zeroized buffer of {} bytes", count)
}
```

Or, in (rare) cases where the parameter doesn't implement `Debug` or `Display`, fix it by doing

```rust
pub fn create_buffer(_count: usize) -> Vec<u8> {
  // note that the parameter _count has a leading underscore; this is to silence unused variable warnings.
  // your implementation should use this parameter, so you will need to remove that underscore.
  unimplemented!()
}
```

coriolinus (approved):

PR #1102, opened 2021-02-01
restore our initial concept enumeration

efx:
Restores https://github.com/exercism/v3/pull/222/files to this
repository.

I wanted a copy of this enumeration handly while polishing our initial
set of exercises.

### DIFF
--- a/reference/README.md
+++ b/reference/README.md
@@ -1,0 +1,190 @@
+# Rust reference
+
+## Concepts
+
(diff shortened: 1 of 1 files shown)


coriolinus (approved):
Seems fine as a WIP reference, but I feel that over time as we finish up concepts and other v3 requirements, we should tend to remove documents from the reference instead of adding them.

efx:
@coriolinus I totally agree. I feel that the `concepts/` entries will subsume this document. I'll create an issue to clean this up once that happens.

PR #1106, opened 2021-02-01
write an introduction to destructuring

### LINKED ISSUE #1075: [rust][concept cleanup] rewrite destructuring introduction
https://github.com/exercism/v3/blob/master/languages/rust/concepts/destructuring/introduction.md

efx:
Close #1075.

### DIFF
--- a/concepts/destructuring/introduction.md
+++ b/concepts/destructuring/introduction.md
@@ -1,75 +1,9 @@
-TODO: the content below is copied from the exercise introduction and probably needs rewriting to a proper concept introduction
+Destructuring is a way to break apart values in Rust. It is often used when extracting values from tupes or structs.
 
-Tuples are a lightweight way to group a fixed set of arbitrary types of data together. A tuple doesn't have
-a particular name; naming a data structure turns it into a `struct`. A tuple's fields don't have names;
-they are accessed by means of destructuring or by position.
-
-## Syntax
-
-### Creation
-
-Tuples are always created with a tuple expression:
-
+For example, the below snippet extracts two values from a tuple:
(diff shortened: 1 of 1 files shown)


efx:
I'm using these resources do to this:
- official spec: https://github.com/exercism/docs/blob/main/anatomy/tracks/concepts.md#file-introductionmd 
- A concept from an exemplary track: https://github.com/exercism/fsharp/blob/main/concepts/strings/introduction.md

concepts/destructuring/introduction.md
```diff
@@ -1,75 +1,9 @@
...
-let two_element = (123, "elements can be of differing types");
-```
-
-Tuples can have an arbitrary number of elements.
-
-### Access by destructuring
-
-It is possible to access the elements of a tuple by destruturing. This just means assigning variable
-names to the individual elements of the tuple, consuming it.
+Destructuring is a way to break apart values in Rust. It is often used when extracting values from tupes or structs.
```
coriolinus, line 1:
```suggestion
Destructuring is a way to break apart compound values in Rust. It is often used to extract fields from tuples or structs.
```

Two wording suggestions, one typo fix.

coriolinus (approved):
In your opinion, is the current `about.md` ok, or should we move any of the content we're removing here into that document?

efx:
@coriolinus  great question. My understanding is that `about.md` should be longer with more meat. The `introduction.md` should be quite minimal. But I'm unsure of the correct balance at the moment.
Once I do a few more and get feedback I can revisit the appropriate partitioning. So I'll keep it as is for now.

PR #1113, opened 2021-02-04
fix inaccurate path name

efx:
I noticed this the other day looking through documentation.
I'm not on Windows but if anyone tries it this helps.


coriolinus (approved):
Good spot!

PR #1116, opened 2021-02-05
[v3] Update Practice Exercises to latest spec

ErikSchierboom:
This week we've defined what the file structure looks like for Practice Exercises. Please skim through [this docs PR](https://github.com/exercism/docs/pull/36#issue-566737871) that gives an overview of the changes, and see [the spec](https://github.com/exercism/docs/blob/main/anatomy/tracks/practice-exercises.md) for more information.

This PR applies changes to the track to conform to the latest spec. Each change is made as a separate commit, to help with reviewing. When merging this PR, I would suggest to _not_ squash this to preserve the individual commits.

As having this PR merged is essential to prepare this track for Exercism v3, we'll automatically merge this PR one week after it was opened (if it hasn't been merged already).

Here are the changes made:

## Add practices field to Practice Exercises

While working on adding prerequisites to the Practice Exercises, we found that we really need two keys for Practice Exercises:

- `practices`: Concepts that the Practice Exercise is actively helping a student to practice. These show up in the UI as "Practice this Concept in: TwoFer, Leap, etc`
- `prerequisites`: The actual Concepts that should have been learnt by a student in order for them be able to do that Exercise.

There are lots of advantages to having the two keys instead of using one key for two purposes, not least that having them split out gives the maintainers a lot more control over the order a student should be exposed to exercises (via the prerequisites) and also where they encourage them to try concepts (via practices), which in reality are quite different things.

We have updated [the docs](https://github.com/exercism/docs/blob/master/anatomy/tracks/practice-exercises.md#practices) with more details and explanation on this.

## Remove README.md files

v3 Practice Exercises won't have a (generated) README.md file anymore. This file will be replaced by `.docs/instructions.md` and optionally a `.docs/introduction.md` files. These files are copies of the information defined for their corresponding problem-specifications exercise (if any).

Exercism v3 will build the README.md file that is sent via the CLI from these files contents, along with some other files.
See [this document](https://github.com/exercism/docs/blob/main/anatomy/tracks/presentation.md) for more information.

This PR removes the `README.md` files.

## Create instruction from prob-specs descriptions

The `.docs/instructions.md` file is a copy of the `instructions.md` file of the problem-specifications exercise, or if that does not exists, its `description.md` file.

The new [configlet](https://github.com/exercism/configlet) version will add support for doing this syncing automatically.

This PR adds the `.docs/instructions.md` file from the problem-specifications exercise's `description.md` file.

## Move .meta/hints.md to .docs/instructions.append.md

[...]

PR #1117, opened 2021-02-05
address new configlet lint errors

efx:
`configlet` was updated along with the specification for concept exercises. So we needed to clean up those lints.
#1116 can rebase once this is merged.


coriolinus (approved):
This looks fine in principle. IMO this should target #1116, not `main`: that way, we can merge this into that branch, and then verify that that PR passes CI, rather than ever merging something for which CI does not pass.

efx:
Closed in favor of that approach!

PR #1118, opened 2021-02-05
fix the concept exercises metadata

efx:
- `blurb` is for `concepts/` documentation not concept exercises
- each needs a name. I didn't feel creative so just capitalized the
existing slug

I followed the specification [here](https://github.com/exercism/docs/blob/main/anatomy/tracks/config-json.md).
I expect CI to fail because the needed configlet lint updates are not in `main`. They are in #1116 as i pushed to the fixes from #1090 to there.

## off topic
I also plan to document exploring our JSON in this repo using nu.
`open config.json | get exercises.concept` provides a pretty table.


config.json
```diff
@@ -19,6 +19,7 @@
       {
         "slug": "entry-api",
         "uuid": "e7d8603e-97a9-11ea-9753-0242ac110002",
+				"name": "Entry API",
```
coriolinus, line 22: These render weirdly because it looks like you used tabs instead of spaces, or something. Please update to an appropriate number of spaces.

coriolinus (approved):
Overall looks good. One request, below.

Please don't merge this until #1116 is merged; that way, you can merge master into this PR and see that CI comes up green before merging this.

PR #1122, opened 2021-02-08
rename match pattern matching concept

### LINKED ISSUE #1121: [concept cleanup] uses dashes for concept categorization instead of periods
Concept names should respect the requirements in [config.json](https://github.com/exercism/docs/blob/main/anatomy/tracks/config-json.md#concepts). From what I read, this means they must be:
- lowercase kebab cased string
- uniquely named 

Our concept names mostly follow this pattern but some use a `.` to specify `category.concept`. 
We'll want to normalize ours to be `category-concept`.

The [JavaScript](https://github.com/exercism/javascript/tree/main/concepts) and [F#](https://github.com/exercism/fsharp/tree/main/concepts) use kebab case for their naming too.

efx:
Close #1121.

I do not believe this breaks anything now but would like to avoid future
surprises. I'm also deferring to other track naming standards.


coriolinus (approved):
Like the idea of standardizing on kebab case. Agree that this probably shouldn't break anything. If we merge and it does, then we can still fix it, as v3 isn't out yet, and it'll provide information we can use to add a CI check which will prevent future breakage. That suggests to me that we should just merge it and see what happens.

PR #1123, opened 2021-02-08
remove early iteration of concepts

efx:
We introduced concepts.csv for CI validation in https://github.com/exercism/v3/pull/856.
It is my understanding that `concepts/` encodes our canonical list.

So concepts.csv is dead code and should be removed from this repository.


coriolinus (changes requested):
I like the motivation--you're right, this file is dead code--but as-is, this is premature. `concepts.csv` contains a lot of data on concepts that is not yet encoded into the canonical source.

The canonical source of concepts is the `concepts` list within `config.json`. Enumerating those items, we see only a few entries:

```sh
$ jq '.concepts[] | .slug' -r config.json | sort
Mon 08 Feb 2021 10:59:00 PM CET
destructuring
entry-api
enums-basic
floating-point-numbers
hashmaps
integers
intro-fn
intro-types
option
string-use
structs
```

Even just a very mechanical porting of these concepts into that list would be better than simply removing the file. We can weed out redundancies later. Probably start with

```sh
$ cat reference/concepts.csv | cut -d, -f1 | sort | tr . -
```

Then, for each item in that list, create a uuid, create a title-case name, create an empty blurb, then insert into `.concepts[]` in `config.json`. Then, create a bunch of appropriately-named TODO files. This should be very automatable. Once we've done that, I'd be ok removing this file.

coriolinus (approved):
After #1127 I no longer object to this PR.

PR #1134, opened 2021-02-11
fix CI difficulty check for practice exercises

### LINKED ISSUE #1131: [ci] fix or deprecate verify-difficulty CI step
Similar to #1129.

I noticed that the verify difficulty script errors but the CI passes:
https://github.com/exercism/rust/runs/1872578899?check_suite_focus=true#step:6:4

You can see this locally by running:
```sh
./_test/verify-exercise-difficulties.sh
jq: error (at rust/config.json:1913): Cannot index array with string "deprecated"
jq: error (at rust/config.json:1913): Cannot index array with string "deprecated"
```
I could make it fail correctly and then fix the script, but since it involves v2 I would rather deprecate it.

What would you prefer @coriolinus?

efx:
Close #1131.

I verified it manually for a practice exercise.
This script does not operate on concept exercises at the moment. I'm
unsure of how to combine two top-level properties in jq and treat them
as one list. So improvements welcome!

### DIFF
--- a/_test/verify-exercise-difficulties.sh
+++ b/_test/verify-exercise-difficulties.sh
@@ -10,7 +10,8 @@ es=0
 # ensure every exercise has a difficulty
 no_difficulty=$(
     jq --raw-output '
-        .exercises.practice[] |
+        .exercises | 
+        .concept[], .practice[] |
         select((.status != "deprecated") and (has("difficulty") | not)) |
         .slug
     ' $config
@@ -24,7 +25,8 @@ fi
 # ensure that all difficulties are one of 1, 4, 7, 10
 invalid_difficulty=$(
     jq --raw-output '
-        .exercises.practice[] |
+        .exercises |
+        .concept[], .practice[] |
         select(
             (.status != "deprecated") and
             has("difficulty") and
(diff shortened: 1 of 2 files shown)


_test/verify-exercise-difficulties.sh
```diff
@@ -8,8 +10,8 @@ es=0
 # ensure every exercise has a difficulty
 no_difficulty=$(
     jq --raw-output '
-        .exercises[] |
-        select((.deprecated | not) and (has("difficulty") | not)) |
+        .exercises.practice[] |
```
coriolinus, line 13:
```suggestion
        .exercises | 
        .concept[], .practice[] |
```

_test/verify-exercise-difficulties.sh
```diff
@@ -22,9 +24,9 @@ fi
 # ensure that all difficulties are one of 1, 4, 7, 10
 invalid_difficulty=$(
     jq --raw-output '
-        .exercises[] |
+        .exercises.practice[] |
```
coriolinus, line 27:
```suggestion
        .exercises |
        .concept[], .practice[] |
```

coriolinus (commented):
Assuming concept and practice objects are structured the same, these suggestions should apply the checks to all of them.

coriolinus (approved):

PR #1136, opened 2021-02-12
[v3] Update Markdown files to latest spec

ErikSchierboom:
We've defined a [specification for Markdown files](https://github.com/exercism/docs/blob/main/contributing/standards/markdown.md) to be applied Exercism-wide. This standard includes, amongst others, the following two rules:

- All files must start start with a level-1 heading (`# Some heading text`)
- No heading may decend a level greater than one below the previous (e.g. `## may only be followed by ###, not ####`)

This PR applies the above two rules to the Markdown documents in this repo. 

## Tracking

https://github.com/exercism/v3-launch/issues/17


exercises/practice/nth-prime/.docs/instructions.append.md
```diff
@@ -1 +1,3 @@
+# Instructions append
```
coriolinus, line 1: There are several instances like this; I suspect we don't _actually_ want titles like "Instructions append". As I understand it, those'll be stripped anyway though, so I don't want to block this PR on it.

coriolinus (approved):
Just a quick skim, but seems plausible.

exercises/practice/nth-prime/.docs/instructions.append.md
ErikSchierboom: They will indeed be stripped.

ErikSchierboom:
Thanks for the review!

PR #1137, opened 2021-02-12
address clippy +beta failures

efx:
- also ignoring `clippy.log`

### DIFF
--- a/exercises/practice/spiral-matrix/tests/spiral-matrix.rs
+++ b/exercises/practice/spiral-matrix/tests/spiral-matrix.rs
@@ -22,7 +22,12 @@ fn size_two_spiral() {
 #[test]
 #[ignore]
 fn size_three_spiral() {
-    let expected: Vec<Vec<u32>> = vec![vec![1, 2, 3], vec![8, 9, 4], vec![7, 6, 5]];
+    #[rustfmt::skip]
+    let expected: Vec<Vec<u32>> = vec![
+        vec![1, 2, 3], 
+        vec![8, 9, 4], 
+        vec![7, 6, 5],
+    ];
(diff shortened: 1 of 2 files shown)


exercises/practice/spiral-matrix/tests/spiral-matrix.rs
```diff
@@ -9,46 +9,42 @@ fn empty_spiral() {
...
 }
 
 #[test]
 #[ignore]
 fn size_three_spiral() {
-    let mut expected: Vec<Vec<u32>> = Vec::new();
-    expected.push(vec![1, 2, 3]);
-    expected.push(vec![8, 9, 4]);
-    expected.push(vec![7, 6, 5]);
+    let expected: Vec<Vec<u32>> = vec![vec![1, 2, 3], vec![8, 9, 4], vec![7, 6, 5]];
```
coriolinus, line 25:
```suggestion
    #[rustfmt::skip]
    let expected: Vec<Vec<u32>> = vec![
        vec![1, 2, 3], 
        vec![8, 9, 4], 
        vec![7, 6, 5],
    ];
```

This is the kind of place where I think manual formatting is appropriate: it helps users visualize the spiral in a way that rustfmt doesn't realize would be helpful to preserve. Not essential, but nice.

coriolinus (approved):
Thanks for fixing the clippy warnings! There's more cleanup you could do optionally, but this already achieves its goal of fixing clippy.

exercises/practice/spiral-matrix/tests/spiral-matrix.rs
efx: Oh, cool! TIL `#[rustfmt::skip]`. 
I totally agree. The grid representation seems more readable to me.

PR #1142, opened 2021-02-16
enable another markdown lint rule

efx:
We ran into this in a PR review so I think it is reasonable to enable
it.

As an aside, I researched if markdown CLI let's you enforce reference style links. By default no.
But the underlying library does let you specify custom rules.
See another project's [implementation](https://github.com/webhintio/hint/commit/c412f9aa7ba99eb7ef6c20b7c496d629530f3ecf).

In our case, I find that too much trouble and maintenance for now.

### DIFF
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -1,4 +1,5 @@
 # Example markdownlint YAML configuration with all properties set to their default value
+# See here: https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#rules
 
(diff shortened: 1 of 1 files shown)


coriolinus (approved):
Applying MD011 seems like a good idea to me.

coriolinus:
However, I would oppose a style change requiring only reference links. When working with the un-rendered markdown text, inline links are easier to work with if you ever want to actually follow them.

PR #1143, opened 2021-02-16
Add test for Poker `test_dupicate_hands_always_tie()`

cwhakes:
In the Poker practice exercise, returning the correct number of hands isn't tested. If three duplicate hands are given, only one needs to be returned, currently.

In `test_three_of_a_kind_cascade_ranks()`, the possibility of using multiple decks is allowed, so this corner condition is technically possible. However, the primary motivation is to make sure that the right number of hands are returned in the `Vec`.

Alternatively `hs_from()` function could use a `BinaryHeap` instead of a `HashSet`.

### DIFF
--- a/exercises/practice/poker/tests/poker.rs
+++ b/exercises/practice/poker/tests/poker.rs
@@ -28,6 +28,16 @@ fn test_single_hand_always_wins() {
 
 #[test]
 #[ignore]
+fn test_duplicate_hands_always_tie() {
+    let input = &["3S 4S 5D 6H JH", "3S 4S 5D 6H JH", "3S 4S 5D 6H JH"];
+    assert_eq!(
+        &winning_hands(input).expect("This test should produce Some value",),
+        input
+    )
+}
+
+#[test]
+#[ignore]
 fn test_highest_card_of_all_hands_wins() {
     test(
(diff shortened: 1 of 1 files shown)


exercises/practice/poker/tests/poker.rs
```diff
@@ -26,6 +26,16 @@ fn test_single_hand_always_wins() {
     test(&["4S 5S 7H 8D JC"], &["4S 5S 7H 8D JC"])
 }
 
+#[test]
+#[ignore]
+fn test_duplicate_hands_always_tie() {
+    let input = &["3S 4S 5D 6H JH", "3S 4S 5D 6H JH", "3S 4S 5D 6H JH"];
+    assert_eq!(
+        &winning_hands(input).expect("This test should produce Some value",),
```
coriolinus, line 34:
```suggestion
        &winning_hands(input).expect("This test should produce Some value"),
```

The extra comma is legal there but looks weird to me.

coriolinus (approved):
I like it! Very simple to understand and tests a useful case.

exercises/practice/poker/tests/poker.rs
coriolinus: Extracted this to https://github.com/exercism/rust/issues/1144.

PR #1149, opened 2021-02-17
🤖 Add file patterns to config.json

ErikSchierboom:
**⚠ This PR requires you to make a simple change before merging. ⚠**

---

To save maintainers from having to manually specify the `files` key in their exercises' `.meta/config.json` files, we are providing support for track-level patterns. See [this PR](https://github.com/exercism/docs/pull/58) for details.

This PR adds (**purposefully wrong**) file patterns to the `config.json` file. It is up to you, the track maintainers, to change these patterns to their correct value.

You can use the following placeholders:
- `%{kebab_slug}`: the `kebab-case` exercise slug (e.g. `bit-manipulation`)
- `%{snake_slug}`: the `snake_case` exercise slug (e.g. `bit_manipulation`)
- `%{camel_slug}`: the `camelCase` exercise slug (e.g. `bitManipulation`)
- `%{pascal_slug}`: the `PascalCase` exercise slug (e.g. `BitManipulation`)

We will soon update `configlet` to enable it to automatically populate the `.meta/config.json` file's `files` property, at which point we will then batch-PR updates to all tracks that have merged this PR.

## Tracking
https://github.com/exercism/v3-launch/issues/19

### DIFF
--- a/config.json
+++ b/config.json
@@ -16,16 +16,16 @@
   },
   "files": {
     "solution": [
-      "path/to/%{kebab_slug}.ext"
+      "src/lib.rs"
     ],
     "test": [
-      "path/to/%{kebab_slug}.ext"
+      "tests/%{snake_slug}.rs"
     ],
     "example": [
-      "path/to/%{kebab_slug}.ext"
+      ".meta/example.rs"
     ],
     "exemplar": [
-      "path/to/%{kebab_slug}.ext"
+      ".meta/exemplar.rs"
     ]
   },
   "exercises": {
--- a/config.json
+++ b/config.json
@@ -1400,418 +1400,418 @@
   },
   "concepts": [
     {
-      "uuid": "f7be1969-380e-4348-ac79-9b2834a526fd",
-      "slug": "&str",
-      "name": "&str",
-      "blurb": ""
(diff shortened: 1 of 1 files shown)


efx:
thanks @ErikSchierboom! I added what I think are the correct paths for us.

config.json
```diff
@@ -1386,418 +1400,418 @@
   },
   "concepts": [
     {
-        "uuid": "f7be1969-380e-4348-ac79-9b2834a526fd",
```
efx, line 1389: nit: I guess we had a different level of indentation? Not a blocker but I would prefer we not have extraneous changes like this.

efx (approved):

config.json
```diff
@@ -14,6 +14,20 @@
     "indent_style": "space",
     "indent_size": 4
   },
+  "files": {
+    "solution": [
+      "src/lib.rs"
```
coriolinus, line 19: It would be nice if we could capture `src/**/*.rs`, but that depends on configlet correctly supporting both recursive and nonrecursive globbing. As mentioned [here](https://github.com/exercism/docs/pull/58#issuecomment-780789579). 

If that isn't currently supported, then these patterns look right to me.

coriolinus (approved):

PR #1150, opened 2021-02-17
Clippy beta fixes

efx:
I batch applied the camel case abbreviation changes.
I was not sure about the simple-linked-list lint so exempted it.
- [exemption applied here](https://github.com/exercism/rust/compare/clippy-beta-fixes?expand=1#diff-82b44a277ea9de5d666b7f807136f46e26f46d7ed16a71c4ca3925a342a1e0bcR61)
- [clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#from_over_into)
- [example failure](https://github.com/exercism/rust/runs/1919516153?check_suite_focus=true#step:7:268)

Since the clippy lint is new, it [fails](https://github.com/exercism/rust/runs/1920253485?check_suite_focus=true#step:7:10) for stable clippy runs. If I've taken the correct approach i'm happy to find a way to exempt this in the script executing clippy. 

Otherwise I'm open to refactoring the simple-linked-list exercise per the exercise or another solution to make all CI steps pass.

### DIFF
--- a/exercises/practice/simple-linked-list/example.rs
+++ b/exercises/practice/simple-linked-list/example.rs
@@ -69,7 +69,7 @@ impl<T> FromIterator<T> for SimpleLinkedList<T> {
 
 impl<T> From<SimpleLinkedList<T>> for Vec<T> {
     fn from(mut linked_list: SimpleLinkedList<T>) -> Vec<T> {
-        let mut vec: Vec<T> = vec![];
+        let mut vec: Vec<T> = Vec::with_capacity(linked_list.len());
         while let Some(data) = linked_list.pop() {
             vec.push(data);
         }
--- a/exercises/practice/simple-linked-list/example.rs
+++ b/exercises/practice/simple-linked-list/example.rs
@@ -67,10 +67,10 @@ impl<T> FromIterator<T> for SimpleLinkedList<T> {
     }
 }
 
-impl<T> Into<Vec<T>> for SimpleLinkedList<T> {
-    fn into(mut self) -> Vec<T> {
-        let mut vec = Vec::new();
-        while let Some(data) = self.pop() {
+impl<T> From<SimpleLinkedList<T>> for Vec<T> {
+    fn from(mut linked_list: SimpleLinkedList<T>) -> Vec<T> {
+        let mut vec: Vec<T> = vec![];
+        while let Some(data) = linked_list.pop() {
             vec.push(data);
         }
         vec.reverse();
--- a/exercises/practice/simple-linked-list/src/lib.rs
+++ b/exercises/practice/simple-linked-list/src/lib.rs
@@ -58,7 +58,6 @@ impl<T> FromIterator<T> for SimpleLinkedList<T> {
 // of IntoIterator is that implementing that interface is fairly complicated, and
 // demands more of the student than we expect at this point in the track.
 
-#[allow(clippy::from_over_into)]
 impl<T> Into<Vec<T>> for SimpleLinkedList<T> {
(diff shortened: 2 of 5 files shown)


coriolinus:
I strongly suspect that if you rewrite `simple-linked-list/.meta/example.rs` and `simple-linked-list/src/lib.rs` as

```rust
impl<T> From<SimpleLinkedList<T>> for Vec<T> { ... }
```

Then the lint will be resolved without needing to change any tests or add any exemptions. 

The reason we implemented `Into` instead of `From` is that originally, the orphan rule prevented implementing a foreign trait (`From`) on a foreign type (`Vec<T>`) unconditionally. A few language releases ago--I don't remember which, precisely, but I think it was within the last year or so--they relaxed the orphan rule to permit a foreign trait parametrized by a local type. [The pattern works now](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=5fb799937fd44c3eb702a86680b7e615).

Now that it works, no reason to keep the workaround.

exercises/practice/simple-linked-list/example.rs
```diff
@@ -67,10 +67,10 @@ impl<T> FromIterator<T> for SimpleLinkedList<T> {
     }
 }
 
-impl<T> Into<Vec<T>> for SimpleLinkedList<T> {
-    fn into(mut self) -> Vec<T> {
-        let mut vec = Vec::new();
-        while let Some(data) = self.pop() {
+impl<T> From<SimpleLinkedList<T>> for Vec<T> {
+    fn from(mut linked_list: SimpleLinkedList<T>) -> Vec<T> {
+        let mut vec: Vec<T> = vec![];
```
coriolinus, line 72:
```suggestion
        let mut vec: Vec<T> = Vec::with_capacity(linked_list.len());
```

Tiny, unimportant optimization.

exercises/practice/simple-linked-list/src/lib.rs
```diff
@@ -58,8 +58,8 @@ impl<T> FromIterator<T> for SimpleLinkedList<T> {
 // of IntoIterator is that implementing that interface is fairly complicated, and
 // demands more of the student than we expect at this point in the track.
 
-impl<T> Into<Vec<T>> for SimpleLinkedList<T> {
-    fn into(self) -> Vec<T> {
+impl<T> From<SimpleLinkedList<T>> for Vec<T> {
+    fn from(mut _linked_list: SimpleLinkedList<T>) -> Vec<T> {
```
coriolinus, line 62: Kind of surprised rustc/CI didn't issue a warning about an unnecessary `mut` for this line.

coriolinus (approved):
Thanks for fixing these! A few more notes, but nothing worth forcing a delay.

exercises/practice/simple-linked-list/src/lib.rs
efx: Ah, so it did when I kept it as `linked_list`. So I'm guessing no lints when you have the variable ignored from linting?

exercises/practice/simple-linked-list/example.rs
efx: huh, applying the suggestion failed using "Commit suggestion" and "Add Suggestion to Batch". CLI to the rescue!

exercises/practice/simple-linked-list/src/lib.rs
coriolinus: That's really interesting, and IMO potentially a bug. Not a huge priority for anyone though, I imagine.

PR #1151, opened 2021-02-17
forth: Remove `Option<_>` type requirement

PaulDance:
Hi!

As discussed recently with @coriolinus, the `None` case can never really occur and was not tested either. Just an empty vector actually covers the desired semantics.

Cheers,
Paul.

### DIFF
--- a/exercises/practice/poker/.meta/example.rs
+++ b/exercises/practice/poker/.meta/example.rs
@@ -7,21 +7,24 @@ use counter::Counter;
 ///
 /// Note the type signature: this function should return _the same_ reference to
 /// the winning hand(s) as were passed in, not reconstructed strings which happen to be equal.
-pub fn winning_hands<'a>(hands: &[&'a str]) -> Option<Vec<&'a str>> {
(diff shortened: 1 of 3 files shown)


exercises/practice/poker/.meta/example.rs
```diff
@@ -7,21 +7,24 @@ use counter::Counter;
...
+        .last()
+        .map(|last| {
+            hands
+                .iter()
+                .rev()
+                .take_while(|&item| item.partial_cmp(last) == Some(Ordering::Equal))
+                .map(|hand| hand.source)
+                .collect()
+        })
+        .unwrap()
```
coriolinus, line 27: We shouldn't be demonstrating `unwrap`; it can cause runtime panics.

```suggestion
        .unwrap_or_default()
```

PaulDance: Yeah, sorry, struggling for no reason here... It should be fixed now.

coriolinus (approved):
Thanks!

PR #1152, opened 2021-02-18
Add concepts to practice exercises

coriolinus:
Feel free to exclude ce48033 from your reviews; that's a purely mechanical autoformat. It's first in the commit list to reduce noise in the rest of them.

### DIFF
--- a/config.json
+++ b/config.json
@@ -134,7 +134,9 @@
         "slug": "hello-world",
         "name": "Hello World",
         "uuid": "13ec1ebe-d71b-436f-ab12-25305e814171",
-        "practices": [],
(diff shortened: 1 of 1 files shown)


efx:
for my own learning: how did you format config.json?
I'll note adding prettier or an equivalent in our CI issues.

coriolinus:
TBH it's just VS Code's default thing, I think; I don't recall explicitly installing or requesting autoformatting other than a generic "format on save" option. I think any JSON formatter would have given similar results.

coriolinus:
Stale, and I don't have the capacity right now to move it forward.

PR #1153, opened 2021-02-18
Concepts audit

efx:
I manually audited the v3 reference enumeration in [reference/README.md](https://github.com/exercism/rust/blob/226a11d0d0028ae7edecf8da6f7d3c651b1bd61f/reference/README.md): 

Where we had an existing folder in `concepts/` I annotated the about.md with the notes from README.

## omitted and unclear whether to add `concept` files
- many immutable or one mutable reference

## other

- `generic lifetimes ('a)` should be covered by `lifetimes/`
- removed duplicate `hashmap` since we have `hashmaps` and seem to prefer plural naming

Do we want to keep the "Concept exercise mapping" [section](https://github.com/exercism/rust/blob/226a11d0d0028ae7edecf8da6f7d3c651b1bd61f/reference/README.md#concept-exercise-mapping)?

I would like to remove obsolete documentation to make it more clear for what our canonical list is and how people can contribute.


concepts/derive/about.md
```diff
@@ -1,3 +1,4 @@
 # About
 
 TODO
+- `PartialEq`, `PartialOrd`, `Debug`, `Display`, `Default`
```
coriolinus, line 4:
```suggestion
- `PartialEq`, `Eq`, `PartialOrd`, `Ord`, `Debug`, `Display`, `Default`
```

Probably most of the time we derive partialeq/partialord, it's appropriate to derive the full equality/ordering as well. Obviously they get their own conceptin "total-ordering", but the takeaway here should just be that it's possible to derive these traits.

coriolinus (commented):
Looks good so far. Are you planning on more work, or would you prefer that I approve as-is?

---

> Unclear whether it deserves a concept: mutability

I think mutability deserves a concept. 

> should we delete "concept exercise mapping"?

Probably not wholesale; it's stale, but we can still take advantage of previous thought on what the concepts should be and how they should relate to each other. 

> I would like to remove obsolete documentation to make it more clear for what our canonical list is and how people can contribute.

Yeah, that's a good idea; I just don't like the idea of deleting previous thought without retaining what's current. I think that determining the current canonical list of concepts is a very good idea; the list in `config.json` is a bit messy right now. That will help new contributors, but it will also help me in #1152.

efx:
> Looks good so far. Are you planning on more work, or would you prefer that I approve as-is?

I'll push up mutability but wasn't planning on more work on this branch. So as-is would be great, thank you!

> Probably not wholesale; it's stale, but we can still take advantage of previous thought on what the concepts should be and how they should relate to each other.

Sounds great.

> I think that determining the current canonical list of concepts is a very good idea;

I'm wondering if we should generate a markdown file programmatically from `config.json` to better visualize inconsistencies? 

As to docs, how do you feel about `docs/archived` and I can move all the `reference/` things there and problem_ordering.md?

coriolinus (approved):

coriolinus:
> wondering if we should generate a markdown file programmatically from config.json

We could write a script to do that, but I'm reluctant to commit such a file to the repo, because it's just begging to get stale. Either we have no CI, in which case it will get stale astonishingly fast, or we have CI enforcement that people regenerate it each time they change `config.json`, in which case it's the old `README.md` pain again, or we give CI permissions to add commits to `main`, which is a security can of worms that I'd rather not open.

> how do you feel about `docs/archived`

I like it; it doesn't purport to be authoritative the way `reference/` does, but we can retain all the stale docs until we've mined them for all they're good for.

efx:
Ah, good points. Sounds like more trouble than its worth compared to our other CI needs too.

> we can retain all the stale docs until we've mined them for all they're good for.

Agreed!
Seems like `docs/archived` is the better way to accomplish this now.

efx:
I mispoke as well: we do have mutability concept:
https://github.com/exercism/rust/blob/main/concepts/mutability/about.md

Issue #1141, opened 2021-02-16
[CI] [proposal] low-efforts to standardize scripting


efx:
I have been enjoying learning more shell by patching things here and there in our `bin/` and `_test/` files.
As an outsider, I can see little things that tell the story of these scripts being written over time by many folks.

I propose we standardize a small set of guidelines.
I appreciate Google's [scripting guidelines](https://google.github.io/styleguide/shellguide.html) but feel we probably don't need that degree of specificity now. 

# wishlist
- underscores or dashes for file names
- whether the filename should include `.sh` or not
- executable bit set
- favor a single shell interpreter
    - I'd vote BASH simply because it gives you a tad more expressiveness and seems more plausible. Actual usage here is split between `sh` and `bash`. And we can standardize `set -e`.

coriolinus:
My votes:

> underscores or dashes

Underscores. Makes things simpler when upgrading a script into a proper language.

> whether the filename should include the extension.

Yes. You can do a fair bit with arbitrary files with the `file` command and a shebang parser, but neither of those are as visible as a filename extension.

> executable bit set

Yes for scripts that are intended to be called directly; no for scripts which simply define utility functions. This is analogous to the `lib.rs` vs `main.rs` distinction. 

> favor a single interpreter

Yes, and agree that bash is a reasonable default. We just have to be careful about what features we use: OSX still distributes bash v3.x by default, despite v5.x being current; this means that the scripts can't depend on certain features like maps.

Also, all shebangs should be of the form `#!/usr/bin/env bash`, not `/bin/bash`: letting `env` pick the version means that OSX users who have manually installed a modern version aren't stuck using the system version.

One more preference: we should enforce use of `shellcheck` via CI.

efx:
> we should enforce use of shellcheck via CI.

💯 !
I'll use #1110 to capture adding shellcheck.

PR #1156, opened 2021-02-19
start notes on maintaining process

efx:
Capture some of our discussion in #1141 for a broader audience.
I see this as freer form: we can adjust it as we make decisions.

I'm not wed to the name or location.
I liked the bash track's approach with `POLICIES.md`.
But I feel formal policies are inappropriate at this time.

### DIFF
--- a/docs/maintaining.md
+++ b/docs/maintaining.md
@@ -1,0 +1,22 @@
+# Maintaining Notes
+
+This document captures informal policies, tips, and topics useful to maintaining the Rust track.
+
+## Internal Tooling
+
+We have a number of scripts for CI tests.
+They live in `bin/` and `_test`.
+
+## Internal Tooling Style Guide
+
+This is non-exhaustive.
+
+- Prefer snakecase for file names
+    ```sh
+    hello_world.sh
+    ```
+- Prefer BASH for scripting
+- Adopt a Unix philosophy for tooling
+  - try to accomplish tooling needs using `jq` and pipelines of minimal utilities
+- Scripts should use `#!/usr/bin/env bash` as their shebang
+  - This increases portability on NixOS and macOS because contributors may not have bash installed in `/bin/bash`.
--- a/docs/maintaining.md
+++ b/docs/maintaining.md
@@ -17,6 +17,8 @@ This is non-exhaustive.
     ```
 - Prefer BASH for scripting
 - Adopt a Unix philosophy for tooling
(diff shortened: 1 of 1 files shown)


docs/maintaining.md
```diff
@@ -0,0 +1,22 @@
...
+
+This is non-exhaustive.
+
+- Prefer snakecase for file names
+    ```sh
+    hello_world.sh
+    ```
+- Prefer BASH for scripting
+- Adopt a Unix philosophy for tooling
+  - try to accomplish tooling needs using `jq` and pipelines of minimal utilities
```
coriolinus, line 20:
```suggestion
  - prefer using tools that do one thing well
  - prefer using tools that are ubiquitous: `jq` or `sed` instead of `prettier` or `sd`
  - write scripts to do one thing well
```

Expanding on the philosophy a bit.

docs/maintaining.md
```diff
@@ -0,0 +1,22 @@
...
+- Scripts should use `#!/usr/bin/env bash` as their shebang
+  - This increases portability on NixOS and macOS because contributors may not have bash installed in `/bin/bash`.
```
coriolinus, line 22:
```suggestion
  - This increases portability on NixOS and macOS because contributors' preferred bash may not be installed in `/bin/bash`.
```

Precision of wording change. For example, in my macbook, I have these results:

```sh
$ /bin/bash --version
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin20)
Copyright (C) 2007 Free Software Foundation, Inc.
$ /usr/bin/env bash --version
GNU bash, version 5.0.18(1)-release (x86_64-apple-darwin19.5.0)
Copyright (C) 2019 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
```

OSX installs an ancient bash in `/bin/bash` for what I believe are legal reasons with the licenses of newer versions. However, it's tedious and error-prone to replace it in place with a more modern version. Much simpler to install a new bash alongside it in some other location and configure the system to prefer the new version. `/usr/bin/env` knows how to respect standard system configurations, so it picks the newer version.

coriolinus (approved):
Some nits, but I think this doc is a good idea.

docs/maintaining.md
efx: Ah, thanks for explaining this history! Systems fun.

PR #1158, opened 2021-02-19
link to TRPL example

efx:
I was not planning on any further work on this concept now, but I did research the topic a bit today and found this description helpful.
Instead of trusting my inconsistent note-taking, `concepts/` it is!

### DIFF
--- a/concepts/newtype-pattern/links.json
+++ b/concepts/newtype-pattern/links.json
@@ -1,1 +1,7 @@
-[]
+[
+
+  {
+    "url": "https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#using-the-newtype-pattern-to-implement-external-traits-on-external-types",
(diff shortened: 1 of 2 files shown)


efx:
Failure is a new configlet lint:
> Not a valid tag: "platform/wasm":
> exercism/rust/config.json

I'll fix that in this PR.

ee7:
I saw:

> Not a valid tag: "platform/wasm":
> exercism/rust/config.json

But `platform/wasm` might be a good tag. I've asked about whether we should add it.

coriolinus (approved):

efx:
Shall we:
1. merge PR's that only fail because of this tag?
1. remove the tag for now and create an issue to follow up?

I prefer #2 so I have less to keep track of LOL.

coriolinus:
Option 2 works for me.

PR #1161, opened 2021-02-19
fix count-ignores.sh

### LINKED ISSUE #1160: [CI] count-ignores.sh does not correctly traverse
While working on #1159, I noticed that count-ignores.sh passes even when you remove ignores from various tests.

This happens because the script expects all exercises are in the v2 structure:
https://github.com/exercism/rust/blob/69b077e2d2c7361ae0433badd0fa1395c67c7c9d/_test/count-ignores.sh#L6

So the fix will simply be recursing further using `*/*` as we do elsewhere.
I am doing this in a separate PR because there will be associated fixes in various `.rs` files.

efx:
Close #1160.

### DIFF
--- a/_test/count-ignores.sh
+++ b/_test/count-ignores.sh
@@ -3,7 +3,7 @@
 repo=$(cd "$(dirname "$0")/.." && pwd)
 exitcode=0
 
-for e in $repo/exercises/*; do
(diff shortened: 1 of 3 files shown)


coriolinus (approved):
Good catch!

PR #1163, opened 2021-02-22
update maintaining.md

efx:
Some other qualifications I thought of recently.

### DIFF
--- a/docs/maintaining.md
+++ b/docs/maintaining.md
@@ -29,3 +29,5 @@ This is non-exhaustive.
 You can run CI tools locally. 
 Scripts expect GNU versions of tooling, so you may see unexpected results on macOS.
 [Here](https://github.com/exercism/rust/issues/1138) is one example.
+Windows users can also run tooling locally using [WSL](https://docs.microsoft.com/en-us/windows/wsl/).
+We recommend WSL 2 with the distribution of your choice.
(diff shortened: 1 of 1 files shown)


docs/maintaining.md
```diff
@@ -20,5 +20,12 @@ This is non-exhaustive.
...
   - prefer using tools that are ubiquitous: `jq` or `sed` instead of `prettier` or `sd`
   - write scripts to do one thing well
+  - prefer GNU versions of `sed` and other utilities
 - Scripts should use `#!/usr/bin/env bash` as their shebang
   - This increases portability on NixOS and macOS because contributors' preferred bash may not be installed in `/bin/bash`.
+
+
+## Running CI Locally
+You can run CI tools locally. 
+Scripts expect GNU versions of tooling, so you may see unexpected results on macOS.
```
coriolinus, line 30: If we're helping out the macOS users with a note here, we should probably also tell Windows users that the recommended maintenance environment on that platform is via WSL 2 and the distribution of their choice.

coriolinus (approved):
Looks good. Suggested a change, but it should not be construed as mandatory.

docs/maintaining.md
efx: Oh, good point. I'll add that note in a few!

PR #1165, opened 2021-02-22
finish string-vs-str, for real 

### LINKED ISSUE #1146: finish csv-builder concept exercise (was string-vs-str)
Final polish for the string-vs-str exercise. Continues #1057.

- [x] create an entry for the exercise and mark as active in config.json
- [x] complete the about.md
- [x] double check Erik's feedback from https://github.com/exercism/v3/pull/3058#pullrequestreview-573361130
- [x] hints should be itemized per the sections of the concept exercise (e.g. making them more meaningful for each step)

efx:
Start actually finishing string-vs-str
Close #1146.


coriolinus:
I can't believe I didn't notice the "and stir" comment previously. It's funny, and it's how I pronounce it internally (though usually, within the privacy of my mind, it gets further abbreviated to "n-stir"), but I think we should at least mention that the "proper" pronunciation is "ref string". The `&` operator takes a reference.

concepts/string-vs-str/about.md
```diff
@@ -9,6 +9,63 @@ This document discusses how they overlap and offers some basic advice.
 `&str` is pronounced "and stir"
 `&str` is immutable.
 
+`String` is a potentially-mutable utf8-encoded representation of a sequence of Unicode codepoints. (In Java or C#, this type is called `StringBuilder`.)
+
+`&str` is a read only view of a well-formed utf8 sequence and because it's a reference it is `Copy` and can be shared.
```
coriolinus, line 14:
```suggestion
`&str` is a read only view of a well-formed utf8 sequence. Because it's a reference, it is `Copy` and can be shared.
```

Previous form isn't wrong, but IMO this form reads a bit better.

concepts/string-vs-str/about.md
```diff
@@ -6,9 +6,66 @@ This document discusses how they overlap and offers some basic advice.
 `String` is an owned, vector of `u8` bytes.
 `String` is mutable.
 `&str`, is a borrowed reference to a String slice.
-`&str` is pronounced "and stir"
 `&str` is immutable.
 
+The proper pronunciation of `&str` is "ref string".
+`String` is a potentially-mutable UTF-8-encoded representation of a sequence of Unicode code points. (In Java or C#, this type is called `StringBuilder`.)
```
efx, line 12: Do we need this line referencing Java and C#? I'm personally unfamiliar with either of those languages so this does not teach me anything.
It would seem helpful for someone familiar with those languages.
But we generally avoid comparisons in other `about.md` so I'm included to remove this line.

coriolinus: Yeah, let's avoid referencing the languages. Turns out [this was already mentioned](https://github.com/exercism/v3/pull/3058/files#r561957164), but got lost at some point among the merges.

coriolinus (approved):
Thanks for doing this work! Last I'd seen, this was still in draft status. Still, I don't have anything I really want to see changed; I think this exercise may finally be complete!

efx:
> Thanks for doing this work! 

Sure thing! And thank you for the helpful reviews @coriolinus.
I'll try to keep the number of outstanding PRs lower since I easily distract myself.

PR #1168, opened 2021-02-23
enable a few more markdown lints

efx:
I took the liberty of enabling a few more markdown lints.
I could not find how to lint one sentence per line, which would be nice
to have.
Otherwise I:
- disabled inline HTML since I don't see the reason we should allow
this. Funny enough, the lint caught an example using `Option<T>` without
code ticks.
- no bare URLs
- consistent code fence style

### DIFF
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -26,6 +26,14 @@ MD013: false
 # MD014/commands-show-output - Dollar signs used before commands without showing output
 MD014: false
 
+# MD033/no-inline-html - Inline HTML
+MD033:
+  # Allowed elements
+  allowed_elements: []
+
+# MD034/no-bare-urls - Bare URL used
+MD034: true
+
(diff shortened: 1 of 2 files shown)


efx:
https://github.com/DavidAnson/markdownlint/issues/66 is the support request for one sentence per line.
It requires a custom rule which I am avoiding :)

.markdownlint.yaml
```diff
@@ -34,3 +42,7 @@ MD040: true
 # MD042/no-empty-links - No empty links
 MD042: true
 
+# MD048/code-fence-style - Code fence style
+MD048:
+  # Code fence syle
+  style: "consistent"
```
coriolinus, line 48:
```suggestion
  style: "backtick"
```

As the rule states, a consistent style aids readability. I have never seen tilde-style fences in this repo, and I don't see any particular reason we would want to introduce them.

coriolinus (approved):
Good idea.

Issue #1111, opened 2021-02-03
[proposal] create a minimal CONTRIBUTING.md


efx:
As a new contributor I would like to understand and respect the maintenance standards for this repository.
I propose we create a `CONTRIBUTING.md` to briefly document what to expect contributing to this repository.

I and @coriolinus are strapped for time so I am on the fence about doing this. I am on the fence because I do not want to drain from our primary goal of high-quality concept exercises for the v3 launch.

So, the first step would be to decide whether we add one.
Then, let's define simple content.

I would appreciate even if this was something like:
> The current active maintainers are [insert our handles or link to the code] 
> Their time is limited and so a proper introduction to maintaining the Rust track is out of scope!
>  But we would appreciate contributions using these basic guidelines.

As an aside, I feel like contributing `concept/` entries is quite well-defined and self-contained. So it may lend to us documenting clearly how to do it. But it is technical writing. Writing about Rust versus writing Rust. In fact, I would argue most of the v3 concept exercise development has been this kind of meta. Encoding teaching Rust in Exercism's format!

coriolinus:
IMO this is a good idea, but fairly low priority. If we do it, it should be more thorough than the 3-line suggestion. I don't think we should complain about limited time there.

efx:
Topics I would like to cover:
- branches policy in the repository
- how we use issues and labels
    - Defering to the outcome of exercism/docs/issues/27
- merging philosophy and process
- how to create a new concept exercise
    - I am finding creating a draft PR with only `exemplar` a great strategy. I'll plan on drafting a simple how to guide at some point.

efx:
Exercism's community process for contributing and maintaining lives [here](https://github.com/exercism/docs/tree/main/contributing).

efx:
#1177 introduced basic contributing guidelines.

#1156 introduced notes for maintainers.

There are other ideas I had here that may be nice to do, but I will close this issue since the core topic has been addressed.

PR #1174, opened 2021-02-24
add a trailing whitespace lint

efx:
I was curious about what [util/exercise](https://github.com/exercism/rust/tree/main/util/exercise) and noticed the `Cargo.toml` has trailing whitespace.
So I took the opportunity to add a lint in CI too.
[This](https://github.com/exercism/rust/runs/1973793069#step:8:4) is an example failure.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -36,6 +36,9 @@ jobs:
       - name: Check exercises for authors
         run: ./_test/check-exercises-for-authors.sh
 
+      - name: Ensure relevant files do not have trailing whitespace
+        run: ./bin/lint_trailing_spaces.sh
+
   configlet:
     name: configlet lint
     runs-on: ubuntu-latest
@@ -74,7 +77,7 @@ jobs:
 
   # stolen from https://raw.githubusercontent.com/exercism/github-actions/main/.github/workflows/shellcheck.yml
   shellcheck:
-    name: shellcheck internal tooling
+    name: shellcheck internal tooling lint
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
--- a/bin/lint_trailing_spaces.sh
+++ b/bin/lint_trailing_spaces.sh
@@ -1,0 +1,14 @@
+#!/usr/bin/env bash
+
+# Report any .toml files that have trailing white space
+# so user can fix them
+set -eo pipefail
+
+files=$(find . -name '*.toml' -exec grep -l '[[:space:]]$' {} \;)
+
+if [ -n "$files" ]; then
+	echo "These files have trailing whitespace:"
+	echo "$files"
+	echo "Our conventions disallow this so please remove the trailing whitespace."
+	exit 1
+fi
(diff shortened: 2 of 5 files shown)


.github/workflows/tests.yml
```diff
@@ -74,7 +77,7 @@ jobs:
 
   # stolen from https://raw.githubusercontent.com/exercism/github-actions/main/.github/workflows/shellcheck.yml
   shellcheck:
-    name: shellcheck internal tooling
+    name: shellcheck internal tooling lint
```
efx, line 80: Renamed this for consistency. I should also rename our `markdown_lint.sh` to `lint_markdown.sh`. It seems most of our internal tooling follows a `category-subcategory` naming scheme.

bin/lint_trailing_spaces.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+# Report any .toml files that have trailing white space
+# so user can fix them
+set -eo pipefail
+
+files=$(find . -name '*.toml' -exec grep -l '[[:space:]]$' {} \;)
```
coriolinus, line 7:
```suggestion
files=$(find . -iname '*.toml' -o -iname '*.rs' -o -iname '*.md' -exec grep -l '[[:space:]]$' {} \;)
```

`-name` -> `-iname` just makes the match case-insensitive, in case someone ever creates `IMPORTANT.MD` or something.

I think that we want to lint rust and markdown files as well.

util/exercise/src/main.rs
```diff
@@ -102,7 +102,7 @@ fn process_matches(matches: &ArgMatches<'_>) -> Result<()> {
         }
 
         ("", None) => {
-            println!("No subcommand was used.\nUse init_exercise --help to learn about the possible subcommands.");
+            println!("No subcommand was used.\nUse exercise --help to learn about the possible subcommands.");
```
coriolinus, line 105: This change makes sense, but seems pretty unrelated to the rest of this PR.

coriolinus (approved):
Good idea, makes sense.

bin/lint_trailing_spaces.sh
efx: Awesome! So our current markdown lint will cover `.md` files so perhaps we only do `.rs` and `.sh`?

coriolinus: Sure, works for me.

util/exercise/src/main.rs
efx: I agree. How do you feel about documenting something like a Unix philosophy for our pull requests?

Keep it small.
One logical change or something.
I can take a stab for #1111.

coriolinus: Yeah, "one logical change" is the best expression of what I want to see in a PR. It's not necessarily small--there have been good, useful PRs which added hundreds of lines of code and took months to resolve (looking at you, #653)--but life is easier when PRs are focused.

PR #1182, opened 2021-02-26
🤖 Update tests.toml files to latest spec

ErikSchierboom:
Track maintainers found that they wanted to add comments to the tests.toml file to e.g. indicate _why_ a test was not included.
Unfortunately, running configlet sync would re-generate the entire file so any manually added comments were lost.

In this PR we're updating the format of tests.toml files to support adding comments.
We do this by creating a separate table for each test case which has `description` and `include` fields.
Tracks are then free to add additional fields, like a `comment` field, but also any other fields they feel might be useful to them.

configlet has _not_ yet been updated to support this new format, but we hope to do this soon. Sorry for the inconvenience.

For more information, see this discussion: https://github.com/exercism/configlet/issues/186

## Implementation

The PR expects the tests.toml files to be in their original format:

```toml
# <description>
"<uuid>" = <include>
```

This is transformed to:

```toml
[<uuid>]
description = "<description>"
include = <include>
```

## Example

```toml
# one factor has multiples within limit
"361e4e50-c89b-4f60-95ef-5bc5c595490a" = true
```

becomes

```toml
[361e4e50-c89b-4f60-95ef-5bc5c595490a]
description = "one factor has multiples within limit"
include = true
```

## Existing comments

As some tracks have manually added comments to tests, we try to detect them by assuming they are either:

- Added as a line comment _before_ the description (we'll ignore empty lines)
- Added as an inline comment _after_ boolean include value

For any such manually detected comments, we'll add a `comment` field.

## Tracking

https://github.com/exercism/v3-launch/issues/22


efx (approved):
Cursory read looks good to me.

exercises/practice/all-your-base/.meta/tests.toml
```diff
@@ -1,69 +1,80 @@
...
+description = "invalid positive digit"
+include = true
+comments = [
+  "input base is negative",
+  "this test may not be in the suite but has not been intentionally rejected",
+  "\"54a23be5-d99e-41cc-88e0-a650ffe5fcc2\" = false",
+  "negative digit",
+  "this test may not be in the suite but has not been intentionally rejected",
+  "\"9eccf60c-dcc9-407b-95d8-c37b8be56bb6\" = false"
+]
```
coriolinus, line 72: Looks like the transform script got a bit confused here. IIRC the comment "this test may not be in the suite but has not been intentionally rejected" was itself generated by script, so we should be on the lookout for more instances like that, and potentially fix them by script if there are many.

ErikSchierboom: Oh my, terribly sorry! This is due to the fact that rust has commented test cases:

```toml
# negative digit
# this test may not be in the suite but has not been intentionally rejected
# "9eccf60c-dcc9-407b-95d8-c37b8be56bb6" = false
```

I'll see if I can update my script.

ErikSchierboom: With `configlet` regenerates the file, these comments will be lost. There are a couple of options:

1. Don't regenerate until you've made a decision what to do with these commented test cases
2. Remove the commented test cases. Functionally, this will cause no change in the behavior of configlet, which will treat them as having an unknown status regardless of whether they are in a comment.
3. Create entries for the commented test cases that have `include = false`, and then add a comment reflecting the status being unknown (`comment = "this test may not be in the suite but has not been intentionally rejected"`).

coriolinus: As a practical matter, I think we're going to end up with option 1. The intent for those comments was to simply annotate those test cases for which no decision has yet been made, to make it easier to identify what the decisions are when someone has time to actually review and decide. If configlet's tooling for discovering new test cases is now implemented, option 2 may be better.

ErikSchierboom: @coriolinus configlet doesn't yet support the new format. I'll let you know when that is the case. For now, it may be best to make this a draft PR.

ErikSchierboom: @coriolinus We've updated configlet to support custom properties. The latest version will also not output `include = true`, as that is the default and thus redundant. 

This PR is not in the best of shapes though, as my script failed to deal with fully commented test cases like:

```toml
# count one word
# this test may not be in the suite but has not been intentionally rejected
# "61559d5f-2cad-48fb-af53-d3973a9ee9ef" = false
```

Seeing how comments are ignored by configlet and will be stripped when re-running configlet, how would you like the above case to be handled?

coriolinus: I think the best resolution as far as this PR is just to strip fully-commented test cases entirely. As long as configlet has the capability to provide a list of cases which don't appear in the suite, those comments are now redundant.

ErikSchierboom: > As long as configlet has the capability to provide a list of cases which don't appear in the suite, those comments are now redundant.

If you run `./bin/configlet sync --check` (we'll be changing the default to be check soon-ish BTW), it outputs something like:

```
[warn] reverse-string: missing 6 test cases
       - an empty string (c3b7d806-dced-49ee-8543-933fd1719b1c)
       - a word (01ebf55b-bebb-414e-9dec-06f7bb0bee3c)
       - a capitalized word (0f7c07e4-efd1-4aaa-a07a-90b49ce0b746)
       - a sentence with punctuation (71854b9c-f200-4469-9f5c-1e8e5eff5614)
       - a palindrome (1f8ed2f3-56f3-459b-8f3e-6d8d654a1f6c)
       - an even-sized word (b9e7dec1-c6df-40bd-9fa3-cd7ded010c4c)
[warn] wordy: missing 2 test cases
       - multiple multiplication (312d908c-f68f-42c9-aa75-961623cc033f)
       - multiple division (3c854f97-9311-46e8-b574-92b60d17d394)
...
```

ErikSchierboom: I've updated the PR to the latest spec.

exercises/practice/acronym/.meta/tests.toml
```diff
@@ -1,28 +1,30 @@
...
 
-# apostrophes
-"5118b4b1-4572-434c-8d57-5b762e57973e" = true
+[5118b4b1-4572-434c-8d57-5b762e57973e]
+description = "apostrophes"
 
-# underscore emphasis
-"adc12eab-ec2d-414f-b48c-66a4fc06cdef" = true
+[adc12eab-ec2d-414f-b48c-66a4fc06cdef]
+description = "underscore emphasis"
```
coriolinus, line 30: Nit: configlet should ensure that its output files have a trailing newline.

exercises/practice/allergies/.meta/tests.toml
```diff
@@ -1,189 +1,28 @@
...
-# this test may not be in the suite but has not been intentionally rejected
-# "07ced27b-1da5-4c2e-8ae2-cb2791437546" = false
+[d2d71fd8-63d5-40f9-a627-fbdaf88caeab]
+description = "allergic to everything"
 
-# allergic to eggs and something else
-# this test may not be in the suite but has not been intentionally rejected
-# "5035b954-b6fa-4b9b-a487-dae69d8c5f96" = false
+[cc789c19-2b5e-4c67-b146-625dc8cfa34e]
+description = "allergic to everything"
```
coriolinus, line 12: Looks like the descriptions got a bit mangled here.

exercises/practice/acronym/.meta/tests.toml
ErikSchierboom: Should be fixed now.

exercises/practice/allergies/.meta/tests.toml
ErikSchierboom: This is due to the test cases being nested. There is an issue for this: https://github.com/exercism/configlet/issues/202 We're working on a fix: https://github.com/exercism/configlet/pull/301

coriolinus: Ok, cool.

coriolinus (approved):
I see that `configlet lint` is failing due to new lints being implemented. Appropriate fixes for the linter are out of scope of this PR. Willing to authorize merging despite that CI failure.

Otherwise I'm now happy with this PR. Configlet is getting much more capable!

ErikSchierboom:
> I see that configlet lint is failing due to new lints being implemented. Appropriate fixes for the linter are out of scope of this PR. Willing to authorize merging despite that CI failure.

Thanks!

> Otherwise I'm now happy with this PR. Configlet is getting much more capable!

Yes! And more is to follow!

Issue #1124, opened 2021-02-08
[concept cleanup] write numbers introduction.md


efx:
This task entails:
-  write the introduction.md for the numbers concept: https://github.com/exercism/rust/tree/main/concepts/numbers

That file should be short and sweet. A paragraph at most. 

See vec-macro's [introduction](https://github.com/exercism/rust/blob/main/concepts/vec-macro/introduction.md) for an example.
See [the specification documentation](https://github.com/exercism/docs/blob/main/anatomy/tracks/concepts.md#file-introductionmd) for all the details.

PR #1183, opened 2021-02-26
write numbers intro

sm1215:
👋 Hey there, first time contributing to OS. Tried to do my homework and read up on contributing guides but if I've missed anything, just let me know! 

- I took a crack at writing an intro to numbers for #1124. I saw there was an emphasis on staying concise so I tried to keep that in mind, however it feels like there's a lot going on with numbers! I broke the content down into two sections and based the information off of the Rust book. 
- Not sure if this is what we'd be looking for, but it's a starting point at least and I can edit from here if needed. Feedback welcome, thanks :]

### DIFF
--- a/concepts/numbers/introduction.md
+++ b/concepts/numbers/introduction.md
@@ -1,17 +1,17 @@
 # Introduction
 
-In Rust, there are two main types of numbers availble for use: Integers and Floating-Point Types. 
+In Rust, there are two main types of numbers availble for use: Integers and Floating-Point Types.
 
 ## Integers
-An Integer is a whole number without a fractional component. They come in a number of variants ranging from 8-bit to 128-bit. The more bits you use, the larger the number can be, however, this increases the memory used. Each variant can be Signed or Unsigned. A Signed number allows for negative values whereas Unsigned does not. You can declare an Integer like so: 
-```
+An Integer is a whole number without a fractional component. They come in a number of variants ranging from 8-bit to 128-bit. The more bits you use, the larger the number can be, however, this increases the memory used. Each variant can be Signed or Unsigned. A Signed number allows for negative values whereas Unsigned does not. You can declare an Integer like so:
+```rust
 // Declaring and assigning an integer with a size of 8 bits
 let myInt: u8 = 5;
 ```
 
 ## Floating-Point Types
 A floating-point number includes decimal points and there are two types available in Rust. An `f32` and `f64` - both referring to a float of 32 bits and a float of 64 bits. `f64` is the default type in Rust as working with either size can be done at about the same speed when using a modern CPU. You can declare a Floating-Point Type like so:
-```
+```rust
 // Declaring and assigning a float with a size of 32 bits
 let myDecimal: f32 = 1.5;
 ```
--- a/concepts/numbers/introduction.md
+++ b/concepts/numbers/introduction.md
@@ -1,10 +1,11 @@
 # Introduction
 
-In Rust, there are are many specific types of numbers but they generally fall under one of two catorgies: integers or floating-point numbers.
+In Rust, there are are many specific types of numbers but they generally fall under one of two catorgies: integers or floating-point numbers. 
 
 Initializing an integer:
 ```rust
 let x = 5; // rust defaults to an i32
+```
 
 Initializing a large positive integer:
 ```rust
--- a/concepts/numbers/introduction.md
+++ b/concepts/numbers/introduction.md
@@ -1,3 +1,17 @@
 # Introduction
 
-TODO
+In Rust, there are two main types of numbers availble for use: Integers and Floating-Point Types. 
+
+## Integers
+An Integer is a whole number without a fractional component. They come in a number of variants ranging from 8-bit to 128-bit. The more bits you use, the larger the number can be, however, this increases the memory used. Each variant can be Signed or Unsigned. A Signed number allows for negative values whereas Unsigned does not. You can declare an Integer like so: 
+```
+// Declaring and assigning an integer with a size of 8 bits
+let myInt: u8 = 5;
+```
+
+## Floating-Point Types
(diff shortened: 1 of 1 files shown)


efx:
@sm1215 welcome! 
Thank you for taking this on. This is a great start.
Can you shoot for one paragraph ( 2 - 3 sentences ) for the introduction.md?

I like surveying unsigned and signed numeric types. Those sections can go in the [`about.md`.](https://github.com/exercism/rust/blob/main/concepts/numbers/about.md#numbers)

The tl;dr I'm learning is that the concept's introduction.md should be 1 paragraph or so, like a teaser. Students see `introduction.md` before doing the exercise.

The about.md has the meat. The about is what a student will see *after* completing the exercise so they can refer back to that as well.

coriolinus:
Welcome! Thanks for contributing.

Agree with @efx: this introduction should be very brief. The content you've written looks great for `about.md`. Keep in mind also that we dive into more detail for both [ints](https://github.com/exercism/rust/blob/main/concepts/integers/about.md) and [floats](https://github.com/exercism/rust/blob/main/concepts/floating-point-numbers/about.md) in their own concepts.

sm1215:
Hey, great points guys, thanks for the feedback! This helps me to understand the lesson structure a little better. I'll move the current writeup to the about.md and take another pass on the intro.

concepts/numbers/introduction.md
```diff
@@ -1,17 +1,6 @@
...
-## Integers
-An Integer is a whole number without a fractional component. They come in a number of variants ranging from 8-bit to 128-bit. The more bits you use, the larger the number can be, however, this increases the memory used. Each variant can be Signed or Unsigned. A Signed number allows for negative values whereas Unsigned does not. You can declare an Integer like so:
-```rust
-// Declaring and assigning an integer with a size of 8 bits
-let myInt: u8 = 5;
-```
-
-## Floating-Point Types
-A floating-point number includes decimal points and there are two types available in Rust. An `f32` and `f64` - both referring to a float of 32 bits and a float of 64 bits. `f64` is the default type in Rust as working with either size can be done at about the same speed when using a modern CPU. You can declare a Floating-Point Type like so:
+You can initialize numbers in Rust quickly and easily like this:
```
efx, line 3: I love the brevity! My other initial reaction is that I feel we should briefly summarize that Rust has a few numeric types first, then give concrete examples. 
So doing what you have here, but naming the specific numeric type in the syntax example.
And the first line summarizing the various numeric types Rust has.
Give it a whirl and let me know!

efx (changes requested):

concepts/numbers/introduction.md
sm1215: That's fair! and a good point. It seems like numbers in rust fall into a couple of broad categories based on what you might need. 

When choosing what type of number to use in Rust two of the biggest questions to me are: 
- does this number need to support decimals?
- does this number need to be negative?

The size of the number is also important but probably not so much for an introduction. Maybe the signed / unsigned distinction is also not so important for an intro? Not sure, but trying to hone in on what kind of questions the intro can answer.

Also, I thought there might be some value in showing what types Rust defaults to if a type isn't specified. I did this through a syntax example to try and demonstrate how the compiler might make decisions for you if you don't explicitly set a type, although this isn't stated outright.

concepts/numbers/introduction.md
```diff
@@ -1,3 +1,19 @@
 # Introduction
 
-TODO
+In Rust, there are are many specific types of numbers but they generally fall under one of two catorgies: integers or floating-point numbers.
```
efx, line 3:
```suggestion
In Rust, there are are many specific types of numbers but they generally fall under one of two caetgories: integers or floating-point numbers.
```

concepts/numbers/introduction.md
efx: Thanks for sharing your reasoning @sm1215! The content length you updated feels perfect to me.

> Also, I thought there might be some value in showing what types Rust defaults to if a type isn't specified. 

I like pointing out type inference at work. Seems helpful for introducing students to how the language works.

Small typo fix and then 👍 from me!

concepts/numbers/introduction.md
sm1215: oh nice catch! thanks!

concepts/numbers/introduction.md
```diff
@@ -1,6 +1,6 @@
 # Introduction
 
-In Rust, there are are many specific types of numbers but they generally fall under one of two catorgies: integers or floating-point numbers.
+In Rust, there are are many specific types of numbers but they generally fall under one of two caetgories: integers or floating-point numbers.
```
coriolinus, line 3: Looks like this introduced its own typo: "caetgories" should be "categories".

efx: ugh, my bad!

concepts/numbers/introduction.md
```diff
@@ -1,3 +1,19 @@
 # Introduction
 
-TODO
+In Rust, there are are many specific types of numbers but they generally fall under one of two caetgories: integers or floating-point numbers.
```
efx, line 3:
```suggestion
In Rust, there are are many specific types of numbers but they generally fall under one of two categories: integers or floating-point numbers.
```

efx (approved):

concepts/numbers/introduction.md
```diff
@@ -1,3 +1,19 @@
 # Introduction
 
-TODO
+In Rust, there are are many specific types of numbers but they generally fall under one of two categories: integers or floating-point numbers.
+
+Initializing an integer:
+```rust
+let x = 5; // rust defaults to an i32
+
+Initializing a large positive integer:
```
coriolinus, line 9: Missing close of code block; needs 

    ```

coriolinus (approved):

efx:
thanks for contributing @sm1215!

PR #1184, opened 2021-02-26
finish minimal shell standardization efforts

### LINKED ISSUE #1141: [CI] [proposal] low-efforts to standardize scripting

efx:
This should be enough to close #1141. 
I
- added extra details from the issue to maintaining
- added a filename linter for anything in `bin/` and `test/` that ends with `.sh`
- added that linter to CI so this PR should fail until I fix those :)


bin/lint_tool_file_names.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+# Prohibit internal shell script file names snake_case
```
coriolinus, line 3:
```suggestion
# Require that internal shell script file names use snake_case
```

bin/lint_tool_file_names.sh
```diff
@@ -0,0 +1,17 @@
...
+set -eo pipefail
+
+# find a list of files whose names do not match our convention
+errant_files=$(find bin/ _test/ -iname '*\.sh' -exec basename {} \; | grep '[^a-z_.].*')
```
coriolinus, line 7: I don't entirely understand the regex's intent here, but it looks like what it's actually checking is that there is a single character not in `[a-z_.]`, followed by any number of arbitrary other characters. That feels like an odd requirement to check. I think that if we're just detecting the presence of a forbidden character, we could simplify like

```suggestion
errant_files=$(find bin/ _test/ -iname '*\.sh' -exec basename {} \; | grep '[^a-z_.]')
```

bin/lint_tool_file_names.sh
```diff
@@ -0,0 +1,17 @@
...
+
+if [ -n "$errant_files" ]; then
+    echo "These file names do not follow our snake case convention:"
+    # find them again to print the whole relative path
+    while IFS= read -r file_name; do
+        find . -iname "$file_name"
```
coriolinus, line 13:
```suggestion
        find . -name "$file_name"
```

In this case we can be pretty confident that we have the named file's casing correct, and can avoid the case-conversion expense.

coriolinus (approved):
I think this is good to merge once CI is green.

bin/lint_tool_file_names.sh
efx: > That feels like an odd requirement to check.

Perfect correction to my regular expression naïveté!
I was trying to find any file name that has forbidden characters. So your fix makes it so!

PR #1187, opened 2021-03-03
fix CI error

efx:
This happened because newer versions of Bash (5.x) rightly fail on using
`continue` incorrectly.

I rewrote that section to simply run `rustfmt` if the example is allowed
to compile. I am unsure why that is in a subshell.

As an aside, I tested this locally on macOS by running `brew install
bash`, then changing the shebang to use `/usr/local/bin/bash`.

### DIFF
--- a/bin/format_exercises
+++ b/bin/format_exercises
@@ -16,15 +16,12 @@ format_exercises() {
         if [ -f "$allowed_file" ]; then
             exercise=$(basename "$exercise_dir")
             echo "$exercise's stub is allowed to not compile"
-            # we disable the below shellcheck lint
-            # because we want to exit this iteration
-            # shellcheck disable=SC2106
-            continue
-        fi
-        cargo fmt
-        file_name=".meta/$source_file_name.rs"
-        if [ -f "$file_name" ]; then
-            rustfmt "$file_name"
+        else
+            cargo fmt
+            file_name=".meta/$source_file_name.rs"
+            if [ -f "$file_name" ]; then
+                rustfmt "$file_name"
+            fi
         fi
     )
     done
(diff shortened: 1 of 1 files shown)


coriolinus (approved):
Seems fine to me, with one nit: the opening and closing of the subshell are at different levels of indentation. That's not introduced by this PR, though, so it's not this PR's responsibility to fix it.

I can at least say for sure why the subshell exists: `exercise_dir` is a relative path, not an absolute one, but we `cd` into it. Without the subshell, we'd have to either mess with backing out the directory change manually, or use `pushd`/`popd`, or absolutize all the paths, or something. The subshell is simplest.

Can you explain what the problem with `continue` was? It's not obvious to me.

efx:
> I can at least say for sure why the subshell exists: exercise_dir is a relative path, not an absolute one, but we cd into it. Without the subshell, we'd have to either mess with backing out the directory change manually, or use pushd/popd, or absolutize all the paths, or something. The subshell is simplest.

Oh, thanks! This makes a ton of sense to me.

> Can you explain what the problem with continue was? It's not obvious to me.

Certainly. 
When I run format_exercises with Bash 5.x I see this:
```
./bin/format_exercises
accumulate's stub is allowed to not compile
./bin/format_exercises: line 22: continue: only meaningful in a `for', `while', or `until' loop
error: expected identifier, found `?`
 --> /Users/eli/floss/exercism/rust/exercises/practice/accumulate/src/lib.rs:2:41
  |
2 | pub fn map(input: Vec<i32>, _function: ???) -> Vec<i32> {
  |                                         ^ expected identifier
```

[SC2106](https://github.com/koalaman/shellcheck/wiki/SC2106) also complains that `continue` exits the subshell and does not affect iteration.

My guess is that Bash rejects us using a continue because it is not doing what I intended. I wanted to exit onto the next exercise and directory. But `continue` would only exit the subshell.
So I wager that `exit 0` instead of `continue` is what we want. I'll try that!

PR #1188, opened 2021-03-03
include helpful links on procedural macros

efx:
I learned about procedural macros the other day and found these links
helpful.

Is it accurate to say procedural macros are a kind of meta-programming?


coriolinus (approved):
Yes, procedural macros are a form of metaprogramming.

PR #1194, opened 2021-03-12
[Robot Name] Add test_many_different_robots_have_different_names()

### LINKED ISSUE #1195: [Robot Name] Robot Name example implementation does not check for collisions
Per the instructions "Your solution must ensure that every existing robot has a unique name." Currently, the example implementation does not perform theses checks.

cwhakes:
This test confirms with high probability that a duplicate robot name is not returned.

The number 3529 was chosen with the [simple exponentiation method](https://en.wikipedia.org/wiki/Birthday_problem#A_simple_exponentiation) to provide approximately a 99.99% probability of collision.

Note: the example implementation prior to 3b23e2a fails this test because it does not check for duplicates.

Closes #1195. Replaces #1197.


cwhakes:
Whoops. Commited example fix to the wrong branch.

coriolinus:
I'd be interested to see the math you used to arrive at the number, but that's not essential. Agree that this test case is useful. Are you willing to also update the example such that the new case passes?

coriolinus (approved):
Looks great! Thanks for putting in this work.

cwhakes:
So, the math is thus:

```
let target_probability = 0.0001f64;
let total_names = (26 * 26 * 10 * 10 * 10) as f64;
let odds_of_collision = (total_names - 1.0) / total_names;
let number_of_pairs = target_probability.log(odds_of_collision);
let number_of_names = (2.0 * number_of_pairs).sqrt().ceil();
println!("Number of names required: {}", number_of_names);
```

[Link to  playground.](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=a5acabf6a424b93ff490d3be76ae6649)

PR #1209, opened 2021-03-19
explore simple spell check

efx:
markdownlint bundles a proper noun spellchecker which might be useful:
https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md044

### DIFF
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -46,6 +46,12 @@ MD040: true
 # MD042/no-empty-links - No empty links
 MD042: true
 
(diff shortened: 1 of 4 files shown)


efx:
One drawback I see is someone writing a creative story and getting a CI failure:

> You are programming rust removal software in Rust.

coriolinus:
Agree with your noted drawback. I'm reluctant to accept this change because there isn't a way AFAIK to make it context-aware, and a global rule isn't flexible enough.

PR #1217, opened 2021-03-26
option `--include-ignored` is now available on stable 1.51

Vanille-N:
Stabilization of [`cargo test -- --include-ignore`](https://github.com/rust-lang/rust/pull/80053) reached 1.51.
The pull request incidentally [cites Exercism](https://github.com/rust-lang/rust/pull/80053#issue-540237999) as a use case, it would be a shame not to make users aware of this update.


coriolinus (approved):
I like it!

PR #1218, opened 2021-03-26
fix new lints for scale-generator

coriolinus:
With the new release of rust 1.51, the nightlies are now building rust 1.53. 1.53 includes a new clippy lint which complains about interfaces which return `Result<_, ()>`; they should return `Option<_>` instead. 

In `scale-generator`, this design is intentional; we'd typedef'd `Error` to () with a long comment above so that students could define their own error type. This change produces a new unit struct instead. 

After merging, this PR should cause the `main` branch to successfully pass CI again.


coriolinus:
Per most recent commit message: apparently `unwrap` is also only defined for `E: Debug` in stable, so this isn't a weird new change; it's just one that I'd somehow not noticed until now.

efx (approved):
Thanks for explaining and fixing this!

PR #1220, opened 2021-04-01
use more idiomatic style for Forth default

coriolinus:
We should demonstrate good style to students.

### DIFF
--- a/exercises/practice/forth/src/lib.rs
+++ b/exercises/practice/forth/src/lib.rs
@@ -1,5 +1,5 @@
 pub type Value = i32;
-pub type ForthResult = Result<(), Error>;
+pub type Result = std::result::Result<(), Error>;
 
 pub struct Forth;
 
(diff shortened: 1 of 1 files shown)


exercises/practice/forth/src/lib.rs
```diff
@@ -1,5 +1,5 @@
 pub type Value = i32;
-pub type ForthResult = Result<(), Error>;
```
efx, line 2: curious: `ForthResult` feels repetitive? As in, since we are in the `forth` directory's lib.rs, it is clear from context.

coriolinus: Precisely: it's stuttering to use `foo::FooThing`; we shouldn't encourage that habit.

efx (approved):
My approval is unnecessary but felt like it!

coriolinus:
Approval is not unnecessary. Nobody, including me, gets to just merge changes that haven't been approved, outside of certain emergency circumstances which don't apply to this PR.

PR #1224, opened 2021-04-07
[booleans]: update about.md

bobahop:
First attempt at an `about.md`

### DIFF
--- a/concepts/booleans/about.md
+++ b/concepts/booleans/about.md
@@ -1,3 +1,24 @@
 # About
 
-TODO
+Booleans in Rust are represented by the `bool` type, which values can be either `true` or `false`.
+
+Rust supports one [boolean negation operator][negation operators]: `!` (Not).
+
+Rust supports six [comparison operators][comparison operators]: `==` (Equal), `!=` (Not equal), `>`	(Greater than), `<` (Less than), `>=` 
+(Greater than or equal to), `<=` (Less than or equal to).
+
+Rust supports three [logical binary operators][logical binary operators]: `&` (Logical And), `|` (Logical Or), `^` (Logical Xor).
+
+Rust supports two [lazy boolean operators][lazy boolean operators]: `||` (Or), `&&` (And). They differ from `|` and `&` in that the right-hand operand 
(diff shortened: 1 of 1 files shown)


bobahop (commented):
Deleted trailing spaces

concepts/booleans/about.md
```diff
@@ -1,3 +1,24 @@
 # About
 
-TODO
+Booleans in Rust are represented by the `bool` type, which values can be either `true` or `false`.
+
+Rust supports one [boolean negation operator][negation operators]: `!` (Not).
+
+Rust supports six [comparison operators][comparison operators]: `==` (Equal), `!=` (Not equal), `>`	(Greater than), `<` (Less than), `>=`
```
coriolinus, line 7:
```suggestion
Rust supports six [comparison operators][comparison operators]: `==` (Equal), `!=` (Not equal), `>` (Greater than), `<` (Less than), `>=`
```

Hard tab here also. I guess you copied the text from somewhere? If so, might be necessary to attribute it.

concepts/booleans/about.md
```diff
@@ -1,3 +1,24 @@
...
+(Greater than or equal to), `<=` (Less than or equal to).
+
+Rust supports three [logical binary operators][logical binary operators]: `&` (Logical And), `|` (Logical Or), `^` (Logical Xor).
+
+Rust supports two [lazy boolean operators][lazy boolean operators]: `||` (Or), `&&` (And). They differ from `|` and `&` in that the right-hand operand
+is only evaluated when the left-hand operand does not already determine the result of the expression. That is, `||` only evaluates
+its right-hand operand when the left-hand operand evaluates to `false`, and `&&` only when it evaluates to `true`.
+
+```rust
+true || false // == true
```
coriolinus, line 17:
```suggestion
true || false // == true; the right-hand side was not evaluated
```

Probably worth adding some clarifying text here.

coriolinus (commented):
This also looks quite good, and can be mergeable very quickly, I think.

coriolinus (approved):
Thanks for doing this work!

PR #1225, opened 2021-04-07
[forth] add test ruling out eager expansion of custom arguments.

### LINKED ISSUE #1221: [forth] Add test ruling out eager expansion of custom arguments
I see several implementations which eagerly expand custom arguments such that their dictionary of names contains only primitives. This is a bad idea. Consider the program:

```forth
  : attack-a 0 drop ;
  : attack-b attack-a attack-a ;
  : attack-c attack-b attack-b ;
  // ...
  : attack-z attack-y attack-y ;
```

Given eager argument expansion, `definitions["attack-z"].len() == 2.pow(26)` (26 Mb), which is an exorbitant amount of memory for something defined with two instructions. It would be trivial to extend the attack further to cause an OOM on any arbitrary machine. 

Challenge: we don't want to force students into using any particular data structure for storing their instructions map. 

One approach satisfying the challenge might be to write a custom global allocator for the test module which keeps track of how much memory has been allocated, and provides some interface to query that value at runtime. We could then write a test like the above attack, and check the allocation quantity. As long as it's under, say, 1 Mb, then the test should pass. See [`std::alloc`](https://doc.rust-lang.org/std/alloc/index.html).

coriolinus:
Closes #1221.

Implements the approach suggested in that issue.

### DIFF
--- a/exercises/practice/forth/tests/alloc-attack.rs
+++ b/exercises/practice/forth/tests/alloc-attack.rs
@@ -1,0 +1,96 @@
+//! Certain implementations of Forth naively and eagerly expand sub-definitions
+//! into superior definitions at definition time. While this appears at first to
+//! be a clever trick to solve the problem imposed by the
+//! `can_use_different_words_with_the_same_name` test, it has its own problem: it
+//! can be attacked to cause an OOM condition in an arbitrary machine without
+//! requiring that any instructions are actually executed; only definitions.
+//!
+//! This test doesn't go quite that far, but it does demonstrate the attack and
+//! require that implementations don't just do the naive thing.
+
+use forth::Forth;
+
+#[test]
+#[ignore]
+fn alloc_attack() {
+    let mut f = Forth::new();
+    f.eval(": a 0 drop ;").unwrap();
+    f.eval(": b a a ;").unwrap();
+    f.eval(": c b b ;").unwrap();
+    f.eval(": d c c ;").unwrap();
+    f.eval(": e d d ;").unwrap();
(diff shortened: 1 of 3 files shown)


coriolinus:
Note that the current example solution suffers from the attack this PR is meant to ban. This means that we need to update the example.

I'd ideally use a [student solution](https://exercism.io/mentor/solutions/b6f9f69b03df4b889c9930960cb4a358?iteration_idx=8) which is both compact and readable, but I'm not happy with the ethics of just grabbing it without permission, even with attribution. I think that approach is only viable if the student approves, and I don't have a good way to contact them. I don't think that mentor notes are the appropriate venue for such a request.

I'm happy to contribute [my own solution](https://github.com/coriolinus/exercism-rs/blob/master/forth/src/lib.rs), but that solution has not been reviewed at all and requires an external library. It's very readable (IMO), but I got a little architecture-happy with it.

iHiD:
We don't need permission. Everything a student uploads is licensed to us to use as we want. That's made very explicit during the signup process. 

Once it's done though I can let them know :)

coriolinus:
Thanks @iHiD, I think they'd appreciate knowing that once this is merged.

exercises/practice/forth/tests/alloc-attack.rs
```diff
@@ -0,0 +1,96 @@
+//! Certain implementations of Forth naively and eagerly expand sub-definitions
```
efx, line 1: Love the added context here!

efx (approved):

PR #1227, opened 2021-04-08
[&str] update about.md

bobahop:
It's a start. Tried to cover the likeliest gotchas without going into the weeds.

### DIFF
--- a/concepts/&str/about.md
+++ b/concepts/&str/about.md
@@ -7,7 +7,7 @@ There are two ways that Rust implements an array of characters: [str][str] and [
 To know why, it may help to understand why the following line of code will not compile
 
 ```rust
-const my_const: str = "Hello, World!";
+const my_const: str = *"Hello, World!";
 
 // the size for values of type `str` cannot be known at compilation time
 // the trait `Sized` is not implemented for `str`
--- a/concepts/&str/about.md
+++ b/concepts/&str/about.md
@@ -21,7 +21,7 @@ Fat pointers are of a known size, and so this line will compile
 const my_const: &str = "Hello, World!";
 ```
 
-A mutable string slice (`mut &str`) can have its valued replaced, but not modified in place.
+A mutable binding to a string slice can have its valued replaced, but not modified in place.
 
 Example
 
--- a/concepts/&str/about.md
+++ b/concepts/&str/about.md
@@ -7,7 +7,7 @@ There are two ways that Rust implements an array of characters: [`str`][str] and
 To know why, it may help to understand why the following line of code will not compile
 
 ```rust
-const my_const: str = *"Hello, World!";
+const MY_CONST: str = *"Hello, World!";
 
 // the size for values of type `str` cannot be known at compilation time
 // the trait `Sized` is not implemented for `str`
@@ -18,7 +18,7 @@ But as a _reference_ to a literal (`&str`), it has a "fat pointer" which holds t
 Fat pointers are of a known size, and so this line will compile
 
 ```rust
-const my_const: &str = "Hello, World!";
+const MY_CONST: &str = "Hello, World!";
 ```
 
 A mutable binding to a string slice can have its valued replaced, but not modified in place.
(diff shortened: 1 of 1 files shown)


concepts/&str/about.md
```diff
@@ -1,3 +1,69 @@
...
 
-TODO
+There are two ways that Rust implements an array of characters: [str][str] and [String][String].
+
+`str`, also known as a "string slice", is usually used with a reference as `&str`.
+
+To know why, it may help to understand why the following line of code will not compile
+
+```rust
+const my_const: str = "Hello, World!";
```
coriolinus, line 10:
```suggestion
const my_const: str = *"Hello, World!";
```

We'd need the dereference there to avoid a type error; string literals have type `&'static str`. For example, this is legal:

```rust
const HELLO: [u8; 12] = *b"Hello, world";
```

concepts/&str/about.md
```diff
@@ -1,3 +1,69 @@
...
+
+Even though the literal "Hello, World!" length can be known at compilation time, as a `str` it is a sequence of unknown size.
+But as a _reference_ to a literal (`&str`), it has a "fat pointer" which holds the address and length of the literal in memory.
+Fat pointers are of a known size, and so this line will compile
+
+```rust
+const my_const: &str = "Hello, World!";
+```
+
+A mutable string slice (`mut &str`) can have its valued replaced, but not modified in place.
```
coriolinus, line 24:
```suggestion
A mutable binding to a string slice can have its valued replaced, but not modified in place.
```

Two changes in the suggestion:

- Add the word "binding", because that's what's mutable, not the reference
- remove the type, because `mut &str` isn't really a thing you'll see in rust.

concepts/&str/about.md
```diff
@@ -1,3 +1,69 @@
...
+// mismatched types
+// expected reference `&String`
+// found reference `&str`
+```
+
+A `&str` will not be [coerced][coercion] into a `&String`. But this will compile
+
+```rust
+pub fn main() {
+    let my_string: String = String::from("Hello, World!");
```
coriolinus, line 56:
```suggestion
    let my_string = String::from("Hello, World!");
```

An explicit type there is redundant.

concepts/&str/about.md
```diff
@@ -1,3 +1,69 @@
 # About
 
-TODO
+There are two ways that Rust implements an array of characters: [str][str] and [String][String].
```
efx, line 3:
```suggestion
There are two ways that Rust implements an array of characters: [`str`][str] and [`String`][String].
```

It seems like we are establishing a convention of using code formatting when discussing types. Sound fair?

concepts/&str/about.md
bobahop: The suggested line

```rust
const my_const: str = *"Hello, World!";
```
results in the same error. We need a reference, not a dereference.

concepts/&str/about.md
bobahop: I had it that way at first, but then noticed that it hid the link unless you mouse over it, so I removed the back ticks. But I don't feel strongly about it if you want me to put them back in.

concepts/&str/about.md
bobahop: Yeah, I had it explicit for pedagogic purposes, but I think you're right to keep it best practice.

concepts/&str/about.md
bobahop: Yes, I know what you mean that the dereference is meant to bring the  `&'static str` to a str, but I'm not sure it might be confusing to throw the deref operator in there for students who need help with `&str`. I wanted to avoid getting into derefs in order to keep the focus on `&str`, but I don't feel strongly about putting the `*` in there if you think it won't confuse the matter for newer students.

bobahop: I was tempted to cover `&'static str` but decided to keep lifetimes out of it.

bobahop: I get the type mismatch error on a let, but not a const. Perhaps it should have the deref operator in case the compiler ever changes to squawk  about the mismatch. Sorry for all the thinking out loud...

concepts/&str/about.md
coriolinus: I prefer to render types with code formatting. If exercism's stylesheet hides links in code formatting, that's a stylesheet bug; we should fix the stylesheet instead of leaving out semantic markers to work around it.

coriolinus (approved):
I like it a lot. Thank you very much!

PR #1228, opened 2021-04-09
Update luhn.rs

bobahop:
An example of the code this would catch that's currently slipping through:

```rust
//17 nanoseconds
// Check a Luhn checksum.
 pub fn is_valid(code: &str) -> bool {
     let mut sum = 0;
     let mut i = 0;
     for ch in code.chars().rev() {
         match ch.to_digit(10) {
             Some(digit) => {
                 sum += if i % 2 == 0 {
                     digit
                 } else if 4 < digit {
                     digit * 2 - 9
                 } else {
                     digit * 2
                 };
                 i+=1;
             },
             None => {
                 if ch != ' ' {
                     break
                 }
             },
         }
     }
     i > 1 && sum % 10 == 0
 }
```

### DIFF
--- a/exercises/practice/luhn/tests/luhn.rs
+++ b/exercises/practice/luhn/tests/luhn.rs
@@ -125,3 +125,10 @@ fn test_using_ascii_value_for_nondoubled_nondigit_isnt_allowed() {
 fn test_valid_number_with_an_odd_number_of_spaces() {
     process_valid_case("234 567 891 234", true);
 }
+
+#[test]
+#[ignore]
+/// non-numeric, non-space char in the middle with a sum that's divisible by 10 isn't allowed
+fn test_invalid_char_in_middle_with_sum_divisible_by_10_isnt_allowed() {
+    process_valid_case("59%59", false);
+}


exercises/practice/luhn/tests/luhn.rs
```diff
@@ -125,3 +125,10 @@ fn test_using_ascii_value_for_nondoubled_nondigit_isnt_allowed() {
 fn test_valid_number_with_an_odd_number_of_spaces() {
     process_valid_case("234 567 891 234", true);
 }
+
+#[test]
+#[ignore]
+/// non-numeric, non-space char in the middle with a sum that's divisible by 10 isn't allowed
+fn test_invalid_char_in_middle_with_sum_divisible_by_10_isnt_allowed() {
+    process_valid_case("59%59", false);
```
coriolinus, line 133: The name is a bit confusing; what adjacent subsequence of 5959 sums to a multiple of 10?

coriolinus (approved):
Good catch!

exercises/practice/luhn/tests/luhn.rs
bobahop: I didn't catch it and I can't say I've done more than test it, but here's my attempt to explain what I don't know...

For "59%59"
The first step of the Luhn algorithm is to double every second digit, starting from the right.
So the last 5 gets doubled to 10.
If doubling the number results in a number greater than 9 then subtract 9 from the product.
So 10 - 9 = 1.
If you sum from right to left, then  9 + 1 = 10, which is the sum when you hit the `%`

coriolinus: Hmm. I'm not sure that I follow that, but I'm sure it's fine. I was mainly wondering about the name of the function, not the nature of the failure. It's not a big deal.

PR #1229, opened 2021-04-09
[box] update about.md

bobahop:
first draft. Tried not to bring other concepts into it, but to justify Box, I kinda had to, I think.

### DIFF
--- a/concepts/box/about.md
+++ b/concepts/box/about.md
@@ -68,7 +68,7 @@ fn annoy_me(time_of_day: TimeOfDay) -> impl Annoyance {
 
 ```
 
-The compiler, in all of its helpfullness, tells us exactly what to do, which is to return a boxed trait object. If you don't know what a trait object is,
+The compiler, in all of its helpfulness, tells us exactly what to do, which is to return a boxed trait object. If you don't know what a trait object is,
 a simple way to describe it is a way of returning a specific object via [dynamic dispatch][dynamic dispatch] from a function that could return any object
 implementing a specified trait. Since we need dynamic dispatch, we can try just putting a `dyn` in front of the trait name and see if we get away with it.
 
@@ -86,8 +86,11 @@ fn annoy_me(time_of_day: TimeOfDay) -> dyn Annoyance {
 // unsized locals are gated as an unstable feature
 ```
 
-One of the errors is that the size for `dyn Annoyance` cannot be known at compile time. And we are reminded that one of the uses for `Box` is for a type whose
-size can’t be known at compile time. So now we try using a `Box`ed return value, and we `Box`the values returned from the match expression, like so
+One of the errors is that the size for `dyn Annoyance` cannot be known at compile time. This is because annoy_me does not know what specific trait object
+it will return in advance of when it is called. Structs implementing Annoyance could have different fields and be different sizes. And we are reminded that
+one of the uses for `Box` is for a type whose size can’t be known at compile time. So now we try using a `Box`ed return value, and we `Box`the values returned
+from the match expression, like so
+
 
 ```rust
 fn annoy_me(time_of_day: TimeOfDay) -> Box<dyn Annoyance> {
--- a/concepts/box/about.md
+++ b/concepts/box/about.md
@@ -54,8 +54,8 @@ enum TimeOfDay {
 //return something that implements Annoyance
 fn annoy_me(time_of_day: TimeOfDay) -> impl Annoyance {
     match time_of_day {
-        TimeOfDay::Midnight => WerewolfOfLondon {},
-        TimeOfDay::Morning => Rooster {},
+        TimeOfDay::Midnight => WerewolfOfLondon,
+        TimeOfDay::Morning => Rooster,
     }
 }
 
--- a/concepts/box/about.md
+++ b/concepts/box/about.md
@@ -9,7 +9,7 @@ If the data is of an unknown size at compilation time or has a size that may cha
 space in the heap big enough to store the data, and then the address of where that data starts in memory gets pushed onto the stack. Retrieving data from the
 heap is slower because we need to go through the address on the stack and follow that to the location of tha data on the heap.
 
-`Box` is essentially a smart pointer which allocates data on the heap. The address is on the stack, pointing to the data on the heap. There are several 
+`Box` is essentially a smart pointer which allocates data on the heap. The address is on the stack, pointing to the data on the heap. There are several
 [reasons][reasons] to use `Box` to put data on the heap:
 
 - When you have a type whose size can’t be known at compile time and you want to use a value of that type in a context that requires an exact size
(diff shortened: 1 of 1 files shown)


bobahop:
"One of the errors is that the size for dyn Annoyance cannot be known at compile time."
It probably needs a short explanation of why that is.

concepts/box/about.md
```diff
@@ -1,3 +1,153 @@
...
 
-TODO
+To understand [`Box`][box] it is helpful to understand what the [stack and heap][stack and heap] are.
+
+The stack is an area of memory that works like a last-in-first-out (LIFO) buffer. It is very fast to push data onto the stack and it is fast to pop it off.
+The data can include variables or function addresses, but the data must be of a known fixed size at compilation time.
+
+If the data is of an unknown size at compilation time or has a size that may change, then it must be allocated on the heap. The allocation process must find
+space in the heap big enough to store the data, and then the address of where that data starts in memory gets pushed onto the stack. Retrieving data from the
+heap is slower because we need to go through the address on the stack and follow that to the location of tha data on the heap.
```
coriolinus, line 10: This isn't wrong, but it's misleading. The major slowdown of the heap vs the stack is that allocating memory on the heap is very slow. Reads and mutation, as long as no reallocation is necessary, shouldn't be significantly slower than using the stack.

concepts/box/about.md
```diff
@@ -1,3 +1,153 @@
...
+
+`Box` is essentially a smart pointer which allocates data on the heap. The address is on the stack, pointing to the data on the heap. There are several
```
coriolinus, line 12:
```suggestion
`Box` is a smart pointer which allocates data on the heap. The address is on the stack, pointing to the data on the heap. There are several
```

No need to qualify it; `Box` is a smart pointer as C++ understands it.

concepts/box/about.md
```diff
@@ -1,3 +1,153 @@
...
+[reasons][reasons] to use `Box` to put data on the heap:
+
+- When you have a type whose size can’t be known at compile time and you want to use a value of that type in a context that requires an exact size
```
coriolinus, line 15: Might be worth mentioning inline that this is used in an example below.

concepts/box/about.md
```diff
@@ -1,3 +1,153 @@
...
+
+- When you have a large amount of data and you want to transfer ownership but ensure the data won’t be copied when you do so
```
coriolinus, line 17: So, you're saying that you have 

```rust
type GigArray = [u8; 1024 * 1024 * 1024];
```

In that case, it's true that it's more efficient to write

```rust
fn foo(ga: Box<GigArray>)
```

instead of 

```rust
fn foo(ga: GigArray)
```

However, that's because of a peculiarity of arrays in Rust; [arrays of any size implement `Copy` if the element type does](https://doc.rust-lang.org/std/primitive.array.html). In most other cases I can think of, including the same thing with vectors, Rust doesn't necessarily copy the data when transferring ownership; that was a fairly obvious optimization for the compiler team, and it's been in the language for a while.

I'm not sure that this point is worth including; it's rare to transfer ownership of large arrays.

concepts/box/about.md
```diff
@@ -1,3 +1,153 @@
...
+
+- When you want to own a value and you care only that it’s a type that implements a particular trait rather than being of a specific type
```
coriolinus, line 19: How is this point distinct from the first point?

concepts/box/about.md
```diff
@@ -1,3 +1,153 @@
...
+
+A practical usage of `Box` is when you have multiple structs which you want to handle the same via a trait which they all implement.
+
+Let's say we have two structs. Each struct is of something that can be annoying. We can have each struct implement an Annoyance trait like so
+
+```rust
+struct WerewolfOfLondon {}
+struct Rooster {}
```
coriolinus, line 27:
```suggestion
struct WerewolfOfLondon;
struct Rooster;
```

Unit structs can be more simply defined like this.

concepts/box/about.md
```diff
@@ -1,3 +1,153 @@
...
+enum TimeOfDay {
+    Midnight,
+    Morning,
+}
+
+//return something that implements Annoyance
+fn annoy_me(time_of_day: TimeOfDay) -> impl Annoyance {
+    match time_of_day {
+        TimeOfDay::Midnight => WerewolfOfLondon {},
+        TimeOfDay::Morning => Rooster {},
```
coriolinus, line 58:
```suggestion
        TimeOfDay::Midnight => WerewolfOfLondon,
        TimeOfDay::Morning => Rooster,
```

Assuming you took the previous unit struct optimization, you get a minor usability benefit here.

Also, the same edit in subsequent examples.

concepts/box/about.md
```diff
@@ -1,3 +1,153 @@
...
+    }
+}
+
+// the size for values of type `dyn Annoyance` cannot be known at compilation time
+// the trait `Sized` is not implemented for `dyn Annoyance`
+// all local variables must have a statically known size
+// unsized locals are gated as an unstable feature
+```
+
+One of the errors is that the size for `dyn Annoyance` cannot be known at compile time. This is because annoy_me does not know what specific trait object
```
coriolinus, line 89:
```suggestion
One of the errors is that the size for `dyn Annoyance` cannot be known at compile time. This is because `annoy_me` does not know what specific trait object
```

concepts/box/about.md
```diff
@@ -1,3 +1,153 @@
...
+```
+
+Since my_annoyance is of type `Box<dyn Annoyance>` it must be dereferenced from the `Box` using the deref operator `*`. It is then referenced as a
+`dyn Annoyance` for the be_annoying function to borrow it.
+
+Another way to implement it would be for the be_annoying function to accept a `Box<dyn Annoyance>` and use [deref coercion][deref coercion] to avoid
+having to dereference the `Box` manually.
+
+```rust
+fn be_annoying(annoying_one: Box<dyn Annoyance>) {
```
coriolinus, line 134: Worth noting that this transfers ownership of the box, and you don't get it back.

concepts/box/about.md
```diff
@@ -1,3 +1,153 @@
...
+
+```rust
+"ahhh-OOOOOOO!"
+"errr-errr-errr-errr-ERRRRRRRRR!"
+```
+
+Since my_annoyance is of type `Box<dyn Annoyance>` it must be dereferenced from the `Box` using the deref operator `*`. It is then referenced as a
+`dyn Annoyance` for the be_annoying function to borrow it.
+
+Another way to implement it would be for the be_annoying function to accept a `Box<dyn Annoyance>` and use [deref coercion][deref coercion] to avoid
```
coriolinus, line 130: A third way:

```rust
fn be_annoying<A: Annoyance + ?Sized>(annoying_one: &A) {
    println!("{:?}", annoying_one.annoy());
}
```

[playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=cc6f839bfb2bb6acf281baf656972dd9)

This uses static instead of dynamic dispatch; the compiler has monomorphized `fn be_annoying` in this case, and called the appropriate variant for each variant of `Annoyance`.

concepts/box/about.md
coriolinus: Also just noticed a typo in that line: "tha" for "the".

bobahop (commented):
I think I implemented all suggested changes.

coriolinus (approved):
That's a lot of writing! Thanks again for doing all this work.

PR #1230, opened 2021-04-10
rename `&str` concept to appease `configlet lint`

ee7:
From 4.0.0-alpha.12 onwards, `configlet lint` will check that some JSON
keys have a value that is a kebab-case string, which we define as a
string that matches this regular expression:
`^[a-z0-9]+(-[a-z0-9]+)*$`

A concept's `slug` is one value that must match that pattern, so we must
rename the `&str` concept.

See:
- https://github.com/exercism/docs/blob/main/building/configlet/lint.md
- https://github.com/exercism/docs/pull/113/files

---

See also https://github.com/exercism/configlet/pull/274#issuecomment-816650320.

Hi Rust track folks. The powers that be have decided that `configlet lint` should complain about "kebab strings" that use a character not in `[0-9a-z-]`, and it turns out that the Rust track is the only one that had such a string. Sorry.

Of course, feel free to rename it to something other than `string-slices`.

I wrote the `concepts.name` as `String Slices` rather than `String slices`, which I think is correct. But I don't know if the latter will be disallowed - the format hasn't been fully specified yet.

And I might not have changed it in all the required places, so please check. But searching `git`-tracked files in this repo for `&str`, excluding matches in `.rs` files:
```
$ git grep --break --heading '&str' -- './*' ':!*.rs'
```

```markdown
concepts/string-slices/about.md
5:`str`, also known as a "string slice", is usually used with a reference as `&str`.
17:But as a _reference_ to a literal (`&str`), it has a "fat pointer" which holds the address and length of the literal in memory.
21:const MY_CONST: &str = "Hello, World!";
32:my_string.push_str(", World!"); // there is no equivalent function to modify &str
35:It is less constraining to use `&str` instead of `&String` when defining a parameter for a function, as this will not compile
39:    let my_str: &str = "Hello, World!";
49:// found reference `&str`
52:A `&str` will not be [coerced][coercion] into a `&String`. But this will compile
60:fn say_hi(phrase: &str) {
65:A `&String` will be coerced into a `&str`. So to use a parameter type of `&str` will often be more useful.

concepts/string-vs-str/about.md
3:`String` and `&str` are two fundamental ways of working with textual data in Rust.
8:`&str`, is a borrowed reference to a String slice.
9:`&str` is immutable.
11:The proper pronunciation of `&str` is "ref string".
14:`&str` is a read only view of a well-formed UTF-8 sequence.
17:`String` and `&str` in Rust can not be indexed into like you might in other languages.
46:## Converting a `String` into a `&str`
49:This trait has some compiler special casing: it means that any reference to a `String` can automatically, transparently be coerced into an `&str`.
52:- Any code which expects, for example, an `&str` argument can accept an `&String` instead.
53:- Any method implemented for `&str` can be called on a `String` as well.
55:(If one generically wanted to be able to take any form of reference to a `str` then one could have an argument of type `impl AsRef<str>` but typically typing arguments as `&str` is common.)
57:## Converting a `&str` to a `String`
59:There are many ways to create a `String` from a `&str`:
67:This is why `&str` is available in no_std but `String` is not.
72:## Using String and `&str` in structs
74:This section discusses when to use `String` or `&str` for struct fields.
75:You'll often see structs defined with `String` instead of `&str` for field values.
90:`&str` borrows textual data so will perform better at the trade-off of your struct needing to define lifetime parameters.
96:`&str`'s are a shared reference and so do not require frequent reallocation.

concepts/string-vs-str/introduction.md
3:`String` and `&str` are two fundamental ways of working with textual data in Rust.

concepts/string-vs-str/links.json
20:        "description": "&str primitive documentation"

docs/archived/README.md
34:- [x] `&str`
98:- [x] `String` vs `&str`
154:| `hello-world`            | -                                            | functions, test suites                                   | return the `&str` literal `"hello world"`                                        |
166:| `str-use`                | `char`, `slices`                             | `&str`                                                   | note: immutable!                                                                 |

docs/archived/exercise-concepts/armstrong-numbers.md
23:- `&str`: needed for counting the length of the number since Rust is strongly typed

docs/archived/exercise-concepts/decimal.md
158:    pub fn try_from(input: &str) -> Option<Decimal> {
205:    pub fn try_from(input: &str) -> Option<Self> {

docs/archived/exercise-concepts/luhn.md
9:- `&str` and `char`
10:  - students must grasp the distinction of individual `char` from `&str` in order to correctly exclude or include parts of the provided input
29:pub fn is_valid(code: &str) -> bool {
49:pub fn is_valid(candidate: &str) -> bool {

docs/archived/exercise-concepts/parallel-letter-frequency.md
14:  - `&str` type
95:pub fn frequency(input: &[&str], worker_count: usize) -> HashMap<char, usize> {
134:pub fn frequency(input: &[&str], worker_count: usize) -> HashMap<char, usize> {

docs/archived/exercise-concepts/poker.md
14:- `String` vs `&str`

docs/archived/exercise-concepts/robot-simulator.md
11:  - `&str`
82:    pub fn instructions(self, instructions: &str) -> Self {

exercises/concept/csv-builder/.docs/instructions.md
30:`add` takes immutable string slices, `&str`, as the argument.

exercises/concept/csv-builder/.docs/introduction.md
1:`String` and `&str` are two fundamental ways of working with textual data in Rust.
4:`&str` is a borrowed reference to a String slice and thus immutable.
12:You can also create a `String` by converting a `&str`:

exercises/concept/csv-builder/.meta/design.md
3:Teach the student the differences between String and &str and how to work successfully with both.
7:    Know &str is shared and immutable
8:    Know String has &str methods because of shared traits
32:    [&str primitive documentation](https://doc.rust-lang.org/std/primitive.str.html)

exercises/concept/entry-api/.docs/instructions.md
5:Given `&[&str]` representing the words of a magazine article, and `&[&str]` representing the words of a note you would like to send, can you compose your note by cutting words out of the magazine and pasting them into a letter?
15:pub fn can_construct_note(magazine: &[&str], note: &[&str]) -> bool {
23:let magazine = "two times three is not four".split_whitespace().collect::<Vec<&str>>();
24:let note = "two times two is four".split_whitespace().collect::<Vec<&str>>();

exercises/concept/structs/.docs/hints.md
15:- Notice that the `name` method returns a `&str` when the `name` field on the `User` struct is a `String`. How can we get `&str` and `String` to play nice with each other?
```


coriolinus (approved):
Thanks for doing this work and for the thorough documentation of the rationale!

efx:
Indeed, thank you ee7!

PR #1231, opened 2021-04-11
[collect] update about.md

bobahop:
First draft

### DIFF
--- a/concepts/collect/about.md
+++ b/concepts/collect/about.md
@@ -41,8 +41,8 @@ we want a `Vec<char>` we could insert the type into our call to `collect` like s
 // let rev_chars = "Hello, World!".chars().rev().collect::<Vec<_>>();
 ```
 
-and `rev_chars` will now be a `Vec<char>`, but the double colons and outer angle brackets (also known as the `turbofish`) which passes the type
-information to the `collect` function is noisier syntax than we need. By specifying the type on the binding itself the `collect` can infer what
+The double colons and outer angle brackets (i.e. `::<_>`) are known as the `turbofish`.
+In this case the turbofish is noisier syntax than we need. By specifying the type on the binding itself the `collect` can infer what
 collection it should create without the syntactical noise.
 
 ```rust
--- a/concepts/collect/about.md
+++ b/concepts/collect/about.md
@@ -1,3 +1,85 @@
 # About
 
-TODO
+To understand the `collect` function it's important to know what iterators are. To implement the `Iterator` trait one only has to implement
+a `next` method, which has the following signature
+
+```rust
+pub trait Iterator {
+    type Item;
+
+    fn next(&mut self) -> Option<Self::Item>;
+}
+```
+
+As the `next` method implies, implementing the `Iterator` trait generates a sequence of items, usually from a collection such as a vector,
+array or a string (which is a collection of chars.) Calling `next` for the first time will yield the first item in the sequence if there is
+one, or, if there is nothing in the collection remaining to sequence it will return `None`. However, it is not usual to call the `next` method
+directly. The power of iterators is to use [iterator adaptors][iterator adaptors] which traverse an entire sequence to provide some kind of
+transformation of the collection's items into another iterator. Common iterator adaptors include: [filter][filter], which retains only items from
+a collection that satisfy a condition; [map][map], which transforms the type of the items into another type; and [rev][rev], which reverses the
+order of the items. Since iterator adaptors yield another iterator, they can be chained together. For instance, to reverse the characters in a
+string one might want an iterator of characters and then a reverse iterator, like so
+
+```rust
+let rev_chars = "Hello, World!".chars().rev();
+```
+
+But `rev_chars` does not give us a `String` with the characters of "Hello, World!" reversed. It gives us an `impl Iterator<Item = char>`. It's a
+type which is _prepared_ to yield a sequence of characters, but it doesn't know how to finally assemble them. In this case it doesn't know if we
+want a `Vec<char>` or a `String`.
+
+Even though iterator adaptors do so much, they are _lazy_. But this only means that they are like an arrow drawn in a bow. They are not released
+until they have a target, called a [consuming adaptor][consuming adaptor], which uses up either a single iterator or the last iterator in a chain
+of iterators.
+
+[`collect`][collect] is one such consuming adaptor, and is what can be used to tell `rev_chars` what type of collection we want to end up with. If
+we want a `Vec<char>` we could insert the type into our call to `collect` like so
+
+```rust
+    let rev_chars = "Hello, World!".chars().rev().collect::<Vec<char>>();
+// or, since the collection can only be made of chars if it's a Vec, we don't have to specify them and just use
+// let rev_chars = "Hello, World!".chars().rev().collect::<Vec<_>>();
+```
+
+and `rev_chars` will now be a `Vec<char>`, but the double colons and outer angle brackets (also known as the `turbofish`) which passes the type
+information to the `collect` function is noisier syntax than we need. By specifying the type on the binding itself the `collect` can infer what
+collection it should create without the syntactical noise.
+
+```rust
+    let rev_chars: Vec<char> = "Hello, World!".chars().rev().collect();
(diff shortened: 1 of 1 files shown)


concepts/collect/about.md
```diff
@@ -1,3 +1,85 @@
...
+let rev_chars = "Hello, World!".chars().rev();
+```
+
+But `rev_chars` does not give us a `String` with the characters of "Hello, World!" reversed. It gives us an `impl Iterator<Item = char>`. It's a
+type which is _prepared_ to yield a sequence of characters, but it doesn't know how to finally assemble them. In this case it doesn't know if we
+want a `Vec<char>` or a `String`.
+
+Even though iterator adaptors do so much, they are _lazy_. But this only means that they are like an arrow drawn in a bow. They are not released
+until they have a target, called a [consuming adaptor][consuming adaptor], which uses up either a single iterator or the last iterator in a chain
+of iterators.
```
coriolinus, line 33: While I have _several_ notes here, fundamentally all this information should just live in the `Iterator` documentation. Iterators are absolutely a prerequisite concept for collection, and we shouldn't re-teach the concept here.

Suggest replacing all this with something like:

> Recall that iterators are lazy sequences of items. [Consuming adapters][consuming adaptor] process all items in the iterator to produce some kind of output.

Also, adapter is spelled with an e, not an o.

concepts/collect/about.md
```diff
@@ -1,3 +1,85 @@
...
+
+[`collect`][collect] is one such consuming adaptor, and is what can be used to tell `rev_chars` what type of collection we want to end up with. If
+we want a `Vec<char>` we could insert the type into our call to `collect` like so
```
coriolinus, line 36: Let's be particular about what `collect` is.

```suggestion
[`collect`][collect] a consuming adapter which can be used to produce an instance of any type implementing the [`FromIterator` trait](https://doc.rust-lang.org/std/iter/trait.FromIterator.html). That trait is implemented for all the collections in the standard library. 

Because `FromIterator` is implemented for many types, you need to let Rust know what sort of collection you desire. 
If we want a `Vec<char>` we could insert the type into our call to `collect` like so:
```

concepts/collect/about.md
```diff
@@ -1,3 +1,85 @@
...
+
+```rust
+    let rev_chars = "Hello, World!".chars().rev().collect::<Vec<char>>();
+// or, since the collection can only be made of chars if it's a Vec, we don't have to specify them and just use
+// let rev_chars = "Hello, World!".chars().rev().collect::<Vec<_>>();
```
coriolinus, line 41:
```suggestion
// since the item type of the `String::chars()` iterator can be inferred to be `char`, we can omit that information
// let rev_chars = "Hello, World!".chars().rev().collect::<Vec<_>>();
```

concepts/collect/about.md
```diff
@@ -1,3 +1,85 @@
...
+```
+
+and `rev_chars` will now be a `Vec<char>`, but the double colons and outer angle brackets (also known as the `turbofish`) which passes the type
+information to the `collect` function is noisier syntax than we need. By specifying the type on the binding itself the `collect` can infer what
```
coriolinus, line 45:
```suggestion
The double colons and outer angle brackets (i.e. `::<_>`) are known as the `turbofish`.
In this case the turbofish is noisier syntax than we need. By specifying the type on the binding itself the `collect` can infer what
```

concepts/collect/about.md
```diff
@@ -1,3 +1,85 @@
...
+collection it should create without the syntactical noise.
```
coriolinus, line 46:
```suggestion
collection it should create with less syntactical noise.
```

concepts/collect/about.md
```diff
@@ -1,3 +1,85 @@
...
+// let rev_chars: Vec<_> = "Hello, World!".chars().rev().collect();
+```
+
+If we wanted a `String` instead we could do it like so
+
+```rust
+    let rev_chars: String = "Hello, World!".chars().rev().collect();
+```
+
+But what if we want to return a collection from a function instead of letting a binding? We could do
```
coriolinus, line 60:
```suggestion
Rust's type inference is powerful enough that sometimes we don't need an explicit inline note:
```

concepts/collect/about.md
```diff
@@ -1,3 +1,85 @@
...
+```
+
+The return type of the `get_reversed_chars` function lets the `collect` function infer what collection to return. Here, though, we must specify the
+`Vec<char>` and not `Vec<_>` or we get the compilation error
+
+```rust
+// the type placeholder `_` is not allowed within types on item signatures
+// not allowed in type signatures
+// replace with the correct return type
+```
```
coriolinus, line 75: Not sure this section is essential. Students will discover it for themselves if they encounter the situation, and they'll probably have encountered the rule that Rust never performs type inference in function signatures anyway.

concepts/collect/about.md
```diff
@@ -1,3 +1,85 @@
...
+
+Although you can sometimes use `collect` without specifying the type, at other times the compiler will state it requires the type for a `collect`, so it
+can be a good habit to be aware of how to specify what type you want to collect.
```
coriolinus, line 78: Also not sure this is essential. Rustc will complain if it can't infer something, and students' natural reluctance to add a bunch of syntax will encourage them to discover the minimal limits of what rustc will accept.

coriolinus (commented):
Lots of wording suggestions here. The most important notes are:

- don't re-teach iterators
- do talk about the `FromIterator` trait, which is essential to understand `collect`

concepts/collect/about.md
bobahop: Thank you for your time and attention to this.

Okay to swapping the `Iterator` coverage with your phrase and to cover `FromIterator`.

If you google "adaptor vs adapter" you'll find various opinions on the right spelling. I use "or" for consistency with how it is spelled in TRPL, which I link to, so I opt to keep it.


https://doc.rust-lang.org/book/ch13-02-iterators.html#methods-that-consume-the-iterator
https://doc.rust-lang.org/book/ch13-02-iterators.html#methods-that-produce-other-iterators

concepts/collect/about.md
bobahop: Okay, I'll fiddle with that.

concepts/collect/about.md
bobahop: I make no assumption that they've encountered the rule, and I'd rather give a proactive heads up than have them stumble onto it. For pedagogical purposes I think it does no harm and may be helpful.

concepts/collect/about.md
bobahop: I agree. I basically put it in as a kind of closing thought to keep the ending from seeming abrupt, but I don't feel strongly about it.

concepts/collect/about.md
```diff
@@ -1,3 +1,52 @@
 # About
 
-TODO
+To understand the `collect` function it's important to know that it is used to resolve an iterator. Recall that an iterator is a sequence of items
+that can be chained with other iterators. Iterators are _lazy_, which means they do not execute until they are resolved by a
+[consuming adaptor][consuming adaptor]. Consuming adaptors process all items in the iterator to produce some kind of output.
```
coriolinus, line 5: > they do not execute until they are resolved by a consuming adaptor

I think this is misleading. Consuming adapters are one way to resolve an iterator, but there are others, such as a `for` loop:

```rust
for ch in "hello world".chars() {
  print!("{}", ch);
}
```

concepts/collect/about.md
coriolinus: In that case, don't just show the error they'll get, explain why: Rust doesn't perform type inference for function signatures, not because it's an impossible technical challenge, but because this ensures that someone reading the code can know at a glance what types a function consumes and returns.

concepts/collect/about.md
bobahop: Yes, that's a good point. I'll finesse it.

coriolinus (approved):
I think we got this to a good place! Thanks for putting in all this work.

bobahop:
Thank you for the guidance!

PR #1233, opened 2021-04-13
[functions] update about.md

bobahop:
First draft.

### DIFF
--- a/concepts/functions/about.md
+++ b/concepts/functions/about.md
@@ -1,3 +1,176 @@
 # About
 
-TODO
+Functions are bodies of code containing one or more [statements or expressions][statements or expressions]. A function may optionally
+[return an expression][return value]. (Recall that a statement is an instruction that does not return a value. An expression evaluates to
+a resulting value.)
+
+Rust style is to use [snake case][snake case] for a function name, which is prefaced by the `fn` keyword for a function definition. For example
+
+```rust
+fn do_nothing() {}
+
+```
+
+The empty parentheses are a placeholder for zero or more parameters. A parameter is the definition of a value to be passed into the function.
+Even if there are no parameters the parentheses are still required in the function definition. The actual value passed to a parameter is called
+an argument.
+
+The combination of parameters and return value is known as the function's signature. A function's signature requires that each parameter must
+have its type annotated.
+
+The curly braces enclose the body of the function. Inside the braces are where the function's statements and expressions are placed. Even if the
+function contains no statements or expressions the curly braces are still required for the function definition. Of course there is no point to
+having an empty function.
+
+A function could contain a simple statement. In the following example the function takes in one `i32` argument and prints it
+
+```rust
+fn print_integer(value: i32) {
+    println!("{:?}", value);
+}
+```
+
+Note that the parameter needs to be given a name (`value` in this case) and be annotated with its type.
+
+A function can also return an expression. In the following example the function takes in one `i32` argument and returns it doubled
+
+```rust
+fn double_integer(value: i32) -> i32 {
+    value * 2
+}
+```
+
+The `-> i32` indicates that the function returns an `i32`. Unlike parameters, the returned value is not named. Note that the last expression
+in a function can be returned without the `return` keyword and a semicolon. `return` _is_ needed to return early from a function, like so
+
+```rust
+fn long_function() -> i32 {
+    let some_condition = false;
+    // code snipped
+    if (some_condition) {
+        return 0;
+    }
+    // more code snipped
+    42
+}
+```
+
+A function has its own scope which borrows or takes ownership of arguments passed to it. In this example the last line in `main` causes a
+compile error
+
+```rust
+fn double_integer(value: Box<i32>) -> i32 {
+    *value * 2
+}
+
+pub fn main() {
+    let num = Box::new(42);
+    let num_doubled = double_integer(num); // value moved here
+    println!("{:?}", num_doubled);
+    println!("{:?}", num); // borrow of moved value: `num` value borrowed here after move
+    // move occurs because `num` has type `Box<i32>`, which does not implement the `Copy` trait
+}
+```
+
+`main`, which owns the boxed `num` value, moves `num` as the argument for `value` into `double_integer`, which takes ownership of it. At the
+end of `double_integer`, `value` goes out of scope and is dropped, so `num` is no longer valid for the rest of the `main` function.
+
+Note that `fn main` is prefaced by `pub`, but `fn double_integer` is not. A `pub fn` is public, meaning it can be accessed from a different
+module. Functions defined without `pub` are private and can only be accessed from the current module. `pub(crate) fn` makes a function
+accessible only within the current crate.
+
+`const fn` is used to define a [constant function][constant function]. Going back to `double_integer`, by making it a `const fn` we can now use
+it to set a constant, like so
+
+```rust
+const fn double_integer(value: i32) -> i32 {
+    value * 2
+}
+
+pub fn main() {
+    const DOUBLE_42: i32 = double_integer(42);
+    println!("{:?}", DOUBLE_42);
+}
+```
+
+The constraint for a constant function is that it must be able to be evaluated at compile time. If the function does not satify that constraint
+a compile error will result when using `const fn`, as exampled below
+
+```rust
+const fn multiply_integer(value: i32) -> i32 {
+    use std::time::SystemTime;
+    if SystemTime::now().elapsed().unwrap().as_nanos() == 0 { // this line errors
+        value * 2
+    } else {
+        value * 3
+    }
+// calls in constant functions are limited to constant functions, tuple structs and tuple variants
+}
+```
+
+Parameters can also be defined as functions. In this example we use the [`Fn`][Fn] operator to pass in the `double_integer` function to the
+`call_with_one` function. `call_with_one` takes a function with one `i32` parameter that returns an `i32`.
+
+```rust
+fn double_integer(value: i32) -> i32 {
+    value * 2
+}
+
+fn call_with_one<F>(func: F) -> i32
+where
+    F: Fn(i32) -> i32,
+{
+    func(1)
+}
+
+pub fn main() {
+    let num_doubled = call_with_one(double_integer);
+    println!("{:?}", num_doubled);
+}
+```
+
+Prints `2`
+
+A function can also be the return value from another function. In this example we return either of two functions from `get_log_function`.
+
+```rust
+fn write_to_database(msg: &str) {
+    // implemetation code snipped
+}
+
+fn write_to_file(msg: &str) {
+    // implemetation code snipped
+}
+
+fn is_db_up() -> bool {
+    true
+}
+
+fn get_log_function() -> Box<dyn Fn(&str)> {
+    if is_db_up() {
+        Box::new(write_to_database)
+    } else {
+        Box::new(write_to_file)
+    }
+}
+
+fn log<F>(log_fn: F, msg: &str)
+where
+    F: Fn(&str),
+{
+    log_fn(&msg);
+}
+
+pub fn main() {
+    let msg = "It happened";
+    log(get_log_function(), msg);
+}
+```
+
+There are anonymous functions which are covered in the `closures` topic.
+
+[statements or expressions]: https://doc.rust-lang.org/book/ch03-03-how-functions-work.html#function-bodies-contain-statements-and-expressions
+[return value]: https://doc.rust-lang.org/book/ch03-03-how-functions-work.html#functions-with-return-values
+[snake case]: https://en.wikipedia.org/wiki/Snake_case
+[constant function]: https://doc.rust-lang.org/reference/const_eval.html#const-functions
+[Fn]: https://doc.rust-lang.org/std/ops/trait.Fn.html
--- a/concepts/functions/about.md
+++ b/concepts/functions/about.md
@@ -91,8 +91,9 @@ pub fn main() {
 }
 ```
 
-The constraint for a constant function is that it must be able to be evaluated at compile time. If the function does not satify that constraint
-a compile error will result when using `const fn`, as exampled below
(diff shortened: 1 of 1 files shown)


concepts/functions/about.md
```diff
@@ -1,3 +1,176 @@
 # About
 
-TODO
+Functions are bodies of code containing one or more [statements or expressions][statements or expressions]. A function may optionally
+[return an expression][return value]. (Recall that a statement is an instruction that does not return a value. An expression evaluates to
+a resulting value.)
```
coriolinus, line 5:
```suggestion
[return an value][return value]. 
```

- Macros can return expressions, but functions can only return values.
- People who don't remember the distinction between statements and expressions can just click the link.

concepts/functions/about.md
```diff
@@ -1,3 +1,176 @@
...
+
+Rust style is to use [snake case][snake case] for a function name, which is prefaced by the `fn` keyword for a function definition. For example
+
+```rust
+fn do_nothing() {}
+
+```
+
+The empty parentheses are a placeholder for zero or more parameters. A parameter is the definition of a value to be passed into the function.
```
coriolinus, line 14:
```suggestion
The parentheses surround zero or more parameters, separated by commas. A parameter is a binding to a value of a particular type which is passed into the function.
```

Precision in wording.

concepts/functions/about.md
```diff
@@ -1,3 +1,176 @@
...
+Even if there are no parameters the parentheses are still required in the function definition. The actual value passed to a parameter is called
+an argument.
+
+The combination of parameters and return value is known as the function's signature. A function's signature requires that each parameter must
+have its type annotated.
+
+The curly braces enclose the body of the function. Inside the braces are where the function's statements and expressions are placed. Even if the
```
coriolinus, line 21:
```suggestion
The curly braces enclose the body of the function: its statements and expressions. Even if the
```

Remove some redundancy.

concepts/functions/about.md
```diff
@@ -1,3 +1,176 @@
...
+function contains no statements or expressions the curly braces are still required for the function definition. Of course there is no point to
+having an empty function.
```
coriolinus, line 23: > Of course there is no point to having an empty function.

That's not entirely true. Just about an hour ago, I was working with some code which (paraphrased) worked out to

```rust
impl VeryComplicatedStruct {
  fn create_with_mutate_intermediate_value(mutate_intermediate: impl FnOnce(&mut IntermediateStruct)) -> Self;
}
```

Given that signature, if you want to create a `VeryComplicatedStruct` _without_ actually changing the intermediate value, you need to pass an empty function: `VeryComplicatedStruct::create_with_mutate_intermediate_value(|_intermediate| {})`.

It's not worth adding that entire story to this document, but we should be careful about assumptions about what is useful.

```suggestion
function contains no statements or expressions the curly braces are still required for the function definition. 
```

concepts/functions/about.md
```diff
@@ -1,3 +1,176 @@
...
+
+A function could contain a simple statement. In the following example the function takes in one `i32` argument and prints it
```
coriolinus, line 25:
```suggestion
In the following example the function takes in one parameter of type `i32`, binds it to the name `value`, and prints it.
```

Ironically, `println!` is not a simple statement once you expand the macro.

concepts/functions/about.md
```diff
@@ -1,3 +1,176 @@
...
+
+```rust
+fn print_integer(value: i32) {
+    println!("{:?}", value);
+}
+```
+
+Note that the parameter needs to be given a name (`value` in this case) and be annotated with its type.
```
coriolinus, line 33:
```suggestion
Note the parameter's definition. Each parameter is defined in the format `name: Type`.
```

concepts/functions/about.md
```diff
@@ -1,3 +1,176 @@
...
+
+A function can also return an expression. In the following example the function takes in one `i32` argument and returns it doubled
```
coriolinus, line 35:
```suggestion
A function can also return a value. By default, the output of the final expression is returned. 
In the following example the function has one `i32` parameter and returns its double.
```

concepts/functions/about.md
```diff
@@ -1,3 +1,176 @@
...
+
+```rust
+fn double_integer(value: i32) -> i32 {
+    value * 2
+}
+```
+
+The `-> i32` indicates that the function returns an `i32`. Unlike parameters, the returned value is not named. Note that the last expression
+in a function can be returned without the `return` keyword and a semicolon. `return` _is_ needed to return early from a function, like so
```
coriolinus, line 44: Two concepts, two paragraphs.

```suggestion
The `-> i32` indicates that the function returns an `i32`. Unlike parameters, the returned value is not named. 

It is possible to exit from a function early with the `return` keyword, like so:
```

I want to instill in students the sense that the normal, expected pattern is to terminate a function with an expression returning the final value, and that the `return` keyword should only be used to exit early. That affects how we write about it.

concepts/functions/about.md
```diff
@@ -1,3 +1,176 @@
...
+    let num = Box::new(42);
+    let num_doubled = double_integer(num); // value moved here
+    println!("{:?}", num_doubled);
+    println!("{:?}", num); // borrow of moved value: `num` value borrowed here after move
+    // move occurs because `num` has type `Box<i32>`, which does not implement the `Copy` trait
+}
+```
+
+`main`, which owns the boxed `num` value, moves `num` as the argument for `value` into `double_integer`, which takes ownership of it. At the
+end of `double_integer`, `value` goes out of scope and is dropped, so `num` is no longer valid for the rest of the `main` function.
```
coriolinus, line 76: This information is good but belongs to a different concept: `scope`. Let's not teach it in `functions`.

concepts/functions/about.md
```diff
@@ -1,3 +1,176 @@
...
+
+Note that `fn main` is prefaced by `pub`, but `fn double_integer` is not. A `pub fn` is public, meaning it can be accessed from a different
+module. Functions defined without `pub` are private and can only be accessed from the current module. `pub(crate) fn` makes a function
+accessible only within the current crate.
```
coriolinus, line 80: This information is good but belongs to a different concept. Not sure which precisely; it could go into `scope`, but I can also see an argument for putting it into `module-system`, `visibility`, or something else. Either way we don't have to teach it here.

Also, it's somewhat odd to me to see `pub fn main`, because the `pub` is unnecessary; `clippy` doesn't say anything about it, but in most projects I've seen, it's just `fn main`.

concepts/functions/about.md
```diff
@@ -1,3 +1,176 @@
...
+
+`const fn` is used to define a [constant function][constant function]. Going back to `double_integer`, by making it a `const fn` we can now use
+it to set a constant, like so
```
coriolinus, line 83:
```suggestion
`const fn` is used to define a [constant function][constant function], which can be evaluated at compile time.
Going back to `double_integer`, by making it a `const fn` we can now use it to set a constant, like so
```

Add a note explaining what that can be good for.

concepts/functions/about.md
```diff
@@ -1,3 +1,176 @@
...
+}
+
+pub fn main() {
+    const DOUBLE_42: i32 = double_integer(42);
+    println!("{:?}", DOUBLE_42);
+}
+```
+
+The constraint for a constant function is that it must be able to be evaluated at compile time. If the function does not satify that constraint
+a compile error will result when using `const fn`, as exampled below
```
coriolinus, line 97:
```suggestion
Because constant functions may be evaluated at compile time, they have some restrictions that normal functions do not.
In particular, a `const fn` can only call other functions also marked as `const`. 
Failure to abide by that restriction will result in a compile error.
```

concepts/functions/about.md
```diff
@@ -1,3 +1,176 @@
...
+
+```rust
+const fn multiply_integer(value: i32) -> i32 {
+    use std::time::SystemTime;
+    if SystemTime::now().elapsed().unwrap().as_nanos() == 0 { // this line errors
+        value * 2
+    } else {
+        value * 3
+    }
+// calls in constant functions are limited to constant functions, tuple structs and tuple variants
```
coriolinus, line 107:
```suggestion
```

Tuple structs and enum variants are not functions and are not called.

concepts/functions/about.md
```diff
@@ -1,3 +1,176 @@
...
+}
+```
+
+Parameters can also be defined as functions. In this example we use the [`Fn`][Fn] operator to pass in the `double_integer` function to the
```
coriolinus, line 111: Everything from here to line 168 should be covered in a different concept, probably `fn-traits`. `functions` is one of the first concepts a student will encounter; we don't want to scare them off with detail that they're not ready for yet.

concepts/functions/about.md
bobahop: You know, when I typed that I suspected there was some corner case that would prove that wrong, so your observation is not  unexpected.

concepts/functions/about.md
efx: I'd propose `visibility` since we do not have a `concepts/module-system`.

Though, should we add a new concept `module-system`?

concepts/functions/about.md
bobahop: Of course.

concepts/functions/about.md
bobahop: Yeah, I thought it might be too much. Figured it could be copy/pasted somewhere else if needed.

Thanks for making this better!

concepts/functions/about.md
coriolinus: I see articles every once in a while whose premise is "Rust's module system is confusing; here's how it works." Anecdotally, I feel like we should explicitly teach the module system _somewhere_. I don't have strong feelings about where.

`scope`, `visibility`, and `modules` are very closely coupled concepts. Perhaps for now we should just try to teach all of them in `scope_and_visibility` or something, and then just split things out if that concept gets too big.

concepts/functions/about.md
```diff
@@ -1,3 +1,175 @@
...
 
-TODO
+Functions are bodies of code containing one or more [statements or expressions][statements or expressions]. A function may optionally
+[return an value][return value]. 
+
+Rust style is to use [snake case][snake case] for a function name, which is prefaced by the `fn` keyword for a function definition. For example
+
+```rust
+fn do_nothing() {}
+
```
coriolinus, line 10: Unnecessary blank line.

```suggestion
```

concepts/functions/about.md
bobahop: It's the error msg.

coriolinus: I don't think it's necessary to copy the error message in. Students who try it will see for themselves. 

From a pedagogical standpoint, it's better if we can say "this will fail because of blah blah" and explain the reason for the failure. We're not limited by the same constraints as the compiler; it's not usually hard to come up with a simpler explanation for what's gone wrong than it can provide.

concepts/functions/about.md
```diff
@@ -1,3 +1,93 @@
...
+    if (some_condition) {
+        return 0;
+    }
+    // more code snipped
+    42
+}
+```
+
+`const fn` is used to define a [constant function][constant function], which can be evaluated at compile time. Maing a function constant
+is a way of declaring that the function won't change in ways that are invalid for a `const fn`. Since this may prevent future optimizations,
```
coriolinus, line 58: > Maing a function constant is a way of declaring that the function won't change in ways that are invalid for a `const fn`.

- Typo: "Making" not "Maing"
- This sentence doesn't tell me anything. I recommend removing it.

concepts/functions/about.md
```diff
@@ -1,3 +1,93 @@
...
+making a function `const` should be considered with care. Going back to `double_integer`, by making it a `const fn` we can now use it to set a
```
coriolinus, line 59: > Since this may prevent future optimizations, making a function `const` should be considered with care.

I'm not sure this is accurate. Do you have an example of an optimization prevented by making a function `const`?

concepts/functions/about.md
bobahop: I'm sorry, which sentence? The one with the typo, or the one starting with "Since"?

coriolinus: The one I quoted, starting with the typo.

concepts/functions/about.md
bobahop: Got it from here a reddit discussion. Personally, I haven't had much use for `const fn`, so I'm somewhat at a loss to satisfy your request to add a note explaining what that can be good for. You'd be much better qualified to explain that than I.

https://www.reddit.com/r/rust/comments/iksmgk/psa_what_is_const_fn/#:~:text=const%20fn%20is%20a%20contract,think%20carefully%20before%20using%20it.

coriolinus: The point being made in that PR is that a `const` annotation is part of semantic versioning. If you later refactor a public `const fn` in a way which makes it impossible to implement as `const`, you're going to need to bump your major version to remove that annotation. I don't think that's a relevant concern here. AFAIK we don't even teach semantic versioning in this track.

bobahop: I'm happy to get rid of all of that. Do you have something to put in for what a `const fn` is good for?

coriolinus:
```rust
const fn compute_data_checksum() -> u128 {
  const DATA: &[u8] = include_bytes!("my_big_data_file");
  // checksum implementation is left as an exercise for the reader
}

/// This checksum is used to validate that the user has not tampered with proprietary configuration.
pub const DATA_CHECKSUM: u128 = compute_data_checksum();
```

coriolinus: Note that in that example, the only data that will actually appear in the produced binary is the checksum, not the big data file which produced it.

coriolinus (approved):
Thanks again for putting in the work of writing all this. I'm happy with the state of this now.

bobahop:
Thanks again for making this better.

I'd like to move on to the next one right away, but my wife and I have been feeling increasingly fatigued over the past few days. I plan to pick this up again when I've reclaimed some energy.

PR #1236, opened 2021-04-14
[references] update about.md

bobahop:
First draft. Kind of pointless to explain references without a bit of ownership and borrowing. Hopefully I didn't make a hash of it.

### DIFF
--- a/concepts/references/about.md
+++ b/concepts/references/about.md
@@ -1,7 +1,5 @@
 # About
 
-A [reference][reference] is a pointer: an address in memory where a value is stored.
-
 To understand references it is necessary to understand ownership. Every value has a single owner, which is in a scope. When the scope is exited,
 the value is dropped. For example:
 
--- a/concepts/references/about.md
+++ b/concepts/references/about.md
@@ -6,7 +6,7 @@ To understand references it is necessary to understand ownership. Every value ha
 
 ```rust
 fn log(msg: String) { // msg takes ownership of the value passed to it
-    let formatted_datetime: String;
+    let formatted_datetime = String::new();
     // code  to format current datetime snipped
     println!("{} at {}", msg, formatted_datetime);
 } // msg is dropped here
--- a/concepts/references/about.md
+++ b/concepts/references/about.md
@@ -62,6 +62,9 @@ fn check_shapes(constant: &[u8], linear: &[u8], superlinear: &[u8]) -> (bool, bo
     )
 }
 
+// understanding the implementations of the following functions is not necessary for this example
+// but are provided should you be interested
+
 fn is_constant(slice: &[u8]) -> bool { 
     slice
         .first()
--- a/concepts/references/about.md
+++ b/concepts/references/about.md
@@ -2,7 +2,8 @@
 
 A [reference][reference] is a pointer: an address in memory where a value is stored.
 
-To understand references it is necessary to understand ownership. Every value has a single owner, which is a scope. When the scope is exited, the value is dropped. For example:
+To understand references it is necessary to understand ownership. Every value has a single owner, which is in a scope. When the scope is exited,
+the value is dropped. For example:
 
 ```rust
 fn log(msg: String) { // msg takes ownership of the value passed to it
@@ -19,11 +20,13 @@ pub fn main() {
 ```
 
 A [reference][borrow] represents a _borrow_ of an owned value. Instead of transferring ownership of `my_string`, we can have `log` borrow it and
-then give it back to `my_string` at the end of `log`.
+then give it back to `my_string` in `main` at the end of `log`.
 
-To do this, we define `msg` as a reference by prefixing an ampersand. The type `&String` is read as "reference to a `String`". 
+To do this, we define `msg` as a reference by prefixing an ampersand. The type `&String` is read as "reference to a `String`".
 
-As `log`'s argument is now a reference, we need to pass a reference. To get a reference to `my_string`, prefix it with an ampersand, which is the [borrow operator](https://doc.rust-lang.org/reference/expressions/operator-expr.html#borrow-operators). This lends `my_string` to `log` for the duration of that function.
+As `log`'s argument is now a reference, we need to pass a reference. To get a reference to `my_string`, prefix it with an ampersand, which is
+the [borrow operator](https://doc.rust-lang.org/reference/expressions/operator-expr.html#borrow-operators). This lends `my_string` to `log`
+for the duration of that function.
 
 ```rust
 fn log(msg: &String) { //msg is defined as a reference with an ampersand
@@ -39,8 +42,73 @@ pub fn main() {
 }
 ```
 
-References are immutable, but they can be made mutable by appending `mut` to the ampersand. To access the value, the dereference operator (`*`)
-is prepended to the reference, as exampled below
+## References are immutable by default
+
+A value at any given time can have multiple references to it as long as they are all immutable, as shown below
+
+```rust
+fn check_shapes(constant: &[u8], linear: &[u8], superlinear: &[u8]) -> (bool, bool, bool) {
+    (
+        is_constant(constant),       // an immutable reference
+        is_linear(linear),           // another immutable reference, which is okay
+        is_superlinear(superlinear), // okay, all three references are immutable
+    )
+}
+
+fn main() {
+    let fibonacci = vec![1, 1, 2, 3, 5, 8, 13];
+    println!(
+        "{:#?}",
+        check_shapes(&fibonacci[..=1], &fibonacci[1..=3], &fibonacci[2..],)
+    );
+}
+
+fn is_constant(slice: &[u8]) -> bool { 
+    slice
+        .first()
+        .map(|first| slice.iter().all(|v| v == first))
+        .unwrap_or_default()
+}
+
+fn is_linear(slice: &[u8]) -> bool {
+    let diffs: Vec<_> = slice
+        .windows(2)
+        .map(|window| window[1] - window[0])
+        .collect();
+    is_constant(&diffs)
+}
+
+fn is_superlinear(slice: &[u8]) -> bool {
+    let diffs: Vec<_> = slice
+        .windows(2)
+        .map(|window| window[1] - window[0])
+        .collect();
+    diffs.windows(2).all(|window| window[1] > window[0])
+}
+```
+
+## References can be dereferenced
+
+Sometimes a reference needs to be dereferenced to access its value.
+
+```rust
+fn is_meaning_of_life(value: &u8) -> bool {
+  value == 42
+}
+// oops! compiler error: `==` operator is not defined for `&u8`, `u8`
+```
+
+To access the value, the dereference operator (`*`) is prepended to the reference, as shown below
+
+```rust
+fn is_meaning_of_life(value: &u8) -> bool {
+  *value == 42 // okay, the reference is dereferenced from `&u8` to `u8` which `==` can compare with another `u8`
+}
+```
+
+## References can be mutable
+
+References can be made mutable by appending `mut` to the ampersand. as shown below
 
 ```rust
 fn add_five(counter: &mut i32) { //counter is defined as a mutable reference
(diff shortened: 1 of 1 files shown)


concepts/references/about.md
```diff
@@ -1,3 +1,100 @@
...
+}
+
+pub fn main() {
+    let my_string = "Something happened".to_string();
+    log(&my_string); // my_string is passed as a reference with an ampersand
+    println!("{:?}", my_string);
+}
+```
+
+References are immutable, but they can be made mutable by appending `mut` to the ampersand. To access the value, the dereference operator (`*`)
```
coriolinus, line 41: You haven't demonstrated that there can be many simultaneous immutable references. You could do a lot with something like [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=2cd4e96879021b6994db21e021a7e660):

```rust
fn is_constant(slice: &[u8]) -> bool {
    slice
        .first()
        .map(|first| slice.iter().all(|v| v == first))
        .unwrap_or_default()
}

fn is_linear(slice: &[u8]) -> bool {
    let diffs: Vec<_> = slice
        .windows(2)
        .map(|window| window[1] - window[0])
        .collect();
    is_constant(&diffs)
}

fn is_superlinear(slice: &[u8]) -> bool {
    let diffs: Vec<_> = slice
        .windows(2)
        .map(|window| window[1] - window[0])
        .collect();
    diffs.windows(2).all(|window| window[1] > window[0])
}

fn check_shapes(constant: &[u8], linear: &[u8], superlinear: &[u8]) -> (bool, bool, bool) {
    (
        is_constant(constant),
        is_linear(linear),
        is_superlinear(superlinear),
    )
}

fn main() {
    let fibonacci = vec![1, 1, 2, 3, 5, 8, 13];
    println!(
        "{:#?}",
        check_shapes(&fibonacci[..=1], &fibonacci[1..=3], &fibonacci[2..],)
    );
}
```

coriolinus: Er, this note was meant to attach to line 61, not here, sorry.

concepts/references/about.md
```diff
@@ -1,3 +1,100 @@
 # About
 
-TODO
+A [reference][reference] is a pointer: an address in memory where a value is stored.
```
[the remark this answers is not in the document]
bobahop: I hear you. the reference does describe it that way: "For those familiar with pointers, a reference is just a pointer that is assumed to be aligned, not null, and pointing to memory containing a valid value of T ..." but I hav eno pro blem dropping that line.
https://doc.rust-lang.org/std/primitive.reference.html

concepts/references/about.md
bobahop: Yes, I felt lazy about not showing multi-immutables.  I like this basic structure of an example, but I wonder if the amount of code could be reduced to just snips for the is_X functions  so the volume of code isn't quite as overwhelming. Maybe use your line about leaving implementation as an exercise for the reader and returning a hard-coded bool instead?

coriolinus: I'll leave that to external opinion. @efx?

My take is that these are three very simple, short functions. A hard-coded value plus a comment in each would save fewer than 10 lines of code. The joke works better with the code present, in my opinion.

bobahop: My take is that it deflects focus from the point, which I think is well-represented in `check_shapes`, I mean, when I first saw it I thought, wow! that's a lotta code!

I'm thinking something like

```rust
fn is_constant(slice: &[u8]) -> bool {
    true // implementation is left as an exercise for the reader
}
```
Or maybe to keep it with your implementation, but to put them at the bottom, with `check_shapes` on the top, so it starts with the point, something like

```rust
fn check_shapes(constant: &[u8], linear: &[u8], superlinear: &[u8]) -> (bool, bool, bool) {
    (
        is_constant(constant), // an immutable reference
        is_linear(linear), // another immutable reference, which is okay
        is_superlinear(superlinear), // okay, all three references are immutable
    )
}
```

Yeah, I think that might be best. It keeps your implementation but places the "point" first, at the top.

concepts/references/about.md
```diff
@@ -1,3 +1,100 @@
...
+
+pub fn main() {
+    let my_string = "Something happened".to_string();
+    log(my_string); // passing my_string transfers ownership of its value to msg
+    //println!("{:?}", my_string); // uncommenting this line will error because my_string's value has been dropped while owned by msg
+}
+```
+
+A [reference][borrow] represents a _borrow_ of an owned value. Instead of having `msg` own `my_string`'s value, we can have `msg` borrow it and
+then give it back to `my_string` at the end of `log`.
```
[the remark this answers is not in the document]
bobahop: Yes, the scope is the owner in a way, but TRPL teaches that it's a variable: "Each value in Rust has a variable that’s called its owner." Perhaps a compromise of "then give it back to `my_string` in `main` at the end of `log.`"?

https://doc.rust-lang.org/book/ch04-01-what-is-ownership.html#ownership-rules

concepts/references/about.md
```diff
@@ -1,3 +1,167 @@
...
+A value at any given time can have multiple references to it as long as they are all immutable, as shown below
+
+```rust
+fn check_shapes(constant: &[u8], linear: &[u8], superlinear: &[u8]) -> (bool, bool, bool) {
+    (
+        is_constant(constant),       // an immutable reference
+        is_linear(linear),           // another immutable reference, which is okay
+        is_superlinear(superlinear), // okay, all three references are immutable
+    )
+}
```
coriolinus, line 54: This isn't actually where the overlap occurs. As far as `check_shapes` is concerned, its three parameters are all independent.

The overlap is in the invocation, where overlapping sub-slices of `fibonacci` are passed to `check_shapes`. The whole point of `check_shapes` is to demonstrate that the overlapping references can all live simultaneously, it's not just rustc getting clever about dropping the reference as soon as a particular check is complete.

concepts/references/about.md
efx: I appreciate the topic of these example functions as it assembles teaching overlapping references nicely.

I do feel like it would be too much code to read initially. 
I love the in-depth "how does this work in a discrete code sample" but think they should go in a later section in the about. 

As an aside: we could consider establishing a "How does this actually look and work"? section for various `about.md` documents.

I was also wondering about a student who is reading this document but may not be familiar with iterator adapters or the functional style of programming. In that case, it seems like the implementations of `is_constant` may distract.

concepts/references/about.md
bobahop: Yes, you're absolutely right. My stamina is flagging and I was careless. I'll try to fix it before I knock off for the day.

concepts/references/about.md
bobahop: @efx , how do you like it with the is_X shapes pushed to the bottom? Once I correct the comments might it be okay?

bobahop: Perhaps a comment can be put above the is_X functions something like "understanding the implementation of these functions is not necessary for this example but are provided should you be interested"?

efx: @bobahop Pushed to the bottom sounds great! 
The comment is nice too. We can always update as we get readers.

concepts/references/about.md
```diff
@@ -1,3 +1,168 @@
 # About
 
-TODO
+To understand references it is necessary to understand ownership. Every value has a single owner, which is in a scope. When the scope is exited,
```
coriolinus, line 3: > Every value has a single owner, which is in a scope.

I'm still not happy with that. Even if other documentation puts ownership at the variable level, we have an opportunity here to teach people more precisely what the language is doing.

- Values can have a **binding**. This is what is created with the `let`, `const`, and `static` statements. It identifies the value.
- Function parameters are also bindings which may or may not own the underlying value.
- `const` and `static` bindings can appear at the module level because they are valid for the entire lifetime of the program.
- `let` bindings can only appear within a defined scope.

What counts as a scope?

- functions, including methods and closures
- curly brace pairs which appear within functions

What happens at the end of a scope?

- References get returned to whatever previously held them. 
  - This may or may not be the owner, because a function holding a reference can pass that reference along to an internal call.
  - Due to non-lexical lifetimes, references are sometimes returned before the end of the scope. See [docs](https://doc.rust-lang.org/edition-guide/rust-2018/ownership-and-lifetimes/non-lexical-lifetimes.html).
- Owned items are freed
  - If the owned item implements `Drop`, that function is called first

The major point here is that the _scope_ is what governs when an item is freed, not the binding. A binding may be shadowed, but the shadowed value still lives until the end of the scope. 

Therefore, I strongly desire that we teach that the scope is the fundamental owner, not the binding.

```suggestion
To understand references it is necessary to understand ownership. Every value has a single owner, which is a scope. When the scope is exited,
```

bobahop: Just saw this comment. I'm gonna have to literally sleep on it. Of course the authors of TRPL are not infallible, but my first instinct is to maintain consistency with what is taken as closest to **the** reference of the language. But I'm really too tired to think about it right now.

bobahop: Sorry you had to wait so long for a reply. I seem to be fighting something off that may be more than just fatigue, plus I got my first jab a couple days ago.

It's true that ownership can't exist without bindings _and_ scope. But it seems that the non-lexical lifetime example in the docs indicates that ownership more particularly follows a binding and how it is used than scope per se. As the last line of the "Non-lexical lifetimes" section says: "Instead of pointing to where y goes out of scope, it shows you where the conflicting borrow occurs." That's why we can have two mutable borrows in the same scope, as long as they don't conflict with each other. It has more to do with how two the bindings interact in time than what scope they're in. If I bring a ball onto a playground you could say the playground contains the ball, but I'm holding the ball. If I give the ball to someone else they now hold the ball. It happens in the playground, and the ball is in the playground, but it's the people who hold the ball.

coriolinus: No worries; we're all doing this in what free time we can spare. It's certainly not unreasonable to take half a day to think!

I wrote an [example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=6ce7dfa04bd9a1b81a174ce22564c272) of what I'm talking about. NLL only applies to references within a scope. The ball in playground metaphor is nice, and mostly works most of the time, but it doesn't accurately describe what actually happens.

efx: I'm only skimming this discussion so my apologies for not weighing in.

I am curious about the example.

Is the block scope of `fn main` being torn down the reason why the shadowed variable `"second"` is dropped first, followed by the `"first"`? like Rust is unwinding or something?

efx: @bobahop feel better! 

To the core work here: how do @coriolinus and @bobahop you both feel about approving and merging the current work?

I can see both sides to the discussion. I would vote for one or more follow-up pull requests to clarify our teaching semantics.

coriolinus: I mean, there's a reason I've been using the "comment" feature instead of the "request changes" feature: I don't mean to actively block progress if you (or another maintainer) feels that I'm being obstreperous here. I've intentionally not actively vetoed the current state of things. However, I will not approve it until I feel it is correct.

bobahop: I better understand where you're coming from with your `Drop` example. 

I was intrigued by this statement in the [Drop](https://doc.rust-lang.org/std/ops/trait.Drop.html#you-cannot-call-dropdrop-yourself) reference that `Drop::drop does not take ownership of its input`, and wondered how relevant to ownership `Drop` is vs. its relevance to lifetime and scope (although all three are intertwined.)

I'm not sure I'm sold that if the binding were the owner, we'd see dropping first and then dropping second. I can interpret that as a variable stacking its shadow, so when it goes out of scope, the shadow owner pops off first and then the original owner. But I freely admit I don't know how it's implemented under the covers.

Well, I don't want this to be filioque-type schism here.

I wonder if you'd be okay with `Every value has a single owner. When its scope is exited, the value is dropped.` It only mentions scope and not `variable`, as the TRPL does in `Each value in Rust has a variable that’s called its owner.`

coriolinus: As I understand it, the reason `Drop` doesn't take ownership of its input is to ensure that it's never recursive. 

I'm fine with that proposed phrasing.

coriolinus (approved):
Ok, thanks for putting in all this work! This may be the toughest concept yet written in the track.

bobahop:
You're very welcome! I'm no expert in Rust; I've just explained some simple concepts over and over again for a few months.

I might take the weekend to go back to my Unity course before deciding on the next topic to take on.

PR #1240, opened 2021-04-18
Fizzy tests: take finite number of elements

kuronyago:


coriolinus:
What is the motivation for this change?

kuronyago:
I think we should pass an iterator of a certain size

coriolinus:
Why?

kuronyago:
fn fizz_buzz<T>() -> Fizzy<T>
how can I use this function signature to figure out how to control the iterator?

kuronyago:
`let got = fizz_buzz::<Fizzable>()
        .apply(std::iter::successors(Some(Fizzable(1)), |prev| {
            Some(*prev + 1.into())
        }))
        .take(16)
        .collect::<Vec<_>>();`

and why should Fizzy worry about u8 overflow?

coriolinus:
> how can I use this function signature to figure out how to control the iterator?

Don't worry about the length of the input iterator; just apply to all elements. It is not `apply`'s job to care about how, or whether, the input iterator ever terminates.

> why should Fizzy worry about u8 overflow?

It shouldn't. 

The point of `test_minimal_generic_bounds` is to ensure that `Fizzy` can handle any type implementing the following traits: `'static, Clone, Copy, Debug, Default, PartialEq, From<u8>, Display, Add<Output=Self>, Rem<Output=Self>`. Nothing about that test requires or implies handling `u8` overflow; the test already takes only the first 16 elements from the input iterator.

Student solutions should be capable of handling theoretically infinite iterators; they should work on each element independently. Giving them only finite iterators would cause implementations to pass the test suite which use suboptimal algorithms, such as attempting to collect all results, then returning them one by one. Given that it's [possible](https://github.com/exercism/rust/blob/main/exercises/practice/fizzy/.meta/example.rs) to handle each item independently, I'm not interested in broadening the range of acceptable solutions to include those which collect all the solutions first.

kuronyago:
`.apply(std::iter::successors(Some(Fizzable(1)), |prev| {
            Some(*prev + 1.into())
        }))`

what should happen at step 255?

coriolinus:
Nothing: the iterator will never request more than 16 items, because of the `take(16)` in

https://github.com/exercism/rust/blob/b946cdbccf24b856add18d6c59d2ba0f88777f59/exercises/practice/fizzy/tests/fizzy.rs#L117-L121

If the `apply` method performs more than 16 calls to its interior iterator, then it is implemented wrong.

kuronyago:
> Don't worry about the length of the input iterator; just apply to all elements. It is not `apply`'s job to care about how, or whether, the input iterator ever terminates.

> If the `apply` method performs more than 16 calls to its interior iterator, then it is implemented wrong.

does the implementation of the apply function meet these conditions?

`    pub fn apply<I>(self, items: I) -> impl Iterator<Item = String>
    where
        I: Iterator<Item = T>,
    {
        let mut result: Vec<String> = Vec::new();

        for item in items {
            let mut step: String = String::new();

            for matcher in self.matchers.iter() {
                if let Some(n) = matcher.run(item) {
                    step.push_str(n.as_str())
                }
            }

            if step.is_empty() {
                result.push(item.to_string());
            } else {
                result.push(step);
            }
        }

        result.into_iter()
    }`

coriolinus:
> does the implementation of the apply function meet these conditions?

That implementation is wrong. It does not operate as a streaming iterator adapter. Instead, it attempts to consume the input iterator entirely. This pattern is wrong because it attempts to consume the entire input iterator before producing any output. As you've noted, this can cause overflow issues and other problems for infinite input iterators. For this problem, any solution that ends with `result.into_iter()` is wrong. 

Instead of consuming the entire iterator, operate on each item independently:

```rust
items.map(|item| {
    // your implementation here
})
```

That pattern will process each item independently, but will not have any trouble with infinite input iterators.

PR #1241, opened 2021-04-18
add detailed description of `Box`'s implementation to appropriate concept links

coriolinus:
A useful command to show non-empty `links.json` example:

```sh
fd links.json concepts --size +4b | xargs bat
```


coriolinus:
This link might also be relevant for the references article we worked on recently.

efx (approved):

PR #1250, opened 2021-04-21
Add RPN Calculator exercise

### LINKED ISSUE #1104: vec-stack
> For concept exercises it seems we should be razor focused.

Proposal: to keep things simple, let's get focused. Split the `vec` concept into these concepts:

- `vec-macro` for creating empty vectors, and vectors of particular size, with`vec!`
- `vec-collect` for creating vectors via `collect()` (requires `iter`)
- `vec-stack` for using vectors as an `O(1)` stack (push and pop methods)
- `vec-index` for `O(1)` random access (incl. mutable) within a vector

To avoid overloading ourselves, let's try for these bounds:

- Not more than 15 lines of example implementation
- Not more than 3 paragraphs of documentation
- Not more than 5 tests

_Originally posted by @coriolinus in https://github.com/exercism/v3/issues/2918#issuecomment-756743573_

cwhakes:
I am a lot happier with this exercise for `vec-stack` than Final Electric Bill (#1248). However, it may be less guided than Concept exercises should be, as well as having a bunch of prerequisites.

Supersedes #1248, closes #1104.

### DIFF
--- a/config.json
+++ b/config.json
@@ -129,6 +129,25 @@
         "status": "active"
       },
       {
+        "slug": "rpn-calculator",
+        "name": "RPN Calculator",
+        "difficulty": 4,
+        "uuid": "25cc722b-211d-4271-9381-fdfe16b41301",
+        "concepts": [
+          "vec-stack"
+        ],
+        "prerequisites": [
+          "intro-fn",
+          "numbers",
+          "option",
+          "enums",
+          "match-basics",
+          "loops",
+          "mutability"
+        ],
+        "status": "wip"
+      },
+      {
         "slug": "csv-builder",
         "name": "CSV builder",
         "difficulty": 1,
--- a/config.json
+++ b/config.json
@@ -145,7 +145,7 @@
           "loops",
           "mutability"
         ],
-        "status": "wip"
+        "status": "active"
       },
       {
         "slug": "csv-builder",
--- a/exercises/concept/rpn-calculator/.docs/instructions.md
+++ b/exercises/concept/rpn-calculator/.docs/instructions.md
@@ -35,6 +35,10 @@ In both cases, the expression evaluates to 6.
 ## 2. Example
 
 Lets manually evaluate that complex expression.
+As we learned in the introduction, evaluation of RPN requires a stack.
+This stack is used to hold numeric values that the operators operate on.
+We start our calculator with an empty stack and then evaluate each element one at a time.
+
 First, we encounter a `4`,
 so we push it onto our freshly created stack.
 
--- a/exercises/concept/rpn-calculator/.docs/instructions.md
+++ b/exercises/concept/rpn-calculator/.docs/instructions.md
@@ -1,0 +1,42 @@
+# Instructions
+
+Reverse Polish notation (RPN) is a way of writing mathematical expressions in which operators follow their operands.
+For example, instead of writing:
+```
+2 + 2
+```
+you would write:
+```
+2 2 +
+```
+A major benefit of Reverse Polish notation is that it eliminates the need for parentheses in complex expressions.
+```
+(4 + 8) / (7 - 5)
+```
+can be written as
+```
+4 8 + 7 5 - /
+```
+In both cases, the expression evaluates to 6.
+
+Your goal is to write a calculator to evaluate a list of inputs ordered by Reverse Polish notation.
+You are given the following enum and stubbed function as a starting point.
+
+```rust
+#[derive(Debug)]
+enum CalculatorInput {
+    Add,
+    Subtract,
+    Multiply,
+    Divide,
+    Value(i32),
+}
+  
+pub fn evaluate(inputs: &[CalculatorInput]) -> Option<i32> {
+    unimplemented!(
+		"Given the inputs: {:?}, evaluate them as though they were a Reverse Polish notation expression",
+		inputs,
+	);
+}
+```
+
--- a/exercises/concept/rpn-calculator/.docs/instructions.md
+++ b/exercises/concept/rpn-calculator/.docs/instructions.md
@@ -2,21 +2,29 @@
 
 Reverse Polish notation (RPN) is a way of writing mathematical expressions in which operators follow their operands.
 For example, instead of writing:
+
 ```
 2 + 2
 ```
+
 you would write:
+
 ```
 2 2 +
 ```
+
 A major benefit of Reverse Polish notation is that it eliminates the need for parentheses in complex expressions.
+
 ```
 (4 + 8) / (7 - 5)
 ```
+
 can be written as
+
 ```
 4 8 + 7 5 - /
 ```
+
 In both cases, the expression evaluates to 6.
 
 Your goal is to write a calculator to evaluate a list of inputs ordered by Reverse Polish notation.
--- a/exercises/concept/rpn-calculator/.docs/instructions.md
+++ b/exercises/concept/rpn-calculator/.docs/instructions.md
@@ -1,6 +1,9 @@
 # Instructions
 
-Reverse Polish notation (RPN) is a way of writing mathematical expressions in which operators follow their operands.
+## 1. Overview
+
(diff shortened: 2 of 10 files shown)


exercises/concept/rpn-calculator/.docs/instructions.md
```diff
@@ -0,0 +1,42 @@
+# Instructions
+
+Reverse Polish notation (RPN) is a way of writing mathematical expressions in which operators follow their operands.
+For example, instead of writing:
+```
+2 + 2
+```
```
coriolinus, line 7:
~~~suggestion

```
2 + 2
```

~~~

Markdown formatting: it's easiest when code blocks are surrounded by blank lines. Applies to the other blocks in this document also.

cwhakes:
An alternative template could be something like:

```
#[derive(Debug, PartialEq)]
RpnCalculator(Vec<i32>);

impl RpnCalculator {
    pub fn number(&mut self, number: i32) {}
    pub fn add(&mut self) {}
    pub fn subtract(&mut self) {}
    pub fn multiply(&mut self) {}
    pub fn divide(&mut self) {}
}
```

Which would allow more fine-grained help

exercises/concept/rpn-calculator/.docs/introduction.md
```diff
@@ -0,0 +1,6 @@
+# Introduction
+
+Rust's vector implementation, [`std::vec::Vec`](https://doc.rust-lang.org/std/vec/struct.Vec.html), can be used to simulate a stack by using its `push()` and `pop()` methods.
```
coriolinus, line 3:
```suggestion
Rust's vector implementation, [`std::vec::Vec`](https://doc.rust-lang.org/std/vec/struct.Vec.html), can be used as a stack by using its `push()` and `pop()` methods.
```

It's not really simulating it; it's really a stack.

cwhakes: Fixed in 7f6a9a4

exercises/concept/rpn-calculator/.docs/instructions.md
cwhakes: Fixed in 7f6a9a4

coriolinus:
I prefer keeping the function-based design instead of making an `RpnCalculator` struct with individual methods. As a function accepting a list, it's much more obvious to the student how the RPN theory corresponds to the practice of their implementation.

exercises/concept/rpn-calculator/.docs/introduction.md
```diff
@@ -0,0 +1,6 @@
+# Introduction
```
coriolinus, line 1: Probably worth adding a sentence to this section about what a stack is: like a stack of paper, the last items in are the first ones out.

exercises/concept/rpn-calculator/tests/rpn-calculator.rs
```diff
@@ -0,0 +1,72 @@
...
+fn test_empty_input_returns_none() {
+    let input = vec![];
+    assert_eq!(evaluate(&input), None);
+}
+
+#[test]
+#[ignore]
+fn test_simple_value() {
+    let input = vec![Value(10)];
+    assert_eq!(evaluate(&input), Some(10));
```
coriolinus, line 14: It'll be much easier to read and write the tests if you write a `fn calculator_input(&str) -> Vec<CalculatorInput>` and use it throughout.

exercises/concept/rpn-calculator/.docs/introduction.md
efx: FWIW, the [book](https://doc.rust-lang.org/book/ch04-01-what-is-ownership.html#the-stack-and-the-heap) uses the metaphor of plates to discuss the stack. Perhaps we do that here too for consistency?

efx:
This looks like a fun exercise! Thank you for contributing @cwhakes.

efx:
> I prefer keeping the function-based design instead of making an RpnCalculator struct with individual methods. 

I concur with @coriolinus. As someone still learning Rust I find it easier to ingest functions over struct based design patterns. Just my two cents!

exercises/concept/rpn-calculator/tests/rpn-calculator.rs
cwhakes: Fixed in aaf3df4

cwhakes:
I'll get started on improving the Introduction (and other docs). I admittedly phoned it in because I was unsure if this was the best direction.

exercises/concept/rpn-calculator/.meta/exemplar.rs
```diff
@@ -0,0 +1,47 @@
...
+            Value(value) => stack.push(*value),
+        }
+    }
+
+    let output = stack.pop();
+    if stack.is_empty() {
+        output
+    } else {
+        None
+    }
```
coriolinus, line 46:
```suggestion
    stack.is_empty().then(|| output)
```

Not essential, but fun: `bool` has a [`then` combinator](https://doc.rust-lang.org/std/primitive.bool.html#method.then) to simplify cases like these.

exercises/concept/rpn-calculator/src/lib.rs
```diff
@@ -0,0 +1,19 @@
+// This stub file contains items which aren't used yet; feel free to remove this module attribute
+// to enable stricter warnings.
+#![allow(unused)]
```
coriolinus, line 3: Please remove this. The typical pattern we encourage is to put function parameters into the `unimplemented` section to ensure they're used, or worst case, to prepend a `_` to the name. `#[allow(unused)]` for the whole module is too broad.

cwhakes: That was included because of the CI was not liking the enum. Now that I think of it, it may have been because I forgot to make it `pub`.

Fixed in c413d4e

exercises/concept/rpn-calculator/.meta/exemplar.rs
cwhakes: I am a big fan of `then()`, but I didn't want to have more prerequisites. Should this be taught in `role-playing-game` (the `option` concept exercise).

coriolinus: AFAIK we are allowed to do things in the exemplar that are not covered by the prerequisites. The prerequisites are what is necessary to implement a simple solution, but we aren't restricted by them. 

However, I'm not perfectly confident about my understanding of that policy. @efx what do you think?

exercises/concept/rpn-calculator/.docs/introduction.md
cwhakes: Should be resolved by 2fae730 & ed5ed0d

exercises/concept/rpn-calculator/.meta/exemplar.rs
cwhakes: https://github.com/exercism/v3/blob/main/docs/concept-exercises.md#exemplar-implementation-file

> The implementation should only use language features introduced by the exercise or its prerequisites (and their prerequisites, and so on).

Looks like we shouldn't use things not previously covered.

coriolinus: Thanks for finding the reference! I was wrong; let's keep it as it currently is.

config.json
```diff
@@ -128,6 +128,25 @@
...
+        "prerequisites": [
+          "intro-fn",
+          "numbers",
+          "option",
+          "enums",
+          "match-basics",
+          "loops",
+          "mutability"
+        ],
+        "status": "wip"
```
coriolinus, line 148: I don't think we need to leave this WIP at this piont; may as well switch it to active now.

coriolinus (approved):
Looks great; the recent documentation in particular should be very helpful for students. Thanks for putting in all this work! I won't merge it immediately, mainly so that you can update the status.

PR #1252, opened 2021-04-25
Add test case to rust/anagram to prevent ascii-sum-solutions

apanloco:
Prevent solutions that use the sum of the ascii values of each char
to find anagrams.

### DIFF
--- a/exercises/practice/anagram/tests/anagram.rs
+++ b/exercises/practice/anagram/tests/anagram.rs
@@ -172,3 +172,15 @@ fn test_same_bytes_different_chars() {
 
     process_anagram_case(word, &inputs, &outputs);
 }
+
+#[test]
+#[ignore]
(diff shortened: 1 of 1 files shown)


coriolinus (approved):
I'd be startled to see an implementation which passed all other tests but failed this, but this test makes it absolutely clear that summing over ascii values is an improper strategy. Have you considered also upstreaming this test case to https://github.com/exercism/problem-specifications/?

apanloco:
This implementation fails only this :)
https://exercism.io/tracks/rust/exercises/anagram/solutions/4ceb938349304e4cbb081f1cd2c20775

apanloco:
no, i haven't considered doing that... I'll look into it.

bobahop:
If you do the upstreaming, make sure you use a version 4 UUID. I got bit on that one before.

PR #1263, opened 2021-05-12
Fix configlet lints

coriolinus:
Configlet implemented several new lints recently, and it turns out that this track hadn't previously been in compliance. This PR should fix the deficiencies.

- deletes several unimplemeted todo-only concepts, and lists them in `concepts/planned` instead.
- updates the test files definitions to kebab-case instead of snake-case.
- adds attribution (based on git history) to several concepts.
- adds blurbs to several concepts.


ErikSchierboom (approved):
LGTM!

Issue #1266, opened 2021-05-21
Two-Fer starts with a file that already satisfies all tests


wischi-chr:
I'm not sure if this is a problem with the exercise, v3 or even not a bug at all.

https://exercism.lol/tracks/rust/exercises/two-fer/edit
https://github.com/exercism/rust/tree/main/exercises/practice/two-fer

After starting the exercise there is basically nothing left to do besides pressing F2 ("run tests")

coriolinus:
Two-fer is deprecated and should not be visible in the v3 website:

https://github.com/exercism/rust/blob/519032f2a56902b552d9524fcc6285b21a8b858e/config.json#L1419-L1431

There was a problem, years ago, which broke things if we removed it entirely, so deprecation was the best we could do.

ErikSchierboom:
I think it will be visible to students that started the exercise when it was still not deprecated, right @iHiD?

iHiD:
It will. That's not really a problem in itself as the nature of having starting it means they'll probably have a solution. However, because we have the "reset exercise" functionality in v3, where a student can go back to the start of an exercise if they get stuck, this does become important.

@coriolinus Is there an old stub we can revert to? Or if not, we should probably just make this stub an empty file.

coriolinus:
Previous discussion: https://github.com/exercism/rust/pull/832. 

I note that the exercise has been deprecated for over two years. It seems unlikely to me that any student has opened but not completed that exercise in that span.

See also: https://github.com/exercism/rust/pull/773#issuecomment-490643904, https://github.com/exercism/rust/issues/564, https://github.com/exercism/rust/issues/547, https://github.com/exercism/rust/issues/552.

iHiD:
As there are 22 people in that group, I've opened https://github.com/exercism/rust/pull/1267

PR #1267, opened 2021-05-22
Remove the solution from the stub file of deprecated two-fer

iHiD:
Fixes https://github.com/exercism/rust/issues/1266

We currently have students (22 to be precise) who have an in progress TwoFer. Although it is deprecated it should still work as an exercise for those students. This aims to fix the fact that the solution is provided in the stub by providing a minimal compiling solution.

As I've never written Rust and don't have it installed, I'm entirely guessing at the syntax from a quick Google. If CI passes then I think this is fine. If not, I'll try and iterate until it does compile, but obviously if someone with Rust skills wants to just fix it, that's fine too 🙂

### DIFF
--- a/exercises/practice/two-fer/src/lib.rs
+++ b/exercises/practice/two-fer/src/lib.rs
@@ -1,6 +1,3 @@
 pub fn twofer(name: &str) -> String {
-    match name {
-        "" => "One for you, one for me.".to_string(),
-        _ => format!("One for {}, one for me.", name),
-    }
+    "".to_string()
 }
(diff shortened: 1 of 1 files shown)


exercises/practice/two-fer/src/lib.rs
```diff
@@ -1,6 +1,3 @@
 pub fn twofer(name: &str) -> String {
-    match name {
-        "" => "One for you, one for me.".to_string(),
-        _ => format!("One for {}, one for me.", name),
-    }
+    "".to_string()
```
coriolinus, line 2:
```suggestion
    unimplemented!("how many for {}", name)
```

iHiD: Great. Thanks.

coriolinus (approved):
Feel free to merge when ready.

iHiD:
Great. Thanks @coriolinus.

PR #1270, opened 2021-05-28
🤖 Remove `online_editor.ace_editor_language` key

ErikSchierboom:
We've recently switched from Ace to CodeMirror as our editor. This PR removes the now obsolete `.online_editor.ace_editor_language` key.

## Tracking

https://github.com/exercism/v3-launch/issues/40

### DIFF
--- a/config.json
+++ b/config.json
@@ -13,7 +13,6 @@
   "online_editor": {
     "indent_style": "space",
     "indent_size": 4,
-    "ace_editor_language": "rust",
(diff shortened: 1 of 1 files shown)


coriolinus (approved):
Approving, but please release a configlet version which doesn't complain about this key being missing soon. All CI is going to break until configlet is updated.

ErikSchierboom:
@coriolinus We're planning to release an update later today.

PR #1283, opened 2021-07-07
clippy: remove needless borrows

petertseng:
This was made an error in Clippy recently, so we'll need to fix it if we
want CI to work.

According to https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow

What it does:

Checks for address of operations (`&`) that are going to be dereferenced
immediately by the compiler.

Why is this bad:

Suggests that the receiver of the expression borrows the expression.

Commentary as it applies to the Rust track:

I can get behind the idea that it's better not to mislead the reader of
the code about how many levels of reference the called function needs.

Most of these stem from a misunderstanding of what expressions were
already borrowed. They are listed below. Note carefully inconsistencies
introduced in sublist and tournament, and consider whether we have a
solution for this inconsistency before approving. The concern would be
any confusion it may cause to those learning Rust. It is consistent from
the standpoint of the type we're ultimately passing, but it's
inconsistent from the standpoint of what the call sites look like.

* Dominoes: The input is `&[Domino]`, and check takes `&[Domino]`.
  calling `check(&input)` passes `&&[Domino]` whereas we should call
  `check(input)` to pass `&[Domino]`
* DOT DSL: `iter` iterates over `&T`. Therefore the `name` is a `&&str`
  (recall that string literals in Rust produce `&str`) and we are
  iterating over a `[&str]` producing `&&str`; since `Node::new` takes
  `&str`, just `name` would be better than `&name`, which would pass
  `&&&str`
* grep: The patterns are string literals such as `"Agamemnon"`. Recall
  that string literals in Rust produce `&str`, and grep takes `&str`, so
  no additional `&` should be added. In fact, we should do this for the
  macro too, even though Clippy does not seem to catch those.
* pangram: Recall that string literals in Rust produce `&str`, and
  is_pangram takes `&str`, so no additional `&` should be added.
* sublist: The `v` was declared as `&[u32]`, and `sublist` takes two
  `&[T]`, so no additional `&` should be added. Inconsistency
  introduced: `sublist` is called with `&` in all other instances in
  this file because they are either slice literals or Vec, which do
  require the `&`.
* tournament: Recall that string literals in Rust produce `&str`, and
  `tally` takes `&str`, so no additional `&` should be added.
  Inconsistency introduced: The first four cases use a string literal
  and thus do not require `&`. The next cases use `String` (recall that
  we decided that this was the best way to show the multi-line strings
  in https://github.com/exercism/rust/pull/152#discussion_r69383715),
  and therefore require `&`. Consider carefully whether this may be
  confusing to students that some require `&` and some not.

### DIFF
--- a/exercises/practice/dominoes/tests/dominoes.rs
+++ b/exercises/practice/dominoes/tests/dominoes.rs
@@ -65,7 +65,7 @@ fn check(input: &[Domino]) -> CheckResult {
 }
 
 fn assert_correct(input: &[Domino]) {
-    match check(&input) {
+    match check(input) {
         Correct => (),
         GotInvalid => panic!("Unexpectedly got invalid on input {:?}", input),
         ChainingFailure(output) => panic!(
--- a/exercises/practice/dot-dsl/tests/dot-dsl.rs
+++ b/exercises/practice/dot-dsl/tests/dot-dsl.rs
@@ -130,7 +130,7 @@ fn test_graph_stores_attributes() {
         &["a", "b", "c"]
             .iter()
             .zip(attributes.iter())
-            .map(|(name, &attr)| Node::new(&name).with_attrs(&[attr]))
+            .map(|(name, &attr)| Node::new(name).with_attrs(&[attr]))
             .collect::<Vec<_>>(),
     );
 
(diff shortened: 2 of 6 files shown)


petertseng:
I have determined that if we want to have some consistency in tournament, I originally wanted to avoid multiline string literals having indentation similar to:

```rust
fn test() {
    let scores = "header
line1
line2";
}
```

Perhaps https://doc.rust-lang.org/std/macro.concat.html could be an alternative.

coriolinus (approved):

PR #1285, opened 2021-07-07
🤖 Simplify `hello-world` stub

ErikSchierboom:
Simplify the `hello-world` exercise's stub. The goal of this is to make things easier for students to get started and to introduce as little syntax as possible.

See https://github.com/exercism/v3-launch/issues/46

### DIFF
--- a/exercises/practice/hello-world/src/lib.rs
+++ b/exercises/practice/hello-world/src/lib.rs
@@ -1,5 +1,3 @@
-// The &'static here means the return type has a static lifetime.
-// This is a Rust feature that you don't need to worry about now.
 pub fn hello() -> &'static str {
-    "Goodbye, World!"
+    "Goodbye, Mars!"
 }
--- a/exercises/practice/hello-world/src/lib.rs
+++ b/exercises/practice/hello-world/src/lib.rs
@@ -1,3 +1,4 @@
+// &'static is a "lifetime specifier", something you'll learn more about later
 pub fn hello() -> &'static str {
     "Goodbye, Mars!"
 }


exercises/practice/hello-world/src/lib.rs
```diff
@@ -1,5 +1,3 @@
-// The &'static here means the return type has a static lifetime.
-// This is a Rust feature that you don't need to worry about now.
 pub fn hello() -> &'static str {
-    "Goodbye, World!"
+    "Goodbye, Mars!"
```
ErikSchierboom, line 2: We're standardizing this text. See https://github.com/exercism/v3-launch/issues/46

exercises/practice/hello-world/src/lib.rs
```diff
@@ -1,5 +1,3 @@
-// The &'static here means the return type has a static lifetime.
-// This is a Rust feature that you don't need to worry about now.
```
ErikSchierboom, line 2: I've currently omitted this comment which, while helpful, might not even be that useful for the student. It might be easier for the student to just see the code without the comment. But if you disagree, let me know.

coriolinus: The intent of that comment is to reduce syntax anxiety. If you're used to something like

```python
def hello():
    return "Goodbye, Mars!"
```

then you can probably figure out what `-> ... str` means, but `&'static` feels a bit arcane. I think the note is valuable just to let the student know that they don't need to worry about it for now.

petertseng: Are we able to find a small number of individuals learning Rust and ask them directly whether they would prefer to see this comment or not to see it?

It may otherwise be difficult to speculate.

If we had all the supporting infrastructure to do so, I would have suggested some A/B test where some students see the comment and some don't, but I don't think we are in a position to do that.

I understand that we want to make the exercise as simple as possible - it is currently unavoidable that we put `&'static` there; now the question is how we deal with its necessary presence (should we explain it to preemptively answer any student questions, or should we omit an explanation and expect that students will be able to comfortably ignore it)

ErikSchierboom: @sleeplessbyte @thelostlambda IIRC you both know a bit of Rust. What do you think?

TheLostLambda: I think we could shorten it but keep a note, saying something like: `&'static` is a "lifetime specifier", something you'll learn more about later

That way they know they don't need to understand it right now, but it's also not assuming they know what a "static lifetime" means? That's just my two cents I suppose. It could also be fine without any comments though if we moved that "disclaimer" to the introduction file

ErikSchierboom: @coriolinus @petertseng What do you both think of the above suggestion by @TheLostLambda?

coriolinus: Sure, that's fine.

exercises/practice/hello-world/src/lib.rs
```diff
@@ -1,5 +1,3 @@
...
 pub fn hello() -> &'static str {
```
ErikSchierboom, line 1:
```suggestion
// &'static is a "lifetime specifier", something you'll learn more about later
pub fn hello() -> &'static str {
```

ErikSchierboom: @coriolinus Any remarks on this suggested text?

ErikSchierboom: 👍

coriolinus (approved):

ErikSchierboom:
Thanks all!

PR #1293, opened 2021-07-29
Clarify the instructions for length zero arrays in the Rust proverbs exercise.

ojacobson-sfdc:
This is based on feedback from a colleague trying to learn Rust via exercism. He noted that it wasn't at all clear what was expected following "arbitrary" arrays. There's no singlar obvious right answer, but the test suite is clear about what it expects.

### DIFF
--- a/exercises/practice/proverb/.docs/instructions.md
+++ b/exercises/practice/proverb/.docs/instructions.md
@@ -14,4 +14,4 @@ For want of a battle the kingdom was lost.
 And all for the want of a nail.
 ```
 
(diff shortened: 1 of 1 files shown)


coriolinus (approved):
I'm approving this on the principle that making the docs more explicit is never a bad thing. Still, for the life of me, I'm having a hard time figuring out what other output might be expected given an empty input array.

Thanks for contributing!

ojacobson-sfdc:
Thanks for the merge!

>  Still, for the life of me, I'm having a hard time figuring out what other output might be expected given an empty input array.

I _tend_ to agree, but I actually helped someone informally with this exercise, and that's something they were stuck on. They resorted to trial and error and figured it out that way.

PR #1296, opened 2021-08-09
Add strong testcase in sublist problem

vagrawal:
Most of the solutions to the sublist problem focuses on premature optimizations but are `O(n^2)`. I tested few accepted solutions and those code couldn't finish this case within few minutes.

### DIFF
--- a/exercises/practice/sublist/tests/sublist.rs
+++ b/exercises/practice/sublist/tests/sublist.rs
@@ -78,6 +78,16 @@ fn huge_sublist_not_in_huge_list() {
 
 #[test]
 #[ignore]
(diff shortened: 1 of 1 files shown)


coriolinus:
Good idea! The naive solution is definitely `O(mn)`. As a difficulty 7 problem, I think it's appropriate to add some complexity to the solution to get good performance. You should update the example solution such that it passes this case in CI.

coriolinus:
This PR has gone stale, so I'm closing it. @vagrawal if you wish to reopen it, the major obstacle for now is that CI is experiencing compilation failures given its current state. The example solution needs to be able to solve the exercise.

PR #1300, opened 2021-08-30
Remove concepts without documents

ErikSchierboom:
The concept entries this PR removes from the config.json all did not have any documents. This causes syncing to fail, which previously did not happen because we hadn't yet made the blurb a required field (we do now). There were basically two options:

1. Remove the concept entries from the config.json that did not have any documents
2. Create empty placeholder documents for the concepts

I've opted for option nr. 1, as option nr. 2 would also mean that I'd have to set the `authors` property of the concept's `.meta/config.json` file to a non-empty array and I can't really do that.

Closes https://github.com/exercism/rust/issues/1299

### DIFF
--- a/config.json
+++ b/config.json
@@ -35,7 +35,7 @@
   "exercises": {
     "concept": [
       {
-        "slug": "lucians-luscious-lasagna",        
+        "slug": "lucians-luscious-lasagna",
         "uuid": "29a2d3bd-eec8-454d-9dba-4b2d7d071925",
         "name": "Lucian's Luscious Lasagna",
         "difficulty": 1,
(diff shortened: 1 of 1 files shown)


coriolinus (approved):
It would be nice to preserve the list of removed concepts somewhere, but it's more important to get the track at least minimally functional for v3.

ErikSchierboom:
@coriolinus Here is the list of removed concepts:

anonymous-lifetime
arc
async
borrow-trait
btreemap
casting
cell-and-refcell
channels
closures
conditionals
const-and-static
count-and-sum
deref-coercion
derive
designing-custom-traits
explicit-return
external-crates
external-traits-as-bounds
format-macro
from-into-iterator
futures
generics
hashmap
hashmaps
hashset
higher-order-functions
if-while-let
impl-blocks
implementing-traits
intro-fn
intro-types
iterator-usage
lazy-evaluation
lifetimes
logical-operators
loops
macros-declarative
macros-procedural
map-and-filter
match-basics
match-destructuring
methods
move-semantics
mutability
mutex
newtype-pattern
numeric-traits
overflow-and-underflow
ranges
rc
result
rwlock
scopes-and-expressions
shadowing
slices
static-lifetime
std-thread
string-use
total-ordering
typedefs
unsafe
use
variable-assignment
visibility

Feel free to create an issue, add a file to the repo, or whatever works best for you.

PR #1308, opened 2021-09-02
Fix incorrect variable name in instructions

mlouward:
Found that while doing the exercise, hope it helps !


coriolinus (approved):

coriolinus:
Thanks for spotting this!

PR #1312, opened 2021-09-05
role-playing-game: Add test that forces handing of integer overflow.

hsubra89:
The exemplar for this test suggests the use of `u32.saturating_sub`. However, there are no test cases that capture a use-case for it.

This patch adds in a single test that covers that case. Not sure if there was any specific reason for not adding this case in the first place.


coriolinus (approved):
Looks good! Thanks for adding this.

PR #1315, opened 2021-09-08
Update short-fibonacci.rs

xileftenurb:
the test was not checking the first two element,
this incorrectly make `[0, 0, 0, 0, 0]` valid and also `[1, 2, 3, 5, 8]`.

### DIFF
--- a/exercises/concept/short-fibonacci/tests/short-fibonacci.rs
+++ b/exercises/concept/short-fibonacci/tests/short-fibonacci.rs
@@ -18,6 +18,8 @@ fn test_buffer() {
 fn test_fibonacci() {
     let fibb = fibonacci();
     assert_eq!(fibb.len(), 5);
+    assert_eq!(fibb[0], 1);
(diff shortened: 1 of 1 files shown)


exercises/concept/short-fibonacci/tests/short-fibonacci.rs
```diff
@@ -18,6 +18,8 @@ fn test_buffer() {
 fn test_fibonacci() {
     let fibb = fibonacci();
     assert_eq!(fibb.len(), 5);
+    assert_eq!(fibb[0], 1);
```
coriolinus, line 21:
```suggestion
```

Omitting this line allows us to support the sequences starting with `[0, 1]` or `[1, 1]`.

xileftenurb: but it would also permit weird sequence like `[5, 1, 6, 7, 13]` to pass,

coriolinus (approved):

PR #1326, opened 2021-09-15
clock: make all expected values on the right

Tuxified:
Thanks for all the hard work, I really enjoy using excercism v3 in the browser. This PR stems from a little friction I encountered, feel free to close it as you see fit.

While trying to solve this exercise, I got confused when I ran the tests and sometimes the expected value was on the left side of the comparison and sometimes on the right.
In some cases it is obvious `assertion failed: '(left == right)' left: "12:03", right: "-10:03"`, but others were not `assertion failed: '(left == right)' left: "12:03", right: "11:03"` .. which required me to go back and forth between the test implementation, output of tests and my implementation.

My proposal is to keep all expected values on one side, since the tests started with them being on the right hand side, I sticked with that.

### DIFF
--- a/exercises/practice/clock/tests/clock.rs
+++ b/exercises/practice/clock/tests/clock.rs
@@ -282,67 +282,67 @@ fn test_compare_clocks_with_hour_overflow() {
 #[test]
 #[ignore]
 fn test_compare_clocks_with_hour_overflow_by_several_days() {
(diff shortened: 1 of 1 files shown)


coriolinus (approved):
Sure. This is harmless, and while I haven't checked that it's thorough, if there are any omissions, they can be rectified later.

PR #1328, opened 2021-09-16
Relax precision for assembly-line tests

### LINKED ISSUE #1320: Bug in test suite of "Assembly Line"
Because of floating point inaccuracies, one of the test fails when the factors are multiplied in a specific order.

Here is the solution that fails:

```rust
pub fn production_rate_per_hour(speed: u8) -> f64 {
    success_rate(speed) * 221.0 * (speed as f64) // <------ factor order of that line matters!
}

fn success_rate(speed: u8) -> f64 {
    match speed {
        0      => 0.0,
        1..=4  => 1.0,
        5..=8  => 0.9,
        9..=10 => 0.77,
        _ => panic!("speed must be between 0..=10, but received {} instead.", speed),
    }
}

pub fn working_items_per_minute(speed: u8) -> u32 {
    (production_rate_per_hour(speed) / 60.0) as u32
}

```

The test-helper `process_rate_per_hour` should probably check for a few epsilons and not smaller than a single epsilon.

koushik-ms:
Fixes #1320 
The test helper rounds the actual value down to the nearest hundredth of a unit.

### DIFF
--- a/exercises/concept/assembly-line/tests/assembly-line.rs
+++ b/exercises/concept/assembly-line/tests/assembly-line.rs
@@ -1,6 +1,6 @@
 fn process_rate_per_hour(speed: u8, expected_rate: f64) {
     let actual_rate = assembly_line::production_rate_per_hour(speed);
-    let actual_rate = (actual_rate * 100.0).floor()/100.0;
+    let actual_rate = (actual_rate * 100.0).round()/100.0;
     assert!((actual_rate - expected_rate).abs() < f64::EPSILON);
 }
 
--- a/exercises/concept/assembly-line/tests/assembly-line.rs
+++ b/exercises/concept/assembly-line/tests/assembly-line.rs
@@ -1,6 +1,6 @@
 fn process_rate_per_hour(speed: u8, expected_rate: f64) {
     let actual_rate = assembly_line::production_rate_per_hour(speed);
-    let actual_rate = (actual_rate * 100.0).round()/100.0;
+    let actual_rate = (actual_rate * 100.0).round() / 100.0;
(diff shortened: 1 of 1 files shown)


exercises/concept/assembly-line/tests/assembly-line.rs
```diff
@@ -1,5 +1,7 @@
 fn process_rate_per_hour(speed: u8, expected_rate: f64) {
-    assert!((assembly_line::production_rate_per_hour(speed) - expected_rate).abs() < f64::EPSILON);
+    let actual_rate = assembly_line::production_rate_per_hour(speed);
+    let actual_rate = (actual_rate * 100.0).floor()/100.0;
```
coriolinus, line 3:
```suggestion
    let actual_rate = (actual_rate * 100.0).round()/100.0;
```

If the introduced floating-point error happens to be low instead of high, we'll have more accurate results by rounding instead of flooring.

coriolinus:
You'll also need to run rustfmt to get CI to pass.

coriolinus (approved):

wischi-chr:
Doesn't this (I havn't tested it yet) theoretically suffer from the same issue? The new code does some f64 operations and than again compares with the smallest amount possible. Wouldn't it be more stable and readably to just do that:

```rust
assert!((actual_rate - expected_rate).abs() < 0.0001);
```

In the current implementation there could still be an issue in some situations.

coriolinus:
If we define that numbers are rounded to the nearest hundredth, then student implementations can also just round to the nearest hundredth, avoiding rounding errors entirely.

PR #1324, opened 2021-09-15
Hotfix for sync error due to missing concepts

koushik-ms:
Hotfix for #1323 
This PR creates placeholder documentation for the missing concepts causing the sync-failure.

PR #1330, opened 2021-09-18
Add a test to magazine-cutout

theopomies:
The addition of this test (in PR) makes _those_ (see below) wrong solutions fail, when they currently pass but shouldn't.

```rs
pub fn can_construct_note(magazine: &[&str], note: &[&str]) -> bool {
    magazine.iter().fold(0, |acc, &word| acc + note.contains(&word) as usize) >= note.len()
}
```

### DIFF
--- a/exercises/concept/magazine-cutout/tests/magazine-cutout.rs
+++ b/exercises/concept/magazine-cutout/tests/magazine-cutout.rs
@@ -62,3 +62,15 @@ fn test_magzine_has_more_than_words_available_than_needed() {
     let note = "enough is enough".split_whitespace().collect::<Vec<&str>>();
     assert!(can_construct_note(&magazine, &note));
 }
+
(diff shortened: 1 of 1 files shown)


coriolinus:
You'll need to run rustfmt to get CI to pass, but otherwise this seems like a good idea to me.

coriolinus (approved):
Looks good! Thanks for contributing.

bobahop:
I just downloaded the exercise today. Noticed the same issue. Any idea why this latest test was not included?

bobahop:
It also was not caught in the online editor tests. Something to do with the red "x"?

coriolinus:
At a guess, yes. https://github.com/exercism/rust/pull/1324, when complete, should fix that for everyone, allowing all relevant updates to sync.

PR #1336, opened 2021-09-23
hello world: Update GETTING_STARTED.md

SushantChandla:


coriolinus (approved):

petertseng:
There is another instance a few lines above, isn't there

SushantChandla:
> There is another instance a few lines above, isn't there

Yes, There was one instance, fixed!

PR #1322, opened 2021-09-14
Partial Fix for configlet lint failure

koushik-ms:
Partial fix for #1318 
This fixes all failures due to missing entry in concepts array.

Other than adding missing entries to concepts array, following changes were made to some exercise entries:

1. replace `hashmaps` with `hashmap`
2. replace `numbers` with `integers`
3. replace `string-use` with `string`
4. remove `intro-types`

PR #1345, opened 2021-10-05
remove concepts added by #1322

coriolinus:
This is a temporary reversion. All of the concepts removed here should
be re-added in the future (with the same UUIDs) when suitable documentation is produced.
However, the intent is to fix the configlet lint errors and sync
failures so that the track is once again updateable.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1520,46 +1520,6 @@
       "uuid": "febd2cf9-e058-48ef-bdc7-7583fb67e053",
       "slug": "structs",
       "name": "Structs"
-    },
-    {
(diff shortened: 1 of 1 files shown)


coriolinus:
Note that this still produces lint errors, but the ones resulting from this PR ([demo](https://github.com/exercism/rust/pull/1345/checks?check_run_id=3802430425)) are better than the ones without this PR ([demo](https://github.com/exercism/rust/issues/1323)).

petertseng (approved):

PR #1346, opened 2021-10-05
Re-add hashmap concept

coriolinus:
Add hashmap concept. This PR contains the hashmap portion of the work from #1324, which is stalled. Builds on top of https://github.com/exercism/rust/pull/1345 and must be rebased against `main` after that PR is merged.


concepts/hashmap/about.md
```diff
@@ -0,0 +1,69 @@
+# Hashmap
+
+Hashmap is a container that can be used to store "key-value pairs". This container is also known by the name associative array or Hashtable in other programming langauges. In short, the map contains a set of keys, each of which is mapped to a particular value. The user can retrieve a stored value by passing in the appropriate key. Users can also insert and delete keys and the associated values.
+
+Hashmap is a generic collection (like most of the collections in the standard library) and thus supports a broad variety of types as keys (including user-defined structs and tuples).
```
petertseng, line 5: This talks about the types that can be keys. What about values? It's discussed below in the section on trait bounds, but why not here too?

PR #1350, opened 2021-10-05
Re-add string concept

coriolinus:
Add string concept. This PR contains the string portion of the work from #1324, which is stalled. Builds on top of #1345 and must be rebased against main after that PR is merged.


coriolinus:
After applying the suggestions from the last batch of code review I believe this PR is now ready. However, I cannot approve it myself.

PR #1354, opened 2021-10-05
Update introduction.md of rust/concept/functions

myucesan:
Added some default info that should be sufficient for the exercise.

### DIFF
--- a/concepts/functions/introduction.md
+++ b/concepts/functions/introduction.md
@@ -1,3 +1,39 @@
 # Introduction
 
-TODO
+Sometimes a certain piece of code needs to be used more than once. If that's the case, it might be convenient to put the code into a function. A function generally only performs one specific action.
+
+```rust
+fn say_my_name() {
+    println!("Hello, ....!");
+}
+```
+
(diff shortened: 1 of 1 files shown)


coriolinus (commented):
This is good content, and I appreciate that you've put real work into it, but the policy is that a concept introduction should be one or at most two paragraphs of text; the meat of instruction happens in `about.md`. 

Would you mind trimming this down to that size, please?

myucesan:
İ'll do so later

myucesan:
@coriolinus done.

myucesan:
I think returning a value should be mentioned no? But hard to do with that constraint.

coriolinus:
Possibly the best way to avoid the constraint is to get rid of the examples here. If they show something not already in `about.md`, it's fine to move them there. That trims `introduction.md` down to two paragraphs, and gives you space to add a sentence explaining that in addition to accepting parameters, functions can return values. 

Let me know when you're happy with the text.

myucesan:
@coriolinus Happy with the text after the last commit.

coriolinus:
Thanks for all the work you've put into this!

PR #1355, opened 2021-10-06
Update rust/exercises/concept/semi-structured-logs

myucesan:
Added semicolons

### DIFF
--- a/exercises/concept/semi-structured-logs/src/lib.rs
+++ b/exercises/concept/semi-structured-logs/src/lib.rs
@@ -11,14 +11,14 @@ pub enum LogLevel {
 }
 /// primary function for emitting logs
 pub fn log(level: LogLevel, message: &str) -> String {
-    unimplemented!()
+    unimplemented!();
 }
 pub fn info(message: &str) -> String {
-    unimplemented!()
+    unimplemented!();
 }
 pub fn warn(message: &str) -> String {
-    unimplemented!()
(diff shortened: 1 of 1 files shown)


coriolinus:
Why? This is unnecessary.

myucesan:
> Why? This is unnecessary.

It causes trouble when you don't want to immediately remove the unimplemented!() but rather after you're done, for example,  when you want to keep seeing the goal (as potentially described in unimplemented!("Add two numbers)) of the function. This way it causes zero problems instead of potentially causing a minor annoyance when you add a statement below it. You get an error with rust-analyzer in Visual Studio Code with Exercism CLI saying it expects a semicolon. In this exercise there is no description, though. I thought it might be best keeping that consistent, however.

coriolinus:
> This way it causes zero problems

That's only true because of the `#[allow(unused)]` directive, which is non-standard. Otherwise, if you add statements, rustc complains about unreachable code.

```rust
warning: unreachable statement
  --> src/lib.rs:11:5
   |
10 |     unimplemented!();
   |     ----------------- any code following this expression is unreachable
11 |     let foo = 42;
   |     ^^^^^^^^^^^^^ unreachable statement
   |
   = note: `#[warn(unreachable_code)]` on by default
```

I'm not convinced by this change and will not approve it. I will not veto it if another maintainer chooses to approve this.

PR #1358, opened 2021-10-09
Updated rust/concepts/functions/about.md

myucesan:
Added reference return example


coriolinus:
Why did you add this documentation to the `functions` concept documentation? It seems like it would be a much better fit within the [references documentation](https://github.com/exercism/rust/blob/main/concepts/references/about.md).

myucesan:
Fair, thought it was needed for the concept exercise but that's not the case.

PR #1372, opened 2021-10-15
rework Collatz Conjecture as a concept exercise

coriolinus:
Teaches mutability and loops, reducing the number of CI errors.

@ErikSchierboom how much will this conversion from a practice to a concept exercise break things? Is it better to retain the existing UUID (as here), or to create a new UUID?


hochherz:
I think the removal of options from the function return value is not a good idea because overflow is part of a chaotic system problem ...

olafklingt:
I think the removal of options from the function return value is not a good idea because overflow is part of a chaotic system problem ...

PR #1373, opened 2021-10-16
Update lucians-luscious-lasagna.rs

ccfiel:
elapsed_time_in_minutes(4, 8) should return 40 and elapsed_time_in_minutes(1, 30) should return 12.


petertseng:
My reading of https://github.com/exercism/rust/blob/main/exercises/concept/lucians-luscious-lasagna/.docs/instructions.md#4-calculate-the-elapsed-time-in-minutes would seem to indicate that the existing values are correct here.

PR #1374, opened 2021-10-16
collatz-conjecture: add test for overflows

olafklingt:


coriolinus:
As you pointed out, this is incompatible with #1372. There's nothing wrong with it in principle (aside from the example solution failing against these new tests), but the API in that PR doesn't even use options, for simplicity. 

The track priority right now is clearing up the configlet lint CI issues. That means that #1372 is of a higher priority, which means that this PR can't be merged. 

Thank you for the work, which in other circumstances would be a real improvement. Unfortunately, right now, circumstances don't permit it.

coriolinus:
Looks like the last important failing lint is rustfmt, after which this can be merged. Thanks!

PR #1388, opened 2021-10-31
Fix assert's result that's unexpected.

Jerrody:


coriolinus:
As you can see, this causes the example solution to fail in CI, while it passes in CI with the existing solution. Could you explain why you think this test needs changing?

Jerrody:
> As you can see, this causes the example solution to fail in CI, while it passes in CI with the existing solution. Could you explain why you think this test needs changing?

That's because I cannot understand that test_cast_spell_with_enough_mana ->100 - 3 = 97. But not **6**.

petertseng:
Take a look at the instructions to see what `cast_spell` should return.

Jerrody:
> 
> 
> Take a look at the instructions to see what `cast_spell` should return.

Oh, my bad. Didn't read this until the end.

PR #1392, opened 2021-11-03
gigasecond: update instructions

zaurask:
#1384 replaced chrono with time crate. Reflect this change in instructions.


coriolinus (approved):
Good catch!

PR #1393, opened 2021-11-05
Setting toolchain to nightly as is required by edition 2021

GPavlovic:
Noticed the version was bumped recently, was getting an error so thought I would just throw this PR up to help future users.


coriolinus:
Edition 2021 does not require nightly. It has been stable since Rust 1.56. [link](https://blog.rust-lang.org/2021/10/21/Rust-1.56.0.html).

The Exercism Rust track always targets the latest stable release.

PR #1394, opened 2021-11-06
Relax the precision for assembly-line tests further

Rauno56:
I was surprised when I saw that the solution which was implemented like this:

```rust
pub fn production_rate_per_hour(speed: u8) -> f64 {
    if speed <= 4 {
        return 221.0 * speed as f64;
    }
    if speed <= 8 {
        return 0.9 * 221.0 * speed as f64;
    }
    return 0.77 * 221.0 * speed as f64;
}
```

... was incorrect because the results yielded:

```
9: 1531.5300000000002
10: 1701.7000000000003
```
The bug is subtle(missing paratheses or moving the fraction to the end) and the tests should take into account the expertise level for an average student trying to finish this exercise. I had to print the numbers out to see the mistake.

An alternative could be to add a hint about the order of the factors being important in some cases or rounding.

Also see https://github.com/exercism/rust/pull/1328 - the solution was suggested there as well, but dismissed with "they should just round" which is not mentioned in anywhere in the exercise.

### DIFF
--- a/exercises/concept/assembly-line/tests/assembly-line.rs
+++ b/exercises/concept/assembly-line/tests/assembly-line.rs
@@ -1,7 +1,7 @@
 fn process_rate_per_hour(speed: u8, expected_rate: f64) {
     let actual_rate = assembly_line::production_rate_per_hour(speed);
     let actual_rate = (actual_rate * 100.0).round() / 100.0;
-    assert!((actual_rate - expected_rate).abs() < f64::EPSILON);
+    assert!((actual_rate - expected_rate).abs() < 0.00001);
 }
 
 fn process_rate_per_minute(speed: u8, expected_rate: u32) {


petertseng (dismissed):
I believe we should accept, with thanks.

Unless I misunderstand what epsilon means, there is no number smaller than epsilon. That means, to say that a number is smaller than epsilon is simply to say that it is equal to zero. We know that we didn't want to compare floats using `==`, so we should also not want to compare them by checking if their difference is zero, since those are the same thing.

petertseng:
It is unfortunate that https://rust-lang.github.io/rust-clippy/master/index.html#float_cmp actually suggests comparing against f64::EPSILON, but my judgment says that that should be changed to suggest a better acceptable difference that's not f64::EPSILON.

petertseng:
Actually, here's the thing... I couldn't get the given solution:

```rust
pub fn production_rate_per_hour(speed: u8) -> f64 {
    if speed <= 4 {
        return 221.0 * speed as f64;
    }
    if speed <= 8 {
        return 0.9 * 221.0 * speed as f64;
    }
    return 0.77 * 221.0 * speed as f64;
}
```

to fail the tests. In fact, it passes the current tests for me.

It's true that for 9, production_rate_per_hour produces the value `1531.5300000000002`. However, then `process_rate_per_hour` will multiply that by 100, round it, and divide it by 100 again, which appears to produce a value that does compare equal to the literal 1531.53.

Could you walk me through how that solution fails the tests, please? I'd like to get a better understanding of the situation and be convinced this is necessary.

Rauno56:
Sure! I used the web editor for this exercise.
Check this out: https://www.loom.com/share/e87c6ac29ef144f5a40c963656093589

petertseng:
I see. And for my part, here is the test passing:

```rust
mod assembly_line {
    pub fn production_rate_per_hour(speed: u8) -> f64 {
        if speed <= 4 {
            return 221.0 * speed as f64;
        }
        if speed <= 8 {
            return 0.9 * 221.0 * speed as f64;
        }
        return 0.77 * 221.0 * speed as f64;
    }
}

fn process_rate_per_hour(speed: u8, expected_rate: f64) {
    let actual_rate = assembly_line::production_rate_per_hour(speed);
    let actual_rate = (actual_rate * 100.0).round() / 100.0;
    assert!((actual_rate - expected_rate).abs() < f64::EPSILON);
}

fn main() {
    process_rate_per_hour(9, 1531.53);
    println!("test passed")
}
```

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5170c27cc267db48fc7b1cc32a6011c9

So let's figure out what's different between those two situations. Does the web editor allow you to see what the content of the test is?

petertseng:
and/or does it allow you to see anything else that could explain why this test fails in one situation but doesn't in another?

coriolinus:
There has been no movement here for months. It sure looks like the current solution passes, obviating the need for this PR. Thank you to @Rauno56 and @petertseng for working on this, but for now I will close the issue as stale. Feel free to re-open if new evidence arrives.

PR #1397, opened 2021-11-14
Improve instructions for RPN Calculator

OverHash:
This PR improves the instructions for the RPN exercise, by specifying that your solution should return `None` if there is more than one item left in the stack. This was not mentioned elsewhere in the instructions, but [is included in unit tests](https://github.com/exercism/rust/blob/0c24b8ec71669467ce3a2c59d61b9c2519fbd46f/exercises/concept/rpn-calculator/tests/rpn-calculator.rs#L70-L75).

I could include an example which demonstrates this, but I feel as though this should be pretty clear after this change — and makes the instructions more concise by omitting a new example.

### DIFF
--- a/exercises/concept/rpn-calculator/.docs/instructions.md
+++ b/exercises/concept/rpn-calculator/.docs/instructions.md
@@ -87,6 +87,9 @@ we can say the expression evaluated to 6.
 ## 3. Goal
 
 Your goal is to write a calculator to evaluate a list of inputs ordered by Reverse Polish notation.
+
+If there is more than one element in the stack at the end, return `None`.
+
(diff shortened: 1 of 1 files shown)


petertseng:
Yup, that looks helpful and important to specify.

I notice that there are also other situations in which you might want to return None (operator without enough operands, or zero elements on the stack). Would it be helpful to mention those as well?

Would it also be helpful to say what you should return in normal circumstances?

OverHash:
Good suggestions! I have updated the instructions.

petertseng (approved):
Yes, that seems to cover everything we would want. We accept, with our thanks to you.

PR #1401, opened 2021-11-17
Fix grammar error in lasagna exercise

ConnorMooneyhan:
There was an extra "time" that didn't need to be there.


petertseng (approved):
Yes, that looks better. We accept, with our thanks to you.

PR #1403, opened 2021-11-18
Rust, RPN Calculator, Added test for missing operands

pavka1799:
Now, RPN input like "+ 2 2 *" is not tested.
There are solutions (using fold) that produce "4" for the above example, while it should be "None" due to incorrect input.
Adding one more test case will catch this kind of errors.

### DIFF
--- a/exercises/concept/rpn-calculator/tests/rpn-calculator.rs
+++ b/exercises/concept/rpn-calculator/tests/rpn-calculator.rs
@@ -76,7 +76,14 @@ fn test_too_many_operands_returns_none() {
 
 #[test]
 #[ignore]
-fn test_missing_operands_returns_none() {
+fn test_zero_operands_returns_none() {
+    let input = calculator_input("+");
+    assert_eq!(evaluate(&input), None);
+}
+
+#[test]
+#[ignore]
+fn test_intermediate_error_returns_none() {
     let input = calculator_input("+ 2 2 *");
     assert_eq!(evaluate(&input), None);
 }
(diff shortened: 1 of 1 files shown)


petertseng (commented):
That seems like a good case to test.

Let me make sure I understand why the input `"+ 2 2 *"` was chosen instead of just `"+"`. I imagine those `fold`-based solutions we are talking about simply forgot that the `"+"` had no operands because it was immediately followed by an otherwise-valid input. So, just testing `"+"` would not cover what this test case is trying to cover. Did I get that right?

petertseng:
In this case, I would wonder if it'd be better to split it into two tests!

One test just tests `"+"`. It's specifically testing zero operands.

The second test then tests `"+ 2 2 *"`. It's specifically testing that you should not continue calculations after an error has occurred.

Simply saying that `"+ 2 2 *"` tests zero operands is missing the whole picture.

Thus, this is the suggestion I would like to make for this PR, in order for it to receive approval.

pavka1799:
I agree; I added two tests as suggested.
The original idea indeed was to test that an intermediate error in parsing delivers the `None` result. There are solutions with `fold()` that pass all existing tests but will fail on `test_intermediate_error_returns_none()` including the one with the most stars at the moment.

petertseng (approved):
Awesome, now clear on what these cases are testing, and I think we're all agreed that these test cases are a good addition. Thank you. On behalf of the track I accept this PR with gratitude.

PR #1404, opened 2021-11-27
Add resistor-color exercise

still-flow:
Ported over from the C track, but with a Rust-ic flavor. Hopefully the external crate usage is okay.


exercises/practice/resistor-color/.meta/example.rs
```diff
@@ -0,0 +1,30 @@
...
+    Yellow = 4,
+    Green = 5,
+    Blue = 6,
+    Violet = 7,
+    Grey = 8,
+    White = 9,
+}
+
+pub fn color_to_value(color: ResistorColor) -> u8 {
+    color as u8
```
coriolinus, line 19:
```suggestion
    color.int_value()
```

This improves things in case you ever change the internal representation.

exercises/practice/resistor-color/.meta/example.rs
```diff
@@ -0,0 +1,30 @@
...
+}
+
+pub fn value_to_color_string(value: u8) -> String {
+    if !(0..=9).contains(&value) { return String::from("value out of range"); }
+    // because we've tested the value above, it's safe to unwrap
+    format!("{:?}", ResistorColor::from_int(value).unwrap_or(ResistorColor::Black))
```
coriolinus, line 25:
```suggestion
    ResistorColor::from_int(value).map(|color| format!("{:?}", color)).unwrap_or("value out of range".into())
```

Simplify.

exercises/practice/resistor-color/.meta/example.rs
```diff
@@ -0,0 +1,30 @@
...
+}
+
+pub fn colors() -> Vec<ResistorColor> {
+    (0..=9).map(|i| ResistorColor::from_int(i).unwrap_or(ResistorColor::Black)).collect()
```
coriolinus, line 29: This seems like a place for the [`enum_iterator` crate](https://docs.rs/enum-iterator/0.7.0/enum_iterator/), if we're using external crates anyway.

However, I have to ask: what learning objective is this test supporting?

exercises/practice/resistor-color/src/lib.rs
```diff
@@ -0,0 +1,26 @@
...
+    White,
+    Grey,
+    Violet,
+    Blue,
+    Green,
+    Yellow,
+    Orange,
+    Red,
+    Brown,
+    Black,
```
coriolinus, line 12: Why is this sorted backwards? It should be sorted either forwards, or alphabetically.

coriolinus (commented):
Just took a quick look; it's definitely a good start! Some notes:

- You're going to need to edit the top-level `config.json` to add the exercise there as well.
- What is this exercise meant to teach?
- If we can come up with a good educational goal, then could we structure this exercise as a concept exercise instead of a practice?

exercises/practice/resistor-color/.meta/example.rs
still-flow: I'm not too sure. The C track's version (also Python) all have a test to this end. I guess it's to verify that the enum is represented correctly?

exercises/practice/resistor-color/src/lib.rs
still-flow: I think alphabetically makes sense, so that the student has to read the instructions in order to make correct enum.

exercises/practice/resistor-color/.docs/hints.md
```diff
@@ -0,0 +1 @@
+Use the `int-enum` crate to be able to convert integers into enums. Check out [docs](https://docs.rs/int-enum/0.4.0/int_enum/) to find out how.
```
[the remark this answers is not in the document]
still-flow: I'll fix the link in the instructions.

still-flow:
Re what this exercise teaches: I think it teaches about C-like enums, also about external crate usage. And the neat "{:?}" formatting trick.

As for making this a concept exercise: sure! It's just that I'm new to contributing exercises, so a practice one is certainly easier. However, I'll probably be able to figure things out if necessary.

still-flow:
Re editing top-level `config.json`: where should I read to learn how to properly do it?

still-flow:
Well, here's my attempt, anyway. Please feel free to correct me.

config.json
```diff
@@ -1402,6 +1402,19 @@
...
+        "uuid": "51c31e6a-b7ec-469d-8a28-dd821fd857d2",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 4,
+        "topics": [
+          "enums",
+          "external_crates",
+          "format"
+        ]
+      },
```
coriolinus, line 1417: I see you've placed this last among the non-deprecated exercises. It's not a bad guess, but it should actually go somewhere among the difficulty-4 problems. For practice exercises, the ordering of the exercises in this list determines the "next recommended exercise" flow, so it's best to keep the ordering sensible.

config.json
```diff
@@ -1402,6 +1402,19 @@
...
       },
+      {
+        "slug": "resistor-color",
+        "name": "Resistor Color",
+        "uuid": "51c31e6a-b7ec-469d-8a28-dd821fd857d2",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 4,
+        "topics": [
+          "enums",
```
coriolinus, line 1413: As `enums` is a concept (as defined in the `concepts` section of this file), this should go in the `practices` array instead of `topics`. We have [lots of cleanup](https://github.com/exercism/rust/issues/1407) to do in this repo for that, so let's avoid exacerbating the problem.

config.json
```diff
@@ -1402,6 +1402,19 @@
           "unsafe"
         ]
       },
+      {
+        "slug": "resistor-color",
+        "name": "Resistor Color",
+        "uuid": "51c31e6a-b7ec-469d-8a28-dd821fd857d2",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 4,
```
coriolinus, line 1411: I'm curious: why did you assign this as difficulty 4 when the equivalent C exercise has it as difficulty 1?

exercises/practice/resistor-color/.docs/instructions.md
```diff
@@ -0,0 +1,38 @@
+# Instructions
+
+If you want to build something using a Raspberry Pi, you'll probably use _resistors_.
+For this exercise, you need to know two things about them:
+
+* Each resistor has a resistance value.
+* Resistors are small - so small in fact that if you printed the resistance value on them, it would be hard to read.
```
coriolinus, line 7: Other bulleted lists in this doc use `-` instead of `*` for the bullets.

exercises/practice/resistor-color/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+edition = "2021"
+name = "resistor-color"
+version = "0.0.0"
```
coriolinus, line 4: In general, an exercise which is live on the site should be version `1.0.0` or higher.

exercises/practice/resistor-color/.meta/example.rs
```diff
@@ -0,0 +1,32 @@
...
+
+pub fn value_to_color_string(value: u8) -> String {
+    ResistorColor::from_int(value)
+        .map(|color| format!("{:?}", color))
+        .unwrap_or("value out of range".into())
+}
+
+pub fn colors() -> Vec<ResistorColor> {
+    (0..=9)
+        .map(|i| ResistorColor::from_int(i).unwrap_or(ResistorColor::Black))
```
coriolinus, line 30:
```suggestion
        .map(|i| ResistorColor::from_int(i).unwrap())
```

We know from within this file that the range `0..=9` are all resistor colors, and there are no colors outside that range, so there's no need to consider the case where that isn't true. Might be worth adding a comment to that effect as well, justifying `unwrap`.

exercises/practice/resistor-color/tests/resistor-color.rs
```diff
@@ -0,0 +1,64 @@
+use resistor_color::ResistorColor;
```
coriolinus, line 1:
```suggestion
use resistor_color::{colors, color_to_value, value_to_color_string, ResistorColor};
```

Just makes usage a bit cleaner to import the functions as well.

coriolinus (commented):
Some more notes below.

The biggest comment I have is that we have to decide now if this is going to be a concept exercise; it's not something that can be changed after the exercise goes live. This seems plausible to me for teaching the `external_crates` concept. Upside: it gets a lot more visibility, as the concepts in Rust still need a lot of fleshing out. Downside: creating the concept requires a fair amount of extra work. If you do decide to convert it to a concept exercise, it should use the [`enum-iterator` crate](https://crates.io/crates/enum-iterator). 

[Concept docs](https://exercism.org/docs/building/tracks/concepts), [concept exercise docs](https://exercism.org/docs/building/tracks/concept-exercises).

If you don't want to do the extra work, this is also fine as a practice exercise. Just let me know whether you want to convert this or not.

config.json
still-flow: I think it's *a bit* more difficult because of the extra work expected. Are there difficulty levels between 1 and 4?

still-flow:
Okay, here's all the fixups for now. I think I should try converting this to a concept exercise, but failing that, this should be good for now.

coriolinus:
Cool, I'll look forward to seeing how this looks as a concept!

still-flow:
Alright, here is a rough first version of that.

config.json
still-flow: Changed it to 1 now.

concepts/external-crates/about.md
```diff
@@ -0,0 +1,5 @@
+# About
+
+A crate is a compilation unit in Rust. A crate can be compiled into a binary or into a library. Most of the time, you will be using `cargo` to manage external dependencies, which is as simple as adding a line to your `Cargo.toml` file.
```
coriolinus, line 3:
```suggestion
A crate is a compilation unit in Rust. A crate can be compiled into a binary or into a library. Most of the time, you will be using `cargo` to manage external dependencies. Adding a dependency can be as simple as adding a line to your `Cargo.toml` file.
```

I think splitting that sentence into two simplifies things a bit.

concepts/external-crates/introduction.md
```diff
@@ -0,0 +1,9 @@
+# Introduction
+
+Most programs have dependencies on some libraries. Managing those by hand can be a pain. Luckily, the Rust ecosystem comes standard with `cargo`! `cargo` can manage dependencies for a project.
+
+You will often see external packages being referred to as "crates" in Rust. A crate is a compilation unit in Rust. A crate can be compiled into a binary or into a library.
+
+Most of the time, adding an external dependency is as simple as adding a line to your `Cargo.toml` file.
+
+In the corresponding exercise, such an external dependency was added for you.
```
coriolinus, line 9:
```suggestion
```

Concept docs should stand alone, and there isn't a 1:1 mapping between concepts and concept exercises, so we shouldn't say this.

exercises/concept/resistor-color/.meta/exemplar.rs
```diff
@@ -0,0 +1,30 @@
...
+pub fn color_to_value(color: ResistorColor) -> usize {
+    color as usize
+}
+
+pub fn value_to_color_string(value: usize) -> String {
+    ResistorColor::into_enum_iter()
+        .nth(value)
+        .map(|color| format!("{:?}", color))
+        .unwrap_or_else(|| "value out of range".into())
+}
```
coriolinus, line 26: Why not retain the `IntEnum`-based implementation? I'd expect that to be able to produce a faster lookup than iterating through the entire list. I don't think it's a problem for the concept introducing external dependencies to have two different external dependencies.

exercises/concept/resistor-color/.docs/introduction.md
```diff
@@ -0,0 +1,9 @@
+# Introduction
+
+Most programs have dependencies on some libraries. Managing those by hand can be a pain. Luckily, the Rust ecosystem comes standard with `cargo`! `cargo` can manage dependencies for a project.
+
+You will often see external packages being referred to as "crates" in Rust. A crate is a compilation unit in Rust. A crate can be compiled into a binary or into a library.
+
+Most of the time, adding an external dependency is as simple as adding a line to your `Cargo.toml` file.
+
+In this exercise, such an external dependency was added for you.
```
coriolinus, line 9: It's worth specifying the names of the included external dependencies, and linking to the relevant documentation. IIRC students using the web interface can't ready `Cargo.toml` by default.

exercises/concept/resistor-color/src/lib.rs
```diff
@@ -0,0 +1,28 @@
...
+
+pub fn value_to_color_string(value: usize) -> String {
+    unimplemented!(
+        "convert the value {} into a string representation of color",
+        value
+    )
+}
+
+pub fn colors() -> Vec<ResistorColor> {
+    unimplemented!("return a list of all the colors in order")
```
coriolinus, line 27:
```suggestion
    unimplemented!("return a list of all the colors ordered by resistance")
```

Avoid confusion about i.e. alphabetical order.

coriolinus:
This looks very plausible as a concept to me!

exercises/concept/resistor-color/.meta/exemplar.rs
still-flow: I thought this was what you meant by 

> it should use the enum-iterator crate

To be honest, I'm not sure whether the two crates would interact nicely.

exercises/concept/resistor-color/.docs/introduction.md
still-flow: Isn't it enough to specify that in `instructions.md`?

coriolinus: Oh, you're right, sorry.

exercises/concept/resistor-color/.meta/exemplar.rs
still-flow: Ok, sure, I'll try!

still-flow:
Alright, re-adding the `int-enum` crate seems to have worked.

config.json
```diff
@@ -86,6 +86,20 @@
...
+        "uuid": "51c31e6a-b7ec-469d-8a28-dd821fd857d2",
+        "difficulty": 1,
+        "concepts": [
+          "external-crates"
+        ],
+        "prerequisites": [
+          "functions",
+          "enums"
+        ],
+        "status": "wip"
```
coriolinus, line 101: Note that until the status is `active`, students won't see this exercise. 

The last thing that I think is necessary before setting this active is to ensure that the rust test runner has both of these dependencies in its list of supported external crates.

coriolinus (approved):
Looks good! Thanks for putting in the effort to push this over the finish line.

config.json
still-flow: > Note that until the status is active, students won't see this exercise.

Oh, right. But we can't set it to `active` yet, right? Because some files (e.g. `design.md`) are currently missing.

> The last thing that I think is necessary before setting this active is to ensure that the rust test runner has both of these dependencies in its list of supported external crates.

Is that something I can do?

still-flow: If I understood correctly, [this line](https://github.com/exercism/rust-test-runner/blob/59acc0d59da1e2c2f6778ceca2e1f9c9da4e2e79/local-registry/supported_crates#L46) and [that line](https://github.com/exercism/rust-test-runner/blob/59acc0d59da1e2c2f6778ceca2e1f9c9da4e2e79/local-registry/supported_crates#L75) indeed show that the requisite crates are supported.

coriolinus: Oh very good! It's possible in principle that the versions don't line up--that's a [known issue](https://github.com/exercism/rust-test-runner/issues/37)--but I think we could switch this to active without worrying too much.

Want to do that before I merge, or would you prefer to leave it WIP for some more review first?

still-flow: I think squashing commits would be a good idea before merging (or is that something possible on your end?). But otherwise I'd leave that up to your discretion: perhaps more review is not a bad idea.

coriolinus: Sure, we use the github's squash and merge by default in this repo, no worries!

In my opinion this is ready to set as active now. Worst case we have to add a PR changing it to WIP if students have problems.

still-flow: Alright, if you say it's fine, then let's do this!

still-flow:
It works! Albeit not without trouble: it would seem that the crates have to be re-fetched on every test instance, which slows tests down considerably, even leading to timeouts sometimes. That's probably why there was a rule against external crates?

coriolinus:
Hmm, I'm pretty confident that it's not re-fetching the crates each time, because the test runners run offline. That's why there's a list of supported crates in the first place. However, it's worth looking into the timeout rate.

PR #1405, opened 2021-11-28
Diffie-hellmann: Add test that private key is a prime

dirkkul:
Adds a test that checks that the private key is actually a prime for a simple example. 

All other tests pass if you simply return any random number between 2 and p from the prviate key function. Eg the following code passes:
```
pub fn private_key(p: u64) -> u64 {
    rand::thread_rng().gen_range(3..p)
}

```

### DIFF
--- a/exercises/practice/diffie-hellman/tests/diffie-hellman.rs
+++ b/exercises/practice/diffie-hellman/tests/diffie-hellman.rs
@@ -13,6 +13,15 @@ fn test_private_key_in_range_key() {
 }
 
 #[test]
+fn test_private_key_is_prime() {
+    let primes: Vec<u64> = vec![2, 3, 5, 7, 11, 13];
+    for _ in 0..10 {
+        let private_key: u64 = private_key(15);
+        assert!(primes.contains(&private_key));
+    }
+}
+
+#[test]
 #[ignore]
 fn test_public_key_correct() {
     let p: u64 = 23;


exercises/practice/diffie-hellman/tests/diffie-hellman.rs
```diff
@@ -12,6 +12,14 @@ fn test_private_key_in_range_key() {
     }
 }
 
+
+#[test]
+fn test_private_key_is_prime() {
+    let primes: Vec<u64> = vec![1, 2, 3, 5, 7, 11, 13];
```
petertseng, line 18:
```suggestion
    let primes: Vec<u64> = vec![2, 3, 5, 7, 11, 13];
```

petertseng (commented):
This is a sensible test... what do you think about putting lines 19-20 in a loop that's run, say, 10 times, just to make sure a single run doesn't just get lucky and generate a prime by chance?

I am prepared to accept, with one change... 1 isn't prime.

You should additionally contribute this to https://github.com/exercism/problem-specifications/blob/main/exercises/diffie-hellman/canonical-data.json so other tracks can benefit. Exercism would not be where it is today if each track were selfish and refused to share its cases with other tracks.

The way you express it may have to be a little unsuited for JSON, but hopefully something can be figured out based on what's currently there. (something like adding a test with property `privateKeyIsPrime`)

coriolinus:
I like the idea of running in a loop to reduce the chance of lucky prime guesses.

exercises/practice/diffie-hellman/tests/diffie-hellman.rs
dirkkul: Good ideas, I changed the test.

With regard to json, how about this one?
```
    {
      "uuid": "c3fa6fed-d1b3-45f6-8245-be99d6e593a1",
      "description": "private key is a prime",
      "property": "privateKey",
      "input": {"p" :15},
      "expected": {
        "oneOf": [2, 3, 5, 7, 11, 13]
      }
    },
```

petertseng: Seems reasonable. Others may have opinions on whether it's okay to use the `privateKey` property but it's not currently being used right now, so it would at least not conflict with anything.

petertseng:
Please correct what CI says to correct, and then we can approve and merge it!

petertseng:
Please ask for help if you would like help interpreting what CI is telling you, and we can instruct on how to interpret (so you can help yourself in the future)

petertseng:
I wonder if I've misinterpreted the instructions here. The private key doesn't need to be prime, does it? See https://en.wikipedia.org/wiki/Diffie%E2%80%93Hellman_key_exchange where alice uses `a = 4`, which isn't prime!

dirkkul:
> I wonder if I've misinterpreted the instructions here. The private key doesn't need to be prime, does it? See https://en.wikipedia.org/wiki/Diffie%E2%80%93Hellman_key_exchange where Alice uses `a = 4`, which isn't prime!

Ouch, I think you're right. I'm still a bit confused which number needs to be prime and which not but a seems to be not among them. Sorry for this!

PR #1406, opened 2021-11-28
scale-generator: Require N+1 notes from N intervals

petertseng:
The problem has been explained in
https://github.com/exercism/problem-specifications/issues/1640:

All previous cases only required N notes from N intervals, but that
meant the last interval was completely ignored.
That causes confusion (what is the last interval there for?).

To fix this, all tests with the `interval` property are updated.

The `tonic` and `intervals` inputs have not changed. The `expected` has
changed only in that there is one more note added, indicated by the last
interval, which was previously ignored.

https://github.com/exercism/problem-specifications/pull/1888

### DIFF
--- a/exercises/practice/scale-generator/.meta/example.rs
+++ b/exercises/practice/scale-generator/.meta/example.rs
@@ -346,9 +346,10 @@ impl Scale {
         let mut out = Vec::with_capacity(self.intervals.len());
 
         let mut note = self.tonic;
+        out.push(note.canonicalize(self.lean).to_string());
         for &interval in self.intervals.iter() {
-            out.push(note.canonicalize(self.lean).to_string());
             note += interval;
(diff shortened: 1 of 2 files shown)


petertseng:
I was surprised to see that the `interval` parse of the example solution actually enforces that the intervals result in exactly one octave. That's actually good!

It does have this effect: since `chromatic` is just a special case of `interval`, that meant that the `chromatic` tests also changed to require the last note. I actually didn't change the `chromatic` tests in problem-specifications, but I think I just do that here. I could make the example solution have an extra boolean that is used to drop the last element from a chromatic scale, but it's not the easiest.

coriolinus (approved):

coriolinus:
I don't think we need to do anything special for the chromatic case.

PR #1409, opened 2021-12-04
fix: add gitignore in short-fibonacci

qdelettre:
gitignore seems to be missing


coriolinus:
All files ignored by this `.gitignore` file are already ignored by the [top-level `.gitignore` file](https://github.com/exercism/rust/blob/main/.gitignore). This change is harmless, but it doesn't actually change anything either. I'm inclined not to merge, though I'm willing to hear arguments why I should change my mind.

qdelettre:
@coriolinus 

Yes, indeed from the point of view of this repo i totally agree, top level is ok.

But when you use the exercism cli to download an exercise, the result is that exercism folders "generated" by the cli have a gitignore, some don't.

coriolinus (approved):
Ah, good point.

PR #1412, opened 2021-12-08
luscious-lasagna: remove Python content from text

softmoth:
Defining scope via indentation is a Python feature, and not true for Rust.


coriolinus (approved):
Oops, you're right.

PR #1414, opened 2021-12-16
Update Resister Colour Order

PieterVanZyl-Dev:
Passing the Final Test for the colors() function requires you to reorganize the enum.

Either, update the test to not care about the order, only that all the color strings are there, or update the order of the enum so people don't have to update it manually themselves while working on this problem.

### DIFF
--- a/exercises/concept/resistor-color/src/lib.rs
+++ b/exercises/concept/resistor-color/src/lib.rs
@@ -1,15 +1,15 @@
 #[derive(Debug, PartialEq)]
 pub enum ResistorColor {
     Black,
-    Blue,
     Brown,
-    Green,
-    Grey,
-    Orange,
     Red,
-    Violet,
-    White,
+    Orange,
     Yellow,
+    Green,
+    Blue,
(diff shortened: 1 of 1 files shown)


PieterVanZyl-Dev:
Otherwise as a **third** alternative update challenge text to say that colors() should return in numerical order

coriolinus:
Requiring that students reorder the colors was [intentional](https://github.com/exercism/rust/pull/1404#discussion_r757788072).

mrl5:
hello, I'd like to share my feedback about the intentional disorder of `ResistorColors` enum:

I actually think it's beneficial that they are not in order -> this allowed me to search for a solution that would implement [Ord](https://doc.rust-lang.org/std/cmp/trait.Ord.html)

PieterVanZyl-Dev:
I'm happy to close this as it's intentional 

My recommendation is to mention that ordering is important in the write up for that function. 

Personally I like to know what my function requires and having completed it. I expected the test to pass.

PR #1419, opened 2021-12-28
It's not obvious that ascii is an option in this exercise

gilescope:
It's not obvious to the average student that ascii is an option in this exercise. Some people will know this trick, but a lot of people just wouldn't think of it and it makes the exercise more pleasurable. It also makes collecting the results a little nicer at the end too.

Please feel free to suggest changes in the wording!

### DIFF
--- a/exercises/practice/minesweeper/.docs/instructions.md
+++ b/exercises/practice/minesweeper/.docs/instructions.md
@@ -41,7 +41,7 @@ And your code will transform it into this:
 All the inputs and outputs are in Ascii. Rust `String`s and `&str` are utf8,
 so while one might expect "Hello".chars() to be simple, it actually has to
 check each char to see if it's 1, 2, 3 or 4 `u8`s long. If we know a `&str`
-is ascii then we can call `.as_bytes()` and turn it into a `&[u8]` slice.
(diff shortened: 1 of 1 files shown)


exercises/practice/minesweeper/.docs/instructions.md
```diff
@@ -35,3 +35,14 @@ And your code will transform it into this:
 ·2*2·
 ·111·
 ```
+
+## Performance Hint
+
+All the inputs and outputs are in Ascii. Rust `String`s and `&str` are utf8,
+so while one might expect "Hello".chars() to be simple, it actually has to
+check each char to see if it's 1, 2, 3 or 4 `u8`s long. If we know a `&str`
+is ascii then we can call `.as_bytes()` and turn it into a `&[u8]` slice.
```
gilescope, line 44: We're not changing the underlying data so maybe word it as:
```suggestion
is ascii then we can call `.as_bytes()` and refer to the underlying data via a `&[u8]` slice.
```

coriolinus: This is clearer, yes.

gilescope:
@coriolinus any other changes needed for approval?

coriolinus (approved):
Sorry, no; this just fell off my radar with the holidays.

Issue #1407, opened 2021-11-29
Clean up `topics` vs `practices` in `config.json`


coriolinus:
Per the [current docs](https://exercism.org/docs/building/tracks/config-json), the track-level `config.json`'s per-exercise `topics` key isn't used for anything. This has the advantage that the items there aren't constrained by cross-checks, but the disadvantage that we're not fully exposing to the website the information we have about each exercise.

The general procedure we want to follow goes like this:

- Collect the list of all current concept exercises. Currently:
    ```bash
    $ jq -r '.concepts[] | .slug' config.json | sort -u
    booleans
    box
    char
    collect
    destructuring
    entry-api
    enums
    floating-point-numbers
    fold
    functions
    hashmap
    integers
    loops
    match-basics
    mutability
    option
    references
    strings
    string-slices
    string-vs-str
    structs
    tuples
    vec-macro
    vec-stack
    ````
- For each exercise among the practice exercises:
  - For each item in `topics`
    - If the topic is a current concept, move it from `topics` to `practices`

This process needs to be repeated periodically as we add concepts, so PRs addressing this issue should not close it.

ferminhg:
Hi! I created a PR for it. 

PD: clould you add #hacktoberfest tag? Thanks.

ferminhg:
@coriolinus, is this needs needed?

coriolinus:
Hi @ferminhg, sorry, I stepped down as maintainer of this repo some time ago. I assume that your PR is still useful but I will not be reviewing it.

ferminhg:
Hi @bobahop, do you know who can help me merge it?

bobahop:
I'm sorry that the scope of this is beyond me at the moment. @petertseng may have some thoughts about this.

petertseng:
Wow I thought this was going to be some difficult thing that requires specific knowledge to review, but it's actually a fairly strictly rules-based change, with the instructions given in the issue description. It just says for any concept that exists in https://github.com/exercism/rust/blob/d2a5ba340d89b296d98308c36538acca3ff0a5ab/config.json#L1495, move any instance of it from any practice exercise's `topics` to `practices`. Very rules-based, so reviewing pull requests just means checking that the rules are followed. CI validates that all concepts named in the `practices` array match a given concept so maintainers do not have to worry about that, just have to check that CI passes.

petertseng:
[...]

Issue #1089, opened 2021-01-29
[v3] Add prerequisites to Practice Exercises


ErikSchierboom:
_This issue is part of the migration to v3. You can read full details about the various changes [here](https://github.com/exercism/v3-launch/)._

Exercism v3 introduces a new type of exercise: [Concept Exercises](https://github.com/exercism/v3-docs/blob/main/product/concept-exercises.md). All existing (V2) exercises will become [Practice Exercises](https://github.com/exercism/v3-docs/blob/main/product/practice-exercises.md).

Concept Exercises and Practice Exercises are linked to each other via [Concepts](https://github.com/exercism/v3-docs/blob/main/anatomy/tracks/concepts.md). Concepts are taught by Concept Exercises and practiced in Practice Exercises. Each Exercise (Concept or Practice) has prerequisites, which must be met to unlock an Exercise - once all the prerequisite Concepts have been "taught" by a Concept Exercise, the exercise itself becomes unlocked.

For example, in some languages completing the Concept Exercises that teach the "String Interpolation" and "Optional Parameters" concepts might then unlock the `two-fer` Practice Exercise.

Each Practice Exercise should have one or more prerequisites, where each prerequisite describes the minimal set of concepts the Practice Exercise requires the student to be familiar with.

Although ideally all Concepts should be taught by Concept Exercises, we recognise that it will take time for tracks to achieve that. Any Practice Exercises that have prerequisites which are not taught by Concept Exercises, will become unlocked once the final Concept Exercise has been completed.

## Goal

The `"prerequisites"` field of each element in the `"exercises.practice"` field in the `config.json` file should be updated to contain the prerequisite concepts. See [the spec](https://github.com/exercism/v3-docs/blob/main/anatomy/tracks/config-json.md#practice-exercises).

To help with identifying the prerequisites, the `"topics"` field can be used (if it has any contents). Once prerequisites have been defined for a Practice Exercise, the `"topics"` field should be removed.

Each prerequisite concept should have its own entry in the top-level `"concepts"` array. See [the spec](https://github.com/exercism/v3-docs/blob/main/anatomy/tracks/config-json.md#concepts).

## Example

```json
"practice": [
  {
    "slug": "hello-world",
    "name": "Hello World",
    "uuid": "6c88f46b-5acb-4fae-a6ec-b48ae3f8168f",
    "prerequisites": ["strings"],
    "difficulty": 1
  }
]
```

## Tracking

https://github.com/exercism/v3-launch/issues/6

[...]

PR #1421, opened 2022-01-02
Clean up topics vs practices in config.json

dmaahs2017:
Addresses #1407. I think this also has something to do with #1089. Though I'm not entirely sure.

I also created a python script to automate this task.

### DIFF
--- a/bin/clean_topics_vs_practices.py
+++ b/bin/clean_topics_vs_practices.py
@@ -1,0 +1,42 @@
+#!/usr/bin/env python3
+import json
+
+
+def main():
+    with open("config.json", encoding="utf-8") as f:
+        config = json.load(f)
+
+    concepts = {c['slug'] for c in config['concepts']}
+
+    for practice_exercise in config['exercises']['practice']:
+        if practice_exercise['topics'] is None:
+            continue
+
+        practice_exercise['practices'].extend((topic for topic in practice_exercise['topics'] if topic in concepts))
+        practice_exercise['topics'] = [topic for topic in practice_exercise['topics'] if topic not in concepts]
+
+    for concept in concepts:
+        count = 0
+        for practice_exercise in config['exercises']['practice']:
+            if concept in practice_exercise['practices']:
+                count += 1
+                if count > 10:
+                    practice_exercise['practices'].remove(concept)
+                    practice_exercise['topics'].append(concept)
+
+    for practice_exercise in config['exercises']['practice']:
+        practice_exercise['practices'].sort()
+
+        if practice_exercise['topics'] is not None:
+            practice_exercise['topics'].sort()
+
+
+    with open("config.json", 'w', encoding="utf-8") as f:
+        json.dump(config, f, indent=2, ensure_ascii=False)
+        f.write('\n')
+
+    print("Updated config.json")
+
+
+if __name__ == '__main__':
+    main()
--- a/config.json
+++ b/config.json
@@ -192,11 +192,12 @@
         "slug": "hello-world",
         "name": "Hello World",
         "uuid": "13ec1ebe-d71b-436f-ab12-25305e814171",
-        "practices": [],
+        "practices": [
+          "strings"
+        ],
         "prerequisites": [],
         "difficulty": 1,
         "topics": [
-          "strings",
           "test_driven_development"
         ]
       },
@@ -204,13 +205,14 @@
         "slug": "reverse-string",
         "name": "Reverse String",
         "uuid": "ecf8d1e3-9400-4d1a-8326-2e2820bce024",
-        "practices": [],
+        "practices": [
+          "strings"
+        ],
         "prerequisites": [],
         "difficulty": 1,
         "topics": [
           "iterators",
-          "str_vs_string",
-          "strings"
+          "str_vs_string"
         ]
       },
       {
@@ -228,12 +230,13 @@
         "slug": "clock",
         "name": "Clock",
         "uuid": "543a3ca2-fb9b-4f20-a873-ff23595d41df",
-        "practices": [],
+        "practices": [
+          "structs"
+        ],
         "prerequisites": [],
         "difficulty": 4,
         "topics": [
           "derive",
-          "structs",
           "traits"
         ]
       },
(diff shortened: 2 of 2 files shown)


bin/clean_topics_vs_practices.py
```diff
@@ -0,0 +1,28 @@
+#!python3
+import json
+
+
+def main():
+    with open("config.json") as f:
+        config = json.load(f)
+
+    concepts = [c['slug'] for c in config['concepts']]
```
coriolinus, line 9:
```suggestion
    concepts = {c['slug'] for c in config['concepts']}
```

Making `concepts` a set instead of a list means that searching it is faster: the `in` operator does a linear scan for a list; it can't do a binary search because it can't assume the list is sorted. For a set, it just does a hash comparison.

bin/clean_topics_vs_practices.py
```diff
@@ -0,0 +1,28 @@
...
+    concepts.sort()
```
coriolinus, line 10:
```suggestion
```

Sorting the list of concepts doesn't help, here.

bin/clean_topics_vs_practices.py
```diff
@@ -0,0 +1,28 @@
...
+
+    for practice_exercise in config['exercises']['practice']:
+        if practice_exercise['topics'] is None:
+            continue
+
+        for topic in practice_exercise['topics']:
+            if topic in concepts:
+                practice_exercise['topics'].remove(topic)
+                practice_exercise['practices'].append(topic)
```
coriolinus, line 19: While this only performs a single iteration over the topics, which is good, it's messy to edit the list of topics in the middle of iterating over it. I'd prefer a pair of list comprehensions, so it's clear that the logic is sound.

```suggestion
        practice_exercise['practices'].extend((topic for topic in practice_exercise['topics'] if topic in concepts))
        practice_exercise['topics'] = [topic for topic in practice_exercise['topics'] if topic not in concepts]
```

bin/clean_topics_vs_practices.py
```diff
@@ -0,0 +1,28 @@
...
+
+    with open("config.json", 'w') as f:
```
coriolinus, line 21:
```suggestion
    with open("config.json", 'w', encoding="utf-8") as f:
```

Let's be explicit about the encoding. It wouldn't do for someone to run this on a Windows machine and accidentally encode the file as CP-1252 or something.

bin/clean_topics_vs_practices.py
```diff
@@ -0,0 +1,28 @@
+#!python3
+import json
+
+
+def main():
+    with open("config.json") as f:
```
coriolinus, line 6:
```suggestion
    with open("config.json", encoding="utf-8") as f:
```

Likewise here.

bin/clean_topics_vs_practices.py
```diff
@@ -0,0 +1,28 @@
+#!python3
```
coriolinus, line 1:
```suggestion
#!/usr/bin/env python3
```

We shouldn't assume that `$PATH` exists or is sane.

coriolinus (commented):
Thank you, this is useful work! While I'd slightly prefer it if the script were written in Rust, I do understand the desire for a shorter script for a utility item like this, and it will be fine that it's currently in Python.

bin/clean_topics_vs_practices.py
dmaahs2017: This was something I was also unsure about. I like the suggestion, thank you.

bin/clean_topics_vs_practices.py
dmaahs2017: Good point.

dmaahs2017:
> Thank you, this is useful work! While I'd slightly prefer it if the script were written in Rust, I do understand the desire for a shorter script for a utility item like this, and it will be fine that it's currently in Python.

I do love Rust; it just seemed a bit overkill for this use-case 😃

dmaahs2017:
@coriolinus Is this an important issue? If so, how would you like it to be resolved?

image

bin/clean_topics_vs_practices.py
```diff
@@ -0,0 +1,25 @@
...
+        practice_exercise['topics'] = [topic for topic in practice_exercise['topics'] if topic not in concepts]
+
+    with open("config.json", 'w', encoding="utf-8") as f:
+        json.dump(config, f, indent=2, ensure_ascii=False)
+        f.write('\n')
+
+    print("Updated config.json")
+
+
+main()
```
coriolinus, line 25:
```suggestion
if __name__ == '__main__':
    main()
```

This means it'll only run the `main` function when run as a script; it won't run it on import.

coriolinus:
> @coriolinus Is this an important issue? 

Not sure, I haven't seen it before. I asked the other maintainers. I'll let you know when I hear back.

bin/clean_topics_vs_practices.py
dmaahs2017: Oh right, forgot that!

coriolinus:
The response seems to be that:

- there isn't yet any documentation I can link to about the restriction, unfortunately
- but it's a real limitation we can't bypass. It breaks the website's styling if any concept has more than 10 "practices" exercises

To me this sounds like we need a post-processing step. I think it's safe to assume that entries in `config.json` are sorted in approximate order of difficulty / relevance, so we just need something like

```python
for concept in concepts:
    count = 0
    for practice_exercise in config['exercises']['practice']:
        if concept in practice_exercise['practices']:
            count += 1
        if count > 10:
            practice_exercise['practices'].remove(concept)
```

dmaahs2017:
Thanks for looking into that. I've worked that logic into the script and re-ran it.

coriolinus (commented):
Just two more edits, sorry.

- If we remove a topic from the `practices` array due to count, we should put it back into the `topics` array. That ensures we don't destroy information. I should have thought of that earlier.
- Both the `practices` and `topics` arrays should be sorted for each exercise after all other modifications are complete. The linter will complain otherwise.

dmaahs2017:
Sounds good, added that logic as well

coriolinus (approved):
Thanks for doing this work!

dmaahs2017:
Glad I could help! :smiley:

Issue #1422, opened 2022-01-04
Rust track exercise "xorcism" cannot run tests


jwaddle:
This is due to crates included by default for Xorcism's tests, not due to the user including external crates.

even when submitting old, previously working community solutions to the "Xorcism" problem, Exercism gives the following error:

`We received the following error when we ran your code:
WARNING: student did not upload Cargo.lock. This may cause build errors.
error: failed to select a version for the requirement `hexlit = "^0.3.0"`
candidate versions found which didn't match: 0.5.3
location searched: `/opt/test-runner/local-registry` index (which is replacing registry `crates-io`)
required by package `xorcism v0.1.0 (/mnt/exercism-iteration)`
perhaps a crate was updated and forgotten to be re-vendored?
As a reminder, you're using offline mode (--offli
<img width="1479" alt="Screen Shot 2022-01-03 at 8 41 10 PM" src="https://user-images.githubusercontent.com/5522478/148010494-701bb99c-938b-409b-912a-0c8de7f51dd3.png">
ne) which can sometimes cause surprising resolution failures, if this error is too confusing you may wish to retry without the offline flag.`

<img width="1479" alt="Screen Shot 2022-01-03 at 8 41 10 PM" src="https://user-images.githubusercontent.com/5522478/148010822-06d55435-f187-43df-afba-673f4df70253.png">

iHiD:
Thanks for reporting. cc @exercism/rust

jwaddle:
Does this take a while before it's live?  It still appears to be an issue

coriolinus:
Hmm, it should have updated by now. Is the error identical? In particular, the text of interest in the above screenshot is

> ```
> error: failed to select a version for the requirement `hexlit = "^0.3.0"`
> candidate versions found which don't match:
> 0.5.3
> ```

Is that the same, or has the error changed?

jwaddle:
it looks the same
![Screen Shot 2022-01-05 at 3 14 10 PM](https://user-images.githubusercontent.com/5522478/148303114-c855a8a0-c3ed-4345-bfad-593bee64167c.png)

petertseng:
This is my understanding:

[...]

PR #1423, opened 2022-01-04
xorcism: bump hexlit version

coriolinus:
Closes https://github.com/exercism/rust/issues/1422

The error message complains that the only supported hexlit package (offline) is v0.5.3. Per semver, as a version under 1.0, that's not compatible with v0.3.0. 

This should fix the issue, but I'm not sure how to test the PR against the site before merging. @ErikSchierboom ?

### DIFF
--- a/exercises/practice/xorcism/Cargo.toml
+++ b/exercises/practice/xorcism/Cargo.toml
@@ -10,6 +10,6 @@ edition = "2021"
 io = []
 
 [dev-dependencies]
-hexlit = "0.3.0"
+hexlit = "0.5.0"
 rstest = "0.6.4"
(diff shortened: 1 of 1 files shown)


petertseng (approved):
I didn't know until looking up https://doc.rust-lang.org/cargo/reference/semver.html just now that "Cargo uses the convention that only changes in the left-most non-zero component are considered incompatible." Interesting.

coriolinus:
I'm going to yolo-merge this without waiting for testing instructions based on a combination of confidence that I got it right, and the knowledge that even if I'm wrong, this won't break anything that isn't already broken.

PR #1426, opened 2022-01-14
address clippy lints found in new beta clippy 1.59

coriolinus:
[Failing lints](https://github.com/exercism/rust/runs/4815983072?check_suite_focus=true).

### DIFF
--- a/exercises/practice/decimal/tests/decimal.rs
+++ b/exercises/practice/decimal/tests/decimal.rs
@@ -36,7 +36,7 @@ fn test_ne() {
 fn test_gt() {
     for slice_2 in BIGS.windows(2) {
         assert!(decimal(slice_2[1]) > decimal(slice_2[0]));
-        assert!(!(decimal(slice_2[0]) > decimal(slice_2[1])));
+        assert!(decimal(slice_2[0]) < decimal(slice_2[1]));
     }
 }
 
@@ -45,7 +45,7 @@ fn test_gt() {
 fn test_lt() {
     for slice_2 in BIGS.windows(2) {
         assert!(decimal(slice_2[0]) < decimal(slice_2[1]));
-        assert!(!(decimal(slice_2[1]) < decimal(slice_2[0])));
+        assert!(decimal(slice_2[1]) > decimal(slice_2[0]));
     }
 }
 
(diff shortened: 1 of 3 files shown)


exercises/practice/decimal/tests/decimal.rs
```diff
@@ -36,7 +36,7 @@ fn test_ne() {
 fn test_gt() {
     for slice_2 in BIGS.windows(2) {
         assert!(decimal(slice_2[1]) > decimal(slice_2[0]));
-        assert!(!(decimal(slice_2[0]) > decimal(slice_2[1])));
+        assert!(decimal(slice_2[0]) < decimal(slice_2[1]));
```
petertseng, line 39: Well, at this point, `test_gt` and `test_lt` are testing the same thing, aren't they? May just need to delete this line. if `gt` should remain only about `>`.

petertseng (approved):

exercises/practice/decimal/tests/decimal.rs
```diff
@@ -45,7 +45,7 @@ fn test_gt() {
 fn test_lt() {
     for slice_2 in BIGS.windows(2) {
         assert!(decimal(slice_2[0]) < decimal(slice_2[1]));
-        assert!(!(decimal(slice_2[1]) < decimal(slice_2[0])));
+        assert!(decimal(slice_2[1]) > decimal(slice_2[0]));
```
petertseng, line 48: May just need to delete this line. if `lt` should remain only about `<`.

exercises/practice/decimal/tests/decimal.rs
coriolinus: You're right; I'm not sure how one of those tests could possibly succeed without the other also succeeding. I guess I was trying to guard against a weird `Ord` implementation? That seems like not something to worry about.

Fixed in f3155346.

petertseng (approved):

exercises/practice/decimal/tests/decimal.rs
petertseng: I guess the reasoning there could be: Someone could potentially explicitly define `gt` to always return true, instead of relying on the implementation provided by the trait? If that sounds like something to actually test for, I would also approve something that ignores that particular lint for `test_gt` and `test_lt`

coriolinus: Yeah, I'm not too concerned by that case. It's harder to do that than to implement `Ord` sensibly, so I'm willing to leave this to the mentors.

Issue #1365, opened 2021-10-11
crate availability should be documented in some way on the site


zesty:
wish list of included documentation:
- the fact that they exist
- a listing of or pointer to a list of the crates that are live and/or available
- a way to find out what versions are latest and which are available
- no need for `extern crate` like many top solutions have
- the need to upload Cargo.toml (not Cargo.lock, contrary to the error message) and the solution file at the same time

example 5d ago iteration, where they seemed to give up on using the rand crate, which was available at the time:
https://exercism.org/tracks/rust/exercises/diffie-hellman/solutions/p1way

example 6d ago iteration, where user got a failure, then uploaded Cargo.lock (likely per the error message), then also got a failure when including Cargo.toml with Cargo.lock, and then seemed to give up --- I suspect (but am too lazy to check) that including Cargo.lock caused the failure:
https://exercism.org/tracks/rust/exercises/diffie-hellman/solutions/viet-nv

senekor:
This is important. Problems with dependencies are pretty pervasive. https://github.com/exercism/rust-test-runner/pull/77 should alleviate many of these problems, as the most recent versions should regularly be made available. Once that's merged, we can also include a pointer to that `Cargo.toml` in the documentation, it includes the exact versions of currently available crates.

PR #1429, opened 2022-01-19
diffie-hellman: Add rand 0.8.4 to Cargo.toml

PaulT89:
Problem: If the expectation is for the user to generate a _random_ private key, then there is currently no way to do this in the browser version of this exercise, as the user does not have access to Cargo.toml.

Solution: Allow the use of rand by adding rand 0.8.4 to Cargo.toml dependencies

Caveat: I don't know if  [this post](https://github.com/exercism/rust/issues/1365) indicates that the rand crate _was_ in Cargo.toml at some point and was simply removed for some reason, or whether this is a complaint that all attempts to use rand::Rng were frustrated (because the rand crate was never in Cargo.toml).

### DIFF
--- a/exercises/practice/diffie-hellman/Cargo.toml
+++ b/exercises/practice/diffie-hellman/Cargo.toml
@@ -4,6 +4,7 @@ name = "diffie-hellman"
 version = "0.1.0"
 
 [dependencies]
+rand = "0.8.4"
 
 [features]
 big-primes = []


coriolinus:
A better solution would be to add `Cargo.toml` to the `files.solution` key in `config.json` as documented [here](https://exercism.org/docs/building/tracks/config-json). This should ensure that students always get the chance to edit that file online.

PaulT89:
Fair enough

PaulT89:
Is that something I can do without breaking the exercise?

PR #1430, opened 2022-01-19
diffie-hellman: `use rand::Rng`

PaulT89:
Assuming [my previous PR](https://github.com/exercism/rust/pull/1429) is accepted, this will indicate to the user that the rand crate is ready and available in Cargo.toml


coriolinus:
I don't think this is a good idea. Part of the point of this exercise is for students to practice this sort of external trait usage for themselves.

PR #1431, opened 2022-01-19
assembly-line: Capitalized Headings

PaulT89:
I noticed that all of the other headings were capitalized, so I thought the rest should probably conform to that standard


coriolinus (approved):
Looks good, thanks!

PR #1432, opened 2022-01-19
role-playing-game: Filled introduction.md with content

PaulT89:
I believe I've hit on all the points that the design.md file mentioned.  Hope this is what you're looking for.

### DIFF
--- a/exercises/concept/role-playing-game/.docs/introduction.md
+++ b/exercises/concept/role-playing-game/.docs/introduction.md
@@ -1,2 +1,81 @@
 # Introduction
 
+## Null-References
+
+If you have ever used another programming language (C/C++, Python, Java, Ruby, Lisp, etc.), it is likely that you have encountered `null` or `nil` before. 
+The use of `null` or `nil` is the way that these langugages indicate that a particular variable has no value. 
+However, this makes accidentally calling a variable that points to `null` an easy (and frequent) mistake to make. 
+As you might imagine, trying to call a function that isn't there, or access a value that doesn't exist can lead to all sorts of bugs and crashes. 
+The creator of `null` went so far as to call it his ['billion-dollar mistake'.](https://www.infoq.com/presentations/Null-References-The-Billion-Dollar-Mistake-Tony-Hoare/)
+
+## The Option Type
+
+To avoid these problems, Rust does not use null-references.
+However, it still needs a (safe) way to indicate that a particular variable has no value.
+This is where `Option` comes in.
+Instead of having a variable referencing `null` directly, Rust variables can point to the special `Option` enum.
+This enum, in turn, contains two possible variants: `None` - Rust's null-equivalent; and `Some(T)` - where T is any value of any type.
+
+It looks like this:
+
+```rust
+enum Option<T> {
+    None,
+    Some(T),
+}
+```
+
+You can think of `Option` as an added layer of safety between you and the problems that null-references can cause, while still retaining their conceptual usefulness.
+
+## Using Option
+
+Setting a variable to `None` is fairly straightforward:
+
+```rust
+let nothing: Option<u32> = None;  // Variable nothing is set to None
+```
+
+However, if you wish for the `Option` type to carry a value, you cannot assign this value directly.
+An `Option` type variable and, say, an `i32` type variable are not equivalent.
+You will need to use `Some`:
+
+```rust
+let wrong_way: Option<i32> = -4; // This will not work
+
+let right_way: Option<i32> = Some(-4); // This will work
+let another_right_way = Some(-4); // Compiler infers that this is Option<i32>
+```
+
+It's also for this reason that the following will not work:
+
+```rust
+let number = 47;
+let option_number = Some(15);
+
+let compile_error = number + option_number; // Cannot add an i32 and an Option<i32> - they are of different types
+```
+
+If you wish to get the value that is contained by Some, you will first need to check that it exists:
+
+```rust
+let mut some_words = Some("choose something to say"); // some_words set to something
+
+match some_words {
+    Some(str) => println!("Here, we will {}", str),
+    None => println!("I've got nothing to say"),
+} // Prints "Here, we will choose something to say"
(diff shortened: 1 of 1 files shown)


exercises/concept/role-playing-game/.docs/introduction.md
```diff
@@ -1,2 +1,81 @@
...
 
+## Null-References
+
+If you have ever used another programming language (C/C++, Python, Java, Ruby, Lisp, etc.), it is likely that you have encountered `null` or `nil` before. 
+The use of `null` or `nil` is the way that these langugages indicate that a particular variable has no value. 
+However, this makes accidentally calling a variable that points to `null` an easy (and frequent) mistake to make. 
+As you might imagine, trying to call a function that isn't there, or access a value that doesn't exist can lead to all sorts of bugs and crashes. 
+The creator of `null` went so far as to call it his ['billion-dollar mistake'.](https://www.infoq.com/presentations/Null-References-The-Billion-Dollar-Mistake-Tony-Hoare/)
+
+## The Option Type
```
coriolinus, line 11:
```suggestion
## The `Option` Type
```

exercises/concept/role-playing-game/.docs/introduction.md
```diff
@@ -1,2 +1,81 @@
 # Introduction
 
+## Null-References
+
+If you have ever used another programming language (C/C++, Python, Java, Ruby, Lisp, etc.), it is likely that you have encountered `null` or `nil` before. 
+The use of `null` or `nil` is the way that these langugages indicate that a particular variable has no value. 
+However, this makes accidentally calling a variable that points to `null` an easy (and frequent) mistake to make.
```
coriolinus, line 7:
```suggestion
However, this makes accidentally using a variable that points to `null` an easy (and frequent) mistake to make. 
```

exercises/concept/role-playing-game/.docs/introduction.md
```diff
@@ -1,2 +1,81 @@
...
+If you have ever used another programming language (C/C++, Python, Java, Ruby, Lisp, etc.), it is likely that you have encountered `null` or `nil` before. 
+The use of `null` or `nil` is the way that these langugages indicate that a particular variable has no value. 
+However, this makes accidentally calling a variable that points to `null` an easy (and frequent) mistake to make. 
+As you might imagine, trying to call a function that isn't there, or access a value that doesn't exist can lead to all sorts of bugs and crashes. 
+The creator of `null` went so far as to call it his ['billion-dollar mistake'.](https://www.infoq.com/presentations/Null-References-The-Billion-Dollar-Mistake-Tony-Hoare/)
+
+## The Option Type
+
+To avoid these problems, Rust does not use null-references.
+However, it still needs a (safe) way to indicate that a particular variable has no value.
```
coriolinus, line 14:
```suggestion
However, it still needs a safe way to indicate that a particular variable has no value.
```

exercises/concept/role-playing-game/.docs/introduction.md
```diff
@@ -1,2 +1,81 @@
...
+This is where `Option` comes in.
+Instead of having a variable referencing `null` directly, Rust variables can point to the special `Option` enum.
+This enum, in turn, contains two possible variants: `None` - Rust's null-equivalent; and `Some(T)` - where T is any value of any type.
```
coriolinus, line 17:
```suggestion
Instead of having a variable which lacks a value, Rust variables can use the `Option` enum.
This enum has two variants: `None`, Rust's null-equivalent; and `Some(T)`, where T is a value of any type.
```

The whole point of `Option` is that it is _not_ special; it's just a normal enum, with some handy methods and traits defined on it. You could define your own version of `Option` with its full feature set, if you really wanted to.

The thing which made null pointers into a mistake, instead of a feature, is that they are special: they're a compiler hack, a value with a special meaning. The fix is in making them into a normal instance of a normal type.

exercises/concept/role-playing-game/.docs/introduction.md
```diff
@@ -1,2 +1,81 @@
...
+It looks like this:
+
+```rust
+enum Option<T> {
+    None,
+    Some(T),
+}
+```
+
+You can think of `Option` as an added layer of safety between you and the problems that null-references can cause, while still retaining their conceptual usefulness.
```
coriolinus, line 28:
```suggestion
You can think of `Option` as a layer of safety between you and the problems that null-references can cause, while still retaining their conceptual usefulness.
```

exercises/concept/role-playing-game/.docs/introduction.md
```diff
@@ -1,2 +1,81 @@
...
+
+## Using Option
```
coriolinus, line 30:
```suggestion
## Using `Option`
```

exercises/concept/role-playing-game/.docs/introduction.md
```diff
@@ -1,2 +1,81 @@
...
+    Some(str) => println!("Here, we will {}", str),
+    None => println!("I've got nothing to say"),
+} // Prints "I've got nothing to say"
+```
+
+In Rust, there are other ways besides `match` that you can check for and access values contained within `Option`, but `match` should be familiar to you by now.
+
+Additionally, consider this a demonstration of why Rust uses `Option` instead of a null-reference.
+The point is that **you _must_ check** whether or not the `Option` variable is `Some` (in which case you can go ahead and extract and use the value contained within), or `None`.
+Anything else, and compiler will yell at you - it's trying to keep you safe from `null`.
```
coriolinus, line 81:
```suggestion
Anything else, and your program will not compile; the compiler is keeping you safe from `null`.
```

coriolinus (approved):
This is great! Thank you for putting in the work to write this out. I made several suggestions, but none of them are critical. Apply those which you think improve the text, then let me know when you're done.

PaulT89:
I liked (and committed) all of your suggestions.  I also added one more of my own - line 77 seems to flow a little more nicely than it did before.

PR #1433, opened 2022-01-20
diffie-hellman: Added Cargo.toml to config.json

PaulT89:
Hopefully, this will allow the user to add dependencies (e.g. rand, num_bigint, etc.).  Not sure how to test this for breakage other than simply committing it and testing it on Exercism.

### DIFF
--- a/exercises/practice/diffie-hellman/.meta/config.json
+++ b/exercises/practice/diffie-hellman/.meta/config.json
@@ -19,7 +19,8 @@
   ],
   "files": {
     "solution": [
-      "src/lib.rs"
+      "src/lib.rs",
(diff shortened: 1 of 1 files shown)


coriolinus (approved):
Looks good! I believe that there is not a way to test this without just committing it, but I think this should work and is unlikely to break anything badly.

PaulT89:
@coriolinus I've gone back to the Diffie-Hellman exercise - it all checks out. Cargo.toml appears, rand = "0.8.4" can be added to the dependencies, and rand::thread_rng().gen_range() can be implemented/called.  Nothing broke. 👍

PR #1439, opened 2022-01-26
reverse-string: Added Cargo.toml to config.json

PaulT89:
To complete this exercise in the browser, the user will need access to Cargo.toml in order to add dependencies.


coriolinus (approved):
Good idea.

PR #1445, opened 2022-02-04
tests: `cargo clean` before checking stubs compile

petertseng:
1.59.0-beta.6 (0426998f5 2022-02-02) runs into an internal compiler
error unless we do this.
(This has been reported to rustc in
https://github.com/rust-lang/rust/issues/93131)

It may be a good idea to keep this `clean` in the long term anyway, so
as to ensure that nothing from the "Check exercises" step unduly affects
the "Ensure stubs compile" step.

### DIFF
--- a/_test/cargo_clean_all.sh
+++ b/_test/cargo_clean_all.sh
@@ -1,0 +1,16 @@
+#!/bin/sh
+
+status=0
+repo="$(cd "$(dirname "$0")/.." && pwd)"
+
+for ex in "$repo"/exercises/*/*/; do
+  name=$(grep '^name =' "$ex/Cargo.toml" | cut -d\" -f2)
+  if [ -z "$name" ]; then
+    echo "don't know name of $ex"
+    status=1
+    continue
+  fi
+  cargo clean --manifest-path "$ex/Cargo.toml" --package "$name"
+done
+
+exit $status
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -123,6 +123,9 @@ jobs:
         run: ./_test/check_exercises.sh
         continue-on-error: ${{ matrix.rust == 'beta' && matrix.deny_warnings == '1' }}
 
+      - name: Cargo clean (to prevent previous compilation from unintentionally interfering with later ones)
+        run: ./_test/cargo_clean_all.sh
+
       - name: Ensure stubs compile
         env:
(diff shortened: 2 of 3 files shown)


_test/cargo_clean_all.sh
```diff
@@ -0,0 +1,16 @@
...
+repo="$(cd "$(dirname "$0")/.." && pwd)"
+
+for ex in "$repo"/exercises/*/*/; do
+  name=$(grep '^name =' "$ex/Cargo.toml" | cut -d\" -f2)
+  if [ -z "$name" ]; then
+    echo "don't know name of $ex"
+    status=1
+    continue
+  fi
+  cargo clean --manifest-path "$ex/Cargo.toml" --package "$name"
```
petertseng, line 13: given it only cleans the exercise package and not dependencies, I would understand objections to calling this "clean_all", in which case I would have to think of a better name. `cargo_clean_exercises`?

coriolinus:
On the one hand this should fix the current CI issues. On the other hand, it definitely slows things down, and fast CI is important for everyone.

I believe the linked issue claims that the underlying incremental compilation issue is fixed in the Rust 1.59 release, coming out in (at the time of writing) [17 days](https://www.whatrustisit.com/).

I am reluctant to permanently degrade CI times for an issue that is likely to resolve in under three weeks. I am unwilling to claim that I will remember to revert this in three weeks, if merged. 

Conclusion: I am _entirely uncertain_ as to whether this is a good idea. I reply to acknowledge that I need to think about this, and will attempt to do so in the near future.

petertseng:
Better to quantify the slowdown. Opening a demonstrative PR where all exercises are compiled.

Under ideal circumstances, the slowdown is exactly as much as would have otherwise been necessary. Because the src/lib.rs has changed from the previous compilation, it should have been necessary to recompile that in any case. In cleaning only the exercise package and crucially not any others, dependencies are not recompiled, and the hope was that there was no need to recompile those.

petertseng:
given my seeing no appreciable difference between #1446 and a successful build on `main`, I think I'd go so far as to recommend making this permanent - not even reverting it when 1.59.0 is released.

coriolinus (approved):
Thanks for this! I agree with the methodology of the test performed and the result obtained. We don't need to remember to revert this.

I don't have any problem with the name of the ci step.

PR #1451, opened 2022-02-22
resistor-color: Include Cargo.toml enabling completion via Web-UI

Lewiscowles1986:
Just an idea. Don't make people download and setup rust to learn this Rust concept via Exercism.

* This is guesstimation, no idea how to locally verify this will work, but it seems like including the file should work.

I Have completed this solution and tested locally within docker:

```
docker run --rm -it -v $HOME/Exercism/rust/resistor-color:/resistor-color rustlang/rust:nightly-bullseye bash -c 'cd /resistor-color && cargo t'
```

This gave me some confidence to what was being asked.
Really there is likely more that could and should be done to make this exercise useful for learners.

* The docs for the crates used are abysmal, so may need support documentation or upstream contribution.
* Running this without filling your machine up with `rust` dependencies is non-trivial.
* For some reason Cargo.toml is missing from web editor (this solves)
* Marking tests to be ignored... I think it is to encourage taking things one-step at a time. If this is the case then the tests may also need to be added to the editor.

### DIFF
--- a/exercises/concept/resistor-color/.meta/config.json
+++ b/exercises/concept/resistor-color/.meta/config.json
@@ -7,7 +7,8 @@
   "contributors": [],
   "files": {
     "solution": [
-      "src/lib.rs"
+      "src/lib.rs",
+      "Cargo.toml"
     ],
     "test": [
       "tests/resistor-color.rs"


coriolinus:
I'm moderately in favor of the idea of adding `Cargo.toml` to the web editor. Is there a strong reason to do it for the `resistor-color` exercise specifically instead of in the top-level `config.json` for all exercises?

Lewiscowles1986:
Hey @coriolinus I had no idea about that cascade.

coriolinus:
Cool. Closing in favor of #1458.

PR #1454, opened 2022-02-25
alphametics: add test case for too short sum

pkwiek:
It was possible to pass tests without minding width of sum word. A test
with too short sum word exists, however the equation itself is invalid
due to leading zero, causing a lucky coincidence for some solutions.
This patch adds a test which is invalid only due to sum word length.


coriolinus (approved):
Agree that in the supplied case, there can be no valid solution: `A` and `D` would need to be 0, but in a valid alphametics solution no two letters are assigned to the same digit. Adding valid test cases is always good.

Thanks for spotting this and adding the case!

PR #1455, opened 2022-02-25
simple-cipher: Update Cargo.toml to use rand 0.8

GreenPenguino:
Update to use a more recent version of `rand`

### DIFF
--- a/exercises/practice/simple-cipher/.meta/example.rs
+++ b/exercises/practice/simple-cipher/.meta/example.rs
@@ -1,11 +1,10 @@
-extern crate rand;
 use rand::Rng;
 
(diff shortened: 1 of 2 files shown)


coriolinus (approved):
Looks good, thanks!

coriolinus:
Oops, jumped the gun a little. Looks like `rand` had a breaking change in their API at some point; you'll need to update the example solution such that CI is green. Once that's set, this should be mergeable.

GreenPenguino:
@coriolinus how to restart the CI checks?

coriolinus:
CI starts automatically once you have a commit merged into the repo. I started it for you; as you can see, it works now. Thanks for doing this work!

PR #1456, opened 2022-02-26
poker: add test for flush with pair

mightyiam:
We made a bug that none of the existing tests caught. So, here's one that does.

By the way, we study Rust in the format of mob programming regularly. If anyone is interested, [learn more here](https://mobusoperandi.zulipchat.com/). It's free.

Co-authored-by: Shahar Dawn Or <mightyiampresence@gmail.com>
Co-authored-by: Preet Dua <615318+prabhpreet@users.noreply.github.com>
Co-authored-by: Roland Fredenhagen <dev@modprog.de>

### DIFF
--- a/exercises/practice/poker/tests/poker.rs
+++ b/exercises/practice/poker/tests/poker.rs
@@ -222,3 +222,8 @@ fn test_straight_flush_ranks() {
     // both hands have straight flush, tie goes to highest-ranked card
     test(&["4H 6H 7H 8H 5H", "5S 7S 8S 9S 6S"], &["5S 7S 8S 9S 6S"])
 }
+
+fn test_flush_discriminants() {
+    // both hands have straight flush, tie goes to highest-ranked card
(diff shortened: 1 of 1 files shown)


petertseng (commented):
I will support adding this test case after we've confirmed that the following question has been discussed:

a card with two 2H will only be possible to make by combining two decks. confirm that it is okay to have a test case that implies multiple decks

coriolinus:
>  confirm that it is okay to have a test case that implies multiple decks

I think this is probably fine, but the case would be better if it caught the bug without implying multiple decks. Without seeing the buggy implementation, it's hard to know what bug this actually catches.

@mightyiam could you link to the buggy implementation please, or describe the mode of failure of the buggy code? It's hard to know whether the duplicate 2H is required for the case to catch the bug or not, without knowing what the bug is.

mightyiam:
Will likely respond to this Saturday.

mightyiam:
This is irrelevant in a single-deck game, which this exercise seems to be. So, closing. Thanks.

PR #1459, opened 2022-02-28
put `Cargo.toml` in the web editor for all exercises

### LINKED ISSUE #1458: Add `Cargo.toml` to online editor for all exercises
https://github.com/exercism/rust/blob/4417a29180779257c56a7a51ca849c019aa4a28f/config.json#L22-L24

Adding it to the top-level `solution` key should handle this.

Supersedes #1451, #1453, #1455.
Replaces #1439.

There is no obvious reason why web students should not at least have access to this file. Adding dependencies will remain tricky with regard to timing, but at least students can attempt it.

coriolinus:
Closes #1458.

### DIFF
--- a/config.json
+++ b/config.json
@@ -20,7 +20,8 @@
   },
   "files": {
     "solution": [
-      "src/lib.rs"
+      "src/lib.rs",
+      "Cargo.toml"
     ],
     "test": [
       "tests/%{kebab_slug}.rs"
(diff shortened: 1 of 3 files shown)


coriolinus:
@ErikSchierboom I requested your review on the basis that you probably know better than I do how precisely the various configs json interact, but if there's a better SME, we can just tag them instead. I just want a review from _someone_ who knows precisely how the website handles these.

ErikSchierboom (commented):
Unfortunately this does not work, as the website only looks at the exercise's `.meta/config.json` file at the moment.

Issue #1468, opened 2022-03-17
Easy to pass the Palindrome Products by cheating


FelixGibson:
Hi, guys. I see the [Palindrome Products](https://exercism.org/tracks/rust/exercises/palindrome-products) problem is easy to pass the unit tests by cheating. Many people just left the insert() and value() method  empty. I think the cause is that the expected values in the unit tests rely on methods implemented as part of the exercise. Also note the public value function which is included in the template but not used/validated by any test. A example  of cheating answer is [this](https://exercism.org/tracks/rust/exercises/palindrome-products/solutions/Shawnecy).

FelixGibson:
I post a pr(#1468), which solve the problem. If there is anything about my code, please point out and I'll fix it. Highly appreciate for what the exercism team has done.

coriolinus:
Thanks for pointing this out! Agree that this exercise needs to be improved if it is this easy to complete by cheating.

We can see that the point of this exercise is to practice structs and strings: https://github.com/exercism/rust/blob/7585fdc0026211cf23fc369968247a2fb027bfc5/config.json#L725-L739

Therefore, we need to rework this exercise in a way which requires people to do the right thing. I don't think #1469 goes far enough.

What if we were to restructure the given `lib.rs` like this:

```rust
/// `Palindrome` is a newtype which only exists when the contained value is a palindrome number in base ten.
///
/// A struct with a single field which is used to constrain behavior like this is called a "newtype", and its use is
/// often referred to as the "newtype pattern". This is a fairly common pattern in Rust.
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub struct Palindrome(u64);

impl Palindrome {
    /// Create a `Palindrome` only if `value` is in fact a palindrome when represented in base ten. Otherwise, `None`.
    pub fn new(value: u64) -> Option<Palindrome> { unimplemented!() }
    
    /// Get the value of this palindrome.
    pub fn into_inner(self) -> u64 { unimplemented!() }
}

pub fn palindrome_products(min: u64, max: u64) -> Option<(Palindrome, Palindrome)> { unimplemented!() }
```

We'd then need to rework the tests in a few ways:

- Add some tests for non-palindrome numbers, ensuring that `Palindrome::new` returns `None`.
- Add some tests for palindrome numbers, ensuring that `Palindrome::new` returns `Some` and `Palindrome::new(v).expect("v is a palindrome").into_inner() == v` for all `v` where `v` is a palindrome.
[...]

PR #1469, opened 2022-03-17
Add more strict unit tests (#1468)

FelixGibson:
The palindrome products in previous version is easy to pass the unittests by cheating. The cause is that the expected values in the unit tests rely on methods implemented as part of the exercise. So I modify the unit tests by adding a fixed value. Also, I make use of the value() method which in the previous version  is useless.

Signed-off-by: felix <jjtom5f@gmail.com>


exercises/practice/palindrome-products/tests/palindrome-products.rs
```diff
@@ -3,79 +3,93 @@
...
+fn test_palindrome_new_return_some() {
+    let v = 11;
+    assert_eq!(Palindrome::new(v).expect("11 is a palindrome").into_inner(), v);
+}
+
+#[test]
+#[ignore]
+/// test the Palindrome's new method, an invalid input
+fn test_palindrome_new_return_none() {
+    assert_eq!(Palindrome::new(12), None);
```
coriolinus, line 36:
```suggestion
    for v in [12, 1232, 123432, 12345432, 54321234] {
        assert_eq!(Palindrome::new(v), None);
    }
```

Same thing, may as well test several non-palindrome numbers.

exercises/practice/palindrome-products/tests/palindrome-products.rs
```diff
@@ -3,79 +3,93 @@
...
+fn process_largest_case((from, to): (u64, u64), expected: Option<u64>) {
+    let max = palindrome_products(from, to).map(|(_ , max)| max);
+    assert_eq!(max.map(|newtype| newtype.into_inner()), expected);
 }
 
 #[test]
+/// test the Palindrome's new method, a valid input
+fn test_palindrome_new_return_some() {
+    let v = 11;
+    assert_eq!(Palindrome::new(v).expect("11 is a palindrome").into_inner(), v);
```
coriolinus, line 29:
```suggestion
    assert_eq!(Palindrome::new(v).expect("is a palindrome").into_inner(), v);
```

exercises/practice/palindrome-products/tests/palindrome-products.rs
```diff
@@ -3,79 +3,93 @@
...
-fn process_largest_case(input: (u64, u64), expected: Option<Palindrome>) {
-    let max = palindrome_products(input.0, input.1).map(|(_, max)| max);
-    assert_eq!(max, expected);
+fn process_largest_case((from, to): (u64, u64), expected: Option<u64>) {
+    let max = palindrome_products(from, to).map(|(_ , max)| max);
+    assert_eq!(max.map(|newtype| newtype.into_inner()), expected);
 }
 
 #[test]
+/// test the Palindrome's new method, a valid input
```
coriolinus, line 26:
```suggestion
/// test `Palindrome::new` with valid input
```

Just using more conventional notation to unambiguously identify the method.

coriolinus (commented):
A few suggestions but fundamentally this looks very good! You're going to need to update the example file so that CI will pass.

coriolinus:
You should also make sure to add yourself to the contributors in `.meta/config.json`.

FelixGibson:
Ok, please check it.

exercises/practice/palindrome-products/.meta/example.rs
```diff
@@ -1,69 +1,79 @@
...
     }
-
-    result
 }
-
-#[inline]
-pub fn is_palindrome(n: u64) -> bool {
-    let s = n.to_string().into_bytes();
-    s.iter().zip(s.iter().rev()).all(|(a, b)| a == b)
+fn is_palindrome(v: u64) -> bool {
```
coriolinus, line 70: Haha, this implementation totally bypasses the obvious string implementation. No need to change it.

exercises/practice/palindrome-products/.meta/example.rs
```diff
@@ -1,69 +1,79 @@
...
+            match pmin {
+                None => {
+                    panic!("something wrong");
+                },
+                Some(l) => {
+                    Some((l, r))
+                }
             }
         }
     }
```
coriolinus, line 68: Simplify.

```suggestion
    pmin.zip(pmax)
```

exercises/practice/palindrome-products/.meta/example.rs
```diff
@@ -1,69 +1,79 @@
...
+                    },
+                    Some(t) => {
+                        if t.0 < x.0 {
+                            pmax = Some(x);
                         }
-                        Some((minp, maxp))
                     }
-                };
+                }
+            }
```
coriolinus, line 51: This can be simplified, and in general it's good practice to use more expressive variable names.

```rust
if let Some(palindrome) = Palindrome::new(p) {
    let pmin = pmin.get_or_insert(palindrome);
    *pmin = pmin.min(palindrome);
    let pmax = pmax.get_or_insert(palindrome);
    *pmax = pmax.max(palindrome);
}
```

exercises/practice/palindrome-products/src/lib.rs
```diff
@@ -1,26 +1,25 @@
...
 impl Palindrome {
-    pub fn new(a: u64, b: u64) -> Palindrome {
-        unimplemented!("create a palindrome with factors ({}, {})", a, b)
-    }
-
-    pub fn value(&self) -> u64 {
-        unimplemented!("return the value of this palindrome")
+    /// Create a `Palindrome` only if `value` is in fact a palindrome when represented in base ten. Otherwise, `None`.
+    pub fn new(value: u64) -> Option<Palindrome> {
+        unimplemented!("check if the value {} is a palindrome and return Some otherwise return None", value);
```
coriolinus, line 11:
```suggestion
        unimplemented!("if the value {} is a palindrome return Some, otherwise return None", value);
```

Grammar

coriolinus (commented):
Great progress! There were some areas where we could simplify the example implementation, which should help students looking up the example.

FelixGibson:
Hi, please check it. What about now?

exercises/practice/palindrome-products/.meta/example.rs
```diff
@@ -1,65 +1,37 @@
...
+    let mut pmin: Option<Palindrome> = None;
+    let mut pmax: Option<Palindrome> = None;
+    for i in min..=max {
+        for j in i..=max {
+            if let Some(mut palindrome) = Palindrome::new(i * j) {
+                let pmin = pmin.get_or_insert(palindrome);
+                *pmin = *pmin.min(&mut palindrome);
+                let pmax = pmax.get_or_insert(palindrome);
+                *pmax = *pmax.max(&mut palindrome);
             }
```
coriolinus, line 30: Sorry, when I'd written that suggestion, I hadn't actually tried it out vs the compiler. These values should not need to be mutable. What about this?

```suggestion
            if let Some(palindrome) = Palindrome::new(i * j) {
                pmin = match pmin {
                    None => Some(palindrome),
                    Some(prev) => Some(prev.min(palindrome)),
                };
                pmax = match pmax {
                    None => Some(palindrome),
                    Some(prev) => Some(prev.max(palindrome)),
                };
            }
```

coriolinus (commented):
One more change, which was my fault. Once that's in, this should be mergeable. Thanks for sticking with this!

coriolinus:
Also, as always, CI will need to pass. It looks like the only failing check is the formatting, now; `cargo fmt` will take care of that.

FelixGibson:
Ok, how about now.

coriolinus (approved):
Thanks for bearing with all the revisions!

PR #1471, opened 2022-04-13
modify test: match information given for exercise

xTeKc:


xTeKc:
To make it easier for the reader to follow along by the examples. But, maybe it would make it too easy.

coriolinus:
I'm inclined not to merge this, but will not close it in case other maintainers disagree with me.

coriolinus:
With no further comments in a week I will now close this.

PR #1473, opened 2022-04-18
use f64 for production rate

Stigjb:
A rate can be a positive real number, it doesn't have to be a whole
number of cars. This eliminates a cast as well.


coriolinus (approved):
Sure, this is sensible. Thanks for contributing!

PR #1478, opened 2022-04-28
Add test to exercise luhn

UnlimitedHummus:
I was doing the exercise luhn and noticed that some incorrect code I wrote was passing the tests.
I was using `some_char.is_numeric()` to test whether the char is allowed. Later I used `some_char.to_digit(10).unwrap()` on allowed characters. This program passes all tests but panics, when the received string contains a numeric Unicode character such as `①`.
Where I used `some_char.is_numeric()` I should have used `some_char.is_digit(10)` to test if the Input string contains any disallowed characters. 

I hope that adding this test will prevent other  people from making the same mistake in their code.


coriolinus (approved):
Excellent idea for a test case! Thanks for improving the track.

You might also consider upstreaming this to the [`problem-specifications` repo](https://github.com/exercism/problem-specifications/).

PR #1479, opened 2022-05-01
exercism.org

MenkeTechnologies:


coriolinus (approved):
Sorry it took so long for me to review this; I didn't have time until tonight to double-check that each link produces the same result that it did before. I do note that one of the links points to `/my/solutions/...`, which won't work for anyone except the author, but that's not any more broken than it was previously.

I believe that the CI issues are unrelated to the actual content of this PR, so I'm approving now. I'll merge once the CI fix is in and I've gotten the CI for this PR green.

PR #1481, opened 2022-05-05
improve scripts' ability to ignore things which are not exercises

coriolinus:
Test scripts used to be able to assume that everything in the
`exercises/` directory was in fact an exercise, but with certain
mandatory reorganization in v3 that is no longer a safe assumption.

This change helps the scripts to identify things which are not exercises.

The lack of this PR is causing CI failures on #1479.

### DIFF
--- a/_test/check_exercises.sh
+++ b/_test/check_exercises.sh
@@ -42,7 +42,7 @@ fi
 
 return_code=0
 # An exercise worth testing is defined here as any top level directory with
-# a 'tests' directory
+# a 'tests' directory and a .meta/config.json. 
 for exercise in $files; do
    exercise="$exercise/$target_dir"
 
--- a/_test/check_exercises.sh
+++ b/_test/check_exercises.sh
@@ -42,7 +42,7 @@ fi
 
 return_code=0
 # An exercise worth testing is defined here as any top level directory with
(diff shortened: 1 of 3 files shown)


petertseng (approved):
it sort of implies that these below lines currently in `_test/check_exercises.sh should be changed:

```
# a 'tests' directory
```

https://github.com/exercism/rust/blob/33f6691bc175016e0c8aa6a0975afbacc7b4dac8/_test/check_exercises.sh#L44

petertseng (approved):
I was originally going to say this is a bit of an overcompensation because it reduces our ability to detect exercises that are erroneously created without a .meta/config.json, and we want to be able to error on those. But, upon review I find that configlet checks for that, so it is no loss, and I retract that statement. Although I feel that I would write things differently were I writing these scripts from scratch, this PR is not the time for me to talk about how I would do it differently.

coriolinus:
>  I would write things differently were I writing these scripts from scratch

Agree, I keep finding myself torn between "these scripts could all be so much better in Rust" and "compile time makes Rust a nonstarter for CI unless we can download a prebuilt binary". This implies a new repo with its own CI, which is doable but requires some setup. It won't happen soon, but maybe someday.

PR #1484, opened 2022-05-12
Update lucians-luscious-lasagna.rs Test Cases

nitincodery:
Last two test cases at ln 32, ln 38 had wrong results.


coriolinus:
Would you comment a bit more on why you believe the existing cases are wrong?

Also, I note that the CI is [failing](https://github.com/exercism/rust/runs/6401407395?check_suite_focus=true), due to the changed cases. The [example solution](https://github.com/exercism/rust/blob/ee3b16e8839fde49b178b067254cc353fab87446/exercises/concept/lucians-luscious-lasagna/.meta/exemplar.rs) is used to verify test cases.

nitincodery:
Ok It's alright, I misunderstood the last line of this statement --> "The function should return how many minutes you've worked on cooking the lasagna, which is the sum of the preparation time in minutes, and the time in minutes the lasagna has spent in the oven at the moment." --> as remaining oven time in minutes. Sorry, my bad.

PR #1488, opened 2022-05-18
queen-attack: adding explanation for ranks and files

himalayanZephyr:


coriolinus (approved):
Straightforward and explicit. Thanks for helping to improve the track!

PR #1489, opened 2022-05-24
sublist: Add test for a longer list that is not a superlist

drueck:
Adds one more test case for the `sublist` exercise to cover the case where the first list is longer than the second list, but they do not have a superlist relationship with each other.

This is similar to the case in the problem-specifications for sublist called "second list missing element from first list".  https://github.com/exercism/problem-specifications/blob/main/exercises/sublist/canonical-data.json#L158-L166

I added a test like this when I was working through the exercise myself because my initial TDD (don't write it until the tests force me to) naive implementation returned `Comparison::Superlist` any time the first list was longer than the second without bothering to actually check the contents, and that solution passed all the tests, but was definitely not correct. So I figured I'd suggest adding something along those lines to the official test suite if you all think it is appropriate.

This is my first PR to exercism in a *long* time so let me know if there is any protocol for PRs that I am not doing correctly. Thanks!

### DIFF
--- a/exercises/practice/sublist/tests/sublist.rs
+++ b/exercises/practice/sublist/tests/sublist.rs
@@ -96,6 +96,12 @@ fn superlist_at_end() {
 
 #[test]
 #[ignore]
+fn longer_list_not_superlist() {
+    assert_eq!(Comparison::Unequal, sublist(&[1, 2, 3, 4, 5], &[1, 3, 4]));
+}
+
+#[test]
+#[ignore]
 fn superlist_early_in_huge_list() {
     let huge: Vec<u32> = (1..1_000_000).collect();
 


coriolinus (approved):
Makes sense, thanks for adding this case!

Out of curiosity, why did you create this new case instead of just using the one from the problem specifications?

coriolinus:
Yeah, the theoretical ideal is that the Rust cases are a superset of the problem specifications. It's not a problem if there's some amount of divergence, which is why I just merged the PR as-is instead of requesting a more closely conforming case, but typically there will be some definite reason behind a divergence.

drueck:
The PR with the fix is here: https://github.com/exercism/rust/pull/1490  Thanks for the opportunity to correct my work.

PR #1492, opened 2022-05-26
Capitalize ASCII

rootulp:
ASCII is an abbreviation and stands for American Standard Code for Information Interchange (ref: https://en.wikipedia.org/wiki/ASCII). I think it should be capitalized in all instances in this doc.

### DIFF
--- a/concepts/string-vs-str/about.md
+++ b/concepts/string-vs-str/about.md
@@ -30,7 +30,7 @@ This derives from the following properties:
 - Rust's standard library authors therefore chose not to implement the `Index` trait for strings: while it takes a constant time to index into a byte array, it might take quite a long time to index into a string. This violates the principle of least surprise.
 
 ```rust
-let example = "I'm not entirely ascii! ❤";
+let example = "I'm not entirely ASCII! ❤";
 
 let n_chars = example.chars().count();
 let final_byte = example.as_bytes()[example.len()-1];
--- a/concepts/match-basics/about.md
+++ b/concepts/match-basics/about.md
@@ -69,7 +69,7 @@ fn ascii_type(ch: u8) -> &'static str {
         0..=31 => "non-printable control code",
         32..=126 => "printable character",
         127 => "DEL",
-        _ => "extended ascii; region-specific",
+        _ => "extended ASCII; region-specific",
     }
 }
 ```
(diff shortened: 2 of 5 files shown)


coriolinus:
Looks good, but seems to be incomplete:

```shell
$ rg -iw ascii --type rust --type markdown | rg -v ASCII
Fri 27 May 2022 09:11:43 AM CEST
exercises/practice/luhn/tests/luhn.rs:/// using ascii value for doubled non-digit isn't allowed
exercises/practice/luhn/tests/luhn.rs:/// Convert non-digits to their ascii values and then offset them by 48 sometimes accidentally declare an invalid string to be valid.
exercises/practice/luhn/tests/luhn.rs:/// using ascii value for non-doubled non-digit isn't allowed
exercises/practice/luhn/tests/luhn.rs:/// Convert non-digits to their ascii values and then offset them by 48 sometimes accidentally declare an invalid string to be valid.
concepts/string-vs-str/about.md:let example = "I'm not entirely ascii! ❤";
concepts/match-basics/about.md:        _ => "extended ascii; region-specific",
docs/archived/exercise-concepts/parallel-letter-frequency.md:  - Unicode methods vs ascii methods
docs/archived/exercise-concepts/parallel-letter-frequency.md:- Unicode: performance/feature concessions of ascii vs unicode methods. Benchmarking is skewed when benchmark uses unicode methods and the solution uses ascii methods.
```

concepts/string-vs-str/about.md
```diff
@@ -30,7 +30,7 @@ This derives from the following properties:
 - Rust's standard library authors therefore chose not to implement the `Index` trait for strings: while it takes a constant time to index into a byte array, it might take quite a long time to index into a string. This violates the principle of least surprise.
 
 ```rust
-let example = "I'm not entirely ascii! ❤";
+let example = "I'm not entirely ASCII! ❤";
```
rootulp, line 33: I confirmed that the comments in the below code block are still accurate (because capital ASCII characters take up the same number of bytes as lowercase ASCII characters)

Screen Shot 2022-05-27 at 1 07 44 PM

coriolinus: Good that you checked it, but yes: all ASCII characters take a single byte, whether encoded as ASCII or as UTF-8. Changing capitalization in either of these schemes will not affect the encoded length.

coriolinus (approved):
Thanks!

PR #1493, opened 2022-05-27
Add missing 0% success rate

vinhloc30796:
To comply better with tests. I started out by `panic!` when `speed == 0`


coriolinus:
Why do you specify a 0% success rate? The speed is already explained as a linear multiplier; at speed 0, literally any success rate will produce the correct expected output of 0 cars per hour.

vinhloc30796:
Huh. Indeed. Thank you. I'm closing the PR.

vinhloc30796:
@coriolinus Question:

From the README.md:
> At its lowest speed (`1`), `221` cars are produced each hour.

From the `tests/assembly-line.rs` file:

```rust
#[test]
fn production_rate_per_minute_at_speed_zero() {
    process_rate_per_minute(0, 0);
}
`

Should the test instead be...?

```rust
#[test]
fn production_rate_per_minute_at_speed_zero() {
    process_rate_per_minute(0, !panic); // or smth
}
```

vinhloc30796:
Actually, never mind! I missed the part where it says:

> The assembly line's speed can range from `0` (off) to `10` (maximum).

PR #1500, opened 2022-06-05
rectangles: Add test for three disjoint rectangles

hardselius:
Some solutions to this exercise will look for the top corners of the rectangles, making sure they are connected by a horizontal line, then go on to walk down the left and right side to find corners at the same row, in which case a rectangle is detected. These solutions sometimes fail to check that the bottom two corners are connected by a horizontal line. This test case will catch that.


coriolinus (approved):
Makes sense. Thanks for improving the track! Have you considered contributing this case to the exercism/problem-specifications repo as well?

hardselius:
I have not. Did not know about it. Thanks for that, will check it out!

PR #1511, opened 2022-06-08
small typo corrected. updated brackets to braces

sagarishere:
brackets are for example [], braces are {}

### DIFF
--- a/concepts/functions/introduction.md
+++ b/concepts/functions/introduction.md
@@ -1,6 +1,6 @@
 # Introduction
 
-Sometimes a certain piece of code needs to be used more than once. If that's the case, it might be convenient to put the code into a function. A function generally only performs one specific action. In Rust, the ```fn``` keyword is used to define functions. The code belonging to the function is always between braces i.e. {}. The function named `main` is special because it is the entry point for programs. From that function you can call other functions.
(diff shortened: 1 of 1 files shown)


concepts/functions/introduction.md
```diff
@@ -1,6 +1,6 @@
 # Introduction
 
-Sometimes a certain piece of code needs to be used more than once. If that's the case, it might be convenient to put the code into a function. A function generally only performs one specific action. In Rust, the ```fn``` keyword is used to define functions. The code belonging to the function is always between brackets. The function named `main` is special because it is the entry point for programs. From that function you can call other functions.
+Sometimes a certain piece of code needs to be used more than once. If that's the case, it might be convenient to put the code into a function. A function generally only performs one specific action. In Rust, the ```fn``` keyword is used to define functions. The code belonging to the function is always between braces i.e. {}. The function named `main` is special because it is the entry point for programs. From that function you can call other functions.
```
coriolinus, line 3:
```suggestion
Sometimes a certain piece of code needs to be used more than once. If that's the case, it might be convenient to put the code into a function. A function generally only performs one specific action. In Rust, the ```fn``` keyword is used to define functions. The code belonging to the function is always between braces (i.e. `{}`). The function named `main` is special because it is the entry point for programs. From that function you can call other functions.
```

Putting in the code block helps make clear that this is not part of the normal typography of the text. Putting in the parentheses improves the flow of the sentence.

kotp: Bumping this, but also in addition a comma is probably appropriate after the "i.e." as well.

coriolinus (approved):

PR #1519, opened 2022-06-27
parallel-letter-frequency: update dead links

gekh:


coriolinus (changes requested):
Thanks for spotting these and contributing a fix!

gekh:
@coriolinus replaced the link with the one you suggested.

coriolinus (approved):
Thanks again!

PR #1523, opened 2022-07-09
Revise dominoes example to pop elements with same value first.

mawenqi:


exercises/practice/dominoes/tests/dominoes.rs
```diff
@@ -153,7 +153,7 @@ fn need_backtrack() {
 #[test]
 #[ignore]
 fn separate_loops() {
-    let input = &[(1, 2), (2, 3), (3, 1), (1, 1), (2, 2), (3, 3)];
```
coriolinus, line 156: Instead of editing this test, please add a new test with your input.

mawenqi: Done as you suggest

coriolinus (approved):
Looks good, thanks for contributing!

PR #1529, opened 2022-07-23
resistor-color: changed color values from usize to u32

sinoara:
To better convey the intent, the type for color values should be an int type (I chose ``u32`` here, but any ``u*`` would do) instead of ``usize``, as it's not an index or an address.

> The primary situation in which you’d use isize or usize is when indexing some sort of collection.

from [The Rust Programming Language](https://doc.rust-lang.org/book/ch03-02-data-types.html) book

### DIFF
--- a/exercises/concept/resistor-color/src/lib.rs
+++ b/exercises/concept/resistor-color/src/lib.rs
@@ -12,11 +12,11 @@ pub enum ResistorColor {
     Yellow,
 }
 
-pub fn color_to_value(_color: ResistorColor) -> usize {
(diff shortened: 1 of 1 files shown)


coriolinus:
Sure, this is fine. We could argue that we are in fact creating an index into the enum, but I agree that a 32-bit representation makes even more sense. 

For fun, do you want to change the indices in the [example](https://github.com/exercism/rust/blob/55644b8c27b3492e13a50ab1ef42609ce649764c/exercises/concept/resistor-color/.meta/exemplar.rs) to be the RGBA values associated with each color?

coriolinus (approved):

PR #1531, opened 2022-07-26
concept/semi-structured-logs: add unimplemented message

sinoara:
Added missing _unimplemented_ message to exercise.


petertseng (approved):
I looked through history to see if there is any overriding reason to leave `unimplemented!()` blank for concept exercises, and I do not see any, so I think that means it's time to approve. The Rust track thanks you for taking the time to improve the track.

I see a few follow-up opportunities here.

First is that there are a few other files with `unimplemented!()` that could be considered.

Second is that the `allow(unused)` could be removed if the `unimplemented!` message would reference the arguments passed in to the function. Example would be https://github.com/exercism/rust/blob/main/exercises/practice/acronym/src/lib.rs. It would be good to remove the `allow(unused)`.

PR #1534, opened 2022-07-28
poker: fix broken source url

davidsilvasmith:


coriolinus (approved):

petertseng:
highly advised to make the same change to https://github.com/exercism/problem-specifications/blob/main/exercises/poker/metadata.toml, lest the Rust track lose this change on a sync, or the incorrect link persist in other tracks

PR #1535, opened 2022-07-29
poker: Ensure 222 beats AAKK

davidsilvasmith:
Implementations that add values for pairs could have higher values than low three of a kinds.


coriolinus (approved):

coriolinus:
This test seems likely to be worth upstreaming to the `problem-specifications` repo.

PR #1537, opened 2022-08-05
fix: 2nd assert example from Magazine Cutout exercise

igorluppi:
`thread 'main' panicked at 'assertion failed: can_construct_note(&magazine, &note)` due to "." in the given example from Magazine Cutout exercise.


coriolinus (approved):
Confirmed that the example fails as described. Confirmed that the suggested fix does fix it.

It would be good if this PR also added an explicit test to the suite, but that's not required.

Issue #1513, opened 2022-06-10
Xorcism tests still not running for me


bucatini-coder:
after several attempts, I got past the build dependency errors such as mentioned in #1501 and fixed in #1503, but I am now getting timeouts when the tests run.  Although this might be my solution, it runs ok and very quickly on my laptop, so I'm guessing not?  Is this a further dependency/build issue?

> Your tests timed out. This might mean that there was an issue in our infrastructure, but more likely it suggests that your code is running slowly. Is there an infinite loop or something similar?
> 
> Please check your code, and if nothing seems to be wrong, try running the tests again.

locally running the tests:
> test result: ok. 109 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

I've manually changed Cargo.toml to:
```
[dev-dependencies]
hexlit = "0.5"
rstest = "0.12"
rstest_reuse = "0.3"
```
(note #1503 has `test_reuse="0.1.0"` but that caused dependency complaints for me)
at the time of writing, I cannot see any passing solutions for this exercise in the Community Solutions.

jwaddle:
I was having missing rstest dependency issues (#1501, #1422) and fixed them by updating my Cargo.toml like above and now I get testing timeout, also.

Does this exercise work for people who download it for the first time today?  If so, is there a way to RESET this exercise?

jwaddle:
I was able to get the testing timeout even with uploading my Cargo.lock which was possible after pinning some versions in Cargo.toml:

`[dev-dependencies]
hexlit = "0.5"
quote = "=1.0.18"
proc-macro2 = "=1.0.38"
rstest = "0.12"
rstest_reuse = "0.3.0"
syn = "=1.0.93"
unicode-xid = "=0.2.3"`

locally, "$ time cargo test" shows 0.358s real time

asarkar:
I initially got test failures due to dependency versions, and then after changing them as in this post, timeout. Is there a maintainer that can comment what's going on?

ddrm86:
Same here.
I even tried to copy and paste an old passing solution (gormain's) and it does not work.

vadim2404:
Experience the same issue while uploading the solution. Locally, all tests have passed. However, on the platform, I see the time out. 

Tried to update libraries for test in this PR: #1540

I hope it will help

EarlGray:
I have the same problem of tests timing out. My program is definitely *not* slow or memory-bound:

```console
[...]

PR #1540, opened 2022-08-18
fix: update dependencies for xorcism exercise

vadim2404:
The platform can't build the solution with `rstest_reuse==0.3.0`

Decided to update all libraries to actual versions

### DIFF
--- a/exercises/practice/xorcism/Cargo.toml
+++ b/exercises/practice/xorcism/Cargo.toml
@@ -10,6 +10,6 @@ edition = "2021"
 io = []
 
 [dev-dependencies]
-hexlit = "0.5.0"
-rstest = "0.12.0"
-rstest_reuse = "0.1.0"
+hexlit = "0.5.5"
+rstest = "0.15.0"
+rstest_reuse = "0.4.0"


vadim2404:
#1513 possibly, this update will help with this issue

petertseng:
I opened this exercise in the web editor, used my xorcism implmentatin, and edited Cargo.toml to have these dependencies. the result was
```
WARNING: student did not upload Cargo.lock. This may cause build errors.
error: failed to select a version for the requirement `hexlit = "^0.5.5"`
candidate versions found which didn't match: 0.5.3
location searched: `/opt/test-runner/local-registry` index (which is replacing registry `crates-io`)
required by package `xorcism v0.1.0 (/mnt/exercism-iteration)`
perhaps a crate was updated and forgotten to be re-vendored?
As a reminder, you're using offline mode (--offline) which can sometimes cause surprising resolution failures, if this error is too confusing you may wish to retry without the offline flag.
```

I think until that is solved we shouldn't do this.

I don't see any specific versions mentioned in https://github.com/exercism/rust-test-runner/blob/main/local-registry/supported_crates so I think that's just whatever was available at the time. There was a PR to automatically update crate versions in the registry in https://github.com/exercism/rust-test-runner/pull/40, but a new approach to that one was desired and to my knowledge the new approach has not been started

vadim2404:
Does it mean that we need to rebuild and republish the container with the local registry?

remlse:
The `rstest` dependencies were removed in favor of a simpler declarative macro. I think `rstest` caused the tests to timeout, because compilation took too long. My solution for this exercise from November passed the test runner.

@vadim2404 do you think we can close this?

PR #1541, opened 2022-08-23
series: add test asking for way too many digits

NadCAtarun:
series: add a test going well beyond the edge case of `len == digits.len() + 1`

### DIFF
--- a/exercises/practice/series/tests/series.rs
+++ b/exercises/practice/series/tests/series.rs
@@ -31,3 +31,10 @@ fn test_too_long() {
     let expected: Vec<String> = vec![];
     assert_eq!(series("92017", 6), expected);
 }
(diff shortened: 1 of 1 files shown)


petertseng (approved):
Okay, good test, thank you. Please be advised that you will need to add this to https://github.com/exercism/problem-specifications/blob/main/exercises/series/canonical-data.json. If we only add this test to the Rust track, we are being selfish, and we do not want to gain that sort of reputation.

NadCAtarun:
Thank you for letting me know. I have tried to contribute to the file you
pointed out: https://github.com/exercism/problem-specifications/pull/2093
Since this is my first time touching the specifications, I hope I didn't
botch anything.

PR #1549, opened 2022-09-16
Simplify and correct the Grep exercise

hkBst:
The Grep exercise description suffers from multiple problems:
 - it implies that the task is to search for regular expressions, until one enters the editor.
 - it incorrectly states that a single string should be returned (and also gives examples that show that)
 - it leaves out information about the search order of the files

I have simplified the description and fixed these problems.

### DIFF
--- a/exercises/practice/grep/.docs/instructions.md
+++ b/exercises/practice/grep/.docs/instructions.md
@@ -1,65 +1,25 @@
 # Instructions
 
-Search a file for lines matching a regular expression pattern. Return the line
-number and contents of each matching line.
+Search files for lines matching a search string and return all matching lines.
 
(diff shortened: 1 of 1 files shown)


petertseng (commented):
these changes should be good, but they need to be made in https://github.com/exercism/problem-specifications/blob/main/exercises/grep/description.md. If the Rust track deviates, we'll miss any further improvements problem-specifications makes, disconnecting us from the wider community in both directions: We'll be being selfish and not sharing our improvements to them, and we'll be unable to use their improvements.

hkBst:
I have now created a pull request for the problem specification as suggested: https://github.com/exercism/problem-specifications/pull/2105.

PR #1550, opened 2022-09-16
Update lib.rs

hkBst:
Explain a bit better how and why the PhantomData is there and make it easier to delete by including the full path.

### DIFF
--- a/exercises/practice/circular-buffer/src/lib.rs
+++ b/exercises/practice/circular-buffer/src/lib.rs
@@ -1,11 +1,7 @@
-use std::marker::PhantomData;
-
 pub struct CircularBuffer<T> {
-    // This field is here to make the template compile and not to
-    // complain about unused type parameter 'T'. Once you start
-    // solving the exercise, delete this field and the 'std::marker::PhantomData'
-    // import.
-    field: PhantomData<T>,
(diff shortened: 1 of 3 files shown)


petertseng (commented):
that's a good change. I could merge this as is but before I do that I need to ask about the intention about the rest of the PhantomData instances:

```
exercises/practice/circular-buffer/src/lib.rs:use std::marker::PhantomData;
exercises/practice/circular-buffer/src/lib.rs:    // solving the exercise, delete this field and the 'std::marker::PhantomData'
exercises/practice/circular-buffer/src/lib.rs:    field: PhantomData<T>,
exercises/practice/protein-translation/src/lib.rs:use std::marker::PhantomData;
exercises/practice/protein-translation/src/lib.rs:    // solving the exercise, delete this field and the 'std::marker::PhantomData'
exercises/practice/protein-translation/src/lib.rs:    phantom: PhantomData<&'a ()>,
```

They can be done in this same pull request, or leave them for someone else to do.

hkBst:
I hadn't thought to look more broadly, but I have now included this same change for the other two exercises that you found.

petertseng (approved):

petertseng (commented):
The Rust track thanks you for taking the time to improve the track for all!

PR #1555, opened 2022-10-04
Additional tests for straights

hkBst:
Additional tests for straights with the same tests also for straight flushes

### DIFF
--- a/exercises/practice/poker/tests/poker.rs
+++ b/exercises/practice/poker/tests/poker.rs
@@ -151,6 +151,13 @@ fn test_aces_can_start_a_straight_low() {
 
 #[test]
 #[ignore]
+fn test_no_ace_in_middle_of_straight() {
+    // aces cannot be in the middle of a straight (Q K A 2 3)
+    test(&["2C 3D 7H 5H 2S", "QS KH AC 2D 3S"], &["2C 3D 7H 5H 2S"])
+}
+
+#[test]
+#[ignore]
 fn test_straight_ranks() {
     // both hands with a straight, tie goes to highest ranked card
     test(&["4S 6C 7S 8D 5H", "5S 7H 8S 9D 6H"], &["5S 7H 8S 9D 6H"])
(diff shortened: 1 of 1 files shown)


petertseng (approved):
OK.

Now that you have added these to the Rust track, you are the person in the best position to add these tests to https://github.com/exercism/problem-specifications/blob/main/exercises/poker/canonical-data.json. **Please do so.** We do not want the Rust track to get a reputation of being selfish and refusing to share our tests with other tracks.

exercises/practice/poker/tests/poker.rs
```diff
@@ -222,9 +229,37 @@ fn test_straight_flush_beats_four_of_a_kind() {
...
+#[test]
+#[ignore]
+fn test_aces_can_end_a_straight_flush_high() {
+    // aces can end a straight flush (10 J Q K A)
+    test(&["KC AH AS AD AC", "10C JC QC KC AC"], &["10C JC QC KC AC"])
+}
+
+#[test]
+#[ignore]
+fn test_aces_can_end_a_straight_flush_low() {
```
petertseng, line 241: I'm not sure about "end" in this name, but the same is said of `test_aces_can_end_a_straight_low`, and I guess either end is an end

petertseng:
Amending previous message:

Now that you have added these to the Rust track, you are the person in the best position to add these tests to https://github.com/exercism/problem-specifications/blob/main/exercises/poker/canonical-data.json. **Please do so.** We do not want the Rust track to get a reputation of being selfish and refusing to share our tests with other tracks.

(Why previous message was amended: Previously it said "it is your responsibility", but that doesn't make sense to say that, since the word "responsibility" to me means "someone expects you to do this" and "someone will levy consequences if you do this". That gives the wrong idea here)

exercises/practice/poker/tests/poker.rs
hkBst: It is called "aces can start a straight (A 2 3 4 5)" in the canonical-data.json, which is perhaps better...

hkBst:
As requested, I've proposed these tests also for the common test specification file: https://github.com/exercism/problem-specifications/pull/2117

petertseng (approved):
OK, very good, thank you.

Please note that I have rewritten this main branch. Please take note of the fact that that was done (and take necessary actions) when you next update/use the main branch.

PR #1561, opened 2022-10-06
resistor-color: Use int-enum 0.5.0 for compatilibity with rust-analyzer, and enum-iterator 1.2.0 for routine update

sergey-tihon:
int-enum crate was updated - https://github.com/Juici/int-enum-rs/issues/4
to be compatible with the latest rust-analyzer - https://github.com/rust-lang/rust-analyzer/issues/13347


petertseng (commented):
At first I was worried the versions would cause problems in the test runner, but then I saw that this exercise has the test runner disabled. So I guess there will be no problems updating the versions.

petertseng (approved):
OK, thank you

PR #1565, opened 2022-10-16
parallel-letter-frequency: Suppress nightly feature error

remlse:
Everything works fine from the command line, running the regular tests on the stable toolchain as well as the benchmarks on the nightly toolchain. But I got an error in vscode:

> `#![feature]` may not be used on the stable release channel

So this basically just makes that annoying error message go away.


remlse:
I'm sorry, this actually turns the benches off completely, though I don't know why.

PR #1567, opened 2022-10-20
Update bin/fetch-configlet script

kytrinyx:
The script has been tweaked to work better across all platforms.

Ref https://github.com/exercism/exercism/issues/6591

### DIFF
--- a/bin/fetch-configlet
+++ b/bin/fetch-configlet
@@ -1,5 +1,9 @@
 #!/usr/bin/env bash
 
+# This file is a copy of the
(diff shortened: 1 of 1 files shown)


ErikSchierboom (approved):

kytrinyx:
This repository has two `fetch-configlet` files.

The first is the default (which this PR is updating).

The second is https://github.com/exercism/rust/blob/main/util/exercise/src/cmd/fetch_configlet.rs, which was added in #697. The issue that raised the need for this file is #678, but it doesn't explain why it is needed.

kerolloz (approved):

PR #1569, opened 2022-10-23
resistor-color: Add explanation for disabled tests

remlse:
Issue that lead to the tests being disabled:
https://github.com/exercism/exercism/issues/6353

---

Hi there!

I've been doing a bit of mentoring for Rust lately and I've had three people ask about why the tests for resistor-color aren't running online. I suspect many others are confused, but don't ask via mentoring.

I thought it might be a good idea to include a note in the description.


remlse:
This exercise was removed when learning mode was disabled.

PR #1572, opened 2022-10-30
Added two PartialEq-related tests

kelleyvanevert:
In a silly implementation I accidentally wrote, I was able to pass all the PartialEq-related tests, even though my implementation didn't actually behave.

My implementation was like a floating point implementation, but with the data in a BigInt, but the floating point was incorrectly used in the PartialEq implementation. These tests test for that kind of mistake, where the number of digits in the decimal expansion determine the behavior of the PartialEq implementation.

### DIFF
--- a/exercises/practice/decimal/tests/decimal.rs
+++ b/exercises/practice/decimal/tests/decimal.rs
@@ -41,10 +41,26 @@ fn test_gt() {
 
 #[test]
 #[ignore]
+fn test_gt_2() {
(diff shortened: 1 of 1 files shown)


remlse (changes requested):
Looks good!

nit: I would combine these two added tests and call them something like `test_unequal_number_of_decimal_places`, since the purpose is not to test `gt` or `lt` functionality.

senekor:
superseded by https://github.com/exercism/rust/pull/1726

PR #1573, opened 2022-10-30
Fix typo in magazine test

EliseBriers:
test_magzine_has_more_than_words_available_than_needed() -> test_magazine_has_more_than_words_available_than_needed()


bobahop:
Nice catch!

Perhaps it should be
`test_magazine_has_more_words_available_than_needed`?
instead of
`test_magazine_has_more_than_words_available_than_needed`

EliseBriers:
> Nice catch!
> 
> Perhaps it should be `test_magazine_has_more_words_available_than_needed`? instead of `test_magazine_has_more_than_words_available_than_needed`

Indeed, that makes sense!

bobahop (approved):
Looks good to me.

PR #1581, opened 2022-11-05
Health Statistics Concept Exercise Fix

saurabhmshr:
In this branch:

+ Setter methods for age and weight are renamed to follow [Rust API guidelines](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter). They indicate mutability, which is idiomatic in Rust.

 + Added `#[ignore]` to `test_name` as it was missing (and failing).

### DIFF
--- a/exercises/concept/health-statistics/tests/health-statistics.rs
+++ b/exercises/concept/health-statistics/tests/health-statistics.rs
@@ -5,6 +5,7 @@ const AGE: u32 = 89;
 const WEIGHT: f32 = 131.6;
 
 #[test]
+#[ignore]
 fn test_name() {
     let user = User::new(NAME.into(), AGE, WEIGHT);
     assert_eq!(user.name(), NAME);
@@ -29,7 +30,7 @@ fn test_weight() {
 fn test_set_age() {
     let new_age: u32 = 90;
     let mut user = User::new(NAME.into(), AGE, WEIGHT);
-    user.set_age(new_age);
+    user.age_mut(new_age);
     assert_eq!(user.age(), new_age);
 }
 
@@ -38,6 +39,6 @@ fn test_set_age() {
 fn test_set_weight() {
     let new_weight: f32 = 129.4;
     let mut user = User::new(NAME.into(), AGE, WEIGHT);
-    user.set_weight(new_weight);
+    user.weight_mut(new_weight);
     assert!((user.weight() - new_weight).abs() < f32::EPSILON);
 }
--- a/exercises/concept/health-statistics/tests/health-statistics.rs
+++ b/exercises/concept/health-statistics/tests/health-statistics.rs
@@ -5,7 +5,6 @@ const AGE: u32 = 89;
 const WEIGHT: f32 = 131.6;
 
 #[test]
-#[ignore]
 fn test_name() {
     let user = User::new(NAME.into(), AGE, WEIGHT);
     assert_eq!(user.name(), NAME);
--- a/exercises/concept/health-statistics/.docs/hints.md
+++ b/exercises/concept/health-statistics/.docs/hints.md
@@ -26,7 +26,6 @@ fn foo() -> i32 {
 
 ## 3. Implement the setter methods
 
-- The `set_age()` and `set_weight()` methods are setters, responsible for updating the corresponding field on a struct instance with the input argument.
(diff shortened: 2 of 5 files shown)


bobahop:
I think you may need to look at the [exemplar file ](https://github.com/exercism/rust/blob/main/exercises/concept/health-statistics/.meta/exemplar.rs)and make changes there, too.

Also 
- [hints](https://github.com/exercism/rust/blob/main/exercises/concept/health-statistics/.docs/hints.md)
- [instructions](https://github.com/exercism/rust/blob/main/exercises/concept/health-statistics/.docs/instructions.md)

saurabhmshr:
Done!

exercises/concept/health-statistics/tests/health-statistics.rs
```diff
@@ -5,6 +5,7 @@ const AGE: u32 = 89;
 const WEIGHT: f32 = 131.6;
 
 #[test]
+#[ignore]
```
bobahop, line 8:
```suggestion

```
We don't want to ignore the first test, I believe.

bobahop (changes requested):
It all looks good, except the added `#ignore` for the first test shouldn't be there, I think.

exercises/concept/health-statistics/tests/health-statistics.rs
saurabhmshr: Thanks Bob! The first test is now active.

Note to self: From the [document](https://github.com/exercism/legacy-docs/blob/main/language-tracks/exercises/anatomy/test-suites.md) linked in the [exercise tests](https://github.com/exercism/rust#exercise-tests) section

> In order to provide a better user experience, some language tracks only leave the first test in an exercise test suite active. All subsequent tests are marked as "skipped" or "pending" using whatever method or directive that is provided by the testing framework. This means that the person solving the exercise must manually edit the test suite to activate the tests one by one as they write their solution.

bobahop (approved):
Looks good to me now. I don't currently have write permissions to this repository, so I'm unable to merge this for you.

petertseng:
These `_mut` methods violate conventions. `_mut` methods by convention return `&mut`, which and none of these methods do that. These methods should be named `set_age` and `set_weight`, as https://github.com/rust-lang/rfcs/blob/master/text/0344-conventions-galore.md#gettersetter-apis tells us that a setter for `foo` should be named `set_foo`.

I recommend reversing the changes made in this pull request.

saurabhmshr:
@petertseng can you point me towards official documentation instead of an RFC ? The closest I could find was Rust API guidelines, which I linked in the first comment. From what I've read, on various forums - users.rust-lang, stack overflow, etc. getters and setters are not idiomatic in Rust.

petertseng:
> can you point me towards official documentation instead of an RFC ? 

Yes, I will do that. Here it is: https://doc.rust-lang.org/1.0.0/style/style/naming/README.html#getter/setter-methods-[rfc-344]
Please note that the 1.0.0 refers to the Rust version, and the latest version this URL can be changed to that still includes this style guide is 1.12.0. https://doc.rust-lang.org/nightly/style-guide/ is silent on setters.

---

> The closest I could find was Rust API guidelines, which I linked in the first comment. From what I've read on various forums - users.rust-lang, stack overflow, etc. getters and setters are not idiomatic in Rust.

Those very same guidelines say to consider getters and setters in https://rust-lang.github.io/api-guidelines/future-proofing.html#structs-have-private-fields-c-struct-private:

> Public fields are most appropriate for struct types in the C spirit: compound, passive data structures. Otherwise, consider providing getter/setter methods and hiding fields instead.

---

> Should I make further changes for getter fields to return the `&Object` and setter fields to return `&mut Object` ?

I don't recommend it because let's consider the effect on the tests. Currently, we have `user.set_age(new_age)`. If instead we returned `&mut`, the tests would then look like `*user.age_mut() = new_age` which exposes the field more, bringing the API farther away from following the advice in https://rust-lang.github.io/api-guidelines/future-proofing.html#structs-have-private-fields-c-struct-private.

However, one thing we could consider is that this health data is indeed a compound passive data structure, which means that a direction to consider is using direct field access: `user.age = new_age`.

The thing to consider is what to teach in this exercise. If we expose students to direct field access it would be good to also explain to them what sorts of situations it is or is not appropriate.

PR #1582, opened 2022-11-07
health-statistics: name setters `set_foo`, not `foo_mut`

petertseng:
This reverts commit 6e467864f15b05baf715bb8dfc075c3018d5dfbf.

By the conventions expressed in
https://github.com/rust-lang/rfcs/blob/master/text/0344-conventions-galore.md#gettersetter-apis a setter for `foo` must be named `set_foo`.

Methods named `_mut` by convention return `&mut`, and neither `age_mut` nor `weight_mut` do that. Therefore to name the `_mut` violates convention. We should not violate convention.


bobahop (approved):
LGTM

PR #1583, opened 2022-11-07
low-power-embedded-game: change test order for clarity

antonmry:
In this exercise, it's easy to think you must work with values instead of their position. The README only states it in the title:

> ## 2. Choose even-positioned items from an iterator

And the first tests are about sequences of numbers, so it makes sense. I wasn't the only one who made that mistake:

https://stackoverflow.com/a/71145392
https://stackoverflow.com/questions/69032551/in-rust-how-can-i-restrict-a-generic-t-to-allow-modulus

I propose to change the order of tests, so the test with strings is first. This should be easy, and it doesn't add more tests.

### DIFF
--- a/exercises/concept/low-power-embedded-game/tests/low-power-embedded-game.rs
+++ b/exercises/concept/low-power-embedded-game/tests/low-power-embedded-game.rs
@@ -40,6 +40,15 @@ mod evens {
 
     #[test]
     #[ignore]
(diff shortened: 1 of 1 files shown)


remlse (approved):

senekor:
superseded by https://github.com/exercism/rust/pull/1725

PR #1587, opened 2022-11-09
Update instructions for resistor-color

A5308Y:
Thanks for maintaining!

I wasn't sure what task number one and task number three are exactly. I think numbering them in the list explicitly will increase clarity.


petertseng:
Indeed. Thanks!

PR #1590, opened 2022-11-14
xorcism: remove rstest dependency (exercism#1513)

remlse:
Refers to [this issue](https://github.com/exercism/rust/issues/1513#issuecomment-1314025738).

rstest uses proc macros, which make the tests timeout due to long compile times. Replace rstest with a custom declarative macro.

Drawbacks:
- more indentation
- module structure of tests is flipped around

both of those seem minor to me.

Although declarative macros can be hard to read for those unfamiliar, that was already somewhat the case with `rstest`'s magic in my opinion. So I personally don't think it's worse in terms of the users being able to understand the tests.

The only other alternative I see is to disable the online tests altogether and leave a note about that in the exercise description. That probably wouldn't be that bad, since people solving this hard exercise most likely have a solid local setup. But it would still be cool to run the tests online as well.

Thanks for taking a look!

### DIFF
--- a/exercises/practice/xorcism/Cargo.toml
+++ b/exercises/practice/xorcism/Cargo.toml
@@ -11,5 +11,3 @@ io = []
 
 [dev-dependencies]
 hexlit = "0.5.0"
-rstest = "0.12.0"
-rstest_reuse = "0.1.0"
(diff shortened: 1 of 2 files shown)


remlse:
Oh and about the time, running the tests clean with `rstest` takes about 7.5 seconds on my local machine. With the declarative macro, that's down to about 0.8. So I'm optimistic this will actually fix the issue.

petertseng (approved):
Even though I think the previous module nesting was more meaningful, the compilation time is more important here. We're grateful you've taken the time to make this contribution.

PR #1595, opened 2022-11-17
labels, CODE_OF_CONDUCT, fetch-configlet: sync from org-wide-files

petertseng:
see individual commit messages.

these commits **MUST NOT** be squashed when merging


kytrinyx (approved):

petertseng:
That is a good point!  I will remove the closes: lines because they do not close

PR #1596, opened 2022-11-19
tuples: Fix broken link

atdt:


ben-grossmann (approved):

petertseng (approved):
My opinion is it's a little weird that the first sentence under that heading is "When a tuple struct has only one element...", because it doesn't tell us what a tuple struct is. However, this is not a problem in this context since other materials associated with this concept will tell us what a tuple struct is, or the bulleted list at the top of that same page says what a tuple struct is.

petertseng:
I may be able to get the tests to run by closing and reopening the pull request.

Issue #1585, opened 2022-11-08
Feedback Needed for Concept Exercise: Bird Watcher


ghost:
I'm working on a concept exercise "Bird Watcher" for the Rust track and would like feedback.

This exercise covers vectors and some of its common functionality. It's a combination of the same exercise present in the [Go][1] and [Java][2] tracks.

Specifically, it covers the following tasks:

+ Total bird count
+ Birds in week
+ Fix a counting mistake
+ Birds visited in the first N days
+ Number of busy days
+ Day without birds

I'm creating and executing a test binary locally (using `rustc --test ...`) for the following:

```
fn bird_count(observation_log: Vec<u16>) -> u16 {
    let total: u16 = observation_log.iter().sum();
    total
}

fn birds_in_week(observation_log: Vec<u16>, week: usize) -> u16 {
    let start = (week - 1) * 7;

    // slice from vector
    let period = &observation_log[start..start + 7];
    // let period = &observation_log[start..=start + 6];
    let total: u16 = period.iter().sum();
    total
}

fn birds_in_first_days(observation_log: Vec<u8>, days: usize) -> u8 {
    let birds = observation_log.iter().take(days).sum();
    birds
}

fn fix_bird_count(mut observation_log: Vec<u16>) -> Vec<u16> {
    // bird was in the garden on the first day and every other day thereafter
    // increment the reported count on odd days (even indices)
    for bird_count in observation_log.iter_mut().step_by(2) {
        *bird_count += 1;
    }
    observation_log
}

fn birds_today(observation_log: Vec<u8>) -> u8 {
    // check if Option<f64> returned by last() is Some(val)
    if let Some(birds) = observation_log.last() {
        *birds
    } else {
        0
    }
}

fn day_without_birds(observation_log: Vec<u8>) -> bool {
    observation_log.contains(&0)
}

fn busy_days_number(observation_log: Vec<u8>) -> u8 {
    let mut days = 0;
    for birds in observation_log.iter() {
        if birds >= &5 {
            days += 1;
        }
    }
    days
}

#[test]
fn test_bird_count() {
    let observations = [
        (vec![9, 0, 8, 4, 5, 1, 3], 30, "7 day bird visit log"),
        (vec![2], 2, "1 day bird visit log"),
        (
            vec![17, 15, 20, 12, 14, 21, 23, 20, 21, 19, 16, 22, 26, 11],
            257,
            "14 day bird visit log",
        ),
    ];

    for (log_period, total, description) in observations {
[...]

PR #1597, opened 2022-11-20
Concept Exercise: Bird Watcher

saurabhmishra-io:
Implementation of the issue I raised: https://github.com/exercism/rust/issues/1585#issue-1439299114

Exemplar solution may not be idiomatic. 

Please provide suggestions if any.


saurabhmshr:
Hi @bobahop / @petertseng can you provide me feedback for this PR ?

bobahop:
Thanks for contributing! I'm sorry, but I won't be able to get to this anytime soon.

saurabhmshr:
Sure @bobahop . Do you think its a good idea for me to create PRs based on exercises that exist in other tracks (Go, Python, Java) but not Rust ?

bobahop:
@saurabhmshr Sure, it's fine to transpose an exercise from one language to another, making sure to credit the original author in the config.json. Also, when doing so, the [problem-specifications repository](https://github.com/exercism/problem-specifications) should be used if the exercise is defined there. For bird watcher, it apparently is not.

PR #1601, opened 2022-11-28
armstrong-numbers: add larger testcases

nneonneo:
Add some extra Armstrong number tests for larger numbers. In particular, the ten-digit solutions can cause naive solutions to overflow; it might be reasonable to gate them behind a feature flag or bonus exercise.

The test case 4106098957 is particularly interesting: if the addition is done using `wrapping_add`, this number will appear as an Armstrong number, even though it really isn't. This is because the sum is equal to exactly 2^32 + 4106098957. This might be a worthwhile testcase to add to other languages too to check for overflow bugs. (Rust should panic in debug mode, but this will also catch folks who try to use wrapping_add to get around the panic).

### DIFF
--- a/exercises/practice/armstrong-numbers/tests/armstrong-numbers.rs
+++ b/exercises/practice/armstrong-numbers/tests/armstrong-numbers.rs
@@ -52,3 +52,27 @@ fn test_seven_digit_armstrong_number() {
 fn test_seven_digit_non_armstrong_number() {
     assert!(!is_armstrong_number(9_926_316))
 }
+
+#[test]
+#[ignore]
+fn test_nine_digit_armstrong_number() {
+    assert!(is_armstrong_number(912_985_153));
+}
+
+#[test]
+#[ignore]
+fn test_nine_digit_non_armstrong_number() {
+    assert!(!is_armstrong_number(999_999_999));
+}
+
+#[test]
+#[ignore]
+fn test_ten_digit_non_armstrong_number() {
+    assert!(!is_armstrong_number(3_999_999_999));
+}
+
+#[test]
+#[ignore]
+fn test_ten_digit_fake_armstrong_number() {
+    assert!(!is_armstrong_number(4_106_098_957));
+}


exercises/practice/armstrong-numbers/tests/armstrong-numbers.rs
```diff
@@ -52,3 +52,27 @@ fn test_seven_digit_armstrong_number() {
...
+
+#[test]
+#[ignore]
+fn test_ten_digit_non_armstrong_number() {
+    assert!(!is_armstrong_number(3_999_999_999));
+}
+
+#[test]
+#[ignore]
+fn test_ten_digit_fake_armstrong_number() {
```
petertseng, line 76: let's make it clearer what "fake" means - your commit message already explains it may appear as an armstrong number under `wrapping_add`, but this commit message wouldn't be visible to a student solving this exercise without knowledge of this repo. So I would think this test should be renamed or a comment should be added

petertseng (commented):
in times of peace, these tests should be submitted to https://github.com/exercism/problem-specifications/blob/main/exercises/armstrong-numbers/canonical-data.json. But, https://github.com/exercism/problem-specifications/blob/main/.github/workflows/pause-community-contributions.yml is going to close it and redirect it to https://forum.exercism.org. I do not know the exact process to do that, but I see a previous thread in that vein: https://forum.exercism.org/t/canonical-resistor-color-trio-testing-gap-values-over-1000-which-are-not-kiloohms/1434. I encourage you to seek out this avenue to get your tests added to https://github.com/exercism/problem-specifications/blob/main/exercises/armstrong-numbers/canonical-data.json, but I cannot force you to.

For the Rust track specifically, let's make it clear what "fake" means, and let's go with what you have now.

exercises/practice/armstrong-numbers/tests/armstrong-numbers.rs
nneonneo: Done.

petertseng:
I must have forgotten to check this earlier. https://github.com/exercism/rust/blob/main/exercises/practice/armstrong-numbers/.meta/example.rs will need to be updated to handle these cases. Let's get that done, and that will be all that's needed.

nneonneo:
Done.

petertseng:
thanks!

PR #1602, opened 2022-12-04
Reformat exercise config files

kytrinyx:
This runs the `bin/configlet fmt --update` command to
reformat the exercise configs.

### DIFF
--- a/exercises/concept/assembly-line/.meta/config.json
+++ b/exercises/concept/assembly-line/.meta/config.json
@@ -1,5 +1,4 @@
 {
-  "blurb": "Learn about numbers while working on an assembly line for cars.",
   "authors": [
     "LewisClement",
     "efx"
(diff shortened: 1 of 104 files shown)


ErikSchierboom (approved):

petertseng (approved):

kytrinyx:
@petertseng If you're happy with this, would you merge it?

petertseng:
My apologies, I must have failed to communicate. Let me fix that now and see if it opens up a path for us to work together more smoothly in the future:

The reason I didn't merge it was only because I was used to working in contexts (other organisations, etc.) where the person submitting the pull request is the one responsible for merging it. After all, they know when they want it, and they have the incentive. If I Approve a pull request and have no comments to add I am happy for the submitter to merge it whenever they want

However! I want to be respectful of your time, and in pull requests like this where it's clear you're submitting a pull request to every track, your time is valuable and it is best for us as maintainers to be courteous and avoid spreading you too thin. Specifically, you're working with dozens of tracks, whereas each track is only working with one of you. It's more much efficient for each track to work along your conventions, rather than burdening you with having to remember dozens of tracks' conventions, which may all differ from each other and in some cases might not have even been written down!

PR #1607, opened 2022-12-10
test: add unit test for sum-of-multiples

SethFalco:
I'm learning Rust, and was wondering why my code was working when I realized `reduce` is meant to return `None` when the iterable is empty.

I did `(0..limit).reduce()` in my answer, but there was no test that passed `0` as the limit, so the case that would cause this error never occurred.

Since 0 is a valid u32 integer, perhaps we should have a test that expects a limit of 0 to return 0 rather than be undefined and untested?

Before adding this test, my solution passed all tests but would throw an error on 0.

### DIFF
--- a/exercises/practice/sum-of-multiples/tests/sum-of-multiples.rs
+++ b/exercises/practice/sum-of-multiples/tests/sum-of-multiples.rs
@@ -85,6 +85,12 @@ fn the_only_multiple_of_0_is_0() {
 
 #[test]
 #[ignore]
+fn a_limit_of_0_always_returns_0() {
+    assert_eq!(0, sum_of_multiples(0, &[]))
+}
+
+#[test]
+#[ignore]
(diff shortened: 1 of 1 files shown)


petertseng (commented):
When community contributions are reopened these tests should be submitted to https://github.com/exercism/problem-specifications/blob/main/exercises/sum-of-multiples/canonical-data.json instead of only to the Rust track. The reason is we do not want to gain a reputation for being selfish and hogging improvements only to ourselves. We give back to others and share with others.

petertseng:
When community contributions are reopened these tests should be submitted to https://github.com/exercism/problem-specifications/blob/main/exercises/sum-of-multiples/canonical-data.json instead of only to the Rust track. The reason is we do not want to gain a reputation for being selfish and hogging improvements only to ourselves. We give back to others and share with others.

PR #1608, opened 2022-12-12
Revert the configlet fmt on resistor-color

kytrinyx:
The configlet fmt command removed custom key-value pairs.
This adds the unformatted file back for resistor-color, which
looks like it is the only one where this happened.

See https://forum.exercism.org/t/online-test-timeout-with-recommended-solution/1574


petertseng (approved):
huh you're sure you want to undo the entire act of formatting, as opposed to only adding the key back? Well, I have no reason to prefer one act or the other

PR #1612, opened 2022-12-25
pig-latin: more tests

krolaw:
Adds tests to pig-latin to catch panics in my solutions, where a word is 1 or 0 letters long.


petertseng (commented):
When community contributions are reopened these tests should be submitted to https://github.com/exercism/problem-specifications/blob/main/exercises/pig-latin/canonical-data.json instead of only to the Rust track. The reason is we do not want to gain a reputation for being selfish and hogging improvements only to ourselves. We give back to others and share with others.

PR #1614, opened 2023-01-02
docs: create an introduction for Dig Deeper in the Lucian's Luscious Lasagna exercise

Seven0492:
I don't have any special approach ideas since this is an extremely simple exercise.

Except maybe showcasing different ways to perform basic math operations?

### DIFF
--- a/exercises/concept/lucians-luscious-lasagna/.approaches/introduction.md
+++ b/exercises/concept/lucians-luscious-lasagna/.approaches/introduction.md
@@ -1,0 +1,23 @@
+# Introduction
+
+The key to this exercise are handling basic math operations (Division, addition, etc), and returning their results.
+
+## General guidance
+
+Adding a semi-colon(`;`) at the end of an [expression][expressions-doc-video], changes it to a declaration, or in other words, a statement.
+
+```exercism/note
+Statements do not return anything.
+```
+
+A question to ask yourself when writting your code:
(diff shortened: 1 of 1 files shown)


remlse (commented):
> I don't have any special approach ideas since this is an extremely simple exercise.

I'm wondering what value an approach provides in this case. Most of the content would fit into a hints.md file at best.

Side note: With the entire learning mode, this exercise is currently also disabled.

MatthijsBlom:
Some not too sharply defined thoughts:

- I imagine that a _Dig Deeper_ tab present at more exercises might lead to more students noticing and/or looking for the _Dig Deeper_ feature. It seems to me that currently many students are having trouble with this.

- Approaches to concept exercises could perhaps be used to provide advice that is very commonly given in mentoring sessions. (The Analyzer could/should do this as well.)

- Most of the interesting stuff that might normally go in the Approaches should probably go into the relevant concept's `about.md`.

senekor:
Regarding the problem that students skip over the approaches, I'm not sure the best way to solve that is to simply have more approaches. If the first approach a student reads is uninteresting - because the exercise doesn't lend itself to interesting approaches - then I would expect that student to become even less likely to read approaches in the future. Maybe the feature just needs to be advertised more on those exercises that do have it.

MatthijsBlom:
I agree; 'Approaches' should not be added _just_ for the sake of having more _Dig Deeper_ tabs.

PR #1615, opened 2023-01-04
Exercise-> Practice-> Added Dikjstra Algorithm

Irene-123:
Hello there, I've added an example of the implementation of the Dijkstra algorithm to find the shortest path between 2 nodes. Please review and let me know thank you so much!


petertseng:
When community contributions are potentially opened again in the future, this should be submitted to https://github.com/exercism/problem-specifications, not just to the Rust track. We don't want to be selfish and hog the exercise only to ourselves; we give back to others and share with others.

Please note that because of https://github.com/exercism/problem-specifications/issues/761, when this exercise is eventually submitted to problem-specifications, its name should be reworded to be "shortest path", instead of naming the specific algorithm Dijkstra's.

PR #1616, opened 2023-01-05
Update forth.rs test case

anvad:
Update the `can_use_different_words_with_the_same_name` test so it tests the case where the initial definition is not a number. I see some implementations that pass the existing test because they eagerly check whether the definition maps to a number. These implementations will fail with the updated test.


petertseng:
When community contributions are reopened in the future, this should be submitted to https://github.com/exercism/problem-specifications/blob/main/exercises/forth/canonical-data.json. The reason is we don't want to be selfish and only keep this improved test case to ourselves. We give back to others and share with others.

PR #1618, opened 2023-01-22
hello-world: Show how to run the tests in the instructions

TomasHubelbauer:
Hi, this is an update to the Hello World exercise of the Rust track. The track never introduces how to run tests to the user. It is described in the HELP file but I believe that file is meant for you to check when you get lost solving the exercise? It should continue to contain the more detailed information on how to run tests IMO but the basic command I believe should be a part of the README.

### DIFF
--- a/exercises/practice/hello-world/.docs/instructions.append.md
+++ b/exercises/practice/hello-world/.docs/instructions.append.md
@@ -1,0 +1,1 @@
+In the Rust track, tests are run using the command `cargo test`.


petertseng:
Since instructions.md has to be the exact same text as https://github.com/exercism/problem-specifications/blob/main/exercises/hello-world/description.md, it won't be possible to include language-specific instruction like `cargo test` in here, but I think it would be appropriate to instead create the file `exercises/practice/hello-world/.docs/instructions.append.md` saying something like the following:

In the Rust track, tests are run using the command `cargo test`.

That will get these instructions to appear on the instructions page, and not violate the rule about instructions.md have to be exactly the same as problem-specifications!

petertseng (approved):

PR #1622, opened 2023-02-02
clippy: fix new warnings

remlse:
Most of the warnings relate to idiomatic format strings. I think these warnings were introduced with Rust 1.67.

```
variables can be used directly in the `format!` string
for further information visit
https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args
`#[warn(clippy::uninlined_format_args)]` on by default
```

There is one other warning I fixed about lifetimes that can be elided.

There are still a couple warnings about a `new` function without a default implementation. I didn't do anything about those because it seemed unclear to me what the best thing to do is.


ErikSchierboom:
@remlse I can't re-open this PR. Could you open another PR?

PR #1624, opened 2023-02-11
Add PhantomData explaination to passio

nhawkes:


exercises/practice/paasio/src/lib.rs
```diff
@@ -1,5 +1,9 @@
 use std::io::{Read, Result, Write};
 
+// the PhantomData instances in this file are just to stop compiler complaints
+// about missing generics; feel free to remove them
+
+
```
petertseng, line 6:
```suggestion

```

there's an extra newline that should be removed.

petertseng (approved):

exercises/practice/paasio/src/lib.rs
petertseng: since I cannot remove the extra newline, I cannot merge this

nhawkes:
Removed newline: #1637

PR #1626, opened 2023-02-21
Add secret handshake exercise

dem4ron:


exercises/practice/secret-handshake/.meta/example.rs
```diff
@@ -0,0 +1,18 @@
+const COMMANDS: [&str; 4] = ["wink", "double blink", "close your eyes", "jump"];
+
+pub fn commands(n: u8) -> Option<Vec<&'static str>> {
```
ErikSchierboom, line 3: I'm not sure about returning an `Option` value. I think returning an empty vector in the case of no commands would be fine. Option is usually only used when something can be absent, and I don't know if that is the case here.

ErikSchierboom (commented):
Really nice! One comment on the usage of `Option`. 

I'll ask @petertseng for a proper Rust review.

ErikSchierboom (approved):

config.json
```diff
@@ -1466,6 +1466,15 @@
...
       },
+      {
+        "slug": "secret-handshake",
+        "name": "Secret Handshake",
+        "uuid": "8c044530-9deb-4ff7-a638-98d6c05ebbb8",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 4,
+        "topics": []
+      },
```
petertseng, line 1477: Oh, I thought that https://github.com/exercism/rust/blob/89459bf584f5154e7ec59f8f36b1d603b1a68177/_test/verify_exercise_difficulties.sh#L48-L54 was supposed to ensure that difficulties are sorted, but it got commented out. That is frustrating

exercises/practice/secret-handshake/src/lib.rs
```diff
@@ -0,0 +1,3 @@
+pub fn commands(n: u8) -> Vec<&'static str> {
```
petertseng, line 1: how's everyone feeling about the name `commands` here? It seems to be an awkward fit for the situation of giving a handshake to a friend, since I'm not a general commanding my armies. I am thinking of `actions` instead, because our secret handshake consists of a number of actions.

config.json
petertseng: Okay, since that check is commented out, there is no order requirement. No action needed. This exercise can stay where it is.

exercises/practice/secret-handshake/src/lib.rs
ErikSchierboom: Actions could work well!

dem4ron: Thanks for the feedback. I used the JS exercise as a reference, where they use `commands`, but `actions` do make a lot of sense and probably is a better fit.

petertseng (approved):
In that case, with that being the only applicable and necessary change I saw, let's go then. Thank you for your hard work.

PR #1632, opened 2023-02-27
Sync gigasecond docs with problem-specifications

kytrinyx:
The gigasecond exercise has been overhauled as part of a project
to make practice exercises more consistent and friendly.

For more context, please see the discussion in the forum, as well as
the pull request that updated the exercise in the problem-specifications
repository:

- https://forum.exercism.org/t/new-project-making-practice-exercises-more-consistent-and-human-across-exercism/3943
- https://github.com/exercism/problem-specifications/pull/2206

### DIFF
--- a/exercises/practice/gigasecond/.docs/instructions.md
+++ b/exercises/practice/gigasecond/.docs/instructions.md
@@ -1,6 +1,8 @@
 # Instructions
 
-Given a moment, determine the moment that would be after a gigasecond
-has passed.
+Your task is to determine the date and time one gigasecond after a certain date.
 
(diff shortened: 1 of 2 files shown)


petertseng (approved):
I like this. Are you ready? If some other maintainer discovers a necessary change, then if I merge this too early then you will be forced to make a second pull request to submit that change rather than be able to combine it with this one. If you think the risk of that is sufficiently small, or you think making a second pull request is not a problem for you, I'm ready when you are.

petertseng:
It doesn't currently look like anyone else has submitted any further changes, so at this point I am going to presume it is safe to take this as-is.

PR #1636, opened 2023-03-02
Secret Handshake approach

bobahop:
An approach for Secret Handshake.


bobahop:
@ErikSchierboom: Rust approach for Secret Handshake ready for review.

exercises/practice/secret-handshake/.approaches/iterate-once/content.md
```diff
@@ -0,0 +1,97 @@
...
+        if action == end {
+            break;
+        }
+        if (n & (1 << action)) != 0 {
+            output.push(SIGNS[action as usize])
+        }
+        action += action_incr
+    }
+    output
+}
```
ErikSchierboom, line 24: I must admit that I'm not well-versed in Rust, but to me this is very hard to read. Is this really an idiomatic way to solve the exercise? What about something simple like: https://exercism.org/tracks/rust/exercises/secret-handshake/solutions/aciba90?

bobahop: There are simpler approaches which don't offer as much in pedagogic value. Since I only did one approach, I used one which is not exactly unidiomatic, but which afforded more in the way of explanation.

ErikSchierboom (approved):

exercises/practice/secret-handshake/.approaches/iterate-once/content.md
ErikSchierboom: Right. Okay!

ErikSchierboom: It _is_ an interesting approach

PR #1642, opened 2023-03-08
poker: improve test robustness for ranking pairs/triplets

J-Kappes:
Increased rank of losing hand's lone cards or decreased rank of winning hand's pair/triplet.
This allows the tests to fail if the triplets'/pairs' ranks aren't compared first.

Modified-by: Joel Kappes <jkappes@comtexing.de>


petertseng (commented):
these need to be submitted to https://github.com/exercism/problem-specifications/blob/main/exercises/poker/canonical-data.json, not just to the Rust track. The Rust track cannot afford to gain a reputation of being selfish and only keeping our changes to ourselves. We share with others and give back to others.

PR #1648, opened 2023-03-09
Improve practice exercise generator script

dem4ron:
- [x] Lowercase script-scoped variables to avoid collision with env vars
- [x] `local`ize function-scoped variables for similar reason
- [x] remove `function` from function declarations and follow this convention everywhere so a function looks like this:
```bash
fn_name(){
...
}
```
- [x] Add `.shellcheckrc` - currently not seen by CI
- [x] figure out why
- [x] remove unused imports
- [x] setup shellcheck import paths


ErikSchierboom:
Some warnings: https://github.com/exercism/rust/actions/runs/4379895423/jobs/7666336486

.github/workflows/tests.yml
```diff
@@ -87,7 +87,9 @@ jobs:
         uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
 
       - name: Run shellcheck
-        uses: ludeeus/action-shellcheck@00cae500b08a931fb5698e11e79bfbd38e612a38 # v2.0.0
+        uses: ludeeus/action-shellcheck@master
```
ErikSchierboom, line 90: Could you revert back to using a commit as the version to use?

bin/add_practice_exercise
```diff
@@ -30,51 +37,60 @@ command -v curl >/dev/null 2>&1 || {
...
+author_handle=${3:-$(get_author_handle)}
+message "info" "Using ${yellow}${author_handle}${blue} as author's handle. You can edit this later in the 'authors' field in the ${exercise_dir}/.meta/config.json file"
 
+create_rust_files "$exercise_dir" "$slug" "$has_canonical_data"
 
 # ==================================================
 
-# build configlet
+# Build configlet
 ./bin/fetch-configlet
```
ErikSchierboom, line 84: This is trying to fetch configlet a second time. I guess this is no longer needed?

bin/add_practice_exercise
```diff
@@ -30,51 +37,60 @@ command -v curl >/dev/null 2>&1 || {
...
 }
 
+# Check if configlet is already fetched
+if [ -e bin/configlet ]; then
+    message "success" "Configlet found!"
+else
+    message "info" "Fetching configlet"
+    bin/fetch-configlet
+    message "success" "Fetched configlet successfully!"
+fi
```
ErikSchierboom, line 47: I would always fetch configlet, as that way you're guaranteed to work with the latest version.

petertseng:
These commits needed to be squashed, since. I think the organisation doesn't have an organisation-wide document for when commits should be squashed since last I checked it was track-level policy, so it is my fault for not having written a policy document for the track. Add yet another thing to the long list of things to do, then.

petertseng:
These commits should not have been 120 individual commits but instead combined into fewer commits that each contain only one operation. I think the organisation doesn't have an organisation-wide document for git commit discipline since last I checked it was track-level policy, so it is my fault for not having written a policy document for the track. Add yet another thing to the long list of things to do, then.

dem4ron:
@petertseng I'm very sorry, it is all my fault. I knew I should've squashed the commits, but made a mistake. I thought squash-merge was the default behaviour as it is on most of the repos and clicked the green button out of excitement, but then realised my mistake. I'm very sorry, I feel very bad for messing up the timeline.

PR #1652, opened 2023-03-17
Circular Buffer: added memory leak tests for `overwrite` and `read`

felix91gr:
When solving this with `unsafe`, my mentor pointed out that my `overwrite` method must've been leaking memory. He was correct! And I fixed this issue in my next iteration.

A few days ago however, I remembered that there was a test for memory leaks on the `clear` method. The test in question uses `Rc` and its strong count to check if memory has been freed or not from the buffer.

I decided to base a new couple of tests on the same technique, and well, indeed my earlier buffer fails the `overwrite` test.

I've added these two to make sure the test suite helps students implement a leak-free Buffer :)

### DIFF
--- a/exercises/practice/circular-buffer/tests/circular-buffer.rs
+++ b/exercises/practice/circular-buffer/tests/circular-buffer.rs
@@ -106,6 +106,29 @@ fn clear_actually_frees_up_its_elements() {
 
 #[test]
 #[ignore]
+fn overwrite_frees_up_the_cleared_elements() {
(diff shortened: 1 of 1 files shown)


senekor:
I created a [PR](https://github.com/exercism/rust/pull/1724) to improve the example solution.

But I don't think the tests should be added anymore. The exercise is present in the [problem-specifications](https://github.com/exercism/problem-specifications/) repository, which means new tests should be added there.

Issue #1598, opened 2022-11-21
(Nitpick) Use `todo!` instead of `unimplemented!`.


LouisGariepy:
Currently, the template for not-yet-completed exercises uses the `unimplemented!` macro. I think `todo!` should be preferred.

The std docs state the `unimplemented` means that it *may or may not* be implemented in the future, while `todo!` means that it *should* be implemented. In this case, for the exercise to work, users *should* implement, so I feel like `todo!` is a better fit.

Furthermore, "todo" has a well-known near-universal meaning, while "unimplemented" may not call the user to action as much.

This is very nitpicky, and maybe you have reasons to use `unimplemented` so feel free to close.

felix91gr:
Is anyone working on this? Because I reckon I could whip it up in an afternoon. I have some spare time right now :)

PR #1654, opened 2023-03-18
Changed `unimplemented!` calls to `todo!`

felix91gr:
Solves #1598


Sjoeborg:
rip

senekor:
Hey, sorry for the long wait :-)

I agree with the change. I guess the formatting needs to be fixed.

Also, there are some occurrences of "unimplemented" in markdown docs and shell scripts which should be adjusted accordingly, to make sure things stay consistent.

senekor:
Oh and a reminder to whoever ends up merging this, we should definitely add `[no important files changed]` here. Although I'm not sure if the tests would even run if the source files are changed but not the test files. Better be safe.

senekor:
superseded by https://github.com/exercism/rust/pull/1723

PR #1655, opened 2023-03-21
Added digit 9 test case from the canonical test suite.

felix91gr:
9 is an edge case that is very important to keep testing in the advanced Luhn iteration exercises (Luhn From and Luhn Trait).

I know it because I was optimizing my solutions and after passing all the From and Trait tests, when I put those optimizations in my original Luhn solution, it failed the Luhn tests. Turns out the canonical specification had a test case for this! So I'm adding it to the advanced exercises as well.


senekor (approved):
Looks good, thanks!

PR #1660, opened 2023-03-29
Disable learning mode

ErikSchierboom:
We've had several complaints from students that the learning mode doesn't work well for them. I think it's best to disable it temporarily and then look at ways how to fix it.

### DIFF
--- a/config.json
+++ b/config.json
@@ -3,7 +3,7 @@
   "slug": "rust",
   "active": true,
   "status": {
-    "concept_exercises": true,
+    "concept_exercises": false,
     "test_runner": true,
     "representer": true,
     "analyzer": true


iHiD (approved):

rbtcollins:
@ErikSchierboom where is the feedback being gathered - if someone wanted to think about how to make things better, they will need specifics about what the pain points were.

ErikSchierboom:
Yes we know. We are still determining the best strategy.

iHiD:
@rbtcollins There's various discussions on our forum (forum.exercism.org) and discussions that have been happening on Discord (exercism.org/r/discord) and a multi-hour discussion on last week's community call. If you'd like to help, please get involved in those various places :)

PR #1666, opened 2023-04-05
Add D&D character exercise

dem4ron:
The exercise requires the user to roll 4 dice, so I added a test case to cover this scenario. This wasn't covered in the canonical data.

Additionally, I added another test case to confirm that the user is rolling the dice for each ability score. While there is a possibility for this test to fail, it's quite low since the probability of rolling the same score with each roll is also quite low.

Please let me know what you think about these changes

I branched out from the `generator-script-improvements`, so I can test the updates of the improvement.


exercises/practice/dnd-character/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+        unimplemented!("Return the wisdom score of the character")
+    }
+
+    pub fn charisma(&self) -> u8 {
+        unimplemented!("Return the charisma score of the character")
+    }
+
+    pub fn hitpoints(&self) -> u8 {
+        unimplemented!("Return the hitpoints of the character")
+    }
```
ErikSchierboom, line 46: I'm slightly torn whether these methods are a good or a bad idea. 

- The plus: users are free to implement the struct as they see fit
- The downside: it seems like quite boilerplatey

dem4ron: true about the downside, but I think it is challenging to avoid boilerplate code in this exercise, given its classic object-oriented nature. 

I personally enjoy having these getter methods, as it's allowing me to access attributes like `Character.strength()`. I included these methods in the stub file to ensure that the rust-analyzer doesn't show any errors when processing the test file.

on the other hand it is a perfect spot to write a macro, as I just did in the updated example file.

ErikSchierboom: I'll let @petertseng weigh in. The methods seem like overkill to me over regular struct access, but 🤷

exercises/practice/dnd-character/.meta/example.rs
```diff
@@ -0,0 +1,45 @@
...
+            rolls[i] = rng.gen_range(1..=4)
+        }
+        rolls
+    }
+
+    pub fn calculate_ability_score(ability_dice_rolls: [u8; 4]) -> u8 {
+        let mut ability_dice_rolls = ability_dice_rolls;
+        ability_dice_rolls.sort();
+        ability_dice_rolls[1..].iter().sum()
+    }
```
ErikSchierboom, line 44: Did you consider having just one `random_ability`/`generate_ability`/`random_ability_score` function?

exercises/practice/dnd-character/src/lib.rs
```diff
@@ -0,0 +1,27 @@
...
+}
+
+pub struct Character {
+    pub strength: u8,
+    pub dexterity: u8,
+    pub constitution: u8,
+    pub intelligence: u8,
+    pub wisdom: u8,
+    pub charisma: u8,
+    pub hitpoints: u8,
```
ErikSchierboom, line 12: Much cleaner!

exercises/practice/dnd-character/.meta/example.rs
ErikSchierboom: These functions could also be moved outside of the `Character`'s `impl`, as they don't rely on the character.

exercises/practice/dnd-character/tests/dnd-character.rs
```diff
@@ -0,0 +1,183 @@
...
+    let input = 18;
+    let expected: f32 = 4.0;
+    assert_eq!(modifier(input), expected);
+}
+
+#[test]
+#[ignore]
+fn test_random_ability_is_within_range() {
+    let score = Character::calculate_ability_score(Character::roll_four_dice());
+    assert!((3..=18).contains(&score));
```
ErikSchierboom, line 134: You could also consider adding a test that calls that function N number of times (100?) to see if it doesn't always return the same value.

PR #1669, opened 2023-04-08
✨  (concepts) Add concept exercise `high-school-sweethearts`

devkabiir:
This concept is inspired from the same in csharp track. Marked as WIP since we don't have any exercise with "strings" concept. Nor do we have any with "time" or "date".


senekor:
The author of this PR has stopped responding in the discussions planning the work on the syllabus.

PR #1675, opened 2023-04-09
✨  (concept) Add new concept exercise `cars-assemble`

devkabiir:
This is inspired by the same in csharp track. Provides a gentle introduction to variable assignment, if-statements and numbers.


exercises/concept/cars-assemble/.meta/exemplar.rs
```diff
@@ -0,0 +1,39 @@
+const PRODUCTION_RATE_PER_HOUR_FOR_DEFAULT_SPEED: i32 = 221;
+
+pub fn production_rate_per_hour(speed: i32) -> f64 {
+    let result = production_rate_per_hour_for_speed(speed) * success_rate(speed);
+
+    round_to_1_decimal(result)
+}
+
+fn round_to_1_decimal(input: f64) -> f64 {
+    return (input * 10.0).round() / 10.0;
```
MatthijsBlom, line 10:
```suggestion
    (input * 10.0).round() / 10.0
```

exercises/concept/cars-assemble/.meta/exemplar.rs
```diff
@@ -0,0 +1,39 @@
...
+
+    if speed >= 5 {
+        return 0.9;
+    }
+
+    if speed <= 0 {
+        return 0.0;
+    }
+
+    return 1.0;
```
MatthijsBlom, line 38: We can avoid explicit `return` entirely.

```suggestion
    if speed == 10 {
        0.77
    } else if speed == 9 {
        0.8
    } else if speed >= 5 {
        0.9
    } else if speed <= 0 {
        0.0
    } else {
        1.0
    }
```

exercises/concept/cars-assemble/src/lib.rs
```diff
@@ -0,0 +1,11 @@
+pub fn production_rate_per_hour(_speed: u8) -> f32 {
+    unimplemented!("Implement production_rate_per_hour")
+}
+
+pub fn working_items_per_minute(_speed: u8) -> u8 {
+    unimplemented!("Implement working_items_per_minute")
+}
+
+pub fn success_rate(_speed: u8) -> f32 {
```
remlse, line 9: This function signature does not match the `exemplar.rs`

exercises/concept/cars-assemble/src/lib.rs
```diff
@@ -0,0 +1,11 @@
+pub fn production_rate_per_hour(_speed: u8) -> f32 {
```
remlse, line 1: This function signature does not match the `exemplar.rs`

exercises/concept/cars-assemble/src/lib.rs
```diff
@@ -0,0 +1,11 @@
...
+    unimplemented!("Implement production_rate_per_hour")
+}
+
+pub fn working_items_per_minute(_speed: u8) -> u8 {
```
remlse, line 5:
```suggestion
pub fn working_items_per_minute(_speed: i32) -> f64 {
```

Returning a `f64` will allow us to avoid talking about type casting with `as` entirely in this exercise, which would be consistent with the book.

exercises/concept/cars-assemble/.meta/exemplar.rs
```diff
@@ -0,0 +1,39 @@
...
+
+    round_to_1_decimal(result)
+}
+
+fn round_to_1_decimal(input: f64) -> f64 {
+    return (input * 10.0).round() / 10.0;
+}
+
+fn production_rate_per_hour_for_speed(speed: i32) -> f64 {
+    (PRODUCTION_RATE_PER_HOUR_FOR_DEFAULT_SPEED * speed) as f64
```
remlse, line 14:
```suggestion
    (PRODUCTION_RATE_PER_HOUR_FOR_DEFAULT_SPEED * speed).into()
```

MatthijsBlom:
@devkabiir I don't think there is a need for force-pushes. The commits will probably be squashed at the end anyway. On the other hand force-pushes make it slightly harder for observers to follow developments in the PR.

exercises/concept/cars-assemble/.docs/introduction.md
```diff
@@ -0,0 +1,185 @@
+# Introduction
+
+## Numbers
+
+There are two different types of numbers in Rust:
+
+- Integers: numbers with no digits behind the decimal separator (whole numbers). Examples are `-6`, `0`, `1`, `25`, `976` and `500000`.
+- Floating-point numbers: numbers with zero or more digits behind the decimal separator. Examples are `-2.4`, `0.1`, `3.14`, `16.984025` and `1024.0`.
+
+The two default numeric types in Rust are `i32` and `f64`. An `i32` is a 32-bit integer and a `f64` is a 64-bit floating-point number.
```
remlse, line 10: They aren't used in the exercise, but it might be appropriate to include a note about the other number types? Something like:

> There are other number types with different precision, like `f32` and `i16`. Unsigned (always positive) integers start with a `u` instead of `i`. For example, a byte is expressed as `u8`. Another number type you'll come across is `usize`, its precision depends on your computer (usually `64` bits) and is used for lengths and indexes.

remlse (changes requested):
On a general note, this exercise seems a bit overloaded with concepts to me. I'm counting five in total:
- number types
- constants
- variable declaration
- type conversion
- arithmetic operators
- if statements

I wonder if one or more other exercises should come before this, introducing a subset of these concepts to the students can make smaller steps.

senekor:
The author of this PR has stopped responding in the discussions planning the work on the syllabus. But there's been a lot of work put into reviews already, so I'm keeping it open in case it can be salvaged in a future attempt to create a good syllabus.

PR #1676, opened 2023-04-09
✨  (concepts) Add new concept exercise `phone-number-analysis`

devkabiir:
This is inspired by the same in csharp track. Provides a gentle introduction to tuples.


exercises/concept/phone-number-analysis/.meta/exemplar.rs
```diff
@@ -0,0 +1,13 @@
...
+
+    let (prefix_code, last_4_with_dash) = number.split_at(4);
+
+    let last_4 = last_4_with_dash.split_at(1).1;
+
+    return (dial_code == "212", prefix_code == "-555", last_4);
+}
+
+pub fn is_fake(info: (bool, bool, &str)) -> bool {
+    return info.1;
```
MatthijsBlom, line 12:
```suggestion
    info.1
```

exercises/concept/phone-number-analysis/.docs/instructions.md
```diff
@@ -0,0 +1,32 @@
...
+
+Phone numbers passed to the routines are guaranteed to be in the form
+NNN-NNN-NNNN e.g. 212-515-9876.
+
+## 1. Analyze a phone number
+
+Your analysis should return 3 pieces of data
+
+1. An indication of whether the number has a New York dialing code ie. 212 as the first 3 digits
+2. An indication of whether the number is fake having 555 as a prefix code in positions 5 to 7 (numbering from 1)
```
noah-clements, line 15: I would switch feature ##1 Analyze a phone number and ##2 to Detect fake prefix code.
Simply because the second part of feature 1 relies on feature 2.

exercises/concept/phone-number-analysis/.docs/instructions.md
```diff
@@ -0,0 +1,32 @@
...
+3. The last 4 digits of the number.
+
+Implement the function `analyze()` to produce the phone number info.
+
+```rust
+analyze("631-555-1234");
+// => (false, true, "1234")
+```
+
+## 2. Detect if a phone number has a fake prefix code (555)
```
noah-clements, line 25: Many people would consider that feature 2 would be a standalone test and that feature 1 would reference that test in constructing the feature 1 tuple. It seems that we are looking for feature 2 to return the second element of the feature 1 tuple (i.e. analyze().1) , rather than returning the value of the separate test, simply because the second part of feature 1 relies on feature 2.

Perhaps we should recognize that in the instructions and say (as we do in different exercises) that although you could simply return the value of the standalone test for fake 555 prefix, this exercise asks you to return the value based only on the phone number info produced in task 1.

senekor:
The author of this PR has stopped responding in the discussions planning the work on the syllabus. But there's been work put into reviews already, so I'm keeping it open in case it can be salvaged in a future attempt to create a good syllabus.

PR #1682, opened 2023-04-10
✨  (concepts) Add concept exercise `bird-watcher`

devkabiir:
This is inspired by the same in csharp track. Provides introduction to for loops, arrays and a bit of iterators.


ErikSchierboom:
> Provides introduction to for loops, arrays and a bit of iterators.

I've only briefly looked at the code, but I feel like this might be a bit much to introduce. My hunch would be to see if we can introduce iterators later on. Note that it is perfectly fine for this exercise to expect the student to write things in a somewhat cumbersome, less idiomatic way simply because they don't yet know all the concepts (like iterators here).

noah-clements:
> > > Provides introduction to for loops, arrays and a bit of iterators.
> I understand that you have to use either ranges or iterators in for loops. 
> 
> Update: Removed Iterators completely.

I like the way this ended up, matching [pages 57-58 in The Book](https://doc.rust-lang.org/stable/book/ch03-05-control-flow.html#looping-through-a-collection-with-for). Introducing how to, without going too deep into how does it work.

senekor:
The author of this PR has stopped responding in the discussions planning the work on the syllabus. But there's been work put into reviews already, so I'm keeping it open in case it can be salvaged in a future attempt to create a good syllabus.

PR #1691, opened 2023-04-28
Streamline introduction.md of luscious-lasagna

remlse:
Many topics that were covered previously are not relevant to the actual tasks, they should be covered in a separate exercise.

This was discussed [here](http://forum.exercism.org/t/rust-syllabus-which-exercises-prs-cover-only-concepts-from-chapter-3/5347/17).

### DIFF
--- a/exercises/concept/lucians-luscious-lasagna/.docs/introduction.md
+++ b/exercises/concept/lucians-luscious-lasagna/.docs/introduction.md
@@ -1,17 +1,9 @@
 # Introduction
 
-In Rust, assigning a value to a name is referred to as a _binding_. Bindings are immutable unless declared with the `mut` keyword. As Rust is a statically-typed language, each binding has a type known at compile-time.
+## Functions
 
-Bindings are most commonly defined using the `let` keyword. Specifying a binding's type is optional for most bindings, as Rust's _type inference_ can usually infer the type based on their value. A binding looks like this:
-
-```rust
-// Automatically inferred type
-let fingers = 10;
-```
-
-Functions are _items_. Where bindings typically refer to a particular value, items refer to a unit of code organization, typically a function or a module, which is available throughout the lifetime of the program. A function automatically returns the result of its last expression. A function may have 0 or more parameters, which are bindings with a lifetime of the function call.
-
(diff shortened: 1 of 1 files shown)


exercises/concept/lucians-luscious-lasagna/.docs/introduction.md
```diff
@@ -1,104 +1,51 @@
...
-    // this anonymous scope. Outside this scope, the previous binding still
-    // applies.
-    let a = 2;
-    let b = 3;
-    dbg!(a, b); // 2, 3
-}
-// can't use `b` anymore because it is out of scope
-// dbg!(b);
+The return type is specified after an arrow `->`.
+It is optional if the function doesn't return anything.
```
MatthijsBlom, line 24: One might wonder: if the function does not return anything, how could one give its (presumably nonexistent) return type?

exercises/concept/lucians-luscious-lasagna/.docs/introduction.md
```diff
@@ -1,104 +1,51 @@
...
-// leaving us with our original binding.
-dbg!(a); // 1
-```
+Lastly, the body of the function is enclosed in curly braces `{}`.
+A function automatically returns the result of its last expression.
+Simple functions like the one above can therefore consist of just an expression.
 
-Rust items are often organized in modules. Each crate is implicitly a module, but it can define inner sub-modules of arbitrary depth. A module groups related functionality and is defined using the `mod` keyword.
+Invoking a function is done by specifying its name followed by parentheses.
+If the function requires parameters, they must be specified within the parentheses.
```
MatthijsBlom, line 31:
```suggestion
If the function requires arguments, they must be specified within the parentheses.
```

remlse: thx!

exercises/concept/lucians-luscious-lasagna/.docs/introduction.md
remlse: Good catch. I decided to change the wording so it seems like that's not possible. No need to confuse students with the empty tuple at this point.

senekor:
The team effort to get the syllabus in a good shape again died down. I'm keeping this open in case it is useful at some point, but I think future work on the syllabus should not hesitate to close this and do something new from scratch.

PR #1695, opened 2023-06-01
Clarify constructor return types for RNA Transcription exercise

mnquintana:
I recently completed https://exercism.org/tracks/rust/exercises/rna-transcription, but couldn't quite figure out what the error type in the signatures of `DNA::new()` and `RNA::new()` was supposed to mean from looking at the tests. Unfortunately, I interpreted it as "the number of different nucleotides used before encountering the first invalid char", which just so happened to pass the test cases, even though it was wrong. (I can submit a separate PR for fixing up those test cases.)

It turns out though that the `usize` in `Result<Self, usize>` was meant to mean "the first invalid character index": https://exercism.org/tracks/rust/exercises/rna-transcription

This PR clarifies the meaning of the error value in the return type.


ErikSchierboom (approved):
Nice!

PR #1709, opened 2023-07-20
Remove impls of deprecated exercises

senekor:
I missed [one more clippy warning](https://github.com/exercism/rust/actions/runs/5612515528/job/15206434155) in the stubs.

It actually occurred in the stub of a deprecated exercise, which had a full implementation as a stub because of https://github.com/exercism/rust/pull/832.

I turned them back into regular stubs, so we don't have to maintain implementations of deprecated exercises.


ErikSchierboom (approved):
Makes sense

PR #1712, opened 2023-07-21
Document available crates

### LINKED ISSUE #1365: crate availability should be documented in some way on the site

senekor:
closes #1365

I'm not sure if the `docs/TESTS.md` file is the correct place to add this, but the [building documentation](https://exercism.org/docs/building/tracks/docs) has an example where that is the case.

I slightly updated the rest of that file as well.

### DIFF
--- a/docs/TESTS.md
+++ b/docs/TESTS.md
@@ -1,8 +1,12 @@
 # Writing the Code
 
-Write your code in `src/lib.rs`. Some exercises come with a stub file in `src/lib.rs` that will show you the signatures of the code you'll need to write. If the exercise does not come with a `src/lib.rs` file, create one.
+Write your code in `src/lib.rs`.
+The exercises come with a stub file in `src/lib.rs` that will show you the signatures of the code you'll need to write.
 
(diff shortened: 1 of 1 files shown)


ErikSchierboom (approved):
It's not an unreasonable place I feel.

PR #1715, opened 2023-08-21
Add test case for early returns in `anagram`

maugier:
I've seen during mentoring an incorrect solution using `return` instead of `continue` to skip the trivial solution, which would cause the function to miss genuine anagrams coming after the trivial solution.

None of the existing tests catch that bug; here is a new test case that does.

### DIFF
--- a/exercises/practice/anagram/tests/anagram.rs
+++ b/exercises/practice/anagram/tests/anagram.rs
@@ -184,3 +184,15 @@ fn test_different_words_but_same_ascii_sum() {
 
     process_anagram_case(word, &inputs, &outputs);
 }
+
+#[test]
+#[ignore]
+fn test_does_not_return_early_after_own_anagram() {
+    let word = "hello";
+
+    let inputs = ["nothing", "hello", "banana", "olleh", "zombies"];
+
+    let outputs = vec!["olleh"];
+
+    process_anagram_case(word, &inputs, &outputs);
+}


MatthijsBlom:
It looks like this is already covered in `problem-specifications`:

https://github.com/exercism/problem-specifications/blob/b4f7ee538e5935b1d71428a0c2a606184dd557ca/exercises/anagram/canonical-data.json#L198-L208

Do you disagree?

maugier:
You're right, it's already covered in the specification; it's the test suite that is not in sync with the specification.

I didn't know the tests were derived from a specification; it's indicated neither in `CONTRIBUTING.md`, nor the test file itself; it's only hinted at in `README.md`, in the section about contributing new exercises.

If that is useful, I can instead re-sync the test suite to the specification.

senekor:
I would gladly merge such a PR.

maugier:
@senekor I'm ready to send the PR, I just have a few questions:

- Some of the existing tests (regarding unicode behavior) are not in the spec, should I preserve them?
- Some of the existing tests are not in the spec and redundant with the spec test, should I remove them?
- If yes, should I also submit a PR to the spec repository to add them? They may be hard to pass in other languages.

Cheers,

MatthijsBlom:
> - Some of the existing tests (regarding unicode behavior) are not in the spec, should I preserve them?

I'd say _yes_.

> - Some of the existing tests are not in the spec and redundant with the spec test, should I remove them?

Which ones, and which cases cover them?

Tip: in your upcoming PR you could place comments on the tests you suspect of being redundant. If you post the uuids as well then it becomes very easy to compare them.

> - If yes, should I also submit a PR to the spec repository to add them? They may be hard to pass in other languages.

I expect it would be appreciated if you asked on the forum first.

Tracks are not required to implement all canonical tests. They can pick and choose.

I would welcome (more?) unicode test cases in `problem-specifications`. I have missed them in several places.

senekor:
Agree with @MatthijsBlom. If the PR is opened we can discuss directly on the code.

I'm less enthusiastic about adding unicode tests to the problem specs. Many languages are not equipped to handle unicode out of the box / with their standard library. Which would basically force students of these languages to either use libraries, which I think many test runners don't even support, or manually implement unicode support. I'd say that's beyond the scope of our exercises.

Rust is unique in that it comes with good unicode support already and the test runner comes bundled with a library to handle grapheme clusters.

But it's not my call and a forum post would be the best place to discuss this.

PR #1716, opened 2023-08-24
Saddle-points: fix formatting of instructions

JDeeth:
First line of demo grid is interpreted as a code block and the rest as paragraph text. Wrapping the whole thing in code tags forces it to be shown in fixed-width text instead.


senekor (approved):
I think this document originally comes from problem specs. Let's fix it now though instead of waiting until I or someone else finds the time to sync properly.

senekor:
@JDeeth Thanks!

PR #1723, opened 2023-09-10
Use todo instead of unimplemented

### LINKED ISSUE #1598: (Nitpick) Use `todo!` instead of `unimplemented!`.

senekor:
The difference is one of intent, and the actual panic message.
See the docs why todo is more appropriate:
https://doc.rust-lang.org/stable/std/macro.todo.html

[no important files changed]

closes #1598

### DIFF
--- a/concepts/structs/about.md
+++ b/concepts/structs/about.md
@@ -32,7 +32,7 @@ Lastly, methods can be defined on structs inside of an `impl` block:
 impl Item {
     // initializes and returns a new instance of our Item struct
     fn new() -> Self {
-        unimplemented!()
+        todo!()
     }
 }
 ```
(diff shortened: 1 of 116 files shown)


senekor:
* auto-sync
  * **PR #1723** <a href="https://app.graphite.dev/github/pr/exercism/rust/1723" target="_blank">Graphite</a>  👈
    * **PR #1725** <a href="https://app.graphite.dev/github/pr/exercism/rust/1725" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/exercism/rust/1723?utm_source=stack-comment).

ErikSchierboom (approved):
This is great

PR #1724, opened 2023-09-10
Improve example solution of circular-buffer

senekor:
The trait bounds `Default + Clone` are semantically incorrect. A proper circular buffer needs to handle all kinds of elements.

The reason these bounds are here is because it allows to avoid `unsafe` while enjoying the most efficient memory layout.
However, there is another performance downside:
The implementations of `Default` and `Clone` may be expensive (e.g. cause heap allocations.)

As came up in this discussion, there are other weird side effects, for example for `Rc` which has a non-standard implementation of `Clone`: https://github.com/exercism/rust/pull/1652

The approach I chose here get's rid of the trait bounds and wraps every element in an `Option`.
In the worst case, this may lead to twice the memory consumption. (e.g. `Option<u64>` takes 16 bytes because of alignment and padding) This approach is semantically correct, but not the most performant.

The most performant solution would be to use `unsafe`. I'd like to avoid that in example solutions.

### DIFF
--- a/exercises/practice/circular-buffer/.meta/example.rs
+++ b/exercises/practice/circular-buffer/.meta/example.rs
@@ -4,20 +4,21 @@ pub enum Error {
     FullBuffer,
 }
 
-pub struct CircularBuffer<T: Default + Clone> {
-    buffer: Vec<T>,
-    size: usize,
+pub struct CircularBuffer<T> {
+    /// Using Option leads to less efficient memory layout, but
+    /// it allows us to avoid using `unsafe` to handle uninitialized
+    /// mempory ourselves.
+    data: Vec<Option<T>>,
     start: usize,
     end: usize,
 }
 
-impl<T: Default + Clone> CircularBuffer<T> {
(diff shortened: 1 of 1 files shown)


senekor:
* auto-sync
  * **PR #1723** <a href="https://app.graphite.dev/github/pr/exercism/rust/1723" target="_blank">Graphite</a> 
    * **PR #1727** <a href="https://app.graphite.dev/github/pr/exercism/rust/1727" target="_blank">Graphite</a> 
      * **PR #1726** <a href="https://app.graphite.dev/github/pr/exercism/rust/1726" target="_blank">Graphite</a> 
      * **PR #1725** <a href="https://app.graphite.dev/github/pr/exercism/rust/1725" target="_blank">Graphite</a> 
      * **PR #1724** <a href="https://app.graphite.dev/github/pr/exercism/rust/1724" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/exercism/rust/1724?utm_source=stack-comment).

ErikSchierboom (approved):

PR #1736, opened 2023-09-13
Enforce idiomatic error handling in bash

senekor:
Single commands that may error can stil be achieved with `|| true`.
With this change, we'll only see a limited number of clippy warnings
in CI. I think that's fine, the one fixing it can still remove these
flags locally while working on the fix to see all warnings.

### DIFF
--- a/bin/check_exercises.sh
+++ b/bin/check_exercises.sh
@@ -1,11 +1,5 @@
 #!/usr/bin/env bash
-
-# In DENYWARNINGS or CLIPPY mode, do not set -e so that we run all tests.
-# This allows us to see all warnings.
-# If we are in neither DENYWARNINGS nor CLIPPY mode, do set -e.
-if [ -z "$DENYWARNINGS" ] && [ -z "$CLIPPY" ]; then
-    set -e
-fi
+set -eo pipefail
 
 # can't benchmark with a stable compiler; to bench, use
 # $ BENCHMARK=1 rustup run nightly bin/check_exercises.sh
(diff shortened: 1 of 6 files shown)


senekor:
* main
  * **PR #1736** <a href="https://app.graphite.dev/github/pr/exercism/rust/1736" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/exercism/rust/1736?utm_source=stack-comment).

senekor:
I became aware that this is an issue when I saw that CI passed on another branch, even though it shouldn't have:
https://github.com/exercism/rust/actions/runs/6175508949/job/16762524913?pr=1723

This first commit reveals the problem.

bin/check_exercises.sh
```diff
@@ -33,39 +28,39 @@ return_code=0
 # An exercise worth testing is defined here as any top level directory with
 # a 'tests' directory and a .meta/config.json.
 for exercise in $files; do
-   exercise="$exercise/$target_dir"
+    exercise="$exercise/$target_dir"
```
ErikSchierboom, line 31: The formatting changes seem to be unrelated to the thing the PR is fixing. I'm happy to approve, but would recommend keeping unrelated in separate PRs in the future. It makes reviewing easier.

ErikSchierboom (approved):

bin/check_exercises.sh
senekor: alright, will do!

PR #1746, opened 2023-09-18
Cleanup test scripts

### LINKED ISSUE #1738: Improve example solution testing scripts
Currently, the scripts for testing the example solution create some backup files, copy the `example.rs` to `lib.rs`, modify the test file (remove `#[ignore]`), run the tests and then restore the previous state of the exercise with the backup files.

This is annoying, because running all of these tests sequentially means git shows these diffs while the tests are running. GUI git clients constantly update with these changes. Working with git at the same time as running the tests is difficult. It has happened to me that I accidentally committed some of these changes.

It would be better for the test scripts to copy the exercise folder to another location that is in `.gitignore` and execute the tests there. That directory can be modified by the test scripts freely without interrupting other work.

senekor:
Most importantly, move all required files for testing to a tmp dir. This means the actual files in the repository are not modified anymore. Running these scripts locally does not interrupt the git workflow now.

The output is also cleaned up in various situations.

closes #1738

### DIFF
--- a/bin/check_exercises.sh
+++ b/bin/check_exercises.sh
@@ -1,66 +1,45 @@
 #!/usr/bin/env bash
 set -eo pipefail
 
-# can't benchmark with a stable compiler; to bench, use
-# $ BENCHMARK=1 rustup run nightly bin/check_exercises.sh
-if [ -n "$BENCHMARK" ]; then
-    target_dir=benches
-else
-    target_dir=tests
-fi
-
 repo=$(git rev-parse --show-toplevel)
 
 if [ "$GITHUB_EVENT_NAME" = "pull_request" ]; then
     git fetch --depth=1 origin main
     files="$(
         git diff --diff-filter=d --name-only remotes/origin/main |
-        grep "exercises/" || true |
-        cut -d '/' -f -3 |
-        sort -u |
-        awk -v repo="$repo" '{print repo"/"$1}'
+        grep "exercises/" |
+        cut --delimiter '/' --fields -3 |
+        sort --unique || true
     )"
 else
     files="$repo/exercises/*/*"
 fi
 
-return_code=0
 # An exercise worth testing is defined here as any top level directory with
 # a 'tests' directory and a .meta/config.json.
 for exercise in $files; do
-    exercise="$exercise/$target_dir"
-
-    # This assumes that exercises are only one directory deep
-    # and that the primary module is named the same as the directory
-    directory=$(dirname "${exercise}")
+    slug=$(basename "$exercise")
 
     # An exercise must have a .meta/config.json
-    metaconf="$directory/.meta/config.json"
+    metaconf="$exercise/.meta/config.json"
     if [ ! -f "$metaconf" ]; then
         continue
     fi
 
-    release=""
-    if [ -z "$BENCHMARK" ] && jq --exit-status '.custom?."test-in-release-mode"?' "$metaconf"; then
+    # suppress compilation output
+    cargo_args="--quiet"
+
+    if [ -z "$BENCHMARK" ] && jq --exit-status '.custom?."test-in-release-mode"?' "$metaconf" > /dev/null; then
         # release mode is enabled if not benchmarking and the appropriate key is neither `false` nor `null`.
-        release="--release"
+        cargo_args="$cargo_args --release"
     fi
 
-    if [ -n "$DENYWARNINGS" ]; then
-        # Output a progress dot, because we may not otherwise produce output in > 10 mins.
-        echo -n '.'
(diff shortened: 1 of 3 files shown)


bin/check_exercises.sh
```diff
@@ -1,66 +1,45 @@
...
 repo=$(git rev-parse --show-toplevel)
 
 if [ "$GITHUB_EVENT_NAME" = "pull_request" ]; then
     git fetch --depth=1 origin main
     files="$(
         git diff --diff-filter=d --name-only remotes/origin/main |
-        grep "exercises/" || true |
-        cut -d '/' -f -3 |
-        sort -u |
-        awk -v repo="$repo" '{print repo"/"$1}'
```
senekor, line 21: This was an oopsie on my side. The `|| true` belongs at the end of the pipe chain, otherwise the rest of the commands won't run if `grep` actually succeeds.

bin/check_exercises.sh
```diff
@@ -1,66 +1,45 @@
...
+        grep "exercises/" |
+        cut --delimiter '/' --fields -3 |
+        sort --unique || true
     )"
 else
     files="$repo/exercises/*/*"
 fi
 
-return_code=0
```
senekor, line 27: We `set -e` now, no need to manually remember exit codes to return at the end now.

bin/check_exercises.sh
```diff
@@ -1,66 +1,45 @@
...
 # An exercise worth testing is defined here as any top level directory with
 # a 'tests' directory and a .meta/config.json.
 for exercise in $files; do
-    exercise="$exercise/$target_dir"
-
-    # This assumes that exercises are only one directory deep
-    # and that the primary module is named the same as the directory
-    directory=$(dirname "${exercise}")
+    slug=$(basename "$exercise")
```
senekor, line 21: I made it so the `./bin/test_exercise.sh` script accepts a slug instead of a path. Makes it a little more convenient to test specific exercises one is working on.

bin/test_exercise.sh
```diff
@@ -1,112 +1,81 @@
...
+done
+if [ -z "$exercise_path" ]; then
+    echo "Could not find exercise path for $slug"
+    exit 1
 fi
 
 # what cargo command will we use?
 if [ -n "$BENCHMARK" ]; then
-    command="bench"
+    command="+nightly bench"
```
senekor, line 28: `+nightly` is not necessary in CI, where only the nightly toolchain is installed for the benchmark job. But this allows running benchmarks locally as well, where multiple toolchains are probably installed.

bin/test_exercise.sh
```diff
@@ -1,112 +1,81 @@
...
-            mv "$exercise/$dir.orig" "$exercise/$dir"
-        fi
-    done
-}
-
-# cause the reset to execute when the script exits normally or is killed
-trap reset EXIT INT TERM
+# setup temporary directory for the exercise
+tmp_dir=$(mktemp -d)
+trap 'rm -rf $tmp_dir' EXIT INT TERM
```
senekor, line 39: This is the main motivation of this PR. Executing tests somewhere else in a tmp directory means files in the repository are not modified, so the git workflow is not disturbed.

bin/test_exercise.sh
```diff
@@ -1,112 +1,81 @@
...
+    "tests"
+    "benches"
+    "Cargo.toml"
+)
+for c in "${contents_to_copy[@]}"; do
+    if [ ! -e "$exercise_path/$c" ]; then
+        continue
     fi
+    cp -r "$exercise_path/$c" "$tmp_dir"
 done
```
senekor, line 53: We shouldn't copy over the entire directory, there may be a large `target` folder.

bin/test_exercise.sh
```diff
@@ -1,112 +1,81 @@
...
-if [ -f "$exercise/.meta/Cargo-example.toml" ]; then
-    cp -f "$exercise/.meta/Cargo-example.toml" "$exercise/Cargo.toml"
+if [ -f "$exercise_path/.meta/Cargo-example.toml" ]; then
+    cp -f "$exercise_path/.meta/Cargo-example.toml" "$tmp_dir/Cargo.toml"
 fi
 
-# If deny warnings, insert a deny warnings compiler directive in the header
 if [ -n "$DENYWARNINGS" ]; then
-    sed -i -e '1i #![deny(warnings)]' "$exercise/src/lib.rs"
+    export RUSTFLAGS="$RUSTFLAGS -D warnings"
```
senekor, line 69: better way to deny warnings than injecting annotations into the source code

ErikSchierboom (approved):

PR #1747, opened 2023-09-19
Fix clippy warnings in example solutions

senekor:
CI errors are now surfacing because of [this semi-conscious change](https://github.com/exercism/rust/pull/1746/files#diff-c97b02ca592b9137046a9842fd1d13e04f1baaba1085b41aee9856bea0cfcf6dL97-L100).

Previously, clippy was only run over the test files. Now it is run over the example solutions as well. I like it this way and applied the relevant fixes.

### DIFF
--- a/exercises/concept/csv-builder/.meta/exemplar.rs
+++ b/exercises/concept/csv-builder/.meta/exemplar.rs
@@ -16,9 +16,9 @@ impl CsvRecordBuilder {
             self.content.push(',');
         }
 
-        if val.contains(",") || val.contains(r#"""#) || val.contains("\n") {
+        if val.contains(',') || val.contains('"') || val.contains('\n') {
             self.content.push('"');
-            self.content.push_str(&val.replace(r#"""#, r#""""#));
+            self.content.push_str(&val.replace('"', r#""""#));
             self.content.push('"');
         } else {
             self.content.push_str(val);
@@ -28,5 +28,11 @@ impl CsvRecordBuilder {
     /// Consumes the builder and returns the CSV record.
     pub fn build(self) -> String {
         self.content
+    }
+}
+
+impl Default for CsvRecordBuilder {
+    fn default() -> Self {
+        Self::new()
     }
 }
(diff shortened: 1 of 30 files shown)


exercises/practice/decimal/.meta/Cargo-example.toml
```diff
@@ -4,5 +4,5 @@ name = "decimal"
 version = "0.1.0"
 
 [dependencies]
-num-bigint = "0.1.40"
-num-traits = "0.1.40"
+num-bigint = "0.4.4"
+num-traits = "0.2.16"
```
senekor, line 8: Got some compiler warnings that these old crates contained code which only compiled due to an unsoundness in rustc. They will stop compiling in the future.

exercises/practice/decimal/.meta/example.rs
```diff
@@ -102,7 +102,9 @@ macro_rules! auto_impl_decimal_ops {
             fn $func_name(mut self, mut rhs: Self) -> Self {
                 Decimal::equalize_precision(&mut self, &mut rhs);
                 Decimal::new(
+                    #[allow(clippy::redundant_closure_call)]
                     $digits_operation(self.digits, rhs.digits),
```
senekor, line 106: The macro pastes the closure here, which is immediately called. clippy complains not to create closures and immediately call them.

exercises/concept/csv-builder/.meta/exemplar.rs
```diff
@@ -30,3 +30,9 @@ impl CsvRecordBuilder {
         self.content
     }
 }
+
+impl Default for CsvRecordBuilder {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
senekor, line 38: A type with a `fn new() -> Self` function should also have a `Default` implementation.

exercises/practice/scale-generator/.meta/example.rs
```diff
@@ -355,3 +335,23 @@ impl Scale {
         out
     }
 }
+
+#[cfg(test)]
+mod test {
```
senekor, line 340: Test module should appear at the end of a file. There is a pretty fine-grained module structure in this example file. In a real project, this would've been split into several files. But for us, it's most convenient to keep everything in one file.

ErikSchierboom (approved):

PR #1751, opened 2023-09-24
series: Clarify edge case of zero-length substrings

senekor:


MatthijsBlom:
The intuition is:

```text
 t e x t
^ ^ ^ ^ ^ slices can begin/end at these spots
```

When the slice length is 0, the first slice starts & ends at the first of these spots, the second one starts & ends at the second spot, the third one at the third, etc.

exercises/practice/series/.docs/instructions.append.md
```diff
@@ -0,0 +1,6 @@
+# Instructions append
+
+What do you expect the result to be if the length of the substrings is zero?
+There is no one right answer, but we expect you to output a number of empty strings:
+one more than the length of the input string.
+While not intuitive to everyone, this upholds the property that the number of substrings is one more than the length of the input string minus the length of the substrings.
```
iHiD, line 6: Maybe simplify it and make it a little less apologetic (when someone doesn't have the context of this being controversial, it probably seems confusing to explain it in such depth, rather than just telling them the requirement).

```suggestion
Different languages on Exercism have different expectations about what the result should be if the length of the substrings is zero.
For Rust, we expect you to output a number of empty strings, which will one greater than the length of the input string.
```

filakhtov (approved):
Thanks a lot. 🙇🏻‍♂️  Even though we can agree this is not ideal, at least we have a clear expectations documented now, which is a reasonable trade-off considering we don't want to break all existing solutions.

iHiD (approved):

PR #1756, opened 2023-10-17
Fix clippy warnings

senekor:
[no important files changed]

Seems like clippy::needless_borrow was renamed, or maybe these are separate lints where the other one now takes precedence. Either way, we want to ignore it for this exercise as already documented.


senekor:
superseded by https://github.com/exercism/rust/pull/1757

PR #1757, opened 2023-10-17
Don't run beta clippy in CI

senekor:
Running the beta version of clippy in CI creates unnecessary churn.
This time, a lint was renamed and it is not easy to satisfy *both* the stable and beta version.
Within six weeks, our CI would probably be broken again.
I also vaguely remember changing a bunch of stuff because of a beta lint, which didn't end up being stabilized after all.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -121,10 +121,6 @@ jobs:
     name: Clippy
     runs-on: ubuntu-latest
 
-    strategy:
-      matrix:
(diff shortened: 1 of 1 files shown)


senekor:
@ErikSchierboom Here's the PR where I tried to fix CI. I would close that PR if this one is merged. It's not like it's impossible to fix, but at this point I think the cost of having to please beta clippy outweigh the benefits.

ErikSchierboom (approved):
Good idea

PR #1764, opened 2023-11-13
Sync tournament with problem-specifications

senekor:


senekor:
Tests all look a little different because I used the test case generator.

ErikSchierboom (approved):

PR #1772, opened 2023-11-14
Sync scrabble-score with problem-specifications

senekor:


senekor:
There seem to be two additional tests here not present upstream. (pinata and strasse)

What we could do:
- upstream these tests (makes exercise harder for languages without built-in unicode support)
- delete them (makes exercise easier)
- use this exercise as a proving ground for a system to extend the tests in problem-specifications. (iirc python has a system for that, something like `.meta/additional-tests.json`.)

senekor:
generating additional, track-specific tests is implemented in https://github.com/exercism/rust/pull/1774

senekor:
I used `additional-tests.json` for the unicode cases.

ErikSchierboom (approved):

PR #1773, opened 2023-11-14
Sync say with problem-specifications

senekor:


senekor:
Similar to https://github.com/exercism/rust/pull/1772, there are two additional test cases here we might want to keep only for the Rust track.

senekor:
generation of the additional tests based on https://github.com/exercism/rust/pull/1774

ErikSchierboom (approved):

PR #1775, opened 2023-11-14
Sync saddle-point with problem-specifications

senekor:


senekor:
There are some different test cases here between the Rust track and problem-specifications. But it's not obvious to me that the tests on the Rust track cover some edge case that is not covered by problem-spec and should be upstreamed.

ErikSchierboom (approved):

PR #1777, opened 2023-11-17
Fix renamed clippy lint

senekor:


senekor:
The rename was stabilized in Rust 1.74, which was released yesterday.

ErikSchierboom (approved):

PR #1779, opened 2023-11-17
Sync run-length-encoding with problem-specifications

senekor:
[no important files changed]

### DIFF
--- a/exercises/practice/run-length-encoding/.docs/instructions.md
+++ b/exercises/practice/run-length-encoding/.docs/instructions.md
@@ -2,8 +2,7 @@
 
 Implement run-length encoding and decoding.
 
-Run-length encoding (RLE) is a simple form of data compression, where runs
-(consecutive data elements) are replaced by just one data value and count.
+Run-length encoding (RLE) is a simple form of data compression, where runs (consecutive data elements) are replaced by just one data value and count.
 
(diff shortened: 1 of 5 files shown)


ErikSchierboom:
Great. Before I approve the sync PRs, please keep in mind that if the test file has changed but its changes wouldn't result in any test result changes (e.g. because the test file changes are only cosmetic), please make sure to merge the PR with `[no important files changed]` in the **commit body**. An easy way to do this is by adding it to the description bit when doing a squash merge. See https://exercism.org/docs/building/tracks#h-avoiding-triggering-unnecessary-test-runs for more information.

senekor:
Will do!

Would you consider changing the settings of this repository, so the PR description is automatically suggested as the message body of the squash commit? That way I can write it in the description and you will know I have already thought about it. And there is no risk of me then forgetting to actually put it in the commit message, like it happenend when I acidentally reran the tests for every single exercise...

image

ErikSchierboom:
> Would you consider changing the settings of this repository, so the PR description is automatically suggested as the message body of the squash commit? That way I can write it in the description and you will know I have already thought about it. And there is no risk of me then forgetting to actually put it in the commit message, like it happenend when I accidentally reran the tests for every single exercise...

Good idea. Done!

ErikSchierboom (approved):

PR #1786, opened 2023-11-19
Sync raindrops with problem-specifications

senekor:
[no important files changed]


senekor:
There is one test case that was removed:

```
assert_eq!("12121", raindrops::raindrops(12_121));
```

I can't find out what specific requirement / plausible bug this is testing against, so it seems OK to delete it.

ErikSchierboom (approved):

PR #1797, opened 2023-11-22
Improve exercise generator

senekor:
- Add a clap-based CLI
  All user-input can now be entered upfront with cli flags
  or later by being prompted for it interactively.

- Limit output of rustfmt passed onto user
  (There are usually many duplicates when generating test cases.)

- Handle absence of canonical data
  (e.g. for custom exercises) not from problem-specifications)

piggyback:

- Make exercise difficulty in track config type-safe


senekor:
Apart from a few internal refactorings I'm quite happy about, the main win here for me is the fact that one can run
```bash
just update-exercise --slug accumulate
```
to synchronize an exercise.

Until now, one would always be prompted for the exercise slug interactively. When I'm tweaking an exercise's test template, I run the generator very often and it can get tedious to enter the slug interactively every time.

ErikSchierboom (approved):

PR #1816, opened 2023-11-27
Sync triangle with problem-specifications

senekor:
There was one challenge implementing the test template. The problem specification of this exercise states that it is up to the individual language tracks how to handle invalid triangles. However, this property is not easy to detect from the specification of the test cases. The test cases with invalid triangles were excluded in tests.toml and reimplemented in additional-tests.json with a custom property of "invalid".

It was also necessary to separate the different properties into individual modules. Some descriptions - and therefore the generated test case names - are duplicated across different properties, so they have to be in different namespaces.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -152,6 +152,10 @@ Feel free to add your own in the crate `rust-tooling`.
 Custom filters added there will be available to all templates.
 How to create such custom filters is documented int he [tera docs][tera-docs-filters].
 
+For a rather complicated example, check out the test template of `triangle`.
+It organizes the test cases in modules and dynamically detects which tests to put behind feature gates.
+That exercise also reimplements some test cases from upstream in `additional-tests.json`, in order to add more information to them necessary for generating good tests.
+
 [tera-docs]: https://keats.github.io/tera/docs/#templates
 [word-count-tmpl]: /exercises/practice/word-count/.meta/test_template.tera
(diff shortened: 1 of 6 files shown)


senekor:
In general, the previous test suite seems to have drifted pretty far from problem-specifications. I have noticed two test cases from the previous suite that are now missing, they test whether invalid triangles with floating-point sides are handled correctly. Such a test case does not exist in problem-specifications.

I couldn't think of a case where invalid integer triangles are handled correctly, while invalid floating-point triangles are not, so I think these tests don't provide any value. If I'm missing something and they do provide value, they should probably be upstreamed instead of added to `additional-tests.json`. There was apparantly an effort to upstream these cases, but the [PR](https://github.com/exercism/problem-specifications/pull/1598) was closed due to becoming stale.

ErikSchierboom:
> In order to be able to feature-gate the floating point tests, the
word "float" is detected in the test case description. This is not
perfect, but likely robust enough, even if there are more such tests
in the future.

Maybe it would be worth adding the [`floating-point` scenario](https://github.com/exercism/problem-specifications/blob/main/SCENARIOS.txt#L5) to the [test cases](https://github.com/exercism/problem-specifications/blob/main/exercises/triangle/canonical-data.json)?

ErikSchierboom (approved):

senekor:
Sorry this slipped under the desk a bit. Thanks for suggesting the scenario, I wasn't aware of that feature in `canonical-data.json` yet. I opened a [PR](https://github.com/exercism/problem-specifications/pull/2349) in the problem-spec repo. I'll wait with this PR for that and simplify accordingly.

PR #1822, opened 2023-12-13
Sync series with problem-specifications

senekor:


senekor:
One test case that's being excluded could be controversial. There is a forum post to discuss it [here](https://forum.exercism.org/t/feedback-wanted-removing-controversial-test-from-series-exercise/8659).

senekor:
There was general agreement with the removal of the test on the forum.

ErikSchierboom (approved):

PR #1828, opened 2023-12-15
Add test group description to single test cases

senekor:
This is intended to help identify which group a single test case belongs to, in order to enable group-specific logic in the test template.


senekor:
I originally implemented this for robot-simulator, but it turned out [it wasn't necessary](https://github.com/exercism/rust/pull/1829/commits/a4d8d51571f2d45dedcf7d137d0928dc41bfb655). Closing this, should be easy enough to reimplement if it turns out to be needed later.

Issue #1826, opened 2023-12-15
Test in CI that stubs don't generate clippy warnings


senekor:
With the current plan to run clippy in our analyzer, it would be benefitial to make sure stubs don't trigger clippy warnings.

Currently, an annoyance that impacts many exercises is `clippy::new_without_default`:

> you should consider adding a `Default` implementation

This lint is definitely good in general settings, but not for us. We don't want to force students to write constructors in terms of the Default trait. They may not even be at a stage of their learning where they are comfortable with traits. And an exercise is not a library to be published where it's important to have the standard traits implemented.

We might consider suppressing this lint (and possibly others) via the recently stabilized [lint configuration through cargo](https://blog.rust-lang.org/2023/11/16/Rust-1.74.0.html#lint-configuration-through-cargo). This will keep the code of the student's solution clean and provide a good experience, be that locally or online.

senekor:
Implementation idea: Add a new test in `rust-tooling/ci-tests`. Run clippy over each exercise in parallel by spawning a child process for each. Create a temporary directory for each exercise and point cargo to that as the target directory, to avoid modifying any files in the repo. Don't let child processes inherit stdout (might get messed up due to parallelism), instead collect outputs and collect dem into a nicely formatted logfile (e.g. in the repo root, added to gitignore).

PR #1830, opened 2023-12-16
Test against clippy warnings in stubs

senekor:


senekor:
closes #1826

senekor:
I'm not 100% sure right now what exactly the bash scripts are already testing, there is some clippy stuff in there too. (e.g. [here](https://github.com/exercism/rust/blob/main/bin/test_exercise.sh#L79))

Might be worth it moving more of the CI tests into rust-tooling. That would simplify the CI configuration and reduce the risk of duplication across Bash & Rust. But that's something for another PR I'd say.

PR #1837, opened 2023-12-17
Test parameter naming convention

senekor:


senekor:
The implementation of the test is a little primitive and comes with fine print.

Some usage of `_parameter` is actually ok, mostly when generics are used. This could technically be detected with AST parsing, but that seems overkill for this purpose. I just hard coded a list of exceptions. If a new exercise is created where `_parameter` is ok, it has to be added to the list of exceptions. That might be considered a stumbling block for future maintainers. I added a banner to the failed test output, notifying about the possibility to add an exception.

Another option would be to scrap the test and document the convention in `CONTRIBUTING.md`. It's not the end of the world either if some exercise gets added that doesn't follow the convention.

But the test also helped me to systematically go through all exercises. Doing that by hand would've been tedious. So I do think the test has value.

ErikSchierboom (approved):

PR #1842, opened 2023-12-22
Add a test to the CircularBuffer exercise to check if elements in the buffer are dropped when the buffer is dropped

LarsKue:
A common solution to this exercise is to use `MaybeUninit<T>` in the buffer. However, `MaybeUninit<T>` explicitly never calls `T`s drop code.

A sound solution should drop elements in the buffer when it is dropped. This test checks for that.

### DIFF
--- a/exercises/practice/circular-buffer/tests/circular-buffer.rs
+++ b/exercises/practice/circular-buffer/tests/circular-buffer.rs
@@ -143,6 +143,18 @@ fn overwrite_replaces_the_oldest_item_remaining_in_buffer_following_a_read() {
 
 #[test]
 #[ignore]
+fn dropping_the_buffer_drops_its_elements() {
+    let element = Rc::new(());
+    {
+        let mut buffer = CircularBuffer::new(1);
+        assert!(buffer.write(Rc::clone(&element)).is_ok());
+        assert_eq!(Rc::strong_count(&element), 2);
+    }
+    assert_eq!(Rc::strong_count(&element), 1);
+}
+
+#[test]
+#[ignore]
 fn integer_buffer() {
     let mut buffer = CircularBuffer::new(2);
     assert!(buffer.write(1).is_ok());


LarsKue:
As per request on the forum post, here is a solution that passes the current test suite but not the new test:

```rust
#![feature(non_null_convenience)]

use std::mem::MaybeUninit;
use std::ptr::NonNull;

#[derive(Debug)]
pub struct CircularBuffer<T> {
    origin: Box<[MaybeUninit<T>]>,
    reader: NonNull<MaybeUninit<T>>,
    writer: NonNull<MaybeUninit<T>>,
    len: usize,
    cap: usize,
}

#[derive(Debug, PartialEq, Eq)]
pub enum Error {
    EmptyBuffer,
    FullBuffer,
}

impl<T> CircularBuffer<T> {
    pub fn new(capacity: usize) -> Self {
        let mut origin = Self::new_origin(capacity);
        let reader = unsafe { NonNull::new_unchecked(origin.as_mut_ptr()) };
        let writer = reader.clone();

        Self {
            origin,
            reader,
            writer,
            len: 0,
            cap: capacity,
        }
    }

    fn new_origin(capacity: usize) -> Box<[MaybeUninit<T>]> {
        let layout = std::alloc::Layout::array::<T>(capacity).unwrap();
        let ptr = unsafe { std::alloc::alloc(layout) as *mut MaybeUninit<T> };
        unsafe { Box::from_raw(std::slice::from_raw_parts_mut(ptr, capacity)) }
    }

    pub fn write(&mut self, element: T) -> Result<(), Error> {
        if self.is_full() {
            return Err(Error::FullBuffer);
        }

        unsafe {
            // write element
            // safety: we know the writer points into the buffer
            self.writer.write(MaybeUninit::new(element));

            // advance writer
            // safety: use wrapping offset so the writer remains in the buffer
            self.writer = NonNull::new_unchecked(self.writer.as_ptr().wrapping_offset(1));
        }

        self.len += 1;

        Ok(())
    }

    pub fn read(&mut self) -> Result<T, Error> {
        if self.is_empty() {
            return Err(Error::EmptyBuffer);
        }

        let element;
        unsafe {
            // read element
            // safety: we know the reader points into the buffer
            // safety: we know the memory at the reader is initialized because the buffer is not empty
            element = self.reader.as_ptr().read().assume_init();

            // advance reader
            // safety: use wrapping offset so the reader remains in the buffer
            self.reader = NonNull::new_unchecked(self.reader.as_ptr().wrapping_offset(1));
        }

        self.len -= 1;

        Ok(element)
    }

    pub fn clear(&mut self) {
        while let Ok(_) = self.read() { }
    }

    pub fn overwrite(&mut self, element: T) {
        if !self.is_full() {
            self.write(element).unwrap();
            return;
        }

        unsafe {
            // write element
            // safety: we know the writer points into the buffer
            let _ = std::mem::replace(&mut *self.writer.as_ptr(), MaybeUninit::new(element));

            // advance writer
            // safety: use wrapping offset so the writer remains in the buffer
            self.writer = NonNull::new_unchecked(self.writer.as_ptr().wrapping_offset(1));

            // advance reader
            // safety: use wrapping offset so the reader remains in the buffer
            self.reader = NonNull::new_unchecked(self.reader.as_ptr().wrapping_offset(1));
        }
    }

    pub fn is_full(&self) -> bool {
        self.len == self.cap
    }

    pub fn is_empty(&self) -> bool {
        self.len == 0
    }

    pub fn len(&self) -> usize {
        self.len
    }

    pub fn capacity(&self) -> usize {
        self.cap
    }
}

/// Without this the new test is not passed
// impl<T> Drop for CircularBuffer<T> {
//     fn drop(&mut self) {
//         self.clear();
//     }
// }
```

senekor:
[forum post for reference](https://forum.exercism.org/t/add-a-test-to-the-circularbuffer-exercise-to-check-if-elements-in-the-buffer-are-dropped-when-the-buffer-is-dropped/8817)

senekor (approved):

PR #1844, opened 2023-12-23
Sync anagram with problem-specifications

senekor:


ErikSchierboom (approved):

ler0sQ:
@senekor According to the problem's description there'll be ONLY ASCII alphabetic characters. What am I missing? (I'm no developer here, just a bit frustrated problem solver, who found himself in the middle of multiple changes)

senekor:
@ler0sQ you're right, there's a mismatch between the instructions and the unicode tests. I can't _change_ the instructions, they're coming from upspream. But I can make track-specific additions at the end. Would it fix the confusion to add this sentence at the end of the instructions:

> The Rust track extends the possible letters to be any unicode character, not just ASCII alphabetic ones.

PR #1847, opened 2024-01-04
Fix already-solved lifetime challenge in anagram

senekor:
This is basically a revert of #1755.

The original "fix" was a reaction to this forum post: https://forum.exercism.org/t/the-function-signature-for-anagrams-in-rust-is-missing-a-lifetime/7677

It seems the exercise instructions, which explicitly mention that this missing lifetime annotation is intentional, went unnoticed.


senekor:
For reference, I was made aware of this on [Discord](https://discord.com/channels/854117591135027261/1191925044885073940/1192208387421773995).

ErikSchierboom (approved):

PR #1848, opened 2024-01-04
Update instructions append for anagram

senekor:
- Remove some performance considerations. These are basically giving away the implementation. Also, not every student might want to focus on performance. Such considerations may be better suited for approaches articles.

- Remove non-Rust-related hints. These are pretty much covered by the instructions. I guess these were lacking at some point, but updated in the meantime.

- Move the most important hint about lifetimes to the top.

### DIFF
--- a/exercises/practice/anagram/.docs/instructions.append.md
+++ b/exercises/practice/anagram/.docs/instructions.append.md
@@ -1,7 +1,4 @@
-# Hints
+# Instructions append
 
 You are going to have to adjust the function signature provided in the stub in order for the lifetimes to work out properly.
 This is intentional: what's there demonstrates the basics of lifetime syntax, and what's missing teaches how to interpret lifetime-related compiler errors.
-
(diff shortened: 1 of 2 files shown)


ErikSchierboom (approved):

exercises/practice/anagram/.docs/instructions.append.md
```diff
@@ -1,11 +1,7 @@
 # Hints
```
ErikSchierboom, line 1: BTW, it is possible to have a `.docs/hints.md` file that contains, well, hints.

ErikSchierboom (approved):

exercises/practice/anagram/.docs/instructions.append.md
senekor: Good idea! I moved part of it over there.

ErikSchierboom (approved):

PR #1852, opened 2024-01-11
Maintain the signature of the recursive function identical to that of the loop

herobs:


MatthijsBlom:
The code you changed is replicated in `recursion/content.md`. I suspect you'd want to change it there too.

senekor:
What's the motivation for this? Some forum post I missed? A description would've been nice.

Anyway, I like this change. It's cleaner, simpler. A slight downside might be that while Rust doesn't guarantee tail recursion optimization, as is mentionend in the text, it is now impossible. (right?) But since using recursion in Rust over a loop is not a choice one makes for performance reasons, I'm not worried about it.

cc @bobahop, your take would be welcome since you wrote the approach.

herobs:
Sorry for not providing a description and discussing before creating a pull request. As the title states, there is no need to change the recursive function's signature to accommodate the offset argument.

As for tail recursion, they aren't simple tail recursions for both versions.

senekor (approved):

PR #1872, opened 2024-03-17
Use `as_bytes()` instead of `chars()` for ASCII strings

BillQK:
This pull request updates the code to use `as_bytes()` instead of `chars()` for processing ASCII strings. Since all the inputs and outputs are in ASCII, using `as_bytes()` allows us to work directly with the underlying byte representation of the string as a `[u8]` (byte slice). This change improves performance by avoiding the need to check for variable-length UTF-8 characters when iterating over the ASCII data. ASCII bytes are always one byte long, making iteration over a slice of ASCII bytes more efficient.


senekor:
CI needs fixes, otherwise ok to merge

senekor (approved):

PR #1877, opened 2024-03-27
Add Exercise Matrix

jpal91:
Per forum post - [Adding Matrix to the Rust Track](http://forum.exercism.org/t/adding-matrix-to-the-rust-track/10480)

Here's what I did:

- Added Matrix exercise from problem-specifications using `rust-tooling/generate`
- Created example.rs
- Added all required tests
- Created boilerplate in lib.rs

This is my first contribution so there's a chance I probably forgot something but it looks like all cargo and CI tests are running on my end, so hopefully we should be good.

Also, I increased the difficulty to 4 as depending on the implementation you could be working with nested iterators/map/collect, but I was going back and forth on this. If you think 1 is still more appropriate, just let me know and I'll switch it back. 

Thanks!

### DIFF
--- a/exercises/practice/matrix/.docs/instructions.append.md
+++ b/exercises/practice/matrix/.docs/instructions.append.md
@@ -2,7 +2,7 @@
 
 ## Challenges
 
-- Can you implement this solution with using **ANY** `for` loops? (ie using [iterators][iterators] instead)
+- Can you implement this solution without using any `for` loops? (i.e. using [iterators][iterators] instead)
 - Although it would be relatively straight-forward to have two data structures in `Matrix` (one representing rows and another for columns), can you implement this solution without having to hold two separate copies of the input in your struct?
 
 ## Helpful Methods
--- a/exercises/practice/matrix/.docs/instructions.append.md
+++ b/exercises/practice/matrix/.docs/instructions.append.md
@@ -1,0 +1,20 @@
+# Instructions append
+
+## Challenges
+
+- Can you implement this solution with using **ANY** `for` loops? (ie using [iterators][iterators] instead)
+- Although it would be relatively straight-forward to have two data structures in `Matrix` (one representing rows and another for columns), can you implement this solution without having to hold two separate copies of the input in your struct?
+
+## Helpful Methods
+- [`map()`][map]
+- [`filter()`][filter]
+- [`lines()`][lines]
+- [`split()`][split]
+- [`parse()`][parse]
+
+[iterators]: https://doc.rust-lang.org/book/ch13-02-iterators.html
+[map]: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.map
+[filter]: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter
+[lines]: https://doc.rust-lang.org/std/primitive.str.html#method.lines
+[split]: https://doc.rust-lang.org/std/primitive.str.html#method.split
+[parse]: https://doc.rust-lang.org/std/primitive.str.html#method.parse
--- a/exercises/practice/matrix/.meta/additional-tests.json
+++ b/exercises/practice/matrix/.meta/additional-tests.json
@@ -7,7 +7,10 @@
             "string": "1 2 3\n4 5 6\n7 8 9",
             "index": 4
         },
-        "expected": null
+        "expected": null,
+        "comments": [
+            "Additional test to ensure the method can handle invalid input"
+        ]
     },
     {
         "uuid": "304d71d4-cf26-41d4-9b74-cc04441fec8c",
@@ -17,6 +20,9 @@
             "string": "1 2 3\n4 5 6\n7 8 9",
             "index": 4
         },
-        "expected": null
+        "expected": null,
+        "comments": [
+            "Additional test to ensure the method can handle invalid input"
+        ]
     }
 ]
--- a/exercises/practice/matrix/.meta/additional-tests.json
+++ b/exercises/practice/matrix/.meta/additional-tests.json
@@ -1,22 +1,22 @@
-{
-    "cases": [
-        {
-            "description": "cannot extract row with no corresponding row in matrix",
-            "property": "row",
-            "input": {
-                "string": "1 2 3\n4 5 6\n7 8 9",
-                "index": 4
-            },
-            "expected": null
+[
+    {
+        "uuid": "9ca9c869-7220-4783-93ba-3e4f96b81732",
+        "description": "cannot extract row with no corresponding row in matrix",
+        "property": "row",
+        "input": {
+            "string": "1 2 3\n4 5 6\n7 8 9",
+            "index": 4
         },
-        {
-            "description": "cannot extract column with no corresponding column in matrix",
-            "property": "column",
-            "input": {
-                "string": "1 2 3\n4 5 6\n7 8 9",
-                "index": 4
-            },
-            "expected": null
-        }
-    ]
-}
+        "expected": null
+    },
+    {
+        "uuid": "304d71d4-cf26-41d4-9b74-cc04441fec8c",
+        "description": "cannot extract column with no corresponding column in matrix",
+        "property": "column",
+        "input": {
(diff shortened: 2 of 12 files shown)


senekor (changes requested):
Haven't done a detailed review yet, but it looks like you're not using the Tera template to generate the tests. Please do that so we can keep the test suite up-to-date easily. Fastest way to regenerate tests from the template: `just update-exercise --slug word-count` (need `just` installed)

You can find some tips about writing the Tera template in `CONTRIBUTING.md`.

senekor:
I'd say difficulty 4 is fine. Looking through the config file, there are many other exercises at difficulty 4 which I would categorize as "light data structure handling".

jpal91:
Sorry about that. I realized when I forked the repo the `problem-specifications` submodule didn't pull in correctly so the tests never wrote when I ran the generate command the first time. I thought I had to write all the tests from scratch :laughing:.

I think I got the hang of it and everything should be correct now.

exercises/practice/matrix/src/lib.rs
```diff
@@ -0,0 +1,17 @@
+pub struct Matrix {
+    // Implement your Matrix struct
+}
```
senekor, line 3: Did you consider any other design than struct + methods? as far as the upstream test cases are concerned, it could just be two plain functions without an intermediate data structure. The way the tests are written now, this struct + methods approach cannot be avoided.

exercises/practice/matrix/.meta/test_template.tera
```diff
@@ -0,0 +1,18 @@
...
+{% endif -%}
+fn {{ test.description | slugify | replace(from="-", to="_") }}() {
+    let matrix = Matrix::new({{ test.input.string | json_encode() }});
+    assert_eq!(matrix.{{ test.property }}({{ test.input.index }}), Some(vec!{{ test.expected | json_encode() }}));
+    {% if loop.index == 4 -%}
+    assert_eq!(matrix.column({{ test.input.index }}), None);
+    {% endif -%}
+    {% if loop.index == 7 -%}
+    assert_eq!(matrix.row({{ test.input.index }}), None);
+    {% endif -%}
```
senekor, line 16: This is well intentioned, but we should stick to problem-specifications as closely as possible. If we find that some test cases are missing, it is better to add them upstream. (In rare cases where Rust-specific tests make sense, there can be an `additional-tests.json` file that is merged with the standard test cases.)

I don't think we need to check explicitly that this returns `None`, what else could it possibly do? Maybe an empty vector?

I am surprised upstream has not a single error handling test case. maybe that it's worth upstreaming at least one.

exercises/practice/matrix/.docs/instructions.append.md
```diff
@@ -0,0 +1,20 @@
+# Instructions append
+
+## Challenges
+
+- Can you implement this solution with using **ANY** `for` loops? (ie using [iterators][iterators] instead)
```
senekor, line 5:
```suggestion
- Can you implement this solution without using any `for` loops? (i.e. using [iterators][iterators] instead)
```

senekor (changes requested):

exercises/practice/matrix/src/lib.rs
jpal91: I did not just because I was basing the design on the Python and JavaScript implementations (only other languages I'm comfortable with) and both implemented classes for the exercise. 

I will say that the tests on the other languages would agree with your thought process. None of them are really in need of having to keep the state of the input because it's always just checking either row or column but not both. The only thing I would say goes in favor of having a struct is if this were a real life program, I would want to use a struct so I can keep the state of the input and reuse it when needed.

I can re-write as two functions, though, if you think it would be more beneficial to the track overall.

exercises/practice/matrix/.meta/test_template.tera
jpal91: I did find it strange as well, it really feels like there should be some kind of error handling test. Most other exercises assume at least one test is going to have invalid input. 

However, pushing it upstream would cause breaking changes in other repos I imagine? Obviously I'm new here so if you think it's a better option to just switch it to an empty vec, or keep it the same and not specifically test for a `None` output, just let me know.

exercises/practice/matrix/.docs/instructions.append.md
jpal91: I'm a dummy, I forgot to add `no important changes`. I'll remember next time, sorry.

exercises/practice/matrix/.meta/test_template.tera
senekor: > However, pushing it upstream would cause breaking changes in other repos I imagine?

That's not a problem. Updating an exercise with new tests from upstream is generally a manual process per track (though my long-term goal with the generator is to automate much of that work for the Rust track). If some language track wants to update and absolutely does not want to do error handling, a language track can always exclude that specific test in `.meta/tests.toml`.

Would you mind opening a forum post about such a new test case and presumably follow that up with a PR?

I think the `Option<Vec<_>>` return type is definitely idiomatic, let's keep it. We can either wait for the error test to be upstreamed or delete these special cases to keep this PR unblocked. Up to you.

exercises/practice/matrix/.docs/instructions.append.md
senekor: You don't need to worry about that. It's only needed when updating an existing exercise, so it's irrelevant for this PR. There is also an automation workflow which comments on PRs to remind people when it's relevant. Lastly, we always squash PRs and the `[no important files changed]` label only matters on those squash commits - so, when merging a PR.

exercises/practice/matrix/.meta/test_template.tera
jpal91: > If some language track wants to update and absolutely does not want to do error handling, a language track can always exclude that specific test in .meta/tests.toml.

Totally forgot about this!

I would be happy to open the forum post and I'll wait to see what the feedback is before proceeding on this piece.

senekor: I guess Isaac's argument makes sense to not have the input validation at all. The fact that Rust has such a focus on handling errors explicitly leaves us in a bit of an akward position though. `-> Option<Vec<_>>` is the idiomatic return type, I wouldn't change it. Otherwise we're forcing bad code on users. But it's weird to not have a single test case with `None`. I think two options make sense at this point, I'll leave the choice up to you:

- Not do any input validation at all, as intended by upstream.
- Add custom tests with `None` to `.meta/additional-tests.json`.

jpal91: I agree that he made valid points, and I definitely understand what he's saying. I'm personally still in the mindset of some kind of invalid input test makes sense for this particular exercise, but can see where it may cause larger issues.

I did opt to add in some additional tests -

- I added the `additional-tests.json` file
- I created two new UUIDs (hopefully that was what I was supposed to do)
- Added comments
- Regenerated

Let me know if that works.

exercises/practice/matrix/.meta/additional-tests.json
```diff
@@ -0,0 +1,28 @@
...
+        "description": "cannot extract row with no corresponding row in matrix",
+        "property": "row",
+        "input": {
+            "string": "1 2 3\n4 5 6\n7 8 9",
+            "index": 4
+        },
+        "expected": null,
+        "comments": [
+            "Additional test to ensure the method can handle invalid input"
+        ]
```
senekor, line 13: I'd like this comment to answer two questions:
- Why was the test case not upstreamed / why do other languages not need this test?
- What additional value does the test provide?

suggestion:
```json
[
  "Additional test to ensure the method can handle invalid input",
  "Upstream does not want to add this test to avoid every exercise being about input validation.",
  "Rust puts a lot of emphasis on error handling, hence the idiomatic Option return type.",
  "With Option in the API, it makes sense to test for None at least once."
]
```

jpal91: That makes sense and good to know for the future!

I added the comments into both. We should be good to go now.

senekor (approved):
Thank you for the great work!

PR #1879, opened 2024-03-27
Add test to ensure tera templates are in sync

### LINKED ISSUE #1878: Add test that templates match generated test suite
Every exercise that already has a `.meta/test_template.tera` file should be tested in CI that the template actually matches the test file on disk. This should help with adding new exercises so people don't forget to use the template and write the test suite manually. Also, it will help when refactoring the exercise generator, to prevent mistakes that break test generation for some exercise. With such a test, that would be noticed only when the affected exercise would be updated the next time.

senekor:
closes #1878

### DIFF
--- a/exercises/practice/knapsack/tests/knapsack.rs
+++ b/exercises/practice/knapsack/tests/knapsack.rs
@@ -1,6 +1,16 @@
 use knapsack::*;
 
 #[test]
+fn test_no_items() {
+    let max_weight = 100;
+    let items = [];
+    let output = maximum_value(max_weight, &items);
+    let expected = 0;
+    assert_eq!(output, expected);
+}
+
+#[test]
+#[ignore]
 fn test_one_item_too_heavy() {
     let max_weight = 10;
     let items = [Item {
--- a/exercises/practice/queen-attack/.meta/test_template.tera
+++ b/exercises/practice/queen-attack/.meta/test_template.tera
@@ -1,4 +1,4 @@
-use queen_attack::{ChessPiece, ChessPosition, Queen};
+use queen_attack::*;
 
 {% for test in cases %}
 #[test]
--- a/exercises/practice/variable-length-quantity/.meta/test_template.tera
+++ b/exercises/practice/variable-length-quantity/.meta/test_template.tera
@@ -7,7 +7,7 @@ use variable_length_quantity as vlq;
 fn {{ test.description | slugify | replace(from="-", to="_") }}() {
 {%- if test.property == "encode" %}
     let input = &{{ test.input.integers | json_encode() }};
-    let output = vlq::{{ fn_names[0] }}(input);
+    let output = vlq::to_bytes(input);
     let expected = vec![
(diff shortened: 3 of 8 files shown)


exercises/practice/knapsack/tests/knapsack.rs
```diff
@@ -1,6 +1,16 @@
...
+fn test_no_items() {
+    let max_weight = 100;
+    let items = [];
+    let output = maximum_value(max_weight, &items);
+    let expected = 0;
+    assert_eq!(output, expected);
+}
+
+#[test]
+#[ignore]
```
senekor, line 13: I think this is a normal update from problem-specifications.

exercises/practice/queen-attack/.meta/test_template.tera
```diff
@@ -1,4 +1,4 @@
-use queen_attack::{ChessPiece, ChessPosition, Queen};
+use queen_attack::*;
```
senekor, line 1: This was inconsistent, must've been a mistake of mine. The change doesn't have any effect though.

exercises/practice/variable-length-quantity/.meta/test_template.tera
```diff
@@ -19,7 +19,7 @@ fn {{ test.description | slugify | replace(from="-", to="_") }}() {
             0x{{ byte | to_hex }},
         {%- endfor -%}
     ];
-    let output = vlq::{{ fn_names[1] }}(input);
+    let output = vlq::from_bytes(input);
```
senekor, line 22: `fn_names[1]` was failing for some reason, the explicit function name is more readable anyway.

exercises/practice/wordy/.meta/test_template.tera
```diff
@@ -3,8 +3,11 @@
 {% if loop.index != 1 -%}
 #[ignore]
 {% endif -%}
+{% if test.property == "exponentials" -%}
+#[cfg(feature = "exponentials")]
+{% endif -%}
 fn {{ test.description | slugify | replace(from="-", to="_") }}() {
-    let input = {{ test.input | json_encode() }};
+    let input = {{ test.input.question | json_encode() }};
```
senekor, line 10: Weird mistake, the test generation fails completely without `.question`.

exercises/practice/wordy/.meta/additional-tests.json
```diff
@@ -0,0 +1,28 @@
...
+    "comments": [
+      "This test case was added a long time ago.",
+      "Upstreaming it would make the exercise more difficult."
+    ],
+    "property": "exponentials",
+    "input": {
+      "question": "What is 2 raised to the 5th power?"
+    },
+    "expected": 32
+  },
```
senekor, line 14: These cases were in the finished test file, but test generation would remove them, because their not in `canonical-data.json`. I decided to keep them around, but not upstream them.

ErikSchierboom (approved):

PR #1884, opened 2024-03-30
generator: strip first ignore-annotation in Rust

senekor:
[no important files changed]


rust-tooling/generate/src/lib.rs
```diff
@@ -121,7 +121,13 @@ fn generate_tests(slug: &str, fn_names: Vec<String>) -> String {
...
 
     let rendered = template.render("test_template.tera", &context).unwrap();
-    let rendered = rendered.trim_start();
+
+    // Remove ignore-annotation on first test.
+    // This could be done in the template itself,
+    // but doing it here makes all templates more readable.
+    // Also, it is harder to do this in the template when the template
+    // generates test functions inside a macro for modules.
+    let rendered = rendered.replacen("#[ignore]\n", "", 1);
```
senekor, line 130: This is the interesting part - the rest is just removing the equivalent logic from all tera templates.

exercises/practice/allergies/tests/allergies.rs
```diff
@@ -335,7 +335,6 @@ fn allergic_to_everything_cats() {
 
 #[test]
 #[ignore]
-
```
senekor, line 338: The previous version of the template for `allergies` seems to have generated unnecessary whitespace. This change fixes that (by accident).

ErikSchierboom: Funnily enough, I just noticed that yesterday!

ErikSchierboom (approved):

Issue #1824, opened 2023-12-14
Remove all util functions from test files


senekor:
Some test files include some common code in the form of a funtion. Individual tests then call that function. This is a good instinct to do in normal code: reduce duplication (DRY). But our test runner should at some point (hopefully soon) gain the ability to extract the code of a test to show it to students when a test fails. In that case, it is very bad to call a separate function in the test body. Every test function should contain all of the logic required to understand what it's doing.

Here's an [example](https://github.com/exercism/rust/blob/a666018acf3a431aa1c7dd0a188f893e02517c3f/exercises/practice/word-count/tests/word-count.rs#L3) of such a util function.

Since we're now using a test generator anyway, we don't need to worry about having to keep all this copied logic in sync manually.

Make sure to update [the documentation](https://github.com/exercism/rust/blob/main/docs/CONTRIBUTING.md#tera-templates) where these util funtions are encouraged.

senekor:
Apart from util functions, declarative macros should not be used either for the same reason.

senekor:
On second thoughts, macros might be fine. If we're going to get the source code of the tests by AST, macros should already be resolved by that point.

And using tera templates for exercises which don't come from problem-specifications might be inconvenient. Macros can still be a good option in that case.

senekor:
Might also be a good idea to document this in `CONTRIBUTING.md`, since it's not really possible to write a test against this.

senekor:
For reference, here is an [earlier discussion](https://github.com/exercism/rust/issues/392#issuecomment-343865993) about using table-driven tests for this track.

senekor:
> On second thoughts, macros might be fine. If we're going to get the source code of the tests by AST, macros should already be resolved by that point.

This is incorrect, AST parsing with `syn` does not resolve macro invocations. There are only five exercises with macro definitions in test files left, we should just get rid of them.

senekor:
There are still a few of these around. Not every piece of testing code has been moved into a test function body. But I think for the remaining occurrences, it makes sense and the test functions are still understandable on their own.

PR #1886, opened 2024-03-31
generator: remove custom context

senekor:
`crate_name` and `fn_names` were previously injected into the context for rendering a test template to allow something like the following in a template:

```
let output = {{ crate_name }}::{{ fn_names[0] }}(input);
```

This made both the templates and Rust code for the generator more complicated. The benefit is only a few characters less typing when creating the template once. In hindsight, this is not worth it in my opinion.

[no important files changed]


docs/CONTRIBUTING.md
```diff
@@ -134,18 +134,10 @@ let expected = {% if test.expected is object -%}
 {%- endif %};
 ```
 
-If every test case needs to do some crunching of the inputs,
-you can add utils functions at the top of the tera template.
-See [`word-count`'s template][word-count-tmpl] for an example.
```
senekor, line 139: This is a piggy-back, I removed this recommendation because of reasons explained in #1824

ErikSchierboom (approved):

PR #1891, opened 2024-04-02
forth: sync

senekor:
[no important files changed]


senekor:
The two test files are generally quite different, although there are a couple more tests in the new version. I glossed over most of it and I couldn't find anything specific that's missing in the new test suite. But I honestly wasn't that thorough, it's a big diff. I added the "no important files changed" tag because I'm not aware of any new edge case that's supposed to be covered.

ErikSchierboom (approved):

PR #1895, opened 2024-04-03
protein-translation: sync

senekor:
[no important files changed]

### DIFF
--- a/exercises/practice/protein-translation/.meta/additional-tests.json
+++ b/exercises/practice/protein-translation/.meta/additional-tests.json
@@ -1,0 +1,182 @@
+[
+  {
+    "uuid": "ecafab60-89f2-4680-9286-56b423e8a7b9",
+    "description": "methionine",
+    "comments": [
+      "The original design of the exercise deviates from problem-specifications",
+      "by having a name_for function for individual codons.",
+      "Upstreaming is impossible as it would change the design of the exercise.",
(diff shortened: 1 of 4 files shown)


rust-tooling/generate/templates/default_test_template.tera
```diff
@@ -3,7 +3,7 @@ use crate_name::*;
 {% for test in cases %}
 #[test]
 #[ignore]
-fn {{ test.description | slugify | replace(from="-", to="_") }}() {
+fn {{ test.description | snake_case }}() {
```
senekor, line 6: piggyback

senekor:
blocked on [this discussion](https://forum.exercism.org/t/incorrect-test-in-protein-translation/10616)

senekor:
I struggled with the separation of the `name_for` and `of_rna` methods, which is not intended by problem-specifications. I considered to map all tests with only one output amino acid to the `name_for` method, but that doesn't work because of some edge cases:
- one amino acid and then immediately a stop codon (should test `of_rna`)
- invalid codons returning `None` (should test `name_for`)

So the solution I came up with for now is to add all previous `name_for` tests as custom tests. This leads to duplicated tests (same input once for `name_for` and once for `of_rna`) but preserves the existing exercise design.

ErikSchierboom (approved):

PR #1897, opened 2024-04-07
sieve: improve range start in approaches

senekor:


senekor:
As suggested on the Forum:
https://forum.exercism.org/t/minor-error-in-dig-deeper-solution-to-sieve-in-rust/10668

ErikSchierboom (approved):

PR #1898, opened 2024-04-07
collatz-conjecture: sync

senekor:
The additional test cases for overflow are incorrect even with u64 in the API. u128 can be used as needed to calculate the intermediate values without overflow.

[no important files changed]


senekor:
Motivated by [this report on the forum](https://forum.exercism.org/t/pull-request-fixed-collatz-conjecture-tests-that-disallowed-correct-answers/10672) which points out that the overflow test cases are actually incorrect.

ErikSchierboom (approved):

PR #1912, opened 2024-04-28
improve exercise order

senekor:
I believe some work went into the exercise order by the previous maintainers, so I didn't completely redo it. I only fixed the obviously bad aspects of the order:

- difficulty 10 exercises being way too early in the track
- new, easy exercises being at the end of the exercise list

forum discussion:
https://forum.exercism.org/t/rust-track-exercise-order/10855

### DIFF
--- a/config.json
+++ b/config.json
@@ -310,59 +310,6 @@
         ]
       },
       {
-        "slug": "parallel-letter-frequency",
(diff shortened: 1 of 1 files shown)


iHiD:
Nice. Thanks! :)

ErikSchierboom (approved):
Thanks!

PR #1915, opened 2024-04-30
sum-of-multiples: update "deep dive"

chezwicker:
Update the solution and description for the "sum from factors" approach as discussed in [the forum](https://forum.exercism.org/t/sum-of-multiples-dig-deepter-possible-improvement/10994/8).


senekor (changes requested):
lgtm, can you also update the [snippet.txt](https://github.com/exercism/rust/blob/main/exercises/practice/sum-of-multiples/.approaches/from-factors/snippet.txt)?

chezwicker:
Didn't see that, done now.

PR #1933, opened 2024-06-18
Update instructions.append.md for Triangle

mckzm:
The current version of ` instructions.append.md` gives an inaccurate definition of a right triangle:

`an isosceles triangle whose equal sides are separated by 90 degrees`

(a right triangle need not be isosceles)

This PR attempts to fix this while trying not to add too much additional text that's incidental to the exercise - hopefully the result is readable!

The PR also tries to clarify a little the phrasing about  the representation of irrationals, at the end of the same paragraph.


senekor (approved):
as discussed on the [forum](https://forum.exercism.org/t/pr-to-update-instructions-append-md-for-triangle/11752)

PR #1940, opened 2024-07-09
Simplify protein-translation

### LINKED ISSUE #1937: protein-translation: slightly confusing setup
I'm not really sure why the `protein-translation` exercise has an additional `parse` function that returns a `pub struct CodonsInfo<'a>`. It feels a bit weird to me. Maybe a better approach would be to model the codons as an enum?

senekor:
closes #1937


ErikSchierboom (approved):
Lovely! Thanks for that

Issue #1938, opened 2024-07-09
grep exercise fails with implementation using clap on the grader


dan-seol:
```error: no matching package named `clap_derive` found
location searched: registry `crates-io`
required by package `grep v1.3.0 (exercism-iteration)`
As a reminder, you're using offline mode (--offline) which can sometimes cause surprising resolution failures, if this error is too confusing you may wish to retry without the offline flag.```

this was quite surprising given

> /// In the real world, it's common to use crates such as [`clap`] or
/// [`structopt`] to handle argument parsing, and of course doing so is
/// permitted in this exercise as well, though it may be somewhat overkill.

could we fix the grader to accept clap based solutions?

senekor:
That is indeed a problem, thanks for the report. I'm pretty sure that `clap_derive` will be too slow to compile and exercises using it would hit the timeout. Here are the options I can think of:

- Suggest people use the builder API of clap. It's less convenient, but nonetheless an "officially blessed" way to write clap apps. Compile times would hopefully be fast enough. (Note to self: would have to add clap to the test runner in this case.)

- Suggest to use a different crate (lexopt, pico-args...)

- Remove the suggestion to use libraries entirely (sad)

What do you think?

dan-seol:
Hi, I think the first option sounds ideal -- but if the suggestion is already removed it's all good

PR #1942, opened 2024-07-10
grep: remove suggestion of unavailable crates

senekor:
clap and structopt are not available in the test runner. This is because their derive-API (which is the most popular) takes too long to compile.


senekor:
closes #1938

ErikSchierboom (approved):

PR #1947, opened 2024-07-26
Add error handling tests to PaaS I/O

### LINKED ISSUE #992: PaaS I/O test suit does not account for failing io operations.

EMBradley:
Adds two tests to the PaaS I/O exercise to check that I/O errors are propagated rather than unwrapped/expected. [Fixes #992]


senekor:
[forum thread](https://forum.exercism.org/t/rust-paas-i-o-test-cases-for-error-handling/12229) for reference

senekor (approved):
Great, thanks!

PR #1949, opened 2024-08-09
Cleanup new_without_default workarounds

senekor:


bin/ensure_stubs_compile.sh
```diff
@@ -61,7 +61,7 @@ for dir in $changed_exercises; do
         # since we're generally not going to test it.
         if ! (
             cd "$dir" &&
-            cargo clippy --lib --tests --color always -- --allow clippy::new_without_default 2>clippy.log
```
senekor, line 64: I double-checked that exercise stubs don't produce clippy warnings as a preparation for running clippy in our analyzer. Removing this exception ensures that no new exercises with this problem are introduced. The best way to handle exercise-specific lint-exceptions is by specifying them in Cargo.toml.

ErikSchierboom (approved):

PR #1954, opened 2024-08-09
accumulate: sync

senekor:


senekor:
This is my first manual sync. The problem specification just doesn't lend itself to test generation. One would essentially have to parse and transpile the pseudo code in canonical-data...

I made sure that every test in `.meta/tests.toml` has a corresponding one in the rust test file and adjusted the names accordingly. I left the additional tests as they are, they are important to force the users to get the type signature right.

ErikSchierboom (approved):

PR #1957, opened 2024-08-10
Eliuds Eggs => Eliud's Eggs

keiravillekode:
[no important files changed]

This is similar to various tracks needing `Pascals Triangle` => `Pascal's Triangle`,  which was discussed in the [forum](http://forum.exercism.org/t/48in24-exercise-08-27-pascals-triangle/12056/9).

### DIFF
--- a/config.json
+++ b/config.json
@@ -553,7 +553,7 @@
       },
       {
         "slug": "eliuds-eggs",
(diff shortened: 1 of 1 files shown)


senekor (approved):
Thanks!

PR #1969, opened 2024-08-15
generator: replace snake_case with make_ident

senekor:
[no important files changed]

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -156,8 +156,8 @@ In that case, you can use the group description to organize the tests in modules
 There are some custom tera filters in [`rust-tooling`](/rust-tooling/generate/src/custom_filters.rs).
 Here's the hopefully up-to-date list:
 - `to_hex` formats ints in hexadecimal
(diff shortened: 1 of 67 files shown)


senekor:
A lot of template files changed here, but the output generally stays the same. A couple actual test files changed (only the names of the test functions), the diff is [here](https://github.com/exercism/rust/pull/1969/commits/015a79e7f60701a67f22d9799789742178e34ce6).

senekor:
I noticed that the "no important files changed" workflow didn't do its job here. Wild guess: it doesn't work because there are _multiple_ test files changed. In the workflow, this line will be executed multiple times:

```
echo "important_files_changed=${changed}" >> "$GITHUB_OUTPUT"
```

So the variable `GITHUB_OUTPUT` is defined multiple times? I adon't know how github handles this, it may do something random like concatinate all values with commas or similar. If that was the case, the following check won't work anymore:

```
steps.check.outputs.important_files_changed == 'true'

ErikSchierboom (approved):

PR #1973, opened 2024-08-15
crypto-square: sync

senekor:
part of https://github.com/exercism/rust/issues/1824


senekor:
Not adding `[no important files changed]` because the tests changed quite significantly, to the point where the example solution failed the new tests. I replaced it with my own.

ErikSchierboom (approved):

PR #1974, opened 2024-08-15
dominoes: sync

senekor:
[no important files changed]

part of https://github.com/exercism/rust/issues/1824

### DIFF
--- a/exercises/practice/dominoes/.meta/test_template.tera
+++ b/exercises/practice/dominoes/.meta/test_template.tera
@@ -1,0 +1,61 @@
+{% for test in cases %}
+#[test]
+#[ignore]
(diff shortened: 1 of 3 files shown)


senekor:
There is complicated validation logic here, which didn't make sense to copy-paste into every test case. But I didn't want to completely delete it either, because it produces nice error messages that can point users in the right direction about what went wrong.

As a compromise, I added a technically redundant `assert!(output.is_some())` above the calls to `assert_correct`. Once the test code is included in the test runner, users will at least know whether `Some` or `None` was expected. How exactly the error message was produced - for that they'd need to open the test file.

ErikSchierboom (approved):

PR #2004, opened 2025-01-28
Sublist: use args in todo! instead of naming with _

ellnix:
This necessitated adding `std::fmt::Debug` as a constraint to `T`. All of the types in the tests are `i32` so this does not cause the tests to break.

Corresponding forum post: https://forum.exercism.org/t/warn-explain-or-remove-variables/15285

### DIFF
--- a/exercises/practice/sublist/src/lib.rs
+++ b/exercises/practice/sublist/src/lib.rs
@@ -6,6 +6,6 @@ pub enum Comparison {
     Unequal,
 }
 
-pub fn sublist<T: PartialEq>(_first_list: &[T], _second_list: &[T]) -> Comparison {
-    todo!("Determine if the first list is equal to, sublist of, superlist of or unequal to the second list.");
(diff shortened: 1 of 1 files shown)


senekor:
What do you think about dropping the generics and just having `&[i32]` in the sigature? No point in overcomplicating things if the tests are only about integers. There are other exercises using generics.

If you think we should keep the generics, maybe at least import `std::fmt::Debug` to not clutter the function signature too much.

ellnix:
> What do you think about dropping the generics and just having `&[i32]` in the sigature? No point in overcomplicating things if the tests are only about integers. There are other exercises using generics.

It's fine by me, I don't see how a student can use a i32 method that's not in PartialEq to solve this.

senekor:
Yeah that's fine. I prefer force pushing myself. (check out [jj](https://github.com/jj-vcs/jj) if you like clean history 🙂) Exercism enforces squash merging so it doesn't matter much.

senekor (approved):

PR #2005, opened 2025-01-29
palindrome-products: require factors in solution

ellnix:
Apologies if I did something wrong, struggled a little bit with the rust-tooling.

Please let me know of any mistake, miss, or decisions you disagree with, no matter how small. **No hard feelings.** 

# The HashSet

Used a `HashSet<(u64, u64)>` for factors to avoid dealing with ordering of the factors and simultaneously communicate that the order doesn't matter. I thought leaving the inner tuple sorted made sense since that's how it appears in the instructions:

> [...] The largest palindrome product is 9. Its factors are (1, 9) and (3, 3).

A sorted `Vec` would probably be fine, and the change can be made easily enough if you think it would be more appropriate.

# The extra tests

There were two extra tests on the now-defunct `Palindrome::new` function. I felt those only made sense if this exercise still contained the newtype pattern. I can add them again if appropriate.

Corresponding forum post: https://forum.exercism.org/t/palindrome-products-description-mentions-factors/

### DIFF
--- a/exercises/practice/palindrome-products/.meta/test_template.tera
+++ b/exercises/practice/palindrome-products/.meta/test_template.tera
@@ -1,42 +1,25 @@
 use palindrome_products::*;
 
-{#
-    These first two custom test cases are for the object-oriented design of the exercise.
-    They don't fit the structure of the upstream tests, so they're implemented here.
-#}
-
-#[test]
-#[ignore]
-/// test `Palindrome::new` with valid input
-fn palindrome_new_return_some() {
-    for v in [1, 11, 121, 12321, 1234321, 123454321, 543212345] {
-        assert_eq!(Palindrome::new(v).expect("is a palindrome").into_inner(), v);
-    }
-}
-
-#[test]
-#[ignore]
-/// test `Palindrome::new` with invalid input
-fn palindrome_new_return_none() {
-    for v in [12, 2322, 23443, 1233211, 8932343] {
-        assert_eq!(Palindrome::new(v), None);
-    }
-}
-
 {% for test in cases %}
 #[test]
 #[ignore]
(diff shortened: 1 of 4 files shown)


exercises/practice/palindrome-products/.meta/test_template.tera
```diff
@@ -1,42 +1,25 @@
...
+      assert!(output.is_some());
+
+      {% if test.property == "smallest" %}
+        let (pal, _) = output.unwrap();
+      {%- else %}
+        let (_, pal) = output.unwrap();
+      {%- endif%}
+      assert_eq!(pal.value(), {{ test.expected.value }});
+      assert_eq!(pal.factors(), &HashSet::from([
+        {{- test.expected.factors | join(sep=", ") | replace(from="[", to="(") | replace(from="]", to=")") -}}
```
ellnix, line 21: I also had this instead of the one-liner:
```suggestion
        {%- for factor in test.expected.factors -%}
          ({{ factor | join(sep=", ") }}){% if not loop.last %}, {% endif %}
        {%- endfor -%}
```

I didn't really like either of them, but thought I'd drop this here just in case. There's probably some jinja-fu that can be done instead.

ellnix:
I was missing the HashSet dependency in the test, I added it but it doesn't seem to update since the PR is closed.

senekor:
Ugh, it won't even let me reopen it because of the force-push. I think you might have to create a new PR? I'll request to add you as maintainer so this doesn't happen again.

ellnix:
> Ugh, it won't even let me reopen it because of the force-push. I think you might have to create a new PR? I'll request to add you as maintainer so this doesn't happen again.

Well, I learned something new today then xD

New PR: https://github.com/exercism/rust/pull/2006

PR #2006, opened 2025-01-30
palindrome-products: require factors in solution

ellnix:
Apologies if I did something wrong, struggled a little bit with the rust-tooling.

Please let me know of any mistake, miss, or decisions you disagree with, no matter how small. **No hard feelings.** 

# The HashSet

Used a `HashSet<(u64, u64)>` for factors to avoid dealing with ordering of the factors and simultaneously communicate that the order doesn't matter. I thought leaving the inner tuple sorted made sense since that's how it appears in the instructions:

> [...] The largest palindrome product is 9. Its factors are (1, 9) and (3, 3).

A sorted `Vec` would probably be fine, and the change can be made easily enough if you think it would be more appropriate.

# The extra tests

There were two extra tests on the now-defunct `Palindrome::new` function. I felt those only made sense if this exercise still contained the newtype pattern. I can add them again if appropriate.

Corresponding forum post: https://forum.exercism.org/t/palindrome-products-description-mentions-factors/

### DIFF
--- a/exercises/practice/palindrome-products/.meta/test_template.tera
+++ b/exercises/practice/palindrome-products/.meta/test_template.tera
@@ -1,42 +1,26 @@
 use palindrome_products::*;
-
-{#
-    These first two custom test cases are for the object-oriented design of the exercise.
-    They don't fit the structure of the upstream tests, so they're implemented here.
-#}
-
-#[test]
-#[ignore]
-/// test `Palindrome::new` with valid input
-fn palindrome_new_return_some() {
-    for v in [1, 11, 121, 12321, 1234321, 123454321, 543212345] {
-        assert_eq!(Palindrome::new(v).expect("is a palindrome").into_inner(), v);
-    }
-}
-
-#[test]
-#[ignore]
-/// test `Palindrome::new` with invalid input
-fn palindrome_new_return_none() {
-    for v in [12, 2322, 23443, 1233211, 8932343] {
-        assert_eq!(Palindrome::new(v), None);
-    }
-}
+use std::collections::HashSet;
 
 {% for test in cases %}
 #[test]
 #[ignore]
 fn {{ test.description | make_ident }}() {
-    {%- if test.property == "smallest" %}
-        let output = palindrome_products({{ test.input.min }}, {{ test.input.max }}).map(|(min, _)| min.into_inner());
-    {%- else %}
-        let output = palindrome_products({{ test.input.min }}, {{ test.input.max }}).map(|(_, max)| max.into_inner());
-    {%- endif%}
+    let output = palindrome_products({{ test.input.min }}, {{ test.input.max }});
+
     {%- if test.expected.error is defined or not test.expected.value %}
-        let expected = None;
+        assert!(output.is_none());
     {%- else %}
-        let expected = Some({{ test.expected.value }});
+      assert!(output.is_some());
+
+      {% if test.property == "smallest" %}
+        let (pal, _) = output.unwrap();
+      {%- else %}
+        let (_, pal) = output.unwrap();
+      {%- endif%}
+      assert_eq!(pal.value(), {{ test.expected.value }});
+      assert_eq!(pal.factors(), &HashSet::from([
+        {{- test.expected.factors | join(sep=", ") | replace(from="[", to="(") | replace(from="]", to=")") -}}
+      ]));
     {%- endif%}
-    assert_eq!(output, expected);
 }
 {% endfor -%}
--- a/exercises/practice/palindrome-products/.meta/test_template.tera
+++ b/exercises/practice/palindrome-products/.meta/test_template.tera
@@ -18,7 +18,7 @@ fn {{ test.description | make_ident }}() {
         let (_, pal) = output.unwrap();
       {%- endif%}
       assert_eq!(pal.value(), {{ test.expected.value }});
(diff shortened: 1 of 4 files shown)


exercises/practice/palindrome-products/.meta/test_template.tera
```diff
@@ -1,42 +1,26 @@
...
+      assert!(output.is_some());
+
+      {% if test.property == "smallest" %}
+        let (pal, _) = output.unwrap();
+      {%- else %}
+        let (_, pal) = output.unwrap();
+      {%- endif%}
+      assert_eq!(pal.value(), {{ test.expected.value }});
+      assert_eq!(pal.factors(), &HashSet::from([
+        {{- test.expected.factors | join(sep=", ") | replace(from="[", to="(") | replace(from="]", to=")") -}}
```
ellnix, line 22: From the last PR:

I also had this instead of the one-liner:
```suggestion
        {%- for factor in test.expected.factors -%}
          ({{ factor | join(sep=", ") }}){% if not loop.last %}, {% endif %}
        {%- endfor -%}
```

I didn't really like either of them, but thought I'd drop this here just in case. There's probably some jinja-fu that can be done instead.

senekor:
> struggled a little bit with the rust-tooling

Yeah that's something I cooked up on my own. What were you struggling with? It could point us to potential improvements or at least missing documentation. Did you stumble on `docs/CONTRIBUTING.md#tooling`?

exercises/practice/palindrome-products/.meta/test_template.tera
senekor: I'm not too worried about the beauty of these templates. It seems like a write-only thing to me. If the exercise is changed substantially, there's a good chance the test template will have to be done basically from scratch.

exercises/practice/palindrome-products/src/lib.rs
```diff
@@ -1,19 +1,15 @@
-/// `Palindrome` is a newtype which only exists when the contained value is a palindrome number in base ten.
-///
-/// A struct with a single field which is used to constrain behavior like this is called a "newtype", and its use is
-/// often referred to as the "newtype pattern". This is a fairly common pattern in Rust.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub struct Palindrome(u64);
+use std::collections::HashSet;
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Palindrome;
```
senekor, line 4:
```suggestion
pub struct Palindrome {
    // TODO
}
```

While solving exercises, I sometimes wasn't sure what I was supposed to change and what to leave as it is. So I think it could remove some friction to mark this explicitly.

exercises/practice/palindrome-products/src/lib.rs
```diff
@@ -1,19 +1,15 @@
...
-    pub fn new(value: u64) -> Option<Palindrome> {
-        todo!("if the value {value} is a palindrome return Some, otherwise return None");
+    pub fn value(&self) -> u64 {
+        todo!("return the value of the palindrome")
     }
 
-    /// Get the value of this palindrome.
-    pub fn into_inner(self) -> u64 {
-        todo!("return inner value of a Palindrome");
+    pub fn factors(&self) -> &HashSet<(u64, u64)> {
```
senekor, line 11: I wonder if we should return on owned hashset here? Having the reference there forces the internal representation of `Palindrome` to store the hashset. I think that's a good approach, but maybe we shouldn't force it.

Returning an owned hashset might lead to additional heap allocations, but that doesn't matter in my view. If desired, we could do something like `fn into_factors(self) -> HashSet`, maybe that would yield the most flexibility.

exercises/practice/palindrome-products/src/lib.rs
ellnix: Fair point, added.

exercises/practice/palindrome-products/src/lib.rs
ellnix: I like the idea, done! :+1:

exercises/practice/palindrome-products/.meta/test_template.tera
ellnix: Fair enough, as someone who has tried to make tera templates pretty in the past, I can say you have the right attitude.

senekor (approved):

senekor:
I'll let you merge yourself so you can decide on the squashed commit message. The default setting is to use the PR description. This is useful, because you can add the `[no important files changed]` tag at the beginning in the PR description and then you won't forget to add it at merge-time. (Not saying the tag should be added on _this_ PR, it's not applicable.)

ellnix:
> I'll let you merge yourself so you can decide on the squashed commit message. The default setting is to use the PR description. This is useful, because you can add the `[no important files changed]` tag at the beginning in the PR description and then you won't forget to add it at merge-time. (Not saying the tag should be added on _this_ PR, it's not applicable.)

Thanks, but I don't have write access for now. I'm guessing I should wait to be added as a maintainer.

senekor:
Jeremy says he added you, so it should work now 👍

ellnix:
> Jeremy says he added you, so it should work now 👍

Github says no:
image

senekor:
@ellnix Jeremy says you should've gotten an invite that you need to accept?

PR #2008, opened 2025-02-01
Remove undefined offset variable from `tracks/rust/exercises/binary-search/dig_deeper`

jacobhq:
In the recursion section, `offset` is not defined. I updated it to match https://exercism.org/tracks/rust/exercises/binary-search/approaches/recursion.


senekor (approved):
Thanks!

PR #2013, opened 2025-02-09
Inaccurate description in comments

yhuang912:
Just corrected a minor inaccurate description that may cause confusion.


senekor:
Thanks, that looks like an oversight.

senekor (approved):

PR #2014, opened 2025-02-10
Luhn trait: update rust book url

0xwetzo:
Current rust book url is no longer valid. Updated with the active url of the trait page.


kotp:
I  can not re-open, but good link!  Thank you!

senekor (approved):
thanks!

PR #2019, opened 2025-02-14
accumulate: deprecate

ellnix:
Should add
- list-ops which replaces accumulate
- bottle-song which replaces beer-song
- diffie-hellman is deprecated without a replacement

accumulate and bottle-song seem pretty clear-cut to me, however an argument could be made for keeping diffie-hellman since the discussion around its deprecation explicitly mentions that track maintainers may choose to keep it:
- [diffie-hellman PR](https://github.com/exercism/problem-specifications/pull/2149)
  - [elixir](https://github.com/exercism/elixir/pull/1356) deprecated it while [the JS track](https://github.com/exercism/problem-specifications/pull/2149) seem to be planning to continue supporting it
- [beer-song PR](https://github.com/exercism/problem-specifications/pull/2097)
- [accumulate PR](https://github.com/exercism/problem-specifications/pull/1918)

I only noticed this because I was planning to make a PR adding a missing test in diffie-hellman and changing its kind of ridiculous difficulty of `1`, so I thought I'd check if we want to keep it before doing that work.

### DIFF
--- a/config.json
+++ b/config.json
@@ -333,7 +333,8 @@
         "topics": [
           "case",
           "vectors"
-        ]
+        ],
+        "status": "deprecated"
       },
       {
         "slug": "difference-of-squares",
@@ -495,7 +496,8 @@
         "difficulty": 1,
         "topics": [
           "math"
-        ]
+        ],
+        "status": "deprecated"
       },
       {
         "slug": "series",
@@ -1079,7 +1081,8 @@
           "generics",
           "higher_order_functions",
           "move_semantics"
-        ]
+        ],
+        "status": "deprecated"
       },
       {
         "slug": "affine-cipher",
--- a/config.json
+++ b/config.json
@@ -324,10 +324,7 @@
         "slug": "beer-song",
         "name": "Beer Song",
         "uuid": "bb42bc3a-139d-4cab-8b3a-2eac2e1b77b6",
-        "practices": [
-          "loops",
-          "strings"
-        ],
+        "practices": [],
         "prerequisites": [],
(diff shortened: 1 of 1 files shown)


senekor (changes requested):
lol, I just checked my solution of diffie-hellman and it contains a comment "pretty much copy pasted from wikipedia" 😄 so yeah, I'm fine with removing it.

The bottles song can be replaced, that's also fine.

Accumulate and list-ops I'm not 100% sure about. I think I looked at it at some point in the past. The existing accumulate exercise has only one function IIRC and the challenge is to get the function signature right. The list-ops exercise has many different functions to implement. It's not a trivially equivalent replacement. So if we want to do that, we should first have a rough plan of how we're going to design the list-ops exercise to make sure we won't lose anything by getting rid of accumulate.

Maybe a good idea to split this into three PRs.

ellnix:
I'll keep this PR for accumulate and create new ones for the other two. I think accumulate would be the longest running discussion so it'll have the full context.

> Accumulate and list-ops I'm not 100% sure about

Maybe we can defer making a decision after trying to implement list-ops? They could co-exist if they present different challenges.

I'll try to work on list-ops as soon as possible.

senekor:
> Maybe we can defer making a decision after trying to implement list-ops? They could co-exist if they present different challenges.

Yeah, totally. My intuition says that it will make sense to keep both. So feel free to design list-ops independently, don't feel the need to "supersede" accumulate. If the design ends up very similar anyway, then we might decide to deprecate accumulate.

Issue #1735, opened 2023-09-13
Document / reconsider `topics` field in track config


senekor:
The field has no purpose from Exercism's viewpoint. Consider deleting it or documenting its rust-track-specific purpose. Probably in the track config data structure of the rust-tooling crate, where it could be an optional field with appropriate documentation.

---

Why is the `topics` key added? It is only really there for inspiration for the `practices`/`prerequisites` fields, but it serves no purpose from Exercism's viewpoint.

_Originally posted by @ErikSchierboom in https://github.com/exercism/rust/pull/1722#discussion_r1324276837_

senekor:
Found the documentation, it's in [this issue](https://github.com/exercism/rust/issues/1407).

It seems like the purpose this field serves right now is that for any future work on the syllabus, these topics might help guide which exercise should be marked to practice which topic. One of the topics is `"goofy_bowling_logic"` so that list is definitely a draft that may support, not a well-crafted list of concepts ready to use.

PR #2021, opened 2025-02-14
beer-song: deprecate

ellnix:


senekor:
Let's do the deprecation of the old exercise together with the implementation of the new one.

exercises/practice/bottle-song/.meta/test_template.tera
```diff
@@ -9,7 +9,12 @@ use bottle_song::*;
 fn {{ test.description | make_ident }}() {
     assert_eq!(
         recite({{ test.input.startBottles }}, {{ test.input.takeDown }}),
```
ellnix, line 11: I saw that beer-song called `trim()` on this value, but that just feels like cheating...

It's up to the student to use `trim()` if they need it, in my opinion.

config.json
```diff
@@ -324,16 +324,23 @@
...
+        "status": "deprecated"
+      },
+      {
+        "slug": "bottle-song",
+        "name": "Bottle Song",
+        "uuid": "31e8185b-39d8-48c6-88a0-f302e2864f16",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 1,
+        "topics": []
```
ellnix, line 343: I would be lying if I said I knew what this `topics` key does.

ellnix:
Thought `concat!` would make the tests easier to read, if you disagree I can revert the commit easily.

config.json
BNAndras: Topics were used during Exercism v2 instead of practices but no longer signify anything.

senekor: Yeah the topics key is present on all the exercises, we have a separate [issue](https://github.com/exercism/rust/issues/1735) about cleaning that up at some point.

exercises/practice/bottle-song/.meta/test_template.tera
```diff
@@ -0,0 +1,21 @@
...
+{% for test in subgroup.cases %}
+
+#[test]
+#[ignore]
+fn {{ test.description | make_ident }}() {
+    assert_eq!(
+        recite({{ test.input.startBottles }}, {{ test.input.takeDown }}),
+        concat!(
+            {% for line in test.expected %}
+            "{{line}}{% if not loop.last %}\n{% endif %}",
```
senekor, line 14: nit: keep the trailing newline

ellnix: I went through other repos after receiving this feedback, I had only looked at Elixir which does not have the newline. Ruby seems to have the newline, and python, javascript, and go seem to return some sort of array of strings.

Should we just go for a Vec of Strings?

I'll keep the trailing new line in the meantime, even though I think that the challenge makes more sense without it. But it's no big deal.

senekor: > Should we just go for a Vec of Strings?

Nah, that seems weird to me.

In my mind, text that consists of multiple lines should always have a trailing newline. But it's not important. Neither alternatives result in a more or less interesting / challenging exercise.

Maybe that's what the [trim](https://github.com/exercism/rust/pull/2021#discussion_r1962418286) call was for. Allow students to implement their preference, with or without trailing newline.

ellnix: I can add trim if you want, or else I guess we can merge this as is?

senekor: yeah whatever you want 🙂

ellnix: Added `trim()`, feel free to merge if you have no reservations.

PR #2022, opened 2025-02-15
Update test for anagram.rs

aure31:
the "αβγ" should be in the right output of the test because it's the same a the precedent with a different order


senekor:
The word `αβγ` is the same as `ΑΒΓ` but in lowercase, that's why it's not included.

PR #2023, opened 2025-02-15
list-ops: add exercise

ellnix:
@senekor I'm starting to agree that accumulate has a place for teaching students how to come up with the closure trait. I don't think it would be a good idea to replace the trait with `???` in this exercise because it would be impossible to work on the functions one at a time.

This is just a first draft of course (at least the battle with Tera is won), but any suggestions are welcome.

P. S.: The fold functions are floats because the tests would fail with integer division, there is another set of tests that work with integer division but for some reason they were not getting passed to Tera. I'm not sure if it's the tooling removing duplicate descriptions or where they went, I just made it work with floats.


senekor:
```
"function": "(acc, el) -> el + acc"
```

Ughhh that must've been the reason why I gave up on this exercise when I had a look at it. I can't imagine what went through people's heads when they decided it was a great idea to force downstream to implement an expression parser in a templating language 🙄

Some tests have a `reimplements` key that points to another test case. The pointed-to test case is considered deprecated, so the exercise generator doesn't pass it to tera. So, intended behavior.

ellnix:
> Ughhh that must've been the reason why I gave up on this exercise when I had a look at it. I can't imagine what went through people's heads when they decided it was a great idea to force downstream to implement an expression parser in a templating language 🙄

Believe it or not that was far from the biggest issue. I got it figured out in seconds. The floats tho... could not find a way to make them not-ugly.

> Some tests have a reimplements key that points to another test case. The pointed-to test case is considered deprecated, so the exercise generator doesn't pass it to tera. So, intended behavior.

Yeah, it looks like they have `included = false` in `tests.toml` too. Didn't realize.

config.json
```diff
@@ -1551,6 +1551,19 @@
...
+        "name": "List Ops",
+        "uuid": "4b411941-d272-4b95-8c40-5816fecf27a2",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 4,
+        "topics": [
+          "generics",
+          "higher_order_functions",
+          "move_semantics"
+        ]
```
senekor, line 1566: The location of this entry in the list determines the order in which the exercises are shown to students. At least, it should be ordered by difficulty. Ideally, one would even think about where in the list of exercises with equal difficulty this makes sense.

Don't worry too much about it now, we might end up with a different difficulty as we iterate on the exercise design.

Of course, if we gain a syllabus, there will be better ways to determine the order of exercises.

exercises/practice/list-ops/.meta/config.json
```diff
@@ -0,0 +1,16 @@
+{
+  "authors": ["ellnix"],
+  "files": {
+    "solution": [
+      "src/lib.rs",
+      "Cargo.toml"
```
senekor, line 6: I saw a version where this wasn't filled. Did the tooling not do that for you? Would be nice if it did.

exercises/practice/list-ops/src/lib.rs
```diff
@@ -0,0 +1,33 @@
+use std::fmt::Debug;
+
+pub fn append(a: Vec<i32>, b: Vec<i32>) -> Vec<i32> {
+    todo!("add all items in {b:?} to the end of the {a:?}");
+}
+
+pub fn concat<T: Debug>(list: Vec<Vec<T>>) -> Vec<T> {
```
senekor, line 7: Since we need to use the prefix underscore approach for closures, I'd prefer to use that everywhere and get rid of the `Debug` bounds for consistency.

senekor (changes requested):
I have to say I find the design based on vectors boring and unidiomatic. Students probably get plenty of exercise with basic vector operations from all the other exercises. What do you think about cranking the difficulty up a notch and basing it on iterators? `foldr` and `reverse` could make use of [DoubleEndedIterator](https://doc.rust-lang.org/stable/std/iter/trait.DoubleEndedIterator.html)

At the end of the day, this is reimplementing std lib functions. But I think that can be quite valuable for understanding how the std lib works.

senekor:
Btw since you're a maintainer now, you can push branches directly to this repo and don't need to use a fork anymore.

config.json
senekor: huh, they aren't even sorted right now. I thought I had fixed that at some point.

senekor: Ah, here's the [forum discussion](https://forum.exercism.org/t/rust-track-exercise-order/10855/5) I vaguely had in mind. So it's not a hard requirement to sort by difficulty. The reasoning that a little variation is better for pacing still makes sense to me. So we should just put it somewhere it's not obviously out-of-place.

ellnix: I didn't know about the sorting, I thought the server sorted by difficulty then alphabetically. I think after accumulate would make sense? Unless you think it's best to space out two exercises that are so similar.

exercises/practice/list-ops/.meta/config.json
ellnix: I can't see a line in the `generate` crate that would do that. It wasn't a big deal to copy-paste it but if you think it's worth adding to the update function I can give it a shot.

exercises/practice/list-ops/src/lib.rs
ellnix: Fair enough.

ellnix:
> I have to say I find the design based on vectors boring and unidiomatic. Students probably get plenty of exercise with basic vector operations from all the other exercises. What do you think about cranking the difficulty up a notch and basing it on iterators? foldr and reverse could make use of [DoubleEndedIterator](https://doc.rust-lang.org/stable/std/iter/trait.DoubleEndedIterator.html)

I think Iterators as arguments could work well, but I think returning anything other than a vector is sort of pointless. After all I could just solve it with a vec and call `into_iter()` at the end.

Unless you want to change this exercise to be about finding the right iterator methods to transform from an iterator to another, but that would miss out on teaching the student the underlying logic in something like map. The Elixir track for instance instructs students not to use any `map`/`filter`/`fold` functions from their stdlib.

About the arguments, would you like `impl IteratorTrait` directly or `IntoIterator`?

> Btw since you're a maintainer now, you can push branches directly to this repo and don't need to use a fork anymore.

Would a branch on this repo be more convenient for you? I'm using my own fork just out of inertia.

senekor:
> I think Iterators as arguments could work well, but I think returning anything other than a vector is sort of pointless. After all I could just solve it with a vec and call `into_iter()` at the end.

Well yes, there is still the _possibility_ of students implementing a suboptimal solution. But going with vectors in the signature essentially _forces_ the unidiomatic implementation. With iterators, at least students _can_ do the idiomatic and performant thing. There is also the opportunity of mentors pointing out this potential improvement.

> Unless you want to change this exercise to be about finding the right iterator methods to transform from an iterator to another, but that would miss out on teaching the student the underlying logic in something like map. The Elixir track for instance instructs students not to use any `map`/`filter`/`fold` functions from their stdlib.

I agree, we should not do that. The instructions from problem-specifications already tell students not to use any of these methods, we should stick to it. I'm thinking it would be nice if students realized how all these awesome iterator methods can be implemented under the hood (even if maybe the std lib implementation is more obtuse for performance reasons.)

Here's what I have in mind:

```rs
pub fn map<T, I, F>(iter: I, function: F) -> impl Iterator<Item = T>
where
    I: Iterator<Item = T>,
    F: Fn(T) -> T,
{
    struct Map<I, F> {
        iter: I,
        f: F,
    }

    impl<T, I, F> Iterator for Map<I, F>
    where
        I: Iterator<Item = T>,
        F: Fn(T) -> T,
    {
        type Item = T;

        fn next(&mut self) -> Option<Self::Item> {
            let next = self.iter.next()?;
            Some((self.f)(next))
        }
    }

    Map::<I, F> { iter, f: function }
}
```

> About the arguments, would you like `impl IteratorTrait` directly or `IntoIterator`?

I think `Iterator` should be fine. `IntoIterator` is nice for library users, because they don't have to call `.into_iter()` everywhere. But "we" (the ones writting the tests) are basically the only users of this "library", so it doesn't matter. `Iterator` is simpler and let's students focus on the learning goal of this exercise.

> Would a branch on this repo be more convenient for you? I'm using my own fork just out of inertia.

Only slightly. I could checkout your branches directly without `gh pr checkout <PR number>`. No need to change your setup.

senekor:
Let's also make all the functions generic, otherwise it's a little confusing why some of them are for `i32`, others for `f64` and still others generic.

senekor:
Oh and the discussion in #2025 reminds me, it would be nice to group the tests in modules so it reflects the grouping in canonical-data.

senekor:
sry wrong button

config.json
ellnix: I think the difficulty should be like 6-8 depending on what we decide for the starting `lib.rs`, so I no longer think it should be right after accumulate :laughing:

ellnix:
I'll have to refactor `example.rs` and the template as much as I can, but thought we should have a discussion about how much boilerplate the student is provided with.

I'm thinking we should either add the structs inside the functions, or add the structs outside the functions and return the structs instead of `impl Iterator`s.

exercises/practice/list-ops/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+    // () doesn't implement Iterator
+    std::iter::empty()
+}
+
+/// Returns an iterator of all items in iter for which `predicate(item)` is true
+pub fn filter<I, T, F>(_iter: I, _predicate: F) -> impl Iterator<Item = T>
+where
+    I: Iterator<Item = T>,
+    F: Fn(T) -> bool,
+    T: Clone,
```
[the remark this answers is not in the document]
senekor: Looks like a type inference problem. If the closure is inline, it works:
```rs
let output = filter(list, |x| x % 2 == 1);
```

Then the `Clone` bound is not necessary anymore and F can be `Fn(&T) -> bool`

exercises/practice/list-ops/src/lib.rs
```diff
@@ -0,0 +1,70 @@
+/// Yields each item of a and then each item of b
+pub fn append<T, I>(_a: I, _b: I) -> impl Iterator<Item = T>
+where
+    I: Iterator<Item = T>,
+{
+    // this empty iterator silences a compiler complaint that
+    // () doesn't implement Iterator
+    std::iter::empty()
+}
+
```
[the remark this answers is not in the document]
senekor: I wouldn't add anything. It's not our job to force users to implement what we think is the best solution. If we do that, we also destroy any potential "Aha!" moments when a user figures it out themselves.

Simple solutions based on vectors are still possible and I think that's good. Some people may want to learn Rust without any interest in performance optimization. There's no reason to force them to.

But I think it would be appropriate to add a hint that this exercise _can_ be solved without any allocations. That should be enough to make many users curious and motivated to figure it out on their own.

Other students might be able to improve their solution in a conversation with a mentor, which is also a good experience to enable.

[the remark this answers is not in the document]
senekor: [instructions append](https://exercism.org/docs/building/tracks/practice-exercises#h-file-docs-instructions-append-md) would be an appropriate place to add this hint I think.

exercises/practice/list-ops/src/lib.rs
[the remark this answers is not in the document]
ellnix: Inlined the function, so this should be fixed.

exercises/practice/list-ops/src/lib.rs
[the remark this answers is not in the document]
ellnix: I wrote an instructions.append.md file, which probably needs some changes.

ellnix:
@senekor I don't think I have anything left to do, I'll wait for your feedback to implement any changes you suggest.

exercises/practice/list-ops/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+    I: Iterator<Item = T>,
+    F: Fn(U, T) -> U,
+{
+    todo!("starting with initial, fold (reduce) each iter item into the accumulator from the left")
+}
+
+pub fn foldr<I, F, T, U>(mut _iter: I, _initial: U, _function: F) -> U
+where
+    I: DoubleEndedIterator<Item = T>,
+    F: Fn(U, T) -> U,
```
senekor, line 129:
```suggestion
    F: Fn(U, T) -> U,
```
I think with foldr, the argument order is usually reversed compared with foldl, right?
e.g. https://hoogle.haskell.org/?hoogle=foldr

ellnix: I have never used Haskell, but it's not reversed in problem-specifications. It just loops from right to left:

```json
        {
          "uuid": "8066003b-f2ff-437e-9103-66e6df474844",
          "reimplements": "be396a53-c074-4db3-8dd6-f7ed003cce7c",
          "description": "direction dependent function applied to non-empty list",
          "comments": [
            "Reimplemented to remove ambiguity about the parameters of the    ",
            "function input. Expects / to preserve fractions. Integer division",
            "will not work here, since it would compute 4 / 24 = 1 / 6. Use   ",
            "the original test values (be396a53-c074-4db3-8dd6-f7ed003cce7c)  ",
            "if integer division is expected / required."
          ],
          "property": "foldr",
          "input": {
            "list": [1, 2, 3, 4],
            "initial": 24,
            "function": "(acc, el) -> el / acc"
          },
          "expected": 9
        }
```

ellnix: Gets the right answer without reversing the arg order:
```
>> n = 24.0
=> 24.0
>> n = 4 / n
=> 0.16666666666666666
>> n = 3 / n
=> 18.0
>> n = 2 / n
=> 0.1111111111111111
>> n = 1 / n
=> 9.0
```

senekor: I just checked how other languages do it. Most have the same order, like problem-specifications. Only Haskell reversed it, presumably to stay true to their standard library 🙂 Rust doesn't have foldr, so not much risk of causing confusion. Let's leave it as is.

exercises/practice/list-ops/src/lib.rs
```diff
@@ -0,0 +1,69 @@
+/// Yields each item of a and then each item of b
+pub fn append<T, I>(_a: I, _b: I) -> impl Iterator<Item = T>
+where
+    I: Iterator<Item = T>,
+{
+    // this empty iterator silences a compiler complaint that
+    // () doesn't implement Iterator
+    std::iter::empty()
```
senekor, line 8:
```suggestion
    std::iter::from_fn(|| todo!())
```

Let's use something like this as a placeholder iterator, so the "empty" tests don't pass at first.

config.json
senekor: Current location seems fine to me. You can bump it to a 7 if you like, but 4 and 7 are both displayed as medium. Probably doesn't matter...

senekor (approved):
Alright, looks good to me. I would just add the panicking iterator as placeholder and then this is ready to ship. Thanks for being patient with me! 😄

ellnix:
@senekor It's no problem! Are there any rules I should follow about what to add in the squashed commit description when adding a new exercise?

senekor:
> Are there any rules I should follow about what to add in the squashed commit description when adding a new exercise?

Not really. I'd say normal commit message rules apply. Of course, keep the PR link in there, that way discussion that happened here can be tracked down. Other than that, just ask yourself what you'd want the message to be if you were digging in the history sometime in the future. I wouldn't duplicate any iformation that can be found here, overly verbose history can be harder to navigate. IMO you can even leave the body empty.

PR #2024, opened 2025-02-15
sync problem-specifications

senekor:


exercises/practice/grep/.meta/test_template.tera
```diff
@@ -132,7 +132,7 @@ impl<'a> Files<'a> {
     }
 }
 
-impl<'a> Drop for Files<'a> {
+impl Drop for Files<'_> {
```
senekor, line 135: This is the only manual change, seems there's a new / improved clippy warning against lifetimes that can be elided.

ellnix: I have yet to do the grep exercise, but if it's a clippy warning then this is a good change.

ellnix (approved):
Seems like a bunch of automatic changes, I can't see anything wrong with them.

LGTM.

senekor:
Yup, all generated from the existing templates.

PR #2025, opened 2025-02-16
Fix stub generation for new exercises

ellnix:
Previously, exercises where the canonical_data was grouped were causing Tera to crash, which meant that GeneratedExercises could not be generated. This prevented basically every file from being generated.

Reproduce the error:
`just add-exercise -s list-ops`

With canonical_data cases nested two levels deep :face_holding_back_tears: :
`just add-exercise -s bottle-song`

### DIFF
--- a/rust-tooling/generate/src/custom_filters.rs
+++ b/rust-tooling/generate/src/custom_filters.rs
@@ -8,6 +8,7 @@ pub static CUSTOM_FILTERS: &[(&str, Filter)] = &[
     ("to_hex", to_hex),
     ("make_ident", make_ident),
     ("fmt_num", fmt_num),
+    ("recursive_flat_map", recursive_flat_map),
 ];
 
 pub fn to_hex(value: &Value, _args: &HashMap<String, Value>) -> Result<Value> {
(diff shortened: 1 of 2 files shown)


senekor:
Not sure I like this. I think my tooling was working like this at some point. The problem is that in some cases, you _need_ the nesting to generate the tests. I think there was at least one exercise where the tests had identical descriptions across different groups. Even if it's not necessary, it is often nice to make use of the nesting to group tests in modules. So I don't like removing that information by default.

What's the problem with the generator crashing? Maybe the dev experience could be improved another way?

senekor:
What if, instead of crashing, we fallback to some default like

```rs
#[test]
fn invalid_template() {
    panic!("The exercise generator failed to produce valid tests from the template.");
}
```

and emit the error from the `generate_tests` function as a warning in the terminal? Would that be a good development experience for you?

ellnix:
> and emit the error from the generate_tests function as a warning in the terminal? Would that be a good development experience for you?

I feel like if canonical_data changes majorly in an exercise it would fail with a warning and we might miss it.

For me it's not really a problem (for list-ops I ended up writing the files myself which wasn't a big deal), the add-exercise script is just leaving the exercise stub in an unfinished state. I think it would be a problem to anyone who isn't familiar with the repo and the tooling.

The reason I went for this solution was that I think there is value in keeping the add workflow as close to the update workflow as possible. 
- If we write a template file that cannot generate tests and a test file they would sort of be out of sync
- If we don't write the template at all we are not encouraging the templates to be used

I'm not at all attached to this solution, if you feel something is better we can do that instead.

senekor:
My idea was that the test file (<slug>/tests/<slug>.rs) would _always_ be overwritten. If the template fails, the content would be the placeholder. So even if you miss the warning in the terminal, you'd get the obviously incorrect test file in the diff. This would even be caught by CI, because the placeholder test immediately fails when run.

So we still encourage the use of the template by generating it and the only way to make CI pass is write a working template OR delete the template file intentionally.

Making the "add" and "update" workflows as similar as possible is a good idea.

senekor:
With #2027 merge, would you like to keep the custom tera filter `recursive_flat_map`? In case you want to use it ad-hoc for another new exercise.

ellnix:
@senekor I don't think it's in any way superior to just having multiple `for` loops. If anything, multiple for loops are better since they allow using fields from the group(s).

I think it's best to err on the side of keeping less code to maintain, if an exercise needs a function like this I could come back to this and copy it.

I can definitely think of filters that would be more useful that this but did not add them to keep templates easy to read without having to read the source code of all of the filters. 

I'll close the PR so it doesn't clutter the list, but feel free to keep the conversation going if you disagree.

PR #2026, opened 2025-02-17
simplify generated gitignore files

senekor:
Cargo has simplified its output a while ago, this is making us more consistent with that. Cargo doesn't add Cargo.lock anymore at all, neither for libraries nor for binaries. It still makes sense for us though, since our users shouldn't submit Cargo.lock to the test runner.

---

I noticed this on the PR for list-ops, but thought it's better to update everything at once in a separate PR.


rust-tooling/models/src/problem_spec.rs
```diff
@@ -13,6 +13,7 @@ pub struct CanonicalData {
 #[derive(Clone, Debug, Serialize, Deserialize)]
 #[serde(deny_unknown_fields)]
 #[serde(untagged)]
+#[allow(clippy::large_enum_variant)]
```
senekor, line 16: unrelated piggyback :)

ellnix (approved):
Looks good to me!


For anyone looking at this in the future, relevant PRs on cargo:
https://github.com/rust-lang/cargo/pull/12382
I hadn't really considered the `git bisect` argument but it's a good one. I don't think it applies to us though.

https://github.com/rust-lang/cargo/pull/7647

The crates.io guide page in the diff also seems to have moved:
https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html

rust-tooling/models/src/problem_spec.rs
ellnix: clippy is too smart for its own good

PR #2027, opened 2025-02-17
generate: don't crash on invalid templates

senekor:
This ensures that all other files are still written to disk as expected. There is no risk of an invalid template being merged, because the generated fallback tests will always fail in CI.


senekor:
@ellnix this is my attempt to solve the DX problem discussed in #2025

PR #2028, opened 2025-02-17
generate: fill list of files in exercise config

senekor:


ellnix (approved):
The generation works well, and is simple to understand. One downside is that we cannot accurately fill in the "authors" field, so this file still needs to be manually edited every time.

I think in a separate PR we could add a prompt to add-exercise for "author" just like we have for exercise name and difficulty? Could also pre-fill it with `git config user.name`.

senekor:
> I think in a separate PR we could add a prompt to add-exercise for "author" just like we have for exercise name and difficulty? Could also pre-fill it with `git config user.name`.

great idea!

Issue #1839, opened 2023-12-19
Move more CI tests to rust tooling


senekor:
Some CI tests exist as bash scripts, others as rust tests in `rust-tooling/ci-tests`. Let's move more if not all of the bash stuff to rust. There is especially some overlap with clippy. Both bash scripts and rust tests run clippy over _some_ parts of an exercise. That should be made more consistent.

PR #2030, opened 2025-02-20
replace ensure_stubs_compile with faster rust test

senekor:
part of #1839


senekor:
The Rust test `stubs_are_warning_free` basically does everything `ensure_stubs_compile.sh` does, just in parallel (waaay faster). The only differences I found when comparing were:
- The bash script also checked tests, so I added the `--tests` flag to the Rust test
- On PRs, the bash script only runs on the exercises that were changed. Since all exercises are checked in parallel now, I think we can drop that complexity. It's not needed anymore for speed.

ellnix (approved):
Read the files multiple times, I don't think anything fell through the cracks. LGTM!

The parallel test in indeed way faster.

PR #2031, opened 2025-02-20
run format_exercises.sh in parallel

senekor:


ellnix (approved):
Can't see a difference apart from the lack of printing (and obviously the speed), both in the code and in practice.

LGTM!

PR #2033, opened 2025-02-21
update exercise's package manifest

senekor:
scripted changes:
- update to edition 2024
- bring name, version, edition in customary order
- make sure all names use underscores instead of dashes
- add comment to dependencies section pointing to available libraries
- apply formatting of 2024 edition

manual interventions:
- fix crate names of all-your-base and two-fer
- generator: update manifest template
- fix exercises for the 2024 edition


senekor:
It should be noted that the 2024 edition is really new. I updated the test runner and analyzer specifically to support it. I expect that many students who are currently solving Rust exercises will not yet have Rust 1.85 installed (minimum version where edition 2024 is stable). The next exercise they download will not compile with a rather rude error message that their toolchain is outdated.

There is an argument to be made that we should wait with this until most people are probably running a recent enough toolchain.

I personally think it's good to push people to use the newest toolchain. It's just one `rustup update` away. The new edition has lots of changes that will improve the experience of Rust newbies, especially related to the ownership system. (temporary scopes of `if let` and tail expressions) If we _don't_ update the edition, we are by default opting our students into the older, worse semantics.

@ellnix what do you think? Should we hold off on this for a little longer?

For reference, the [blog post announcing Rust 2024](https://blog.rust-lang.org/2025/02/20/Rust-1.85.0.html).

ellnix (approved):
I think since we are changing crate names to be snake_cased, we could do the same to the test file names, if you feel it's right.

I can see how a breaking change like this would be problematic, but I don't see how it will get any better if we wait. Solutions are going to break and students will have to update their tooling.

I think the only difference between doing it now and doing it at an unspecified later date is that students don't get to benefit from the new changes.

The comment on the `Cargo.toml` files should pretty much cut in half the volume of forum posts in Rust :laughing: 

It might be a better idea to link to somewhere more stable than the `local-registry/Cargo.toml`, something like a documentation heading or maybe a forum post where we can also explain why the packages are limited. That's a minor gripe though I don't imagine that link will break anytime soon.

senekor:
> I think since we are changing crate names to be snake_cased, we could do the same to the test file names, if you feel it's right.

Yeah, that's a good idea. Snake case is more idiomatic for Rust. I'll do it in a separate PR.

> It might be a better idea to link to somewhere more stable than the `local-registry/Cargo.toml`

Yeah, the link hasn't changed since I got involved and I don't see a reason why it would. I'm also not aware of documentation that would be better to link to.

PR #2034, opened 2025-02-22
update rust-tooling to 2024 edition

senekor:


ellnix (approved):
Looks good, `cargo fix --edition` didn't make any code changes (from main), I did a few smoke tests and everything seemed good. I don't think there are any logic regressions.

PR #2035, opened 2025-02-22
generate: format tests with edition 2024

senekor:


senekor:
There's one test that makes sure test templates are in sync, that one is actually not run in CI (because it's slow, we might reconsider this) That test fails currently because the test generator tries to revert the formatting of the 2024 edition.

ellnix (approved):
LGTM. Confirmed that it was bugged before and it's fixed after.

> because it's slow, we might reconsider this

It doesn't seem that slow, it finished under 5s on my machine. Would probably be much faster in CI.

senekor:
> > because it's slow, we might reconsider this
> 
> It doesn't seem that slow, it finished under 5s on my machine. Would probably be much faster in CI.

I think you're right. With a clean build, running tests in `rust-tooling/` instead of `rust-tooling/ci-tests/` increases total time locally from 15 to 25 seconds, so it doesn't seem excessively bad.

PR #2037, opened 2025-02-22
activate test tera_templates_are_in_sync in ci

senekor:
It wasn't active before, because it was thought to be too slow. Turns out it's not that bad.


senekor:
as discussed [here](https://github.com/exercism/rust/pull/2035#pullrequestreview-2634529712)

senekor:
hmm, works on my machine 🤔

ellnix:
Works on my machine as well.

ellnix (approved):
@senekor I should have known from the start that it was just problem-specifications missing, but wasn't familiar with github actions and especially this repo's actions. Live and learn I guess.

Unfortunately the first exercise with a template also had additional tests, which had me thinking that only some of the canonical-data tests were being rendered. Of course, it ended up just being additional-tests.json.

senekor:
Ah, makes sense.

For the record, the problem-specifications stuff is one of the things I did I'm not totally happy with. It's not entirely my fault, I think there should be a way to point configlet at a specific version of prolem-specifications (e.g. a submodule), but there isn't. It only wants to work with its own cache. So I bent over backwards to get some kind of reproducibility.

I'm open to ideas about how to do that better.

I _personally_ also don't need the `.prob-spec` symlinks anymore that are strewn about all the exercise directories. I used to work with a file tree where these were useful, but now I mostly use a fuzzy file picker. If you don't need them either, we could remove them.

For next time, I'd prefer if you push your own work to a separate branch instead of mine 🙂

PR #2038, opened 2025-02-22
run shellcheck in local test script

senekor:
`bin/` is the only directory where we have shell scripts. This will fail if shellcheck is not installed, but even the Ubuntu repos have it, so nobody should have trouble installing it.


ellnix (approved):
Shell scripting is not my area, but Fedora has it as well, I don't see a problem with this.

PR #2041, opened 2025-02-25
Fix invalid file path

yjhtry:
image
image


senekor:
Thanks, that's a regression from 88c13fb21eacd6d4101b98f900e525fe1be3d1ca.

Did you test that this is enough? I also see a file `macros/tests/invalid/Cargo.toml` with the old file names, is that not causing trouble?

senekor:
Yeah the tests are just "passing" because all of the invalid examples are failing to compile unconditionally.

senekor:
I pushed a version where the invalid tests actually do their job [here](https://github.com/exercism/rust/pull/2042)

PR #2043, opened 2025-02-25
doubly-linked-list: improve leak check

senekor:
Previously the check may fail due to an underflow in the subtraction, rather than the assert.

Forum discussion:
https://forum.exercism.org/t/fix-operand-order-in-leaked-bytes-check-in-leak-test-doubly-linked-list-rust-exercise/15884

### DIFF
--- a/exercises/practice/doubly-linked-list/tests/step_4_leak_test_2.rs
+++ b/exercises/practice/doubly-linked-list/tests/step_4_leak_test_2.rs
@@ -19,8 +19,7 @@ fn drop_no_leaks() {
     drop(list);
 
     let allocated_after = ALLOCATED.load(SeqCst);
-    let leaked_bytes = allocated_before - allocated_after;
-    assert!(leaked_bytes == 0);
+    assert_eq!(allocated_before, allocated_after);
 }
 
 // Defines a wrapper around the global allocator that counts allocations


exercises/practice/doubly-linked-list/tests/step_4_leak_test_2.rs
```diff
@@ -19,8 +19,7 @@ fn drop_no_leaks() {
     drop(list);
 
     let allocated_after = ALLOCATED.load(SeqCst);
-    let leaked_bytes = allocated_before - allocated_after;
-    assert!(leaked_bytes == 0);
```
ellnix, line 23: Was there any particular reason for choosing subtraction over equality in the past? It seems like a really strange decision.

ellnix: Looking at it more closely, it just looks like a logic error. Even if there was a reason for the subtraction, the order should have been reversed. There's no conceivable reason for `allocated_before` to ever be larger than `allocated_after`.

ellnix (approved):
LGTM!

exercises/practice/doubly-linked-list/tests/step_4_leak_test_2.rs
senekor: Original PR is [here](https://github.com/exercism/rust/pull/653), I can't find any discussion on that.

PR #2044, opened 2025-02-26
doubly-linked-list: check compilation in ci

senekor:
The reason given for why this exercise was allowed to not compile
doesn't seem to apply anymore.


senekor:
just started this exercises and noticed the clippy warnings

ellnix (approved):
LGTM, I don't think anything was missed. In any case it's safe since this only affects CI and shouldn't be seen by any students.

PR #2045, opened 2025-02-28
Sync all docs and metadata

ellnix:
This was mostly a handwritten script that used configlet directly without going through rust-tooling. Afterward I reviewed the git diff.

I'll try to update the tests as well when I have time. With that completed everything will be in sync. I did check just in case there was a test change in addition to the docs change with any of these exercises, but that wasn't the case.


senekor:
I think `just configlet sync --update --yes --docs --metadata` would've done the same thing 🙂

senekor (approved):

ellnix:
That was essentially what I did, except with a for loop because I wanted to be able to easily review the changes just in case.

PR #2046, opened 2025-03-01
matching-brackets: Sync tests and add test template

ellnix:
Some shenanigans was going on where tests were missing in `tests.toml` but were actually implemented in the test file. With this everything is consistent.

Specifically
fn early_incomplete_brackets() {}
fn early_mismatched_brackets() {}


ellnix:
I don't think any solutions that were working will now break, but as there is technically a new test I think the tests should run again.

senekor (approved):

PR #2047, opened 2025-03-01
difference_of_squares: Sync tests & add template

ellnix:
I considered adding the [tera_text_filters](https://docs.rs/tera-text-filters/latest/tera_text_filters/) crate but decided on if statements to avoid adding more dependencies.

With a snake_case filter, the template could be written without the `function` variable.

P. S.: I didn't forget to run `update-exercise` there is just no diff on the test file.

[no important files changed]

### DIFF
--- a/exercises/practice/difference-of-squares/.meta/test_template.tera
+++ b/exercises/practice/difference-of-squares/.meta/test_template.tera
@@ -1,0 +1,26 @@
+use difference_of_squares as squares;
+
+{% for group in cases %}
+{% for test in group.cases %}
+
+{% if test.property == "squareOfSum" %}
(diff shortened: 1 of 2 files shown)


senekor (approved):

senekor:
I don't think getting rid of the if statement is worth adding a dependency. You could implement a simple version of such a filter yourself if you wanted to.

ellnix:
> I don't think getting rid of the if statement is worth adding a dependency. You could implement a simple version of such a filter yourself if you wanted to.

Agreed, one if statement is no big deal, if it comes up in another template I'll probably make a short naive `snake_case`. I'm glad I got your opinion on it first :smile:

PR #2050, opened 2025-03-03
parallel-letter-frequency: Sync tests & add missing

ellnix:
The tests are unique and not really in sync with problem-specifications, however they test the same behavior so it should be fine? I thought I'd make a PR before making a template since it might make sense to keep the bespoke tests that we have now. **Would like any input on this.**

The tests I added were missing, but they are just convenience tests for students to be able to iterate their solution, so I think this is appropriate:

[no important files changed]

### DIFF
--- a/exercises/practice/parallel-letter-frequency/.meta/tests.toml
+++ b/exercises/practice/parallel-letter-frequency/.meta/tests.toml
@@ -1,0 +1,49 @@
+# This is an auto-generated file.
+#
+# Regenerating this file via `configlet sync` will:
+# - Recreate every `description` key/value pair
+# - Recreate every `reimplements` key/value pair, where they exist in problem-specifications
+# - Remove any `include = true` key/value pair (an omitted `include` key implies inclusion)
+# - Preserve any other key/value pair
+#
+# As user-added comments (using the # character) will be removed when this file
+# is regenerated, comments can be added via a `comment` key.
+
+[c054d642-c1fa-4234-8007-9339f2337886]
+description = "no texts"
+
+[818031be-49dc-4675-b2f9-c4047f638a2a]
+description = "one text with one letter"
+
+[c0b81d1b-940d-4cea-9f49-8445c69c17ae]
+description = "one text with multiple letters"
+
+[708ff1e0-f14a-43fd-adb5-e76750dcf108]
+description = "two texts with one letter"
+
+[1b5c28bb-4619-4c9d-8db9-a4bb9c3bdca0]
+description = "two texts with multiple letters"
+
+[6366e2b8-b84c-4334-a047-03a00a656d63]
+description = "ignore letter casing"
+
+[92ebcbb0-9181-4421-a784-f6f5aa79f75b]
+description = "ignore whitespace"
+
+[bc5f4203-00ce-4acc-a5fa-f7b865376fd9]
+description = "ignore punctuation"
+
+[68032b8b-346b-4389-a380-e397618f6831]
+description = "ignore numbers"
+
+[aa9f97ac-3961-4af1-88e7-6efed1bfddfd]
+description = "Unicode letters"
+
+[7b1da046-701b-41fc-813e-dcfb5ee51813]
+description = "combination of lower- and uppercase letters, punctuation and white space"
+
+[4727f020-df62-4dcf-99b2-a6e58319cb4f]
+description = "large texts"
+
+[adf8e57b-8e54-4483-b6b8-8b32c115884c]
+description = "many small texts"
--- a/exercises/practice/parallel-letter-frequency/.meta/tests.toml
+++ b/exercises/practice/parallel-letter-frequency/.meta/tests.toml
@@ -41,7 +41,6 @@ rust_fn = "numbers_dont_count"
 
 [aa9f97ac-3961-4af1-88e7-6efed1bfddfd]
 description = "Unicode letters"
-rust_fn = "all_three_anthems_1_worker"
 
 [7b1da046-701b-41fc-813e-dcfb5ee51813]
 description = "combination of lower- and uppercase letters, punctuation and white space"
--- a/exercises/practice/parallel-letter-frequency/.meta/tests.toml
+++ b/exercises/practice/parallel-letter-frequency/.meta/tests.toml
@@ -26,24 +26,30 @@ description = "two texts with multiple letters"
 
 [6366e2b8-b84c-4334-a047-03a00a656d63]
 description = "ignore letter casing"
+rust_fn = "case_insensitivity"
 
 [92ebcbb0-9181-4421-a784-f6f5aa79f75b]
 description = "ignore whitespace"
 
 [bc5f4203-00ce-4acc-a5fa-f7b865376fd9]
 description = "ignore punctuation"
+rust_fn = "punctuation_doesnt_count"
 
 [68032b8b-346b-4389-a380-e397618f6831]
 description = "ignore numbers"
+rust_fn = "numbers_dont_count"
 
 [aa9f97ac-3961-4af1-88e7-6efed1bfddfd]
 description = "Unicode letters"
+rust_fn = "all_three_anthems_1_worker"
 
 [7b1da046-701b-41fc-813e-dcfb5ee51813]
 description = "combination of lower- and uppercase letters, punctuation and white space"
+rust_fn = "all_three_anthems_1_worker"
 
 [4727f020-df62-4dcf-99b2-a6e58319cb4f]
 description = "large texts"
 
 [adf8e57b-8e54-4483-b6b8-8b32c115884c]
 description = "many small texts"
+rust_fn = "many_times_same_text"
--- a/exercises/practice/parallel-letter-frequency/tests/parallel_letter_frequency.rs
+++ b/exercises/practice/parallel-letter-frequency/tests/parallel_letter_frequency.rs
@@ -97,6 +97,13 @@ fn many_empty_lines() {
 
 #[test]
 #[ignore]
+fn ignore_whitespace() {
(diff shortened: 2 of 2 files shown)


senekor:
I don't really understand. Where are the three additional tests coming from? Do you propose a second PR later with a template?

In general, I'd be in favor of adopting the tests from problem-specifications after checking that the tested behavior is the same.

Making the TDD experience more granular for students doesn't seem like a compelling enough argument for me to introduce custom tests. Usually that makes sense if they are inherently Rust-specific or somehow needed to preserve backwards-compatibility with an existing exercise design.

ellnix:
> I don't really understand. Where are the three additional tests coming from? Do you propose a second PR later with a template?

I should have been clearer, those tests are in problem-specifications but not in Rust. I would probably just add a commit to this PR for the template, don't want to add too much noise to the repo.

This is the rename:
```toml
[818031be-49dc-4675-b2f9-c4047f638a2a]
description = "one text with one letter"
```

These are the new tests:
```toml
[c0b81d1b-940d-4cea-9f49-8445c69c17ae]
description = "one text with multiple letters"

[708ff1e0-f14a-43fd-adb5-e76750dcf108]
description = "two texts with one letter"

[1b5c28bb-4619-4c9d-8db9-a4bb9c3bdca0]
description = "two texts with multiple letters"
```

> In general, I'd be in favor of adopting the tests from problem-specifications after checking that the tested behavior is the same.

Alright, so I'm assuming I should go ahead and make a template for this test?

senekor:
Ah, got it. Yeah, a template sounds good.

ellnix:
@senekor I attempted this a couple days ago and I don't honestly think a template is worth it, there's tests with very long text, and also a test with repeating text which would add like 30 lines alone.

<details>
<summary>Test with repeating text</summary>

```json
{
      "uuid": "adf8e57b-8e54-4483-b6b8-8b32c115884c",
      "description": "many small texts",
      "scenarios": ["concurrent"],
      "property": "calculateFrequencies",
      "input": {
        "texts": [
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc",
          "abbccc"
        ]
      },
      "expected": {
        "a": 50,
        "b": 100,
        "c": 150
      }
    }
```

</details>

Keeping the text from being long and unwieldy would require as much code in Tera as there is in the test file itself. I think the way we have it with the texts hoisted at the top makes the most sense.

senekor:
That's ok, but in that case we need to make sure that `tests.toml` actually matches our manually-maintained tests. For example, `tests.toml` currently contains the following:

```toml
[adf8e57b-8e54-4483-b6b8-8b32c115884c]
description = "many small texts"
```

That's the unwieldy test you posted. If we don't include this test, then it must be marked as excluded with a justification in `tests.toml`.

One advantage the templates give us is that they automatically make sure all the tests that are declared as included in `tests.toml` are in fact included.

If we're gonna maintain the mapping manually, we should add a comment above every test with its UUID or something so we can easily trace that work in the future.

I'm open to alternatives, but what we _shouldn't_ do is add a `tests.toml` file that contains incorrect information about which tests are actually implemented and which aren't.

senekor (changes requested):
so you can re-request a review :)

senekor (changes requested):
I meant re-request when you think I should take another look at the code 😄

ellnix:
Oh lmao, I thought for some reason you wanted to go through the request-response workflow again for some reason, my bad. I did a `git reset --hard` on that attempt to add the template because I felt like I wasn't getting anywhere. I got a notification from github about your request for changes but not the first message you sent, and it scrolled past it :tired_face:
 
> I'm open to alternatives, but what we shouldn't do is add a tests.toml file that contains incorrect information about which tests are actually implemented and which aren't.

That test is actually present, and it's just a call to vec! instead of all those lines.

```rust
#[test]
#[ignore]
fn many_times_same_text() {
    let v = vec!["abc"; 1000];
    let mut hm = HashMap::new();
    hm.insert('a', 1000);
    hm.insert('b', 1000);
    hm.insert('c', 1000);
    assert_eq!(frequency::frequency(&v[..], 4), hm);
}
```

> If we're gonna maintain the mapping manually, we should add a comment above every test with its UUID or something so we can easily trace that work in the future.

I think this would be really confusing since students see those comments. Since `tests.toml` *should* keep all arbitrary pairs, we can add a `rust_fn = many_times_same_text` key or similar. That would also be much easier to parse in CI, wherein we could try to match every entry in `tests.toml` with a function matching the description or `rust_fn`.

senekor:
> Since `tests.toml` _should_ keep all arbitrary pairs, we can add a `rust_fn = many_times_same_text` key or similar.

Great idea!

> That would also be much easier to parse in CI, wherein we could try to match every entry in `tests.toml` with a function matching the description or `rust_fn`.

Also a great idea! But I'm not gonna block this PR on it. If you want to do it, go ahead, otherwise feel free to leave it for later. My main concern is that I don't want maintainers to have to do the manual work of matching up `tests.toml` with the actual tests over and over. With the addtional key in `tests.toml`, that's pretty much covered for me.

exercises/practice/parallel-letter-frequency/.meta/tests.toml
```diff
@@ -0,0 +1,55 @@
...
+description = "ignore numbers"
+rust_fn = "numbers_dont_count"
+
+[aa9f97ac-3961-4af1-88e7-6efed1bfddfd]
+description = "Unicode letters"
+rust_fn = "all_three_anthems_1_worker"
+
+[7b1da046-701b-41fc-813e-dcfb5ee51813]
+description = "combination of lower- and uppercase letters, punctuation and white space"
+rust_fn = "all_three_anthems_1_worker"
```
ellnix, line 47: I felt it was appropriate to match tests based on the behavior under test, not necessarily the exact content of the input.

exercises/practice/parallel-letter-frequency/tests/parallel_letter_frequency.rs
```diff
@@ -129,3 +177,183 @@ fn non_integer_multiple_of_threads() {
     hm.insert('c', 999);
     assert_eq!(frequency::frequency(&v[..], 4), hm);
 }
+
+const DOSTOEVSKY: [&str; 4] = [
```
ellnix, line 181: This text is really large so I thought I'd have it after all the tests that work without it so the file is easier to skim. I can understand the argument for having all of the constants defined at the top though.

I don't feel strongly in either case.

exercises/practice/parallel-letter-frequency/.meta/tests.toml
ellnix: For reference, the input of "combination of lower- and uppercase letters, punctuation and white space" is this:

```json
      "input": {
        "texts": [
          "There, peeping among the cloud-wrack above a dark tower high up in the mountains, Sam saw a white star twinkle for a while. The beauty of it smote his heart, as he looked up out of the forsaken land, and hope returned to him. For like a shaft, clear and cold, the thought pierced him that in the end, the shadow was only a small and passing thing: there was light and high beauty forever beyond its reach."
        ]
      },
```

exercises/practice/parallel-letter-frequency/.meta/tests.toml
```diff
@@ -0,0 +1,54 @@
...
+# - Recreate every `reimplements` key/value pair, where they exist in problem-specifications
+# - Remove any `include = true` key/value pair (an omitted `include` key implies inclusion)
+# - Preserve any other key/value pair
+#
+# As user-added comments (using the # character) will be removed when this file
+# is regenerated, comments can be added via a `comment` key.
+
+[c054d642-c1fa-4234-8007-9339f2337886]
+description = "no texts"
+
```
senekor, line 14: Regarding your idea for a CI test, I assume if there is no `rust_fn` field, the test would assume it to be `snake_case(description)`?

exercises/practice/parallel-letter-frequency/.meta/tests.toml
senekor: I agree. If we're maintaining the tests manually, we might as well grant ourselves the flexibility a template would lack. Behavior is what counts.

exercises/practice/parallel-letter-frequency/.meta/tests.toml
senekor: made a corresponding issue [here](https://github.com/exercism/rust/issues/2059)

exercises/practice/parallel-letter-frequency/tests/parallel_letter_frequency.rs
ellnix: I like the idea, since they are constants anyway we have that flexibility.

ellnix: Implemented.

senekor (approved):
Thanks!

PR #2058, opened 2025-03-09
Add a fourth approach to the rust version of the Bob problem

ArturAssisComp:
The fourth approach added to the benchmark is the finite state machine approach. It is faster because it loops through the message chars only once.

### DIFF
--- a/exercises/practice/bob/.articles/performance/code/main.rs
+++ b/exercises/practice/bob/.articles/performance/code/main.rs
@@ -6,6 +6,7 @@ fn main() {
     println!("Hello, world!");
 }
 
+// Reply using match
 pub fn reply_match(msg: &str) -> &str {
     let message = msg.trim_end();
     if message.is_empty() {
@@ -24,6 +25,7 @@ pub fn reply_match(msg: &str) -> &str {
     }
 }
 
+// Reply using if chain
 pub fn reply_if_chain(msg: &str) -> &str {
     let message = msg.trim_end();
     if message.is_empty() {
@@ -47,6 +49,7 @@ pub fn reply_if_chain(msg: &str) -> &str {
     "Whatever."
 }
 
+// Reply using array
 const ANSWERS: &'static [&'static str] = &[
     "Whatever.",
     "Sure.",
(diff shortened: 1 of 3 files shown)


ArturAssisComp:
If this approach is approved, I can add more details to the `approaches` folder.

ellnix:
I like this solution, it seems like it's interesting and creative. I would be interested to see a full approach with it :+1: 

I would suggest that for the approach you write, you should imagine your audience is not familiar with state machines, since Bob is an easy-level exercise. In your shoes I would try to find a good beginner-friendly outside resource that explains state machines as a "Learn more" type of link.

exercises/practice/bob/.articles/performance/content.md
```diff
@@ -7,19 +7,21 @@ The [approaches page][approaches] lists two idiomatic approaches to this exercis
 1. [Using `if` statements][approach-if].
 2. [Using `match` on a `tuple`][approach-match].
 
-For our performance investigation, we'll also include a third approach that [uses an answer array][approach-answer-array].
+For our performance investigation, we'll also include a third approach that [uses an answer array][approach-answer-array] and
+a fourth that uses a state machine.
```
senekor, line 11:
```suggestion
For our performance investigation, we'll also include a third approach that [uses an answer array][approach-answer-array] and a fourth that uses a state machine.
```

Our convention in markdown is [one sentence per line](https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line).

exercises/practice/bob/.articles/performance/content.md
```diff
@@ -7,19 +7,21 @@ The [approaches page][approaches] lists two idiomatic approaches to this exercis
...
-test multiple_line_question_if     ... bench:          97 ns/iter (+/- 12)
-test multiple_line_question_array  ... bench:         100 ns/iter (+/- 2)
+test multiple_line_question_array         ... bench:         175.80 ns/iter (+/- 5.91)
+test multiple_line_question_if            ... bench:         157.26 ns/iter (+/- 8.69)
+test multiple_line_question_match         ... bench:         152.95 ns/iter (+/- 5.79)
+test multiple_line_question_state_machine ... bench:          89.17 ns/iter (+/- 9.65)
 ```
 
-All three approaches are close in performance, but the `if` statements and `match` approaches may be considered to be more idiomatic.
+All four approaches are close in performance, but the `if` statements and `match` approaches may be considered to be more idiomatic. The state machine approach is the fastest because it loops through the characters of the message only once.
```
senekor, line 24:
```suggestion
All four approaches are close in performance, but the `if` statements and `match` approaches may be considered to be more idiomatic.
The state machine approach is the fastest because it loops through the characters of the message only once.
```

same here, one sentence per line

exercises/practice/bob/.articles/performance/code/main.rs
```diff
@@ -71,20 +74,132 @@ pub fn reply_array(msg: &str) -> &str {
...
 #[bench]
 /// multiple line question for match
 fn multiple_line_question_match(b: &mut Bencher) {
-    b.iter(|| reply_match("\rDoes this cryogenic chamber make me look fat?\rNo."));
+    b.iter(|| {
+        reply_match("\rDoes this cryogenic chamber make me look fat?\rNo.");
+        reply_match("  ");
+        reply_match("Does this cryogenic chamber make me look fat?");
+        reply_match("WHAT'S GOING ON?");
+        reply_match("ZOMG THE %^*@#$(*^ ZOMBIES ARE COMING!!11!!1!");
```
senekor, line 167: A separate function could make sense here, to make it more clear that all benches are running the same code.

senekor (changes requested):
You could also try [match guards](https://doc.rust-lang.org/rust-by-example/flow_control/match/guard.html) to make the match arms more concise. Not a requirement, go with what you prefer.

exercises/practice/bob/.articles/performance/content.md
```diff
@@ -7,19 +7,21 @@ The [approaches page][approaches] lists two idiomatic approaches to this exercis
...
 To benchmark the approaches, we wrote a [small benchmark application][benchmark-application].
 
 ```
-test multiple_line_question_match  ... bench:          96 ns/iter (+/- 17)
-test multiple_line_question_if     ... bench:          97 ns/iter (+/- 12)
-test multiple_line_question_array  ... bench:         100 ns/iter (+/- 2)
+test multiple_line_question_array         ... bench:         175.80 ns/iter (+/- 5.91)
+test multiple_line_question_if            ... bench:         157.26 ns/iter (+/- 8.69)
+test multiple_line_question_match         ... bench:         152.95 ns/iter (+/- 5.79)
+test multiple_line_question_state_machine ... bench:          89.17 ns/iter (+/- 9.65)
```
senekor, line 21: I cannot reproduce this result. On my machine, the state machine approach is consistently the slowest:

```
test multiple_line_question_array         ... bench:         119.61 ns/iter (+/- 1.57)
test multiple_line_question_if            ... bench:          95.64 ns/iter (+/- 2.83)
test multiple_line_question_match         ... bench:          96.57 ns/iter (+/- 5.44)
test multiple_line_question_state_machine ... bench:         151.83 ns/iter (+/- 7.84)
```

exercises/practice/bob/.articles/performance/code/main.rs
```diff
@@ -71,20 +74,132 @@ pub fn reply_array(msg: &str) -> &str {
...
+    });
 }
 
 #[bench]
 /// multiple line question for if statements
 fn multiple_line_question_if(b: &mut Bencher) {
-    b.iter(|| reply_if_chain("\rDoes this cryogenic chamber make me look fat?\rNo."));
+    b.iter(|| {
+        reply_if_chain("\rDoes this cryogenic chamber make me look fat?\rNo.");
```
senekor, line 175:
```suggestion
        reply_if_chain(std::hint::black_box("\rDoes this cryogenic chamber make me look fat?\rNo."));
```

The inputs to the functions under test should be wrapped in `black_box`, to prevent the compiler from optimizing the function based on knowledge about its input. In my testing, `multiple_line_question_array` slows down a little from 120 ns/iter to 140 ns/iter, the other approaches stay the same.

senekor:
This doesn't make any sense at all. I had a fifth idea and added that to the benchmarks, which significantly slowed down the results of the _other_ functions. I think the builtin bencher might just be trash, no wonder its unstable. We could try to setup [divan](https://docs.rs/divan/latest/divan/) or [criterion](https://docs.rs/criterion/latest/criterion/), those should give more reliable results. But not today for me 🙂

ArturAssisComp:
Ok, I will try criterion for best results. If you can add your approach here for me to test it.

ArturAssisComp:
Ok, I will fix the requests.

ArturAssisComp:
I used Criterion and found what was happening. Basically, the approaches that use `message == message.to_uppercase()` take benefit from the following optimization in some cases:
image

Initially, I found it very strange that the approaches that transformed all the message to uppercase (creating a new String) and then compared them using == were faster than an approach that loops through all the chars once and does not perform very heavy operations on each iteration. But I forgot that the compiler and the std lib do a good job on optimizations and that iterating over chars is not so cheap as iterating over the bytes of the string. 

One of the optimizations (the one on the image above) makes the operation `to_uppercase` much faster on pure ascii strings, that is why the state machine approach is slower in its worst cases (when it needs to loop through all the string to find the answer). For tests that have non-ascii chars, it tends to perform better. 

Maybe it is possible to optimize the state machine approach to deal with ascii chars, but I did not have time to try.

## Results
- **For YELL and YELL QUESTION**: one of the state machine's worst case because it needs to loop through all the string to find the answer. Thus, it performed worse than the other approaches for lore ipsum with only ascii chars but, surprisingly, performed better for lore ipsum with ascii and non-ascii chars. The explanation is that the other approaches could not take advantage of the ascii optimization.
- **For normal question and whatever**: state machine performed better, but that happened because the lorem ipsum has lower case chars distributed along the string, including the end of the string, which makes it almost O(1) for finding the answer. It is possible to build a worst case situation for the state machine in which all the chars are uppercase except the first, making it necessary to loop through all the chars. In that situation, we would have something close to the YELL and YELL QUESTION.
- **For the empty answer**: For that specific situation, all the approaches are optimized, but the state-machine performed a little bit worse. That was because it used `trim` instead of `trim_end`, generating a little bit more computation.

ArturAssisComp:
I updated the content.md for the benchmark. I found that maybe this exercise is not the best for introducing the concept of state machines, thus, I decided to keep the state machine approach only in the benchmark. Maybe an exercise for building a tokenizer or something like that would be better to introduce this concept.

senekor:
Looks like you committed the output directory of criterion, please clean up the diff.

ArturAssisComp:
yes, the idea was to provide the report (the output of criterion), given that it takes a lot of time to reproduce the result. But I will clean it.

ArturAssisComp:
Results cleaned and content updated.

PR #2064, opened 2025-04-16
Fix Working with strings in Rust link

Klavionik:
Hey! 

Just fixed the link, the current one shows 404.


senekor (approved):
Out of curiosity, where did you bump into that? The syllabus is supposed to be deactivated.

PR #2065, opened 2025-04-16
Fix clippy warnings

senekor:


ellnix (approved):
Seems strange that it's not just a `vec![0; 36]`, but nothing wrong per se.

senekor:
TBH I find this check annoying, the example solutions aren't shown to users. No need to keep them at 100% code quality. I'm just always too lazy to get rid of this check, fixing the warnings is easier in the moment.

PR #2066, opened 2025-04-17
Fix appended instructions for Say exercise

software-artificer:
The "Say" exercise contains the statement that is not true, at least no longer:

> There is a -1 version of a test case, but it is commented out. If your function is implemented properly, the -1 test case should not compile.

Remove this paragraph to avoid confusion.

### DIFF
--- a/exercises/practice/say/.docs/instructions.append.md
+++ b/exercises/practice/say/.docs/instructions.append.md
@@ -8,8 +8,6 @@ errors for out of range, we are using Rust's strong type system to limit
 input.  It is much easier to make a function deal with all valid inputs,
 rather than requiring the user of your module to handle errors.
 
-Adding 'and' into number text has not been implemented in test cases.
-
 ### Extension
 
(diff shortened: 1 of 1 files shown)


senekor:
The regression was introduced in #1773

exercises/practice/say/.docs/instructions.append.md
```diff
@@ -8,9 +8,6 @@ errors for out of range, we are using Rust's strong type system to limit
 input.  It is much easier to make a function deal with all valid inputs,
 rather than requiring the user of your module to handle errors.
 
-There is a -1 version of a test case, but it is commented out.
-If your function is implemented properly, the -1 test case should not compile.
-
 Adding 'and' into number text has not been implemented in test cases.
```
senekor, line 11: There is also no more test case with `"and"` as input upstream, so let's remove this line as well.

software-artificer: Dropped that one as well, thank you!

PR #2067, opened 2025-04-19
etl: downgrade to easy difficulty

senekor:
forum discussion: https://forum.exercism.org/t/rust-etl-exercise-is-a-bit-odd/16921


ellnix (approved):
Agreed, this exercise is pretty easy. Does not belong in the same category as alphametics.

PR #2072, opened 2025-06-25
Ensure problem-specifications is symlinked

senekor:
This is especially relevant for the add-exercise workflow, which will suggest slugs of unimplemented exercises from problem-specifications only if the symlink exists. Otherwise no suggestions are made without any hint that something went wrong.

### DIFF
--- a/justfile
+++ b/justfile
@@ -23,8 +23,11 @@ test:
     cd rust-tooling && cargo test
     ./bin/format_exercises.sh ; git diff --exit-code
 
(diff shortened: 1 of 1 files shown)


senekor:
The idea is that maybe you're testing some other commit, maybe you have some local changes in problem-specifications that you want to test how they interact with the Rust track. It would be annoying if running these scripts messed with those changes. So, only run the script if the symlink doesn't exist already.

PR #2074, opened 2025-06-30
Create run-configlet-sync.yml

jagdish-15:
### 🛠️ Automating Configlet Sync

This PR introduces a workflow that automates `configlet sync` for test (only an issue is raised), docs, metadata, and filepaths. There are two jobs in the main workflow; we can skip either one of them. Let me know how you'd like to set it up!

⏱️ The workflow is currently scheduled to run on the **15th of every month** via cron.  
Let me know if you'd prefer a different schedule (e.g., weekly, the 1st of each month, etc.).

This setup is already being used across multiple tracks!.

For more context and discussion, see:  
🔗 https://forum.exercism.org/t/automating-syncing-with-github-actions-final-testing-going-on-for-java-track-open-to-more-tracks/17807

### DIFF
--- a/.github/workflows/run-configlet-sync.yml
+++ b/.github/workflows/run-configlet-sync.yml
@@ -1,0 +1,10 @@
+name: Run Configlet Sync
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 0 15 * *'
+
+jobs:
+  call-gha-workflow:
+    uses: exercism/github-actions/.github/workflows/configlet-sync.yml@main


senekor:
Sounds cool, gonna have a closer look later and probably merge.

senekor:
Just for reference, we currently have a system where we pin problem-specifications to a specific commit so we can reliably test if the tests checked into the repo are the same as the ones produced by the test generator. We should probably get rid of that, adopting this PR new action is a different, incompatible approach.

PR #2077, opened 2025-07-03
Pov exercise

devcyjung:
Linked from this forum post:

[new-rust-practice-exercise-suggestion-pov](https://forum.exercism.org/t/new-rust-practice-exercise-suggestion-pov/18128/5)

Disregard the last PR #2076 please, I confused example.rs with lib.rs there

### DIFF
--- a/config.json
+++ b/config.json
@@ -1572,6 +1572,15 @@
           "lists",
           "unsafe"
         ]
+      },
+      {
+        "slug": "pov",
+        "name": "POV",
+        "uuid": "bf7b7309-3d34-4893-a584-f7742502e012",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 10,
+        "topics": []
       }
     ],
     "foregone": [
--- a/exercises/practice/pov/.meta/example.rs
+++ b/exercises/practice/pov/.meta/example.rs
@@ -1,35 +1,95 @@
-#[derive(Clone, PartialEq, Eq)]
-pub struct Tree {}
+use std::ops::{Deref, DerefMut};
 
-impl Tree {
+#[derive(Clone, PartialEq)]
+pub struct Tree<T: Clone + PartialEq> {
+    label: T,
+    children: Vec<Box<Tree<T>>>,
+}
+
+impl<T: Clone + PartialEq> Tree<T> {
     #[must_use]
-    pub fn new(_label: &str) -> Self {
-        todo!("Implement a function that creates a new Tree with given label");
+    pub fn new(label: T) -> Self {
+        Self {
+            label,
+            children: Default::default(),
+        }
     }
 
     #[must_use]
-    pub fn with_children(_label: &str, _children: &[Tree]) -> Self {
-        todo!("Implement a function that creates a new Tree with given label and Children");
+    pub fn with_children(label: T, children: Vec<Self>) -> Self {
+        Self {
+            label,
+            children: children.into_iter().map(Box::new).collect(),
+        }
     }
 
     #[must_use]
-    pub fn get_label(&self) -> String {
-        todo!("Implement getter for label.");
+    pub fn get_label(&self) -> T {
+        self.label.clone()
     }
 
     #[must_use]
     pub fn get_children(&self) -> Vec<&Self> {
-        todo!("Implement getter for children.");
+        self.children.iter().map(Box::deref).collect()
     }
 
     #[must_use]
-    pub fn pov_from(self, _from: &str) -> Option<Self> {
-        todo!("Implement a function that reparents Tree with 'from' as root.");
+    pub fn pov_from(&self, from: &T) -> Option<Self> {
+        // list of (child, parent, child's index in parent.children)
+        let mut lookup = vec![(self, None, None)];
+        let mut stack = vec![self];
+        while let Some(parent) = stack.pop() {
+            if &parent.label == from {
+                return self.reparent(parent, lookup.as_slice()).into();
+            }
+            lookup.extend(
+                parent
+                    .children
+                    .iter()
+                    .map(Box::deref)
+                    .enumerate()
+                    .map(|(i, child)| (child, Some(parent), Some(i))),
+            );
+            stack.extend(parent.children.iter().map(Box::deref));
+        }
+        None
+    }
+
+    // lookup is list of (child, parent, child's index in parent.children)
+    #[must_use]
+    fn reparent(&self, parent: &Self, lookup: &[(&Self, Option<&Self>, Option<usize>)]) -> Self {
+        let mut new_root = parent.clone();
+        let mut current = parent;
+        let mut clone_mut = &mut new_root;
+        let find_parent = |child| lookup.iter().find(|(c, _p, _i)| *c == child);
+        while let Some(&(_, Some(parent), Some(index))) = find_parent(current) {
+            let mut parent_clone = parent.clone();
+            parent_clone.children.swap_remove(index);
+            clone_mut.children.push(Box::new(parent_clone));
+            current = parent;
+            let new_box = clone_mut
+                .children
+                .last_mut()
+                .expect("We just inserted node, this is not empty");
+            clone_mut = new_box.deref_mut();
+        }
+        new_root
     }
 
     #[must_use]
-    pub fn path_to(self, _from: &str, _to: &str) -> Option<Vec<String>>
-    {
-        todo!("Implement a function that returns the list of labels in the shortest path from 'from' to 'to'");
+    pub fn path_to(&self, from: &T, to: &T) -> Option<Vec<T>> {
+        if from != &self.label {
+            return self.pov_from(from).and_then(|pov| pov.path_to(from, to));
+        }
+        if to == &self.label {
+            return Some(vec![self.label.clone()]);
+        }
+        for child in self.children.iter() {
+            if let Some(mut path) = child.path_to(&child.label, to) {
+                path.insert(0, self.label.clone());
+                return Some(path);
+            }
+        }
+        None
     }
 }
--- a/exercises/practice/pov/.meta/example.rs
+++ b/exercises/practice/pov/.meta/example.rs
@@ -1,0 +1,35 @@
+#[derive(Clone, PartialEq, Eq)]
+pub struct Tree {}
+
+impl Tree {
+    #[must_use]
+    pub fn new(_label: &str) -> Self {
+        todo!("Implement a function that creates a new Tree with given label");
+    }
+
+    #[must_use]
+    pub fn with_children(_label: &str, _children: &[Tree]) -> Self {
+        todo!("Implement a function that creates a new Tree with given label and Children");
+    }
+
+    #[must_use]
+    pub fn get_label(&self) -> String {
+        todo!("Implement getter for label.");
+    }
+
+    #[must_use]
+    pub fn get_children(&self) -> Vec<&Self> {
+        todo!("Implement getter for children.");
+    }
+
+    #[must_use]
+    pub fn pov_from(self, _from: &str) -> Option<Self> {
+        todo!("Implement a function that reparents Tree with 'from' as root.");
+    }
+
+    #[must_use]
+    pub fn path_to(self, _from: &str, _to: &str) -> Option<Vec<String>>
+    {
+        todo!("Implement a function that returns the list of labels in the shortest path from 'from' to 'to'");
+    }
+}
--- a/exercises/practice/pov/Cargo.toml
+++ b/exercises/practice/pov/Cargo.toml
@@ -1,0 +1,9 @@
+[package]
+name = "pov"
+version = "0.1.0"
+edition = "2024"
+
+# Not all libraries from crates.io are available in Exercism's test runner.
+# The full list of available libraries is here:
+# https://github.com/exercism/rust-test-runner/blob/main/local-registry/Cargo.toml
+[dependencies]
--- a/exercises/practice/pov/src/lib.rs
+++ b/exercises/practice/pov/src/lib.rs
@@ -1,84 +1,38 @@
-use std::ops::{Deref, DerefMut};
-
 #[derive(Clone, PartialEq)]
 pub struct Tree<T: Clone + PartialEq> {
-    label: T,
-    children: Vec<Box<Tree<T>>>,
+    _remove_this: std::marker::PhantomData<T>,
 }
 
 impl<T: Clone + PartialEq> Tree<T> {
     #[must_use]
-    pub fn new(label: T) -> Self {
-        Self { label, children: Default::default() }
+    pub fn new(_label: T) -> Self {
+        todo!("Implement a function that creates a new Tree with given label");
     }
 
     #[must_use]
-    pub fn with_children(label: T, children: Vec<Tree<T>>) -> Self {
-        Self { label, children: children.into_iter().map(Box::new).collect() }
+    pub fn with_children(_label: T, _children: Vec<Self>) -> Self {
+        todo!("Implement a function that creates a new Tree with given label and Children");
     }
 
     #[must_use]
     pub fn get_label(&self) -> T {
-        self.label.clone()
+        todo!("Implement getter for label.");
     }
 
     #[must_use]
     pub fn get_children(&self) -> Vec<&Self> {
-        self.children.iter().map(Box::deref).collect()
+        todo!("Implement getter for children.");
     }
 
     #[must_use]
-    pub fn pov_from(&self, from: &T) -> Option<Self> {
-        // list of (child, parent, child's index in parent.children)
-        let mut lookup = vec![(self, None, None)];
-        let mut stack = vec![self];
-        while let Some(parent) = stack.pop() {
-            if &parent.label == from {
-                return self.reparent(parent, lookup.as_slice()).into();
-            }
-            lookup.extend(
-                parent.children.iter()
-                    .map(Box::deref).enumerate()
-                    .map(|(i, child)| (child, Some(parent), Some(i)))
-            );
-            stack.extend(parent.children.iter().map(Box::deref));
-        }
-        None
-    }
-    
-    // lookup is list of (child, parent, child's index in parent.children)
-    #[must_use]
-    fn reparent(&self, parent: &Self, lookup: &[(&Self, Option<&Self>, Option<usize>)]) -> Self {
-        let mut new_root = parent.clone();
-        let mut current = parent;
-        let mut clone_mut = &mut new_root;
-        let find_parent = |child| lookup.iter().find(|(c, _p, _i)| *c == child);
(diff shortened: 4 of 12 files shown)


exercises/practice/pov/tests/pov.rs
```diff
@@ -0,0 +1,428 @@
...
+///
+/// If appropriate for your track, you may test that the input tree is not modified.
+///
+/// Note that when rerooting upon a target node that has both parents and children,
+/// it does not matter whether the former parent comes before or after the former children.
+/// Please account for this when checking correctness of the resulting trees.
+/// One suggested method is to only check two things:
+/// 1) The root of the returned tree is the root that was passed in to from_pov.
+/// 2) The sorted edge list of the returned tree is the same as the sorted edge list of the expected tree.
+mod reroot_a_tree_so_that_its_root_is_the_specified_node {
```
senekor, line 13: This module name is too long. In combination with long test names, user's test output will be cluttered. Please make it shorter. If need be, I don't mind hardcoding the module name in the template.

exercises/practice/pov/src/lib.rs
```diff
@@ -0,0 +1,38 @@
...
+
+impl<T: Clone + PartialEq> Tree<T> {
+    #[must_use]
+    pub fn new(_label: T) -> Self {
+        todo!("Implement a function that creates a new Tree with given label");
+    }
+
+    #[must_use]
+    pub fn with_children(_label: T, _children: Vec<Self>) -> Self {
+        todo!("Implement a function that creates a new Tree with given label and Children");
```
senekor, line 14: We prefer to avoid parameter names prefixed with underscore when possible. Please implement `Debug` on `Tree` and use `"{tree:?}"` in the format string. (goes for all functions here)

exercises/practice/pov/tests/pov.rs
```diff
@@ -0,0 +1,428 @@
...
+
+    pub fn tree_option_eq(lhs: Option<Tree<String>>, rhs: Option<Tree<String>>) -> bool {
+        match (lhs, rhs) {
+            (None, None) => true,
+            (Some(l_inner), Some(r_inner)) => tree_eq(&l_inner, &r_inner),
+            _ => false,
+        }
+    }
+
+    pub fn tree_eq(lhs: &Tree<String>, rhs: &Tree<String>) -> bool {
```
senekor, line 411: I would prefer the normal `Eq` traits be used instead of a separate function. Maybe you could implement a standalone function that only sorts the keys? Then you could get the tree from the user's solution, sort it, then compare it normally with the expected tree.

senekor (changes requested):
Thanks, this looks like a great start! This is a quick initial pass, I'll do a more in-depth review later.

Btw. you don't have a close the PR and open another one if something isn't right. Just add your improvements to the branch and push again, the PR will be updated automatically.

rust-tooling/generate/src/custom_filters.rs
```diff
@@ -56,3 +57,24 @@ pub fn fmt_num(value: &Value, _args: &HashMap<String, Value>) -> Result<Value> {
     let pretty_num = String::from_utf8(pretty_digits).unwrap_or_default();
     Ok(Value::String(pretty_num))
 }
+
+pub fn camel_to_snake(value: &Value, _args: &HashMap<String, Value>) -> Result<Value> {
```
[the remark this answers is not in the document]
devcyjung: Oh this was because

For example fromPov -> turned into frompov

But I wanted from_pov.

But it's alright. I can remove if you think it's unnecessary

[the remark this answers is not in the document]
senekor: Oh, now I understand. make_ident only handles kebab-to-snake. In that case, I would prefer if make_ident handled camel-to-snake as well.

exercises/practice/pov/.meta/example.rs
```diff
@@ -0,0 +1,96 @@
...
+use std::ops::{Deref, DerefMut};
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct Tree<T: Clone + Debug + PartialEq> {
+    label: T,
+    children: Vec<Box<Tree<T>>>,
+}
+
+impl<T: Clone + Debug + PartialEq> Tree<T> {
+    #[must_use]
```
senekor, line 11: also remove the `#[must_use]` annotations here in the example.rs

exercises/practice/pov/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.1.0"
+edition = "2024"
+
+# Not all libraries from crates.io are available in Exercism's test runner.
+# The full list of available libraries is here:
+# https://github.com/exercism/rust-test-runner/blob/main/local-registry/Cargo.toml
+[dependencies]
+
+[dev-dependencies]
+pretty_assertions = "1.4.1"
```
senekor, line 12:
```suggestion
pretty_assertions = "*"
```

This will make it less likely for solutions to break if pretty_assertions is updated to 2.0 and the test runner doesn't include 1.x anymore. It can still break if pretty_assertions 2.0 doesn't have a macro `assert_eq!` anymore, but that seems even less likely.

also, nit: add trailing newline

rust-tooling/generate/src/custom_filters.rs
```diff
@@ -55,4 +55,4 @@ pub fn fmt_num(value: &Value, _args: &HashMap<String, Value>) -> Result<Value> {
     pretty_digits.reverse();
     let pretty_num = String::from_utf8(pretty_digits).unwrap_or_default();
     Ok(Value::String(pretty_num))
-}
+}
```
senekor, line 58: nit: add trailing newline

exercises/practice/pov/src/lib.rs
```diff
@@ -0,0 +1,34 @@
...
+
+    pub fn with_children(label: T, children: Vec<Self>) -> Self {
+        todo!("Implement a function that creates a new Tree with given {label:?} and {children:?}");
+    }
+
+    pub fn get_label(&self) -> T {
+        todo!("Implement getter for label.");
+    }
+
+    pub fn get_children(&self) -> Vec<&Self> {
```
senekor, line 21: The [Rust API guidelines](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter) discourage a `get_` prefix for getter names. Rename these to just `label` and `children`.

exercises/practice/pov/tests/pov.rs
```diff
@@ -0,0 +1,421 @@
...
+
+    #[test]
+    fn results_in_the_same_tree_if_the_input_tree_is_a_singleton() {
+        let input = Tree::new("x".to_string());
+        let from = "x".to_string();
+        let result = input.pov_from(&from);
+        let expected = Some(Tree::new("x".to_string()));
+        assert_eq!(
+            result.map(|t| crate::test_util::tree_to_sorted(&t)),
+            expected.map(|t| crate::test_util::tree_to_sorted(&t))
```
senekor, line 14: To make this less verbose, make the `tree_to_sorted` function take an `Option<Tree>` directly. It's only ever used this way. It's not necessarily a beautiful API, but it doesn't matter, because the user doesn't interact with this function. it's more important for the tests themselves to be readable.

Also, import the function at the top of the module:
```rs
mod from_pov {
    use super::test_util::tree_to_sorted;
    // ...
    assert_eq!(tree_to_sorted(result), tree_to_sorted(expected));
```

exercises/practice/pov/src/lib.rs
```diff
@@ -0,0 +1,34 @@
+use std::fmt::Debug;
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct Tree<T: Clone + Debug + PartialEq> {
+    _remove_this: std::marker::PhantomData<T>,
+}
```
senekor, line 6: The exercise is not primarily about generics, so what do you think about dropping them and using something like this for the label:

```rs
// In production code, this would be a type parameter on `Tree`.
pub type Label = &'static str;
```

I'm not 100% sure, maybe it's dumbing it down too much. But right now I'm not seeing the value of making the user implement this with generics.

senekor (changes requested):

exercises/practice/pov/src/lib.rs
devcyjung: I initially started with non-generic Tree, but along the way I realized that making this non-generic doesn't really make the exercise any simpler. So I added generic. It also has expected trait bound of PartialEq and Clone, so users can expect where their solution should be directed to.

Also with the String non-generic Tree I spent a lot of time should I make this return type impl AsRef<str> or just String, or  Cow<>, etc

exercises/practice/pov/src/lib.rs
```diff
@@ -0,0 +1,34 @@
...
+
+    pub fn with_children(label: T, children: Vec<Self>) -> Self {
+        todo!("Implement a function that creates a new Tree with given {label:?} and {children:?}");
+    }
+
+    pub fn label(&self) -> T {
+        todo!("Implement getter for label.");
+    }
+
+    pub fn children(&self) -> Vec<&Self> {
```
senekor, line 21: I don't like that we need this function in the first place. It's only used in the util sorter function... the tests don't require it directly. What if we just tell users that they must ensure tree comparison doesn't fail due to a different order of children? They would have to options of ensuring that:

1. Use a `HashSet` instead of a `Vec` to store children.
2. Always sort children themselves.

---

This is not relevant if we can remove the function, but I think the idiomatic API here would be:

```suggestion
pub fn children(&self) -> impl Iterator<Item = &Self> { todo!() }
```

exercises/practice/pov/tests/pov.rs
```diff
@@ -0,0 +1,404 @@
+/// Reroot a tree so that its root is the specified node.
+mod from_pov {
+    use super::test_util::tree_to_sorted;
+    use pov::*;
+    use pretty_assertions::assert_eq;
+
+    #[test]
+    fn results_in_the_same_tree_if_the_input_tree_is_a_singleton() {
+        let input = Tree::new("x".to_string());
```
senekor, line 9: Even if we keep the generics, why are all labels converted to `String`? Wouldn't `&'static str` work perfectly fine? The tests look rather cluttered with all the `.to_string()`.

senekor (changes requested):

exercises/practice/pov/tests/pov.rs
devcyjung: In that case the Tree will be Tree<'a str> 
I wanted to test against the most typical type of generic structure, which owns T

exercises/practice/pov/src/lib.rs
```diff
@@ -0,0 +1,34 @@
...
+pub struct Tree<T: Clone + Debug + PartialEq> {
+    _remove_this: std::marker::PhantomData<T>,
+}
+
+impl<T: Clone + Debug + PartialEq> Tree<T> {
+    pub fn new(label: T) -> Self {
+        todo!("Implement a function that creates a new Tree with given {label:?}");
+    }
+
+    pub fn with_children(label: T, children: Vec<Self>) -> Self {
```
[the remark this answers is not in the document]
devcyjung: Ok in that case the testcases will be a lot longer, but if you are ok with that I'll do so

rust-tooling/generate/src/custom_filters.rs
```diff
@@ -35,6 +36,14 @@ pub fn make_ident(value: &Value, _args: &HashMap<String, Value>) -> Result<Value
     Ok(Value::String(value))
 }
 
+fn underscore_camel_case(value: &str) -> String {
+    let re = Regex::new(r"([a-z0-9])([A-Z])").unwrap();
+    let value = re.replace_all(value, "${1}_${2}");
+    let re2 = Regex::new(r"([A-Z])([A-Z][a-z])").unwrap();
+    let value = re2.replace_all(&value, "${1}_${2}");
+    String::from(value)
+}
```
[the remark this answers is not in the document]
devcyjung: Ok I'll make no regex version

exercises/practice/pov/src/lib.rs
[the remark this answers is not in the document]
senekor: Yeah, making the tests longer is not that big of a deal. I like the last option because it doesn't prime the user to use a specific data structure. Another way to do that would be to have a parameter `children: impl Iterator<Item = Self>`. That can be collected into any data structure, vec or hash set. Either way is fine by me.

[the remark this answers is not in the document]
senekor: In that case, go with one of these two options:

```
fn with_child(self, child: Self) -> Self {}
fn add_child(&mut self, child: Self) {}
```

I'm against `with_children(self, Vec)` because it gently pushes people toward a specific implementation.

exercises/practice/pov/src/lib.rs
senekor: Answering from [this comment](https://github.com/exercism/rust/pull/2077/files#r2183697487) here on the topic of removing / keeping generics:

> Also designing the data structure itself is a big part of this exercise. That's why I like the generic idea. Data structure is the best use case of generics.

I don't think generics must necessarily be a big part of the exercise just because it's about a data structure. I think the essence of this exercise is how to handle `Option<Box<T>>` with mutation. The choice between Vec and HashSet is neither important nor interesting.

> If this is the ultimate challenge for data structure design, I think we should go for a proper generic support

I do _not_ think this is the ultimate challenge for data structure design. I'm much more interested in a focused exercise about handling `Option<Box<T>>` with mutation.

> The Trait bound that I chose didn't come from own my implementation. If that were the case I would have added Hash so I can use HashMap for trail.

I see this as another argument against generics. With the current bounds, users cannot use a `HashSet` for their implementation. They have to further restrict the bounds with `Eq + Hash`.

> There's no reason not to go generic imo.

I think the fact that the trait bounds are either too restrictive or too loose for any given implementation is a pretty good argument against generics. In addition to that, I maintain that it's needless complexity that distracts from the learning purpose of the exercise.

---

As commented elsewhere, I now see an argument in favor of generics. Please use the following bounds: `T: Debug + Eq + Hash`. This will allow us to add a test case with a type that doesn't implement `Clone` ensuring users don't cheat by cloning nodes.

config.json
```diff
@@ -1572,6 +1572,15 @@
...
           "unsafe"
         ]
+      },
+      {
+        "slug": "pov",
+        "name": "POV",
+        "uuid": "bf7b7309-3d34-4893-a584-f7742502e012",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 10,
```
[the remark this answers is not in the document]
senekor: Here's the list of difficult exercises on the rust track:

- parallel-letter-frequency
- macros
- poker
- forth
- ocr-numbers
- react
- circular buffer
- rectangles
- xorcism
- dominoes
- doubly-linked-list

With the exception of ocr-numbers and rectangles, all of these are more difficult than `pov` in my opinion. I would actually consider downgrading ocr-numbers and rectangles to medium, I don't see what's difficult about them.

A comparable exercise is simple-linked-list, it also involves `Option<Box<T>>`. But it's clearly easier, it doesn't involve any weird manipulation of the tree.

Looking over all the other medium exercises, `pov` is clearly more difficult than most of them. Few if any of the medium exercises actually require dealing with Rust's ownership system in a meaningful sense like `pov`. So, while I still think it's on the lower end of difficult, I'm okay with keeping it there.

exercises/practice/pov/src/lib.rs
senekor: You mixed up the topic of the thread here. This thread is about whether or not we need users to implement the function `children` as a publicly exposed API. The comments you made about generics should be in a different comment thread. I'll repost them there in quoted blocks so we don't get things mixed up. So my answer here is only about the `children` API. Please keep threads on-topic in the future.

> All the language tracks that I've tried this exercise so far don't demand specific sort order.

We don't have to implement things exactly the same way as other language tracks. Rust has two properties that not all other langues have:
1. The implementor of a type provides its own comparison function.
1. Encapsulation is idiomatic, publicly exposing internals for the sake of testing is discouraged.

These two properties make it generally unidiomatic to extend foreign types with custom sorting logic. IMO this justifies adding some additional requirements to the exercise. Asking users to make sure their data type compares equal independent of the order of the children also fits well with the instructions themselves, which describe the trees as undirected. The operation of reparenting wouldn't make sense in a tree where the order of children matters.

exercises/practice/pov/src/lib.rs
```diff
@@ -0,0 +1,34 @@
...
+
+    pub fn label(&self) -> T {
+        todo!("Implement getter for label.");
+    }
+
+    pub fn children(&self) -> Vec<&Self> {
+        todo!("Implement getter for children.");
+    }
+
+    pub fn pov_from(&self, from: &T) -> Option<Self> {
```
senekor, line 25: quoting from [another comment](https://github.com/exercism/rust/pull/2077/files#r2183697487):

> Clone is necessary because we want them to make immutable solution. It's in the canonical json spec comments that language tracks can choose to enforce immutability, and in Rust making mutable solution is arguably more complex anyways.

I hadn't noticed that yet, but I think the solution should _definitely_ use a mutable approach. Without that, marking the exercise as difficult makes no sense. Users should be encouraged to reuse allocations whenever reasonably possible. Rust is a performance-oriented language. That doesn't mean performance has to be the focus of everything, but we shouldn't design our APIs in a way that _forces_ users into an inefficient implementation.

I propose:

```rs
fn pov_from(tree: &mut Option<Box<Tree>>, from: &T) { todo!() }
```

If desired, this can be split in two types `Tree` and `Node`, where the tree only holds a `Option<Box<Node>>` and the node is where the action happens in terms of the data structure.

Now I actually see an argument in favor of generics... With generics, we could add a special test case with a type that doesn't implement `Clone`, to make sure users don't add the `Clone` bound themselves to cheat.

senekor (changes requested):

exercises/practice/pov/src/lib.rs
devcyjung: Can you please explain why tree is `Option`? How about `&mut Box<Tree>`

devcyjung: @senekor ok I'll implement as close to what you suggested, do you expect it will be merged soon by then? It will take a day or two for me I think since the API is totally different

senekor: I cannot say in advance when this will be merged.

senekor: At this point, please don't mark any conversations as resolved, ever. I need the resolved state to keep track of what has been completed and what hasn't. I will mark conversations as resolved when I do a review and find that I won't need to review it again in the future.

senekor:
@devcyjung do you still intent to work on this? I like the exercise idea a lot. If you don't have time anymore, I might pick it up myself to get it over the finish line (preserving your authorship of course).

PR #2079, opened 2025-07-05
flower-field replaces minesweeper

keiravillekode:


senekor:
For reference: https://forum.exercism.org/t/suggestion-deprecate-minesweeper-for-flower-field/

PR #2084, opened 2025-07-15
Remove system for pinning problem-specifications

senekor:
We now have CI workflows that raise issues when configlet finds something that needs updating. This is in conflict with the pinned problem-specifications approach.


senekor:
Yes, that is correct. And the new CI workflow now notifies us if tests are out of date anyway, so the test has less value.

PR #2091, opened 2025-08-15
Fix clippy warnings

senekor:


senekor:
The diff looks ugly, but it's just collapsing if-statements now that [let-chains are stable](https://blog.rust-lang.org/2025/06/26/Rust-1.88.0/#let-chains).

senekor:
ping @ellnix

PR #2098, opened 2025-09-15
migrate from just to mise

senekor:


senekor:
(To be clear: I don't think `mise` has a huge advantage for this repo in particular. I migrated over because mise also works as an all-purpose devtool manager on top of a task runner. I use that a lot, installing misc. project-specific tools with mise. The only reason to migrate here for me is consistency with other projects, not a big deal.)

ellnix (approved):
I'm not at all opinionated in run tools, if it works well for you that's good enough for me.

`mise` works fine and you seem to have ported everything except uuid, which I don't think is necessary anyway.

senekor:
Thanks! Yeah, I dropped `uuid` because the `add-exercise` script already generates a UUID for new exercises anyway, so there's no need for it.

PR #2099, opened 2025-09-30
paasio: Prevent non-unwinding panic in tests

senekor:
Related forum post:
https://forum.exercism.org/t/test-runner-fail-on-paas/19426

[no important files changed]


ellnix (approved):
Interesting bug and creative solution. I don't think we can set the panicked flag inside the BufWriter, although I assume that would work around this.

I think in addition to the comment explaining why this is done, we should include a link to the forum discussion.

senekor:
> I think in addition to the comment explaining why this is done, we should include a link to the forum discussion.

Done 👍

PR #2104, opened 2025-10-15
Fix clippy warnings

senekor:


senekor:
I basically fixed the exact same warning in 5fb310bb616ccf68b28dee82a15244bc65fcb86e, but not for all exercises. CI only checks modified exercises on a PR... now I ran the clippy tests locally for all exercises. Should get used to doing that every time new clippy warnings pop up.

PR #2105, opened 2025-11-02
Add test for Anagram exercise to cover equivalent code points

bmwant:
### Description

[Anagram problem](https://exercism.org/tracks/rust/exercises/anagram) states that

> The Rust track extends the possible letters to be any unicode character, not just ASCII alphabetic ones.

at the same time allowing naive solution to pass all the test cases

```rust
word.chars().flat_map(|c| c.to_lowercase()).collect();
```

This PR adds a test to specifically target [unicode macron](https://en.wikipedia.org/wiki/Macron_(diacritic)) symbol 
that belongs to the `ā` grapheme and should not consider this code point applied to different letters as anagrams.

### DIFF
--- a/exercises/practice/anagram/tests/anagram.rs
+++ b/exercises/practice/anagram/tests/anagram.rs
@@ -186,3 +186,12 @@ fn different_characters_may_have_the_same_bytes() {
     let expected = HashSet::from_iter([]);
     assert_eq!(output, expected);
 }
+
+#[test]
+fn equivalent_code_points_on_different_letters() {
+    let word = "āae";
+    let inputs = &["aeā", "aāe", "ēaa"];
+    let output = anagrams_for(word, inputs);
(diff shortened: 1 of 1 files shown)


bmwant:
@exercism/maintainers-admin can you check this one as it would affect results of all previously submitted solutions.
Or maybe just update description, so it would only require handling ASCII anagrams.

bmwant:
Hello, my apologies for handling things the wrong way. I read [this guide](https://exercism.org/docs/community/being-a-good-community-member/pull-requests) and checked other PRs before opening mine, and I didn’t mean to be rude or to seek attention with my proposed change.

I misunderstood the bot’s comment, which suggested that I should ping someone from the mentioned group if my change might affect others’ solutions. I really enjoy the platform and just wanted to contribute back. As an apology gesture, I’ve also made a donation to support the organization.

Sorry for any inconvenience caused by my actions.

PR #2134, opened 2026-04-11
bin/verify_exercises_in_docker.sh

keiravillekode:
This script from generic-track provides a convenient way to test exercises against the test runner.

We make four changes

- a local test runner image can be specified using -i my-local-image

- allowing for unequal numbers of solution and example files: only copy while we have both source and destination

- the test runner is run as the current user

- file systems are not mounted as read-only

### DIFF
--- a/bin/verify_exercises_in_docker.sh
+++ b/bin/verify_exercises_in_docker.sh
@@ -1,0 +1,102 @@
+#!/usr/bin/env bash
+
+# Synopsis:
+# Verify that each exercise's example/exemplar solution passes the tests
+# using the track's test runner Docker image.
+# You can either verify all exercises or a single exercise.
+
+# Example: verify all exercises in Docker
+# bin/verify-exercises-in-docker
+
+# Example: verify single exercise in Docker
+# bin/verify-exercises-in-docker two-fer
+
+set -eo pipefail
+
+die() { echo "$*" >&2; exit 1; }
+
+required_tool() {
+    command -v "${1}" >/dev/null 2>&1 ||
+        die "${1} is required but not installed. Please install it and make sure it's in your PATH."
+}
+
+required_tool docker
+
+copy_example_or_examplar_to_solution() {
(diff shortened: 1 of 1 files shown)


keiravillekode:
[loop_over_rust_exercises.txt](https://github.com/user-attachments/files/26645107/loop_over_rust_exercises.txt)

I looped over the exercises, and ran the script for each exercise.

The following fail:

```text
alphametics
decimal
diffie-hellman
grep
pig-latin
poker
```

senekor (approved):
Looks very useful!

senekor:
@keiravillekode actually I just tried running this locally and I'm getting:

```
Verifying assembly-line exercise...
Error: failed to write results.json

Caused by:
    Permission denied (os error 13)
cat: /tmp/exercism-verify-assembly-line-Wu1om/results.json: No such file or directory
```

So, the container cannot write to /output? Any ideas what's going wrong?

keiravillekode:
I am not experiencing that, running `Ubuntu 24.04.4 LTS`

Line 42 tells docker to run as the current user, instead of running as root. We can remove that line, but then we have a different error:

At the end we delete the temporary directory using

```bash
trap 'rm -rf "$tmp_dir"' EXIT
```

The `rm` fails if sub-directories like `target/debug/deps` `target/debug/.fingerprint` `target/debug/incremental` are created as root.

senekor:
I also get the same error when I remove `-u "$(id -u):$(id -g)"` 🤔

senekor:
Ugh, sorry about that. I had docker installed from the Fedora repos. Removing and installing with the official Docker instructions fixed it. There must be some weird difference between these two versions.

senekor:
> The following fail:
> 
> ```
> alphametics
> decimal
> diffie-hellman
> grep
> pig-latin
> poker
> ```

For reference, that was because their example solutions have different dependencies than the example `Cargo.toml` file. This is fixed mostly with #2136, but also updating the dependencies of individual exercises to ones that are actually present in the test runner.

PR #2135, opened 2026-04-13
verify_exercises_in_docker: avoid version mismatch

senekor:
The versions are pinned in the test runner. Hopefully they are not too
outdated, but they can easily be a few patch versions behind. If the
solution directory contains a lock file, these requirements often won't
be satisfied in the test runner.


ellnix (approved):
~~It seems fine, but did a conflict like this ever happen? If there's a case, I can't think of it. If it's prophylactic, that's fine.~~

Nevermind, this was the wrong PR to see first :smile: I see the other ones now.

senekor:
It's a valid question, since the lock file might not be created in some (most?) workflows. I do sometimes have them lying around. For quick and dirty changes to exercises, I `cd` into the exercise directory, copy the example solution to lib.rs, make my changes, run `cargo test`, copy the changes back to the example solution, restore lib.rs. So, this did fix a real issue for me :)

PR #2136, opened 2026-04-13
verify_exercises_in_docker: fix handling of example Cargo files

senekor:
Solutions that used different dependencies than the ones specified in the template Cargo.toml would fail in the verify_exercises_in_docker script. To fix that, the file is included in the "filed.example" list, such that it is copied into the temporary solution directory, and moved to Cargo.toml.


ellnix (approved):
Works as intended, good job on these fixes 👍

PR #2137, opened 2026-04-13
alphametics: improve example solution

senekor:
The previous example didn't compile in the test runner, because the dependencies aren't present anymore. I was able to fix it with minor adjustments, but it was very slow. So, I replaced it with a faster solution to speed up runs of the verify_exercises_in_docker script.


ellnix (approved):
LGTM!

PR #2138, opened 2026-04-13
diffie-helman: fix example solution

senekor:
The test runner doesn't have such an old version of `rand` anymore, so I updated it.


ellnix (approved):
LGTM!

PR #2140, opened 2026-04-13
pig-latin: fix example solution

senekor:
The test runner doesn't have the regex crate, because it is so slow to compile that most user-provided solution immediately time out just from including that dependency. regex-lite does the job just as well here.


ellnix (approved):
LGTM!

PR #2147, opened 2026-05-25
Format all `instructions.append.md` to start with an H1 (required but ignored) and H2

IsaacG:
See related discussions [here](https://forum.exercism.org/t/grep-instructions-append-may-need-a-visible-header/53923) as well as https://github.com/exercism/docs/pull/592. Append files should all have an H1. Including an H2 helps break the flow from the non-append file to the append file so it does not read as the same stream of thought.

### DIFF
--- a/exercises/practice/anagram/.docs/instructions.append.md
+++ b/exercises/practice/anagram/.docs/instructions.append.md
@@ -1,6 +1,6 @@
 # Instructions append
 
-## Implementation
+## Unicode and lifetimes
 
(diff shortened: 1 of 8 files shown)


senekor:
I don't like "implementation" as a generic header. I think the h2 headers should be exercise specifics. Do you want to go through the exercises manually and adjust each one? I can do that myself as well.

(also, the whitespace changes should be reverted, to keep the history clean)

IsaacG:
Done

senekor (approved):
thanks!

PR #2151, opened 2026-07-01
rna-transcription: add test to avoid cheesing

senekor:
In the normal tests for this exercise, we compare two RNA instances. If the user defines RNA as an empty struct, those instances will always compare equal. This is prevented with the new test to make sure different RNA don't compare equal.

I had to add the `lower` filter to keep the test identifiers stable. It seems the template and the generated tests have drifted a bit. I'll have to investigate and fix that in a separate patch.

Forum post:
https://forum.exercism.org/t/successful-null-solution-for-rna-transcription/63064

### DIFF
--- a/exercises/practice/rna-transcription/.meta/test_template.tera
+++ b/exercises/practice/rna-transcription/.meta/test_template.tera
@@ -1,4 +1,21 @@
 use rna_transcription::{Dna, Rna};
+
+{#-
+    We usually do not add additional tests directly in the template.
+    In this case however, the structure of the tests is different from the
+    regular ones. Accommodating that in the template would be complicated
+    and unnecessary. We also want to test this early, so that users don't
+    make it through basically all tests without doing anything.
+#}
+
+#[test]
+#[ignore]
+fn different_rna_are_different() {
+    let a = Rna::new("G").unwrap();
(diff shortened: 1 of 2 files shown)


ellnix (approved):
That's a very funny solution, but this exercise sure doesn't have a lot of tests upstream. I suppose the structs being empty was just one way in an entire universe of cheese since you could implement PartialEq however you like.

ellnix:
I went through and checked why the test template generation changed, it seems like the POV exercise had properties that were camel cased, and you (I think) added a helper in makeident to turn them into snake cased first. The commit:

https://github.com/exercism/rust/pull/2094/changes/4e6346bf59c45eea534882199954668f15a462af#diff-6e99d847a66783ee3d4d58d400532ee97bbb6e357026c274bef49cbb8db370b9 from #2094 

Perhaps converting camel cased to snake cased can be its own filter? That way you could apply it selectively to avoid causing issues for exercises like this one.


`exercises/practice/pov/.meta/test_template.tera:47`
```tera
{{ test_group.cases[0].property | make_ident }}
```

senekor:
Ah, good digging, thanks. I have a fix prepared. I'll merge this PR later to get the `| lower` out of the diff.

PR #2152, opened 2026-07-01
mise: add task to update all templated exercises

senekor:
This makes it easier to test changes of the exercise generator against
all exercises that already have a template. Avoiding unnecessary changes
to existing exercises keeps the history clean.


senekor:
I used this to test #2153. The diff isn't completely clean yet, some exercises could use an update. But the diff is much better than when running this script on main.

ellnix (approved):
LGTM

PR #2153, opened 2026-07-01
fix acronym-handling of make_ident filter

senekor:
When this function was added to the make_ident filter, it wasn't tested
with existing templated exercises. For ones with acronyms in the test
descriptions, those acronyms would be split by this function. (example:
RNA -> r_n_a). This fixes that. The function doesn't perform full
conversion to snake case anymore, which we don't need anyway, so it
is renamed to reflect that.


ellnix (approved):
Neat implementation of the function, I think this is closer to a typical implementation of an camel-case to underscore. For instance ActiveSupport's `#underscore` method from the ruby & rails world: https://api.rubyonrails.org/classes/ActiveSupport/Inflector.html#method-i-underscore (does some more domain specific stuff but the relevant logic is treating a sequence of uppercase characters like a singular block)

PR #2154, opened 2026-07-01
tooling: fix serializing of custom metadata

senekor:
The struct `Custom` contains two fields. If one of them is set, the
other will also be serialized with a `null` value, which we don't want.
The attribute `skip_serializing_if = "Option::is_none"` prevents that.


ellnix (approved):
Seems harmless, LGTM!

PR #2155, opened 2026-07-01
bottle-song: bring tests and template in sync

senekor:
The pretty_assertions import was mistakenly added to the test file
directly, instead of going through the template:
7978fdb604d9ea713de0a4105a8d56587ff56d16


ellnix (approved):
LGTM.

PR #2157, opened 2026-07-01
raindrops: regenerate tests

senekor:
The test generator now splits camel case to create identifiers.

[no important files changed]


ellnix (approved):
The names actually look better than before

PR #2158, opened 2026-07-01
eliuds-eggs: format config.json

senekor:
Running `mise run update-exercise --slug eliuds-eggs` will format the
config file in this way.


ellnix (approved):
Seems like the prettifying is a side effect of re-saving the config file after the essential fields are created if they don't exist, so I think it's here to stay since it needs to be written programmatically anyway.

LGTM!

PR #2162, opened 2026-07-28
macros: attach doctests to separate items

senekor:
Cargo emits the item name in the test output for doctests. In the future, we can use this in the test runner to display a better name.

[no important files changed]


ellnix (approved):
Looks fine, I'm guessing this is in preparation for supporting doc tests in the test runner in the future.

